PR #8, opened 2019-12-04
Fix typo in cli

mpkocher:
Fix Unbound error from unassigned var.

### DIFF
--- a/metaflow/main_cli.py
+++ b/metaflow/main_cli.py
@@ -208,7 +208,7 @@ def pull(episode):
     if not episode:
         episodes = get_all_episodes()
     else:
-        episodes = list(episode)
+        episodes = [episode]
         # Validate that the list is valid.
         for episode in episodes:
             validate_episode(episode)
--- a/metaflow/main_cli.py
+++ b/metaflow/main_cli.py
@@ -208,7 +208,7 @@ def pull(episode):
     if not episode:
         episodes = get_all_episodes()
     else:
-        episodes = list(episodes)
+        episodes = list(episode)
         # Validate that the list is valid.
         for episode in episodes:
             validate_episode(episode)


mpkocher:
I would suggest standardizing on a convention for naming the tutorials. The mixing of "episodes" vs tutorial (name/id) is a bit unnecessarily confusing. 

Currently, it's ` metaflow tutorials pull --episode '02-statistics'. 

Whereas, either of these might be more obvious. 

` metaflow tutorials pull --name '02-statistics'
` metaflow tutorials pull --id '02-statistics'

metaflow/main_cli.py
```diff
@@ -208,7 +208,7 @@ def pull(episode):
     if not episode:
         episodes = get_all_episodes()
     else:
-        episodes = list(episodes)
+        episodes = [episode]
         # Validate that the list is valid.
         for episode in episodes:
```
binarycrayon: good find, might as well check episode outside of if-else? i.e.
```
    if not episode:
        episodes = get_all_episodes()
    else:
        episodes = list(episodes)
    # Validate that the list is valid.
    for episode in episodes:
        validate_episode(episode)
```

savingoyal:
Thanks for the PR! Much appreciated! Please refer to the [contributing guide](https://github.com/Netflix/metaflow/blob/master/CONTRIBUTING.md). We are aligning internally on how to make contributing to Metaflow easy and transparent.

metaflow/main_cli.py
romain-intel: That would indeed be even better :).

seeravikiran: I had considered against this for the following reason:
get_all_episodes() anyway looks at the source_dir (`get_tutorials_dir`) to list all the valid episodes, and validate_episode() does the same per episode. So I think moving the loop outside would end up also validating `get_all_episodes()` output which seems unnecessary.
Happy to hear alternate arguments, but just wanted to share the original author's perspective. :)

mpkocher: It would be useful to consider simplify this. The motivation for iterating over a list that will only have one element in it is a bit unclear. Hence, the confusion about the validation block being a potential indent mistake. 

```python
    if not episode:
        episodes = get_all_episodes()
    else:
        validate_episode(episode)
        episodes = [episode]
```

The original intent of this PR was to fix the fundamental bug with pulling a particular tutorial. 

It might be better for these cleanup/enhancements to be done in house with team members that have a clear understanding of priorities of the project (i.e., there could be other more important issues to address).

seeravikiran: I agree with the original motivation of the PR, and therefore I'm fine with it fixing it just that.
I'm happy to follow-up with stylistic / readability improvements when I revisit this area of code. Your point about validating a single episode is very fair - it's possible that got lost across the multiple revisions to that file.

Thanks for the thoughtful response @mpkocher.

seeravikiran: I just realized something post accepting the PR.
The idea behind the loop is because pull is capable of accepting multiple episodes (and not just a single one); and the type of the value would be a tuple.
I’m not sure if list(...) and [...] have the same semantics converting tuple to list; but I will double check at HEAD.

Just thought I’d mention since I remembered just now. :(

mpkocher: Perhaps something did get lost in revision history, but In the current code, the click defined option will only pull one. 

```python
@click.option('--episode', default="", help="Optional episode name "\
              "to pull only a single episode.")
```

I believe you'll need to set 'nargs' to "*" or similar if you want the type to be an N-tuple/list.

```bash
metaflow tutorials pull --episode 00-helloworld 01-playlist
Metaflow (2.0.0)

Usage: metaflow tutorials pull [OPTIONS]
Try "metaflow tutorials pull --help" for help.

Error: Got unexpected extra argument (01-playlist)
```

I'm a bit surprised click doesn't raise an exception here. Apparently, it will use the last value provided and discard previous values. 

```bash
metaflow tutorials pull --episode 00-helloworld --episode 01-playlist
Metaflow (2.0.0)

Pulling episode "01-playlist" into your current working directory.

To know more about an episode, type:
metaflow tutorials info [EPISODE]
```

seeravikiran: When I typed the above response, I think I couldn't view the overall code from my phone, so apologies for that.

Yes it seems the code has evolved from taking multiple episodes to a single one; so the current PR is great. Thanks again @mpkocher.

PR #20, opened 2019-12-05
DataArtifact add size/type method

SteNicholas:
DataArtifact add size/type method to tell the size of the data contained in this artifact and type for this artifact.

### DIFF
--- a/metaflow/client/core.py
+++ b/metaflow/client/core.py
@@ -589,6 +589,7 @@ class MetaflowCode(object):
         self._path = info['location']
         self._ds_type = info['ds_type']
         self._sha = info['sha']
+        self._size = info['size']
         with filecache.get_data(self._ds_type, self._flow_name, self._sha) as f:
             self._tar = tarfile.TarFile(fileobj=f)
             # The JSON module in Python3 deals with Unicode. Tar gives bytes.
@@ -695,11 +696,7 @@ class DataArtifact(MetaflowObject):
         size
             Object size contained in this artifact
         """
-        ds_type = self._object['ds_type']
-        sha = self._object['sha']
-        with filecache.get_data(ds_type, self.path_components[0], sha) as f:
-            obj = pickle.load(f)
-            return sys.getsizeof(obj)
+        return self._object['size']
 
     @property
     def type(self):
--- a/metaflow/client/core.py
+++ b/metaflow/client/core.py
@@ -1,4 +1,5 @@
 from __future__ import print_function
+import sys
 import time
 import tarfile
 import json
@@ -655,6 +656,10 @@ class DataArtifact(MetaflowObject):
     ----------
     data : object
         The unpickled representation of the data contained in this artifact
+    size : int
+        The unpickled size of the data contained in this artifact
+    type : string
+        The specific type for this artifact
     sha : string
         SHA encoding representing the unique identity of this artifact
     finished_at : datetime
@@ -680,13 +685,33 @@ class DataArtifact(MetaflowObject):
             obj = pickle.load(f)
             return obj
 
-    # TODO add
-    # @property
-    # def size(self)
+    @property
+    def size(self):
+        """
+        Unpickled size of the data contained in this artifact.
 
-    # TODO add
-    # @property
-    # def type(self)
+        Returns
+        -------
+        size
+            Object size contained in this artifact
+        """
+        ds_type = self._object['ds_type']
+        sha = self._object['sha']
+        with filecache.get_data(ds_type, self.path_components[0], sha) as f:
+            obj = pickle.load(f)
+            return sys.getsizeof(obj)
+
+    @property
+    def type(self):
+        """
+        Specific type for this artifact.
+
+        Returns
+        -------
+        string
+            Type of this artifact
+        """
+        return self._object['type']
 
     @property
     def sha(self):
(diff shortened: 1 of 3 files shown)


savingoyal:
@SteNicholas Thanks for the PR! Much appreciated! Please refer to the [contributing guide](https://github.com/Netflix/metaflow/blob/master/CONTRIBUTING.md). We are aligning internally on how to make contributing to Metaflow easy and transparent.

metaflow/client/core.py
```diff
@@ -680,13 +685,33 @@ def data(self):
...
+            Object size contained in this artifact
+        """
+        ds_type = self._object['ds_type']
+        sha = self._object['sha']
+        with filecache.get_data(ds_type, self.path_components[0], sha) as f:
+            obj = pickle.load(f)
+            return sys.getsizeof(obj)
+
+    @property
+    def type(self):
```
romain-intel: This function looks fine.

metaflow/client/core.py
```diff
@@ -680,13 +685,33 @@ def data(self):
...
-    # @property
-    # def type(self)
+        Returns
+        -------
+        size
+            Object size contained in this artifact
+        """
+        ds_type = self._object['ds_type']
+        sha = self._object['sha']
+        with filecache.get_data(ds_type, self.path_components[0], sha) as f:
```
romain-intel: I am not a big fan of having to unpickle and then throw away the result just to get the size (and doing it every time). A potentially better approach would be to store the size information along with the sha, ds_type, etc and return that directly.

SteNicholas: @romain-intel I couldn't agree with the approach you mentioned. The sha, ds_type, etc are stored in json. Storing the size information into json is unreasonable. This should calculate the size by `get_data`.

savingoyal: @SteNicholas Why do you think storing the size information is unreasonable?

SteNicholas: @savingoyal In my option, the sha, ds_type etc. information is stored in json file. And the size couldn't be defined the value in json.

savingoyal: Why not?

SteNicholas: @savingoyal Do you mean that the size should be defined the value in json? I will modify this by storing the size in json.

savingoyal: yes, indeed.

SteNicholas: @savingoyal I have already modified the size storing in info. Please review this again.

savingoyal: Thanks, I will take a look in the morning.

romain-intel:
Will look at this a little later today. Apologies for the delay.

SteNicholas:
> Will look at this a little later today. Apologies for the delay.

When you have free time, please review this again. Thanks.

metaflow/datastore/datastore.py
```diff
@@ -525,7 +525,8 @@ def _register_data_artifacts(self, objects, info):
                                   ds_type=self.TYPE,
                                   url=self.object_path(sha),
                                   sha=sha,
-                                  type=info[var]['encoding'])
+                                  type=info[var]['encoding'],
+                                  size=info[var]['size'])
```
romain-intel: I believe the size that is in info corresponds to the size of the pickled object and I believe you want (which I agree with) to return the size of the unpickled object. You may have to modify _save_object in datastore.py to properly return the unpickled size.

SteNicholas: @romain-intel Yeah, I made mistake in the return size of _save_object. And I have modified this unpickled object size. Please review this again.

metaflow/datastore/datastore.py
```diff
@@ -441,9 +441,10 @@ def _save_object(self, transformable_obj, var, force_v4=False):
                                         % var)
                 transformable_obj.transform(lambda x: pickle.dumps(x, protocol=4))
         sha = sha1(transformable_obj.current()).hexdigest()
-        sz = len(transformable_obj.current())
-        self.save_data(sha, transformable_obj)
-        return sha, sz, blobtype
+        pickle_sz = len(transformable_obj.current())
+        path = self.save_data(sha, transformable_obj)
+        unpickle_sz = sys.getsizeof(open(path, 'rb'))
```
romain-intel: Path is not necessarily local (could be on S3) so this won't work. Even if local, this may be a GZip path.

SteNicholas:
> Apologies for the delay in re-reviewing this. My feeling right now is to punt on the size question (ie: not put anything related to size) and just keep the type notion. The reason for this is after a bit more thought, I wonder if we can ever return the unpickled size. There is no good way in Python to get the size of an object in Python (or at least no easy way). The pickled size may be a good approximation but I wonder if that has use to the user. The size on file (so pickled + gzip) may also have some value. In short, I think this may require a bit more thought to determine what "size" means for an artifact.

@romain-intel Therefore, in this interface, whether to provide user with `size` method? And about the size, I thought that the size should be defined that size is unpickled size of artifact object. Because in `DataArtifact`, the `data` method returns unpickled representation of the data contained in this artifact, so the `size` method should return unpickled size. Am I right?

romain-intel:
size was integrated in another PR. This is probably also a bit old since the datastore changed a lot so a new one can be opened if needed.

PR #33, opened 2019-12-07
Use collections.abc for ABC import to fix Python 3.9 compatibility.

tirkarthi:
Importing ABC from collections module will raise `ImportError` in Python 3.9 and raises `DeprecationWarning` from 3.4 . Hence use `collections.abc` and also preserve compatibility with Python 2. Ref : https://github.com/python/cpython/pull/10596

### DIFF
--- a/metaflow/datastore/datastore.py
+++ b/metaflow/datastore/datastore.py
@@ -4,7 +4,10 @@ import sys
 import time
 from hashlib import sha1
 from functools import partial
-from collections import Sequence
+try:
+    from collections.abc import Sequence
+except ImportError:
+    from collections import Sequence
 
 try:
     # Python 2
--- a/metaflow/datastore/datastore.py
+++ b/metaflow/datastore/datastore.py
@@ -4,10 +4,6 @@ import sys
 import time
 from hashlib import sha1
 from functools import partial
-try:
-    from collections.abc import Sequence
-except ImportError:
-    from collections import Sequence
 
 try:
     # Python 2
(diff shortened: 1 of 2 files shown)


savingoyal:
Thanks for the PR! Much appreciated! Please refer to the [contributing guide](https://github.com/Netflix/metaflow/blob/master/CONTRIBUTING.md). We are aligning internally on how to make contributing to Metaflow easy and transparent.

metaflow/datastore/datastore.py
```diff
@@ -4,7 +4,10 @@
 import time
 from hashlib import sha1
 from functools import partial
-from collections import Sequence
+try:
```
romain-intel: Actually, Sequence does not seem to be used anywhere in this file so maybe we can just remove the line and no add anything :).

metaflow/plugins/conda/conda_step_decorator.py
```diff
@@ -11,7 +10,10 @@
     from urlparse import urlparse
 except:
     from urllib.parse import urlparse
-
+try:
```
romain-intel: Perfect :)

metaflow/datastore/datastore.py
tirkarthi: Thanks, I just did a find and replace. Removed the unused import now.

savingoyal:
@tirkarthi Can you resolve the conflicts on this patch so that we can merge it.

PR #55, opened 2019-12-14
Better decorator parsing on CLI

savingoyal:
Decorators can be specified for Metaflow steps via CLI. Some of the
decorator args can include commas, which break the current logic.

For eg, `--with conda:'libraries="{a:x.y,b:x.y}"',python=x.y.z`

This patch fixes the behavior by updating the parsing logic for decospec.

### DIFF
--- a/metaflow/decorators.py
+++ b/metaflow/decorators.py
@@ -1,5 +1,6 @@
 import traceback
 from functools import partial
+import re
 
 from .flowspec import FlowSpec
 from .exception import MetaflowException, InvalidDecoratorAttribute
@@ -99,7 +100,8 @@ class Decorator(object):
             return cls()
         else:
             name, attrspec = top
-            attrs = dict(a.split('=') for a in attrspec.split(','))
+            attrs = dict(a.split('=') for a in 
+                re.split(''',(?=(?:[^'"]|'[^']*'|"[^"]*")*$)''', attrspec))
             return cls(attributes=attrs)
 
     def make_decorator_spec(self):
--- a/metaflow/decorators.py
+++ b/metaflow/decorators.py
@@ -100,8 +100,8 @@ class Decorator(object):
             return cls()
         else:
             name, attrspec = top
-            attrs = dict(a.split('=') for a in 
-                re.split(''',(?=(?:[^'"]|'[^']*'|"[^"]*")*$)''', attrspec))
+            attrs = dict(map(lambda x: x.strip(), a.split('=')) 
+                for a in re.split(''',(?=[\s\w]+=)''', attrspec.strip('"\'')))
             return cls(attributes=attrs)
 
     def make_decorator_spec(self):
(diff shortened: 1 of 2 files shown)


metaflow/decorators.py
```diff
@@ -99,7 +100,8 @@ def _parse_decorator_spec(cls, deco_spec):
             return cls()
         else:
             name, attrspec = top
-            attrs = dict(a.split('=') for a in attrspec.split(','))
+            attrs = dict(a.split('=') for a in 
+                re.split(''',(?=(?:[^'"]|'[^']*'|"[^"]*")*$)''', attrspec))
```
romain-intel: Another suggestion for this regexp would be something of this nature:
`(?=[\s\w]+=)`. It basically matches commas that are "top-level" in the sense that they directly precede another attribute of the type name=value. This would allow you to parse exactly the way you would specify in a step.As an example, we have in the tutorial:
`libraries={'editdistance': '0.5.3', 'pandas' : '0.24.2'}, python=3.8.1` which would be parseable by this method.

Further, I would possibly add a few strips to allow for spaces again to make it more legible for the user:
`attrs = dict(map(lambda x: x.strip(), a.split('=')) for a in re.split(''',(?=[\s\w]+=)''', attrspec.strip('"\'')))`

metaflow/plugins/conda/conda_step_decorator.py
```diff
@@ -79,8 +84,11 @@ def _lib_deps(self):
         base_deps = self.base_attributes['libraries']
         deps.update(base_deps)
         step_deps = self.attributes['libraries']
-        if isinstance(step_deps, collections.Mapping):
-            deps.update(step_deps)
+        if isinstance(step_deps, (unicode, basestring)):
+            step_deps = step_deps.strip('"{}"')
+            if step_deps:
+                step_deps = dict(a.split(':') for a in step_deps.split(','))
```
romain-intel: Similarly, to deal with spaces and whatnot, I suggest something like this:
`dict(map(lambda x: x.strip().strip('"\''), a.split(':')) for a in step_deps.split(','))

metaflow/plugins/conda/conda_step_decorator.py
```diff
@@ -79,8 +84,11 @@ def _lib_deps(self):
         base_deps = self.base_attributes['libraries']
         deps.update(base_deps)
         step_deps = self.attributes['libraries']
-        if isinstance(step_deps, collections.Mapping):
-            deps.update(step_deps)
+        if isinstance(step_deps, (unicode, basestring)):
+            step_deps = step_deps.strip('"{}"')
```
romain-intel: Maybe to be safe, consider the strip `'"{}\''` (the second " is irrelevant in any case).

savingoyal: Done. Thanks for noticing the typo!

metaflow/decorators.py
savingoyal: Done.

metaflow/plugins/conda/conda_step_decorator.py
savingoyal: Done.

Issue #54, opened 2019-12-14
Improve CTRL-C handling


romain-intel:
The current handling of interrupted flows (ie: when the user hits CTRL-C while a flow is running) has two issues:
  - there is a mandatory 1 second wait for each and every subprocess running (ie: 1 per in-flight-task)
  - there is the possibility that the subprocesses are killed before they have a chance to clean up (this is particularly important for subprocesses handling Batch)

It should be possible to:
  - wait for all subprocess to properly shut down (with a timeout)
  - kill any stragglers

This would both eliminate the 1 second minimum time per task and avoid (or at least mitigate) early kills.

savingoyal:
Fixed in https://github.com/Netflix/metaflow/pull/56

PR #56, opened 2019-12-15
Address #54 and improve interrupted shutdown

romain-intel:
Although not fully guaranteeing a clean shutdown every time, this patch makes the shutdown a little more robust (giving time to children process to shutdown), prevents multiple CTRL-C from interrupting the shutdown (in most cases) and is possibly faster in the presence of large fan outs.

### DIFF
--- a/metaflow/runtime.py
+++ b/metaflow/runtime.py
@@ -240,11 +240,7 @@ class NativeRuntime(object):
     def _killall(self):
         # If we are here, all children have received a signal and are shutting down.
         # We want to give them an opportunity to do so and then kill
-        live_workers = []
-        temp_workers = {}
-        for worker in self._workers.values():
-            temp_workers[worker] = 1
-        live_workers = temp_workers.keys()
+        live_workers = list(set(self._workers.values()))
         now = int(time.time())
         self._logger('Terminating %d active tasks...' % len(live_workers),
                      system_msg=True, bad=True)
@@ -258,8 +254,7 @@ class NativeRuntime(object):
                          system_msg=True, bad=True)
             for worker in live_workers:
                 worker.kill()
-        self._logger('Flushing logs -- you may see messages from the tasks that were shutdown...',
-                     system_msg=True, bad=True)
+        self._logger('Flushing logs...', system_msg=True, bad=True)
         # give killed workers a chance to flush their logs to datastore
         for _ in range(3):
             list(self._poll_workers())
--- a/metaflow/runtime.py
+++ b/metaflow/runtime.py
@@ -216,6 +216,13 @@ class NativeRuntime(object):
 
         except KeyboardInterrupt as ex:
             self._logger('Workflow interrupted.', system_msg=True, bad=True)
+            # Ignore new signals
+            import signal
+
+            def print_ignore(signum, frame):
+                self._logger('Shutdown in progress -- ignoring additional CTRL-C',
+                             system_msg=True, bad=True)
+            signal.signal(signal.SIGINT, print_ignore)
             self._killall()
             exception = ex
             raise
@@ -238,7 +245,25 @@ class NativeRuntime(object):
                                         'by the end of flow.')
 
     def _killall(self):
+        # If we are here, all children have received a signal and are shutting down.
+        # We want to give them an opportunity to do so and then kill
+        live_workers = []
+        temp_workers = {}
         for worker in self._workers.values():
+            temp_workers[worker] = 1
+        live_workers = temp_workers.keys()
+        now = int(time.time())
+        self._logger('Pre-killing %d workers' % len(live_workers), system_msg=True, bad=True)
+        while live_workers and int(time.time()) - now < 5:
+            # While not all workers are dead and we have waited less than 5 seconds
+            new_live_workers = []
+            for worker in live_workers:
+                if not worker.pre_kill():
+                    new_live_workers.append(worker)
+            live_workers = new_live_workers
+        self._logger('Have %d remaining workers after %d seconds' % (
+            len(live_workers), int(time.time()) - now), system_msg=True, bad=True)
+        for worker in live_workers:
             worker.kill()
         # give killed workers a chance to flush their logs to datastore
         for _ in range(3):
@@ -381,7 +406,7 @@ class NativeRuntime(object):
                         task = worker.task
                         if returncode:
                             # worker did not finish successfully
-                            if worker.killed or\
+                            if worker.prekilled or \
                                returncode == METAFLOW_EXIT_DISALLOW_RETRY:
                                 self._logger("This failed task will not be "
                                              "retried.", system_msg=True)
@@ -770,6 +795,7 @@ class Worker(object):
 
         self._encoding = sys.stdout.encoding or 'UTF-8'
         self.killed = False
+        self.prekilled = False
 
     def _launch(self):
         args = CLIArgs(self.task)
@@ -826,13 +852,18 @@ class Worker(object):
         return (self._proc.stderr.fileno(),
                 self._proc.stdout.fileno())
 
-    def kill(self):
-        if not self.killed:
+    def pre_kill(self):
+        if self.killed:
+            return True
+        if not self.prekilled:
             for fileobj, buf in self._logs.values():
                 buf.write(b'[KILLED BY ORCHESTRATOR]\n', system_msg=True)
+            self.prekilled = True
+        return self._proc.poll() is not None
+
+    def kill(self):
+        if not self.killed:
             try:
-                # wait for the process to clean up after itself
-                select.poll().poll(1000)
                 self._proc.kill()
             except:
                 pass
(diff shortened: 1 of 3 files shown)


metaflow/runtime.py
```diff
@@ -216,6 +216,13 @@ def execute(self):
 
         except KeyboardInterrupt as ex:
             self._logger('Workflow interrupted.', system_msg=True, bad=True)
+            # Ignore new signals
+            import signal
+
+            def print_ignore(signum, frame):
+                self._logger('Shutdown in progress -- ignoring additional CTRL-C',
```
savingoyal: It will be nice if the second `ctrl-C` instantly terminates the execution instead of blocking on the shutdown procedure.

metaflow/runtime.py
```diff
@@ -238,8 +245,32 @@ def execute(self):
...
+        # If we are here, all children have received a signal and are shutting down.
+        # We want to give them an opportunity to do so and then kill
+        live_workers = []
+        temp_workers = {}
         for worker in self._workers.values():
-            worker.kill()
+            temp_workers[worker] = 1
+        live_workers = temp_workers.keys()
+        now = int(time.time())
+        self._logger('Waiting for %d workers to shutdown...' % len(live_workers),
```
savingoyal: The notion of workers is not exposed to users of Metaflow. Maybe reword this or take this out entirely?

metaflow/runtime.py
```diff
@@ -826,16 +858,22 @@ def fds(self):
         return (self._proc.stderr.fileno(),
                 self._proc.stdout.fileno())
 
-    def kill(self):
-        if not self.killed:
+    def pre_kill(self):
```
savingoyal: maybe `clean` instead of `pre_kill` and `self.cleaned` instead of `self.prekilled`?

metaflow/runtime.py
```diff
@@ -238,8 +245,32 @@ def execute(self):
...
 
     def _killall(self):
+        # If we are here, all children have received a signal and are shutting down.
+        # We want to give them an opportunity to do so and then kill
+        live_workers = []
+        temp_workers = {}
         for worker in self._workers.values():
-            worker.kill()
+            temp_workers[worker] = 1
+        live_workers = temp_workers.keys()
```
savingoyal: isn't `live_workers = self._workers.values()`?

metaflow/runtime.py
```diff
@@ -238,8 +245,32 @@ def execute(self):
...
+        now = int(time.time())
+        self._logger('Waiting for %d workers to shutdown...' % len(live_workers),
+                     system_msg=True, bad=True)
+        while live_workers and int(time.time()) - now < 5:
+            # While not all workers are dead and we have waited less than 5 seconds
+            new_live_workers = []
+            for worker in live_workers:
+                if not worker.pre_kill():
+                    new_live_workers.append(worker)
+            live_workers = new_live_workers
```
savingoyal: minor syntactic nit (up to you to incorporate it) - `live_workers = [worker for worker in live_workers if not worker.pre_kill()]`

metaflow/runtime.py
savingoyal: Similar for L:266 and L:272.

metaflow/runtime.py
savingoyal: Also, is the intended behavior that every worker gets the `pre_kill` signal at least once, even if it takes more than 5 seconds?

metaflow/runtime.py
romain-intel: I think before that's what it did. I wanted to try to prevent multiple CTRL-C from preventing us from shutting down. I can remove this though if you feel that behavior is better.

metaflow/runtime.py
romain-intel: Not quite; in workers, they appear twice (one for each FD). This dedups them.

metaflow/runtime.py
romain-intel: I can maybe say something like "Terminating %d running steps...". I do want to give some feedback to make sure the user feels something is happening.

metaflow/runtime.py
romain-intel: Yes, I thought of checking every time in the loop but decided against it because it would be less clean and nothing in the loop is blocking (they are all polls) so it should be relatively fast in all cases so it may take slightly more than 5 seconds but I doubt it will go longer.

metaflow/runtime.py
savingoyal: Yeah, let's remove it.

metaflow/runtime.py
savingoyal: so, `live_workers = set(self._workers.values())`

metaflow/runtime.py
```diff
@@ -238,8 +245,28 @@ def execute(self):
...
+            # While not all workers are dead and we have waited less than 5 seconds
+            live_workers = [worker for worker in live_workers if not worker.clean()]
+        if live_workers:
+            self._logger('Killing remaining %d tasks after having waited for %d seconds -- '
+                         'some tasks may not exit cleanly' % (len(live_workers),
+                                                              int(time.time()) - now),
+                         system_msg=True, bad=True)
+            for worker in live_workers:
+                worker.kill()
+        self._logger('Flushing logs -- you may see messages from the tasks that were shutdown...',
```
savingoyal: Should we remove this logging message?

metaflow/runtime.py
```diff
@@ -238,8 +245,28 @@ def execute(self):
...
+            temp_workers[worker] = 1
+        live_workers = temp_workers.keys()
+        now = int(time.time())
+        self._logger('Terminating %d active tasks...' % len(live_workers),
+                     system_msg=True, bad=True)
+        while live_workers and int(time.time()) - now < 5:
+            # While not all workers are dead and we have waited less than 5 seconds
+            live_workers = [worker for worker in live_workers if not worker.clean()]
+        if live_workers:
+            self._logger('Killing remaining %d tasks after having waited for %d seconds -- '
```
savingoyal: For consistency with L:256 - maybe just - `Killing %d remaining tasks...`

metaflow/runtime.py
romain-intel: I wanted to keep it to help users understand why they see printout after things have been waited for and terminated.

savingoyal: I am worried that this log is too verbose. Maybe just? - `Flushing logs...`

metaflow/runtime.py
seeravikiran: Could you also detail what `temp_workers` is for? If it's for the sole purpose of creating the set, I'd instead do what @savingoyal suggests w.r.t set.

metaflow/runtime.py
```diff
@@ -238,8 +238,28 @@ def execute(self):
...
         for worker in self._workers.values():
-            worker.kill()
+            temp_workers[worker] = 1
+        live_workers = temp_workers.keys()
+        now = int(time.time())
+        self._logger('Terminating %d active tasks...' % len(live_workers),
+                     system_msg=True, bad=True)
+        while live_workers and int(time.time()) - now < 5:
+            # While not all workers are dead and we have waited less than 5 seconds
+            live_workers = [worker for worker in live_workers if not worker.clean()]
```
seeravikiran: Feel free to reject this suggestion - 

This seems like a good use-case for `filter`:
live_workers = filter(lambda worker: not worker.clean(), live_workers)
I don't know the mutability guarantees for list iteration, so I'm not sure if using a function like `filter` is better.

As bonus it might work on set(s) or list(s).

metaflow/runtime.py
```diff
@@ -238,8 +238,23 @@ def execute(self):
                                         'by the end of flow.')
 
     def _killall(self):
-        for worker in self._workers.values():
-            worker.kill()
+        # If we are here, all children have received a signal and are shutting down.
+        # We want to give them an opportunity to do so and then kill
+        live_workers = list(set(self._workers.values()))
```
seeravikiran: Why do you need this be a list? It seems you just need iteration, and the below suggested `filter`.

metaflow/runtime.py
```diff
@@ -770,6 +785,7 @@ def __init__(self, task, max_logs_size):
 
         self._encoding = sys.stdout.encoding or 'UTF-8'
         self.killed = False
+        self.cleaned = False
```
seeravikiran: Could you add a short comment on what self.killed=True represents vs self.cleaned=True?
It also seems if self.killed = True, then self.cleaned must be True?

Defining the state transition perhaps in a short comment helps understand why you are tracking two states.

metaflow/runtime.py
romain-intel: I read a bit that filter may be a tad slower (not a huge deal here though) but I find list comprehension to be more readable so if OK, I would prefer to leave it as such.

metaflow/runtime.py
romain-intel: sure, I can change it. It made more sense for it to be a list when I had a loop with add but now I can put just a set.

PR #59, opened 2019-12-16
Revamp AWS configuration module

savingoyal:
Redo `metaflow configure aws`. Goes hand-in-hand with https://github.com/Netflix/metaflow-tools/pull/3


metaflow/main_cli.py
```diff
@@ -349,83 +348,79 @@ def sandbox(profile):
     # Persist to a file.
     persist_env(env_dict, profile)
 
-@configure.command(help='Get Metaflow up and running on your own AWS environment.')
+@configure.command(help='Configure Metaflow to access self-managed AWS resources.')
```
seeravikiran: Any reason you chose against the 'Get Metaflow up and running on your self-managed AWS resources.'? Is it getting trimmed?

Currently the language looks slightly inconsistent between `sandbox` and `aws`:
```
Commands:
  aws      Configure Metaflow to access self-managed AWS resources.
  reset    Reset configuration to run locally.
  sandbox  Get Metaflow up and running on our sandbox.
  show     Show existing configuration.
```

metaflow/main_cli.py
```diff
@@ -349,83 +348,79 @@ def sandbox(profile):
...
 @click.option('--profile', '-p', default='',
-              help="Optional user profile to allow storing multiple "
-                   "configurations. Please `export METAFLOW_PROFILE` to "
-                   "switch between profile configuration(s).")
+              help='Configure a named profile. Activate the profile by setting ' 
+                    '`METAFLOW_PROFILE` environment variable.')
```
seeravikiran: Indentation off by 1 on this line.

metaflow/main_cli.py
```diff
@@ -349,83 +348,79 @@ def sandbox(profile):
...
+        echo('\nSetup your access credentials by following this guide: ', nl=False)
+        echo('https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html',
+             fg='cyan')
+    else:
         env_dict = {}
         # Datastore configuration.
-        use_s3 = click.confirm('\nDo you want to use AWS S3 as your datastore?',
+        use_s3 = click.confirm('\nConfigure Amazon '
+                                'S3 as default datastore (required for AWS Batch)?',
                               default=True, abort=False)
```
seeravikiran: Can you add 'Amazon S3' in cyan?

metaflow/main_cli.py
```diff
@@ -349,83 +348,79 @@ def sandbox(profile):
...
-            env_dict['METAFLOW_DATATOOLS_SYSROOT_S3'] = datatools_s3_root
-
-            # Batch configuration (only if S3 is being used).
+            env_dict['METAFLOW_DATASTORE_SYSROOT_S3'] =\
+                click.prompt('\t' + cyan('METAFLOW_DATASTORE_SYSROOT_S3:') + 
+                  ' Amazon S3 url for metaflow datastore (s3://<bucket>/<prefix>)')
+
+            env_dict['METAFLOW_DATATOOLS_SYSROOT_S3'] =\
+                click.prompt('\t' + cyan('METAFLOW_DATATOOLS_SYSROOT_S3:') + 
+                  yellow(' (optional)') + ' Amazon S3 url for metaflow datatools (s3://<bucket>/<prefix>)',
```
seeravikiran: Nit: Do you need this again, since you suggest a default in brackets right after this. I vote for removing it, since it seems pretty clear with the actual default value.
```
METAFLOW_DATATOOLS_SYSROOT_S3: (optional) Amazon S3 url for metaflow datatools (s3://<bucket>/<prefix>) [s3://foo/metaflow/data]
```

metaflow/main_cli.py
```diff
@@ -349,83 +348,79 @@ def sandbox(profile):
...
+                    default='%s/data' % env_dict['METAFLOW_DATASTORE_SYSROOT_S3'],
+                    show_default=True)
+            # AWS Batch configuration (only if Amazon S3 is being used).
             use_batch =\
-                click.confirm('\nDo you want to use AWS Batch for compute?',
+                click.confirm('\nConfigure AWS Batch for compute?',
```
seeravikiran: Same for `AWS Batch` in cyan? The original intent is to given an idea what the next set of configuration is going to be about, with the cyan highlight(s).

metaflow/main_cli.py
```diff
@@ -349,83 +348,79 @@ def sandbox(profile):
...
-
-                env_dict['METAFLOW_BATCH_JOB_QUEUE'] = job_queue
-                env_dict['METAFLOW_BATCH_CONTAINER_IMAGE'] = default_image
-                env_dict['METAFLOW_BATCH_CONTAINER_REGISTRY'] =\
-                    container_registry
-                env_dict['METAFLOW_ECS_S3_ACCESS_IAM_ROLE'] = ecs_s3_role
-
+                env_dict['METAFLOW_BATCH_JOB_QUEUE'] =\
+                    click.prompt('\t' + cyan('METAFLOW_BATCH_JOB_QUEUE:') + 
+                      ' AWS Batch Job Queue')
```
seeravikiran: Could we be a bit more specific here? I thought there was some confusion what entry to have here (the arn or the queue name or what).

metaflow/main_cli.py
```diff
@@ -349,83 +348,79 @@ def sandbox(profile):
...
-                default_val =\
-                    '%s/conda' % env_dict['METAFLOW_DATASTORE_SYSROOT_S3']
-                package_s3root = \
-                    click.prompt('\tPlease enter the bucket prefix for storing '
-                                 'conda packages',
-                                 default=default_val)
-                env_dict['METAFLOW_CONDA_PACKAGE_S3ROOT'] = package_s3root
+            env_dict['METAFLOW_SERVICE_URL'] =\
+                click.prompt('\t' + cyan('METAFLOW_SERVICE_URL:') +
+                  ' URL for Metadata Service (Open to Public Access)')
```
seeravikiran: Could you double check if sentence case or lower case - which looks better - for the info in ()?

metaflow/main_cli.py
```diff
@@ -349,83 +348,79 @@ def sandbox(profile):
...
-                                      default=True, abort=False)
-            if use_conda:
-                echo('\tConda on AWS S3', fg='cyan')
-                default_val =\
-                    '%s/conda' % env_dict['METAFLOW_DATASTORE_SYSROOT_S3']
-                package_s3root = \
-                    click.prompt('\tPlease enter the bucket prefix for storing '
-                                 'conda packages',
-                                 default=default_val)
-                env_dict['METAFLOW_CONDA_PACKAGE_S3ROOT'] = package_s3root
```
seeravikiran: Just to double check - have you decided to retire this configuration for `METAFLOW_CONDA_PACKAGE_S3ROOT`?

metaflow/main_cli.py
savingoyal: Given that all the statements start with an assertive verb, it made sense to modify `Get...` to `Configure...`. I will make a similar language change for sandbox.

metaflow/main_cli.py
savingoyal: I tried that, it ended up being too many colors.

metaflow/main_cli.py
savingoyal: Will fix.

metaflow/main_cli.py
savingoyal: This is to follow a consistent pattern throughout.

metaflow/main_cli.py
savingoyal: Similar argument as above.

metaflow/main_cli.py
savingoyal: Both of them work equally well. For any of the configs, the user can specify ARNs or names as they are interchangeable.

metaflow/main_cli.py
savingoyal: I went with the sentence case for continuity here.

metaflow/main_cli.py
savingoyal: Yes, indeed.

PR #62, opened 2019-12-16
Fix docker registry parsing for AWS Batch decorator

savingoyal:
re.search instead of re.match to disambiguate between valid docker
registry urls


Shago7939:
Code of conduct

savingoyal:
@Shago7939 Could you elaborate?

PR #68, opened 2019-12-18
Fix json loading in datastore

hexa00:
This patch fixes json loading in python 3.5.

Before this patch one would get this error on metaflow resume:

 python3 playlist.py resume
Metaflow 2.0.1 executing PlayListFlow for user:atremblay
Validating your flow...
    The graph looks good!
Running pylint...
    Pylint is happy!
2019-12-18 14:54:41.596 Gathering required information to resume run (this may take a bit of time)...
    Internal error
  File "/home/atremblay/.local/lib/python3.5/site-packages/metaflow/cli.py", line 853, in main
    start(auto_envvar_prefix='METAFLOW', obj=state)
  File "/home/atremblay/.local/lib/python3.5/site-packages/click/core.py", line 764, in __call__
    return self.main(args, kwargs)
  File "/home/atremblay/.local/lib/python3.5/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/atremblay/.local/lib/python3.5/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/atremblay/.local/lib/python3.5/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, ctx.params)
  File "/home/atremblay/.local/lib/python3.5/site-packages/click/core.py", line 555, in invoke
    return callback(args, kwargs)
  File "/home/atremblay/.local/lib/python3.5/site-packages/metaflow/cli.py", line 508, in wrapper
    return func(args, kwargs)
  File "/home/atremblay/.local/lib/python3.5/site-packages/click/decorators.py", line 27, in new_func
    return f(get_current_context().obj, args, kwargs)
  File "/home/atremblay/.local/lib/python3.5/site-packages/metaflow/cli.py", line 557, in resume
    max_log_size=max_log_size  1024  1024)
  File "/home/atremblay/.local/lib/python3.5/site-packages/metaflow/runtime.py", line 123, in __init__
    prefetch_data_artifacts=PREFETCH_DATA_ARTIFACTS)
  File "/home/atremblay/.local/lib/python3.5/site-packages/metaflow/datastore/datastore_set.py", line 35, in __init__
    for step_name, task_id, attempt, data_blob in data_blobs]
  File "/home/atremblay/.local/lib/python3.5/site-packages/metaflow/datastore/datastore_set.py", line 35, in <listcomp>
    for step_name, task_id, attempt, data_blob in data_blobs]
  File "/usr/lib/python3.5/json/__init__.py", line 312, in loads
    s.__class__.__name__))

This patch makes it compatible with python 3.5 by decoding the blob as
utf-8

This is not an issue in 3.6 see: https://docs.python.org/3/whatsnew/3.6.html#json

### DIFF
--- a/metaflow/datastore/datastore_set.py
+++ b/metaflow/datastore/datastore_set.py
@@ -1,5 +1,7 @@
 import json
 
+from metaflow import util
+
 """
 MetaflowDatastoreSet allows you to prefetch multiple (read) datastores into a
 cache and lets you access them.
@@ -30,7 +32,7 @@ class MetaflowDatastoreSet(object):
                                attempt=attempt,
                                event_logger=event_logger,
                                monitor=monitor,
-                               data_obj=json.loads(data_blob),
+                               data_obj=json.loads(util.to_unicode(data_blob)),
                                artifact_cache=artifact_cache)
                       for step_name, task_id, attempt, data_blob in data_blobs]
         if prefetch_data_artifacts:


savingoyal:
@seeravikiran can you take a look?

metaflow/datastore/datastore_set.py
```diff
@@ -30,7 +30,7 @@ def __init__(self,
                                attempt=attempt,
                                event_logger=event_logger,
                                monitor=monitor,
-                               data_obj=json.loads(data_blob),
+                               data_obj=json.loads(isinstance(data_blob,str) and data_blob or data_blob.decode('utf-8')),
```
seeravikiran: Could you verify what happens if we always `data_blob.decode('utf-8')`? If it doesn't we can also leverage the `util` module in Metaflow which simplifies this task for you. 

When `data_obj=None` it seems we are leveraging `load_metadata` which does the decoding unconditionally [1], so we can probably do the same here, since the `data_blob` is directly returned from S3 client `get_many`.

[1] https://github.com/Netflix/metaflow/blob/master/metaflow/datastore/s3.py#L184

hexa00: Yeah my 1st version of the PR was to always decode but it crashes in 3.6  since in python > 3.5 the data_blob is already a binary and there's no decode method on that type..

I'm not sure why the data_blob type is of different type between python versions, something to investigate... ideas welcome there.

I'll checkout the util module... 

Note I'll be away for the holidays so it may take a while for me to take a second look...

seeravikiran: Thanks for the super quick response.

Regarding the util library, what I have in mind is something like `json.loads(util.to_unicode(data_blob))`

I find the current conditional slightly harder to reason about, so seeing if an existing function can handle the compatibility issues across (py) versions.

Also - no rush on this PR; since we anyway plan to merge only when we have a critical mass of work for the next release. Happy holidays!

hexa00: OK cool yeah that would surely be better, I also wasn't found of that conditional...

I'll test it out..

Note also, the current test framework doesn't test for 3.5, I usually like to provide a test with a fix, let me know if that's possible to do or if it's better to wait...

hexa00: @seeravikiran changed as per review

seeravikiran: Sorry I don't remember how this slipped my radar; but your new change looks good to me.

I'll approve once I can test the change myself. Resolving this conversation.

seeravikiran:
@hexa00 I cannot quite reproduce your issue. Could you please see where I might've missed something?

1. Created a conda environment for py3.5:
```
(py3.5) nfml-rchirravuriFPL:metaflow-tutorials rchirravuri$ python --version
Python 3.5.5
```
2. resume works without any changes:
```
(py3.5) nfml-rchirravuriFPL:metaflow-tutorials rchirravuri$ python3 01-playlist/playlist.py resume
Metaflow 2.0.1 executing PlayListFlow for user:rchirravuri
Validating your flow...
    The graph looks good!
Running pylint...
    Pylint not found, so extra checks are disabled.
2020-04-21 20:54:54.933 Gathering required information to resume run (this may take a bit of time)...
2020-04-21 20:54:54.944 Workflow starting (run-id 1587527694933029):
2020-04-21 20:54:54.946 [1587527694933029/start/1] Cloning results of a previously run task 1587527524108284/start/1
2020-04-21 20:54:55.440 [1587527694933029/bonus_movie/2] Cloning results of a previously run task 1587527524108284/bonus_movie/2
2020-04-21 20:54:55.454 [1587527694933029/genre_movies/3] Cloning results of a previously run task 1587527524108284/genre_movies/3
2020-04-21 20:54:56.096 [1587527694933029/join/4] Cloning results of a previously run task 1587527524108284/join/4
2020-04-21 20:54:56.608 [1587527694933029/end/5] Cloning results of a previously run task 1587527524108284/end/5
2020-04-21 20:54:57.141 Done!
```

I'm running this on your own repo (on master):
`git clone https://github.com/hexa00/metaflow.git hexa00_oss`

hexa00:
HI,
  I can't reproduce it either with a conda env with python 3.5... I was doing that on my laptop which I since upgraded so hard to reproduce exactly.

  The core of the issue however seems to be if data_blob is interpreted as a string or a binary, I tested in conda with 3.5 and it's a string.. however somehow on my laptop it must have been a binary.  I'd have to dig more to know in what conditions is is a binary or a string...

 And since json.load only supports binaries starting 3.6 this was causing an issue.

 I'll leave it to your judgement on our to proceed forward or not.

seeravikiran:
> HI,
> I can't reproduce it either with a conda env with python 3.5... I was doing that on my laptop which I since upgraded so hard to reproduce exactly.
>
Yeah I can imagine. 
> The core of the issue however seems to be if data_blob is interpreted as a string or a binary, I tested in conda with 3.5 and it's a string.. however somehow on my laptop it must have been a binary. I'd have to dig more to know in what conditions it is a binary or a string...
> 
I understand, I also liked @tuulos to weigh in - in case he seems a codepath where this was supposed to be binary; or why its working fine now that we try with 3.5.4.
> And since json.load only supports binaries starting 3.6 this was causing an issue.
> 
> I'll leave it to your judgement on our to proceed forward or not.
Happy to spend a bit more time seeing if this is reproducible and might affect more users. Otherwise I'm inclined to leave it as-is and follow-up with a fix if someone reports it again (hopefully with a reproducible use-case)

Let's also wait to hear what @tuulos thinks. I don't have a strong preference either way about this.

romain-intel:
Datastore has been updated. This is probably no longer relevant.

PR #76, opened 2019-12-26
Addition of conditional branching of Tasks

thesillystudent:
Pull request for adding the functionality of branching of tasks based on a condition.

The following changes are made - 

1. With the already existing implementation, a join operation is checked if there is an extra parameter passed to the function. But a join task for split-or takes only one input, the input from the branch being run. Hence another function **traverse_fix_split_or** is run after **split_or** to fix the graph.
2. Conditions are added for scheduling of out tasks of split-or and the scheduling of split-or join.

### DIFF
--- a/metaflow/graph.py
+++ b/metaflow/graph.py
@@ -180,7 +180,7 @@ class FlowGraph(object):
                         traverse(child, seen + [n], split_parents)
 
         vis = set()
-        def traverse2(node, cnt):
+        def traverse_fix_split_or(node, cnt):
             if node.name in vis:
                 return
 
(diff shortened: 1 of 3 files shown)


savingoyal:
@thesillystudent Thanks for the PR. Conditional branching is kind of a mine-field in the sense that it makes reasoning about workflows (and state transitions) non-trivial for users. Do you have specific instances of workflows that necessarily require conditional branching?

thesillystudent:
@savingoyal  I have a few use cases which I'm currently using. Let's take the case of model invocation. Suppose if the result of model-1 is +ve then I may want to run some more steps, like maybe invoking some other models. But if the result of model-1 is -ve I want to move to the end.

hovhanns:
> @savingoyal I have a few use cases which I'm currently using. Let's take the case of model invocation. Suppose if the result of model-1 is +ve then I may want to run some more steps, like maybe invoking some other models. But if the result of model-1 is -ve I want to move to the end.

Hi @thesillystudent 
Do you have any information about merging this PR? I'm creating a Data pipeline where I need conditional branching, it will be good to have this feature in the nearest future.

thesillystudent:
@hovhanns I don’t have any plans as such. I can complete it over some weekends if enough people ask for it and I get a goahead from the devs

savingoyal:
@thesillystudent I am converting this PR to draft since there is no immediate need for this functionality. Please reach out to us at slack.outerbounds.co if you would like to work on this PR.

Issue #100, opened 2020-01-14
Batch job fail when user name has spaces


Dudemullet:
Apple lets user names have spaces but when a user with spaces tries to run a batch job with Metaflow aws does not let job names have spaces. The following error is what a user would see.

```
An error occurred (ClientException) when calling the SubmitJob operation: Error executing request, Exception : Job name should match valid pattern, RequestId: 1db9be95-2aba-4ff0-b19d-a61435a95791
```

savingoyal:
Thanks for the bug report. I will take a look at the PR.

rmeinlsainsburys:
You can fix it by changing this function in the file at "/usr/local/lib/python3.7/site-packages/metaflow/plugins/aws/batch"

`def _job_name(self, user, flow_name, run_id, step_name, task_id, retry_count):
        return '{user}-{flow_name}-{run_id}-{step_name}-{task_id}-{retry_count}'.format(
            **user=user.replace(" ", "").replace(".", ""),**
            flow_name=flow_name,
            run_id=run_id,
            step_name=step_name,
            task_id=task_id,
            retry_count=retry_count,
        )`

Dudemullet:
@rmeinlsainsburys We actually have a PR #101  that will notify of the errors more precisely. Also, I believe the preferred method would be to set your USER environment variable.

- You don't have to update your username at the system level (might be restricted)
- don't have to update source code

savingoyal:
One more data point - https://twitter.com/johancarlin/status/1412700513842470913?s=20

csetzkorn:
Curious, why is the above fix not yet implemented?

mmarqz:
Setting the `METAFLOW_USER` environment variable fixed it for me.

aadhil2k4:
Working on this issue!!

npow:
Closing as stale — this was filed in 2020 and may have been resolved in subsequent releases. If anyone is still hitting this, please open a new issue with reproduction steps and we'll take a look.

PR #101, opened 2020-01-14
Fixes user names possibly having spaces

Dudemullet:
This fixes issue #100 . The issue mainly occurs because Metaflow passes through the systems `user name` directly to aws included in the job name.

Aws does not allow spaces in the batch job name so it fails with an Exception.


savingoyal:
@Dudemullet Thanks for the PR. According to [AWS documentation](https://docs.aws.amazon.com/batch/latest/APIReference/API_SubmitJob.html#API_SubmitJob_RequestSyntax), Batch job names need to respect the following constraint ` first character must be alphanumeric, and up to 128 letters (uppercase and lowercase), numbers, hyphens, and underscores are allowed` - Can you modify your implementation to take care of this constraint.

Dudemullet:
@savingoyal Of course will look into it.

It makes sense to look into all of these considerations, not just the spacing issue.

savingoyal:
PR for #100

metaflow/plugins/aws/batch/batch.py
```diff
@@ -53,15 +54,26 @@ def _search_jobs(self, flow_name, run_id, user):
...
             run_id=run_id,
             step_name=step_name,
             task_id=task_id,
             retry_count=retry_count,
         )
 
+        # AWS Batch job name validation according to
+        # https://docs.aws.amazon.com/batch/latest/APIReference/API_SubmitJob.html#API_SubmitJob_RequestSyntax
+        if not job_name[0].isalnum():
+            raise BatchException('The Batch Jobs first character has to be alphanumeric')
```
savingoyal: Users of metaflow aren't aware of the notion of a `Batch Job Name`. Throwing this error message will most likely confuse the users since they wouldn't necessarily know how to fix it. 

One thought is to have a check in the [Batch decorator class](https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/aws/batch/batch_decorator.py#L111) disallowing usernames which don't begin with an alphanumeric character, have characters which are not dashes, underscores or alphanumeric (along with a helpful error message telling them how to set a new username). The job name that metaflow creates relies on user specified username, flowname and stepname - since flowname and stepname are deduced from the flow itself and are [valid lexical identifiers](https://docs.python.org/3/reference/lexical_analysis.html#identifiers), the check on username should be good enough. 
We can maintain the check on the length of the job name, with a helpful error message asking users to reduce the size of their flow name or the step name.

metaflow/plugins/aws/batch/batch.py
```diff
@@ -53,15 +54,26 @@ def _search_jobs(self, flow_name, run_id, user):
         return jobs
 
     def _job_name(self, user, flow_name, run_id, step_name, task_id, retry_count):
-        return '{user}-{flow_name}-{run_id}-{step_name}-{task_id}-{retry_count}'.format(
-            user=user,
+        job_name = '{user}-{flow_name}-{run_id}-{step_name}-{task_id}-{retry_count}'.format(
+            user=user.replace(" ", ""),
```
savingoyal: This will cause issues for users `a bc` and `ab c`. Instead of removing whitespace, we should disallow it as mentioned in the comment below.

savingoyal:
@Dudemullet Any updates on this PR?

Dudemullet:
Sorry for the delay, was a very busy last month and am currently on vacation (no laptop) . But will look back into this on March 16

Dudemullet:
@savingoyal  Left the `job_name` length validation in the `batch` class and all the `username` validation sent to the `batch_decorator`. 

As for the error messages I intentionally made them verbose but helpful but they're multiline so don't know if I got a little carried away with that.

metaflow/plugins/aws/batch/batch_decorator.py
```diff
@@ -109,6 +109,10 @@ def myStep(self):
     run_time_limit = None
 
     def __init__(self, attributes=None, statically_defined=False):
+        username = util.get_username()
```
savingoyal: If you move L112:115 to step_init() instead, then a much cleaner (prettified) exception gets thrown.

metaflow/plugins/aws/batch/batch_decorator.py
```diff
@@ -109,6 +109,10 @@ def myStep(self):
...
+        if not re.match('^[a-zA-Z0-9][a-zA-Z0-9_-]*$', username):
+            raise BatchException('The username trying to execute this batch job has one or more illegal characters. Make sure the first character in the username is alphanumeric and that it only contains dashes, underscores or alphanumeric characters. You can also set a specific username via the USER environment variable')
```
savingoyal: This seems pretty verbose for the end-user to parse easily. 
Maybe we can throw a `BatchInvalidUserException` (similar to [this](https://github.com/Netflix/metaflow/blob/c2f70c73782549fdce6c418410ecd9b305199b28/metaflow/exception.py#L84)) with the headline - 
`Invalid user name for launching jobs on AWS Batch.` 
and the formatted message 
`User name (overridden via METAFLOW_USER environment variable) should obey the following constraints - `

    <tabbed>1. It should only contain dashes, underscores or alphanumeric characters
    <tabbed>2. The first character should be alphanumeric

metaflow/plugins/aws/batch/batch.py
```diff
@@ -62,6 +62,11 @@ def _job_name(self, user, flow_name, run_id, step_name, task_id, retry_count):
             retry_count=retry_count,
         )
 
+        if len(job_name) > 128:
+            raise BatchException('Cant run a batch job with a name exceeding 128 characters. Try reducing your flow or step names length')
```
savingoyal: Maybe we can reword this error as - `Unable to launch AWS Batch job - job name exceeds 128 characters. Shorten your flow and/or step name by x characters.` Currently, the user has no idea how much to reduce the length by.

metaflow/plugins/aws/batch/batch_decorator.py
savingoyal: Or the message could be simply - `The user name (overridden via METAFLOW_USER environment variable) can only contain lowercase letters, numbers, underscores and dashes (and must start with a letter or number).`

metaflow/plugins/aws/batch/batch_decorator.py
```diff
@@ -122,6 +122,9 @@ def __init__(self, attributes=None, statically_defined=False):
                     self.attributes['image'])
 
     def step_init(self, flow, graph, step, decos, environment, datastore, logger):
+        username = util.get_username()
+        if not re.match('^[a-zA-Z0-9][a-zA-Z0-9_-]*$', username):
+            raise BatchException('The username trying to execute this batch job has one or more illegal characters. Make sure the first character in the username is alphanumeric and that it only contains dashes, underscores or alphanumeric characters. You can also set a specific username via the USER environment variable')
```
savingoyal: You would raise the `BatchInvalidUserException ` here.

metaflow/plugins/aws/batch/batch.py
```diff
@@ -62,6 +68,11 @@ def _job_name(self, user, flow_name, run_id, step_name, task_id, retry_count):
             retry_count=retry_count,
         )
 
+        if len(job_name) > 128:
+            raise BatchException('Unable to launch AWS Batch job - job name exceeds 128 characters. Shorten your flow and/or step name by x characters.')
```
savingoyal: We should replace x by `128 - len(job_name)`.

bhat-ganesh:
hello all, this seem simple enough fix, could we proceed with planning to merge it. I'd be happy to help

PR #109, opened 2020-02-05
Verify s3 puts to guard against corrupt uploads

savingoyal:
Simple size based check to preempt network/storage corruption of data objects


savingoyal:
Since amazon guarantees [no partial uploads](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html), this PR is moot.

Issue #108, opened 2020-01-30
ERROR: Encountered corrupt package tarball


abaspinar:
Hi, 

My step works flawlessly in local mode, but when I tried it with batch mode it failed with the message:
`/bin/sh: 1: metaflow_CleanFlow_linux-64_54816c55859cfd0f8c3c9b2e51678ce87bc33a38/bin/python: not found` 
To understand what might be the reason, I locally ran a docker image(python3:6) and ran all the commands that run on the batch side. 

I noticed when conda creating the environment, some of the packages inside `pkgs` folder were failing to install. I digged deeper into it and noticed that somehow a few tarballs were not fully copied(~80%) to s3 bucket in the beginning and therefore they were incomplete.
I manually downloaded those tarballs and they all started to work fine.
What might be the reason for these incomplete tarball uploads to s3 bucket?

Computer: `Mac OSX: 10.15.1`
Conda: `Anaconda 4.7.12`
Metaflow: `2.0.1`
List of the tarballs that have failed: `chardet-3.0.4-py36_1003.tar.bz2`, `six-1.14.0-py36_0.tar.bz2`, `setuptools-45.1.0-py36_0.tar.bz2`, `pip-20.0.2-py36_1.tar.bz2`

Example error message: ERROR: ```Encountered corrupt package tarball at /root/.aws/metaflow/conda/pkgs/setuptools-45.1.0-py36_0.tar.bz2. Conda has left it in place.  Please report this to the maintainers of your package.  For the defaults channel, please report to https://github.com/continuumio/anaconda-issues```

savingoyal:
Thanks for reporting the issue. We should add checksums on s3 uploads to guard against this.

abaspinar:
Hi @savingoyal,
I have added `payload_signing_enabled:True` from `botocore.config` to `boto3.client` that returns from `metaflow.metaflow_config.get_authenticated_boto3_client` thinking that it resolves this issue.

However, what I've noticed is that incomplete s3 uploads was not the real problem. Somehow, conda sometimes packages with `.tar.bz2` and sometimes with `.conda` suffixes. I checked the jsons for problematic packages under `conda-meta`.

I noticed that `fn` field ends with `.tar.bz2` , but `package_tarball_full_path` field ends with `.conda` for these cases. Metaflow uses `package_tarball_full_path` and upload them with the name `fn`. Therefore `.conda` format package is uploaded there as `.tar.bz2` package.

[...]

PR #118, opened 2020-02-10
Add check on conda-package metadata to guard against .conda packages

savingoyal:
Looks like conda doesn't reliably/correctly update the `env/conda-meta/<package>.json` file which Metaflow relies on to identify which packages should be moved to S3 and their installation order on AWS Batch.

In this specific scenario, when the compute needs to execute on AWS Batch, Metaflow coaxes conda to generate an environment for `linux-64` architectures using only `.tar.bz2` dependencies, irrespective of where conda is executed from (usually a user's macOS or linux instance). In some very specific scenarios, conda correctly picks up the `.tar.bz2` dependencies from upstream repos but fails to update the manifest, incorrectly pointing to `.conda` dependencies instead. This leads to incorrect bundling of packages and the user job fails when executing on AWS Batch since we end up packaging a .conda dependency as a .tar.bz2 dependency.

### DIFF
--- a/metaflow/plugins/conda/conda_step_decorator.py
+++ b/metaflow/plugins/conda/conda_step_decorator.py
@@ -145,10 +145,15 @@ class CondaStepDecorator(StepDecorator):
                                 url.path.lstrip('/'),
                                 package_info['md5'],
                                 package_info['fn'])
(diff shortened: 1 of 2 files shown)


savingoyal:
Fixes #108

PR #128, opened 2020-02-14
Add hook for metaflow R bindings

savingoyal:
Adds necessary hooks for customizing process entry-points to
R executables. This patch preps for the eventual metaflow-R
release.

Addresses #1

### DIFF
--- a/metaflow/package.py
+++ b/metaflow/package.py
@@ -6,6 +6,7 @@ from hashlib import sha1
 from itertools import chain
 
 from .util import to_unicode
+from . import R
 
 try:
     # python2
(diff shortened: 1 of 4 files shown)


metaflow/package.py
```diff
@@ -16,7 +17,7 @@
     import io
     BytesIO = io.BytesIO
 
-DEFAULT_SUFFIXES = ['.py']
+DEFAULT_SUFFIXES = ['.py', '.R']
```
jasonge27, line 20: Do we want to include `.Rmd` by default as well? The flow can be launched from R markdown files.

savingoyal: Good idea! I will add that in.

savingoyal:
Closing this since #214 is now part of master

PR #134, opened 2020-02-18
Initialize datastore_root from cli args in cli.step(...)

savingoyal:
Gitter thread - https://gitter.im/metaflow_org/community?at=5e4b8e24d56ddb68a4aff877

Metaflow initializes datastore_root to fetch user code in `step` command. The behaviour defaults to relying on METAFLOW_DATASTORE_SYSROOT_S3 environment variable to initialize datastore_root. This patch checks if --datastore_root cli arg can be used instead.

### DIFF
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -401,6 +401,7 @@ def step(obj,
          fg='magenta',
          bold=False)
 
+    obj.datastore.datastore_root = obj.datastore_root
     if obj.datastore.datastore_root is None:
         obj.datastore.datastore_root = obj.datastore.get_datastore_root_from_config(obj.echo)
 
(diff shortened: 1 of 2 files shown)


metaflow/cli.py
```diff
@@ -765,6 +766,8 @@ def start(ctx,
                                                   ctx.obj.event_logger,
                                                   ctx.obj.monitor)
     ctx.obj.datastore = DATASTORES[datastore]
+    if datastore_root is None:
+        datastore_root = ctx.obj.datastore.get_datastore_root_from_config(echo)
```
seeravikiran: Discussed offline - I think each of the terminal commands like `step`, `dump` etc. probably handle if `datastore_root is None` so we should be able to safely remove this `if` check.

savingoyal: Made the change.

PR #148, opened 2020-03-03
Feature : Global environment Flow decorator.

valayDave:
A Flow decorator designed to help capture a default global environment. 

It will set default environment variables in every step of the flow. The `@environment` decorator will only add new var / override vars set by `@global_environment`. 

Example code snippet
```
from metaflow import FlowSpec, step, global_environment, environment

@global_environment(vars={'CUSTOM_SET_USER_NAME':"New_User"})
class DummyFlow(FlowSpec):
    
    @step
    def start(self):
        import os
        print("Hello ",os.environ.get('CUSTOM_SET_USER_NAME'))
        self.next(self.end)

    @environment(vars={'CUSTOM_SET_PASSWORD':'random_password'})
    @step
    def end(self):
        import os
        print("Hello ",os.environ.get('CUSTOM_SET_USER_NAME'),os.environ.get('CUSTOM_SET_PASSWORD'))
        print('Done!')

if __name__ == '__main__':
    DummyFlow()
```

### DIFF
--- a/metaflow/plugins/__init__.py
+++ b/metaflow/plugins/__init__.py
@@ -38,7 +38,8 @@ ENVIRONMENTS = [CondaEnvironment]
 # careful with the choice of name though - they become top-level
 # imports from the metaflow package.
 from .conda.conda_flow_decorator import CondaFlowDecorator
-FLOW_DECORATORS = [CondaFlowDecorator]
+from .environment_decorator import GlobalEnvironmentDecorator
+FLOW_DECORATORS = [CondaFlowDecorator,GlobalEnvironmentDecorator]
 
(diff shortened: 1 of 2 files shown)


tuulos:
@valayDave if you want to set an env var for all tasks, you can do it at the top of the module with 
```
import os
os.environ['CUSTOM_SET_USER_NAME'] = "New_User"
```
this will get evaluated before the task function is executed. Would this work for your use case?

valayDave:
> this will get evaluated before the task function is executed. Would this work for your use case?

Yes. Then I can close this PR. Coz that can work anyways.

PR #151, opened 2020-03-04
Extending support for other object datastores.

valayDave:
Added a `@staticmethod` to the `MetaflowDataStore` class pertaining to commands needed for code package download. 

The method needs to be implemented by each child class of a datastore so that commands related to package download can be different according to different child classes. 

This ensures the decoupling of classes relating to the environment and computational resources from the datastore itself.

### DIFF
--- a/metaflow/datastore/datastore.py
+++ b/metaflow/datastore/datastore.py
@@ -607,10 +607,10 @@ class MetaflowDataStore(object):
         return '\n'.join(line for k, line in sorted(lines()))
 
     @staticmethod
-    def package_download_commands(environment,code_package):
+    def package_download_commands(environment, code_package):
         """
         Intended to download the package inside a container for the execution of a step. 
-        Static because it is a utility method WRT to the datastore child class defined for different object store. 
+        Static because it is a utility method WRT the datastore child class defined for different object store. 
         Will be called by the MetaflowEnvironment.get_package_commands
         """
         raise NotImplementedError()
--- a/metaflow/datastore/datastore.py
+++ b/metaflow/datastore/datastore.py
@@ -606,6 +606,15 @@ class MetaflowDataStore(object):
                          .format(key=k, value=v, **self.info[k])
         return '\n'.join(line for k, line in sorted(lines()))
 
+    @staticmethod
+    def package_download_commands(environment,code_package):
+        """
+        Intended to download the package inside a container for the execution of a step. 
+        Static because it is a utility method WRT to the datastore child class defined for different object store. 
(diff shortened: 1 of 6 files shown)


metaflow/datastore/datastore.py
```diff
@@ -606,6 +606,15 @@ def lines():
                          .format(key=k, value=v, **self.info[k])
         return '\n'.join(line for k, line in sorted(lines()))
 
+    @staticmethod
+    def package_download_commands(environment,code_package):
+        """
+        Intended to download the package inside a container for the execution of a step. 
+        Static because it is a utility method WRT to the datastore child class defined for different object store.
```
romain-intel: WRT to -> WRT

metaflow/datastore/local.py
```diff
@@ -241,3 +241,6 @@ def is_done(self):
         filename = self.get_done_filename_for_attempt(self.attempt)
         path = os.path.join(self.root, filename)
         return os.path.exists(path)
+
+    def package_download_commands(environment,code_package):
+        return []
```
romain-intel: Add newline

metaflow/datastore/datastore.py
```diff
@@ -606,6 +606,15 @@ def lines():
                          .format(key=k, value=v, **self.info[k])
         return '\n'.join(line for k, line in sorted(lines()))
 
+    @staticmethod
+    def package_download_commands(environment,code_package):
```
romain-intel: Nit: space after comma -- here and in other places

metaflow/environment.py
```diff
@@ -76,21 +76,15 @@ def get_client_info(cls, flow_name, metadata):
         """
         return "Local environment"
 
-    def get_package_commands(self, code_package_url):
+    def get_package_commands(self, code_package_url,datastore):
         cmds = ["set -e",
                 "echo \'Setting up task environment.\'",
                 "%s -m pip install awscli click requests boto3 \
```
romain-intel: It may make sense to move the awscli and boto3 requirements into the package_download_commands as well since they are not related to the rest here (requests and click are MF requirements independent of AWS/S3)

valayDave: Yes. That makes sense. Do you think abstractions should also apply the Decorators like `BatchDecorator`?

savingoyal:
@valayDave We would need to rebase this PR so that we can support Azure/GCP blob stores.

romain-intel:
Is this something still needed with the neew datastore?

savingoyal:
@romain-intel Not necessarily. We can resurrect the PR as we get around to supporting Azure Blob Store and GCS.

PR #160, opened 2020-03-13
Set thresholds for retrying `describeJobs` in AWS Batch integration

savingoyal:
`describeJobs` API for AWS Batch has particularly low concurrency limits. This PR adds throttling and jitter to the calls we make to aid in higher resiliency.

### DIFF
--- a/metaflow/plugins/aws/batch/batch_client.py
+++ b/metaflow/plugins/aws/batch/batch_client.py
@@ -1,4 +1,5 @@
 from collections import defaultdict, deque
+import random
 import select
 import sys
 import time
@@ -155,19 +156,36 @@ class BatchJob(object):
         return self
 
 
-class limit(object):
-    def __init__(self, delta_in_secs):
+class Throttle(object):
+    def __init__(self, delta_in_secs=1, num_tries=10):
         self.delta_in_secs = delta_in_secs
+        self.num_tries = num_tries
         self._now = None
+        self._reset()
+
+    def _reset(self):
+        self._tries_left = self.num_tries
+        self._wait = self.delta_in_secs
 
     def __call__(self, func):
         def wrapped(*args, **kwargs):
             now = time.time()
-            if self._now is None or (now - self._now > self.delta_in_secs):
-                func(*args, **kwargs)
+            if self._now is None or (now - self._now > self._wait):
                 self._now = now
+                try:
+                    func(*args, **kwargs)
+                    self._reset()
+                except TriableException as ex:
+                    self._tries_left -= 1
+                    if self._tries_left == 0:
+                        raise ex.ex from None
+                    self._wait = (self.delta_in_secs*1.1)**(self.num_tries-self._tries_left) + \
+                        random.randint(0, 3*self.delta_in_secs)
         return wrapped
 
+class TriableException(Exception):
+    def __init__(self, ex):
+        self.ex = ex
 
 class RunningJob(object):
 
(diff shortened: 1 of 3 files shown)


metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -184,16 +202,21 @@ def __repr__(self):
...
 
-    @limit(1)
+    @Throttle()
     def _update(self):
         try:
             data = self._client.describe_jobs(jobs=[self._id])
-        except self._client.exceptions.ClientException:
-            return
+        except self._client.exceptions.ClientError as err:
+            code = err.response['ResponseMetadata']['HTTPStatusCode']
```
romain-intel: Are we sure that all err have this data. Otherwise maybe do something like try: code = err.response[][] except: raise err

metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -155,19 +156,36 @@ def parameter(self, key, value):
...
+            if self._now is None or (now - self._now > self._wait):
                 self._now = now
+                try:
+                    func(*args, **kwargs)
+                    self._reset()
+                except TriableException as ex:
+                    self._tries_left -= 1
+                    if self._tries_left == 0:
+                        raise ex.ex
+                    self._wait = (self.delta_in_secs*1.1)**(self.num_tries-self._tries_left) + \
```
romain-intel: It is a little weird that the jitter portion (from 0 to 3 by default) is larger than the largest exponential wait (1.1^9 = ~2.36)

metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -184,16 +202,21 @@ def __repr__(self):
...
-            return
+        except self._client.exceptions.ClientError as err:
+            code = err.response['ResponseMetadata']['HTTPStatusCode']
+            if code == 429 or code >= 500:
+                raise TriableException(err)
+            raise err
         self._apply(data['jobs'][0])
 
     def update(self):
         self._update()
```
romain-intel: The intention here is to be aggressive when we have no data and to silently ignore update errors later on, correct?

metaflow/plugins/aws/batch/batch_client.py
savingoyal: Changed the multiplier to 1.2.

metaflow/plugins/aws/batch/batch_client.py
savingoyal: ClientError is guaranteed to have this data according to their docs.

metaflow/plugins/aws/batch/batch_client.py
savingoyal: Exactly. We want to block on the first successful data fetch and afterwards ignore the errors till limits are breached.

savingoyal:
@romain-intel Can you bless this PR?

peterwilton-ttam:
FWIW, I'm still hitting this error when attempting to run ~300 Batch jobs concurrently.

wf8:
I'm also still running into this error with 200 concurrent Batch jobs.

corriebar:
I just ran into this error with only 1 concurrent Batch job. Ironically, it happened when trying to start the end task which is empty for my flow... I'm using metaflow 2.2.7

savingoyal:
@seanv507 I am not familiar if boto3 now supports event subscriptions. That will be the most straightforward way to address this. We can raise the number of tries. There is also an effort to integrate with AWS Batch array which will help with this issue as well.

PR #186, opened 2020-04-21
Expose `create_job` and `register_job_definition` methods for SFN integration

savingoyal:


metaflow/plugins/aws/batch/batch.py
```diff
@@ -161,8 +154,49 @@ def launch_job(
...
+        attrs={}
+    ):
+        if queue is None:
+            queue = next(self._client.active_job_queues(), None)
+            if queue is None:
+                raise BatchException(
+                    'Unable to launch Batch job. No job queue '
+                    ' specified and no valid & enabled queue found.'
+                )
+        job = self.create_job(
```
seeravikiran, line 183: Interesting - what is the difference between the types returned by create_job and job.execute()?

savingoyal: `job.execute` returns a reference to the running job on AWS Batch. `create_job` only creates the spec for the job to be executed.

PR #188, opened 2020-04-22
Expose Retry Count in Current Singleton

savingoyal:
current.retry_count now returns the retry attempt (0, 1, .. so on)
for a metaflow task

### DIFF
--- a/metaflow/current.py
+++ b/metaflow/current.py
@@ -6,28 +6,42 @@ class Current(object):
         self._run_id = None
         self._step_name = None
         self._task_id = None
+        self._retry_count = None
         self._origin_run_id = None
         self._namespace = None
         self._username = None
         self._is_running = False
 
     def _set_env(self,
-                 flow_name,
-                 run_id,
-                 step_name,
-                 task_id,
-                 origin_run_id,
-                 namespace,
-                 username):
+                 flow_name=None,
+                 run_id=None,
+                 step_name=None,
+                 task_id=None,
+                 retry_count=None,
+                 origin_run_id=None,
+                 namespace=None,
+                 username=None,
+                 is_running=True):
 
         self._flow_name = flow_name
         self._run_id = run_id
         self._step_name = step_name
         self._task_id = task_id
+        self._retry_count = retry_count
         self._origin_run_id = origin_run_id
         self._namespace = namespace
         self._username = username
-        self._is_running = True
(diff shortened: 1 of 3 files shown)


metaflow/current.py
```diff
@@ -6,28 +6,42 @@ def __init__(self):
...
+        self._retry_count = retry_count
         self._origin_run_id = origin_run_id
         self._namespace = namespace
         self._username = username
-        self._is_running = True
+        self._is_running = is_running
+
+    def _update_env(self, env):
+        for k, v in env.items():
+            setattr(self.__class__, k, property(fget=lambda _, v=v: v))
```
seeravikiran, line 38: How is that different from L:46-48? Is that just code duplication or am I missing something?

metaflow/current.py
```diff
@@ -6,28 +6,42 @@ def __init__(self):
...
+
+    def __contains__(self, key):
+        return getattr(self, key, None) is not None
+
+    def get(self, key, default=None):
+        return getattr(self, key, default)
```
seeravikiran, line 44: Are you planning to use this for some other purpose? It seems unused at least in this PR.
Do you mind also sharing how you plan to use this since I'm wondering why you don't access the attr only if it is non-None.

metaflow/current.py
```diff
@@ -6,28 +6,42 @@ def __init__(self):
...
-                 username):
+                 flow_name=None,
+                 run_id=None,
+                 step_name=None,
+                 task_id=None,
+                 retry_count=None,
+                 origin_run_id=None,
+                 namespace=None,
+                 username=None,
+                 is_running=True):
```
seeravikiran, line 24: I wonder if its worth having the default to False? Any rationale why you chose True instead?
For now - it seems in both places you're being explicit. Another option might be to leave out any default or set to None. Thoughts?

savingoyal: It's set to `True` for parity with nflx-metaflow

metaflow/current.py
savingoyal: Some weirdness in github maybe.

metaflow/current.py
savingoyal: It's used within `@project` which will be a subsequent PR.

metaflow/current.py
seeravikiran: OK.

PR #190, opened 2020-04-23
Add a IncludeMultipleFiles parameter to allow for file Globbing

### LINKED ISSUE #96: multiple files load in IncludeFile
I would like to know is there a way to include multi-files in same directory? As currently it only supports single file. Thanks

romain-intel:
This allows the specification of a parameter that will include
multiple files based on the path glob specified.

Fixes #96

### DIFF
--- a/metaflow/includefile.py
+++ b/metaflow/includefile.py
@@ -1,12 +1,17 @@
 import io
+import glob
 import os
+import random
+import re
+import string
+
 import click
 
 from metaflow.exception import MetaflowException
 from metaflow.parameters import Parameter
 
 
-class InternalFile():
+class InternalFile(object):
     def __init__(self, logger, is_text, encoding, path):
         self._logger = logger
         self._is_text = is_text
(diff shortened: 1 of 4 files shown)


metaflow/includefile.py
```diff
@@ -44,6 +49,35 @@ def size(self):
...
+    def __init__(self, base_name, logger, is_text, encoding):
+        self._base_name = base_name
+        self._logger = logger
+        self._is_text = is_text
+        self._encoding = encoding
+        self._files = {}
+
+    def add_file(self, path):
+        name = os.path.basename(path)
+        # Sanitize the name to make it possible to use as a variable name
```
tuulos, line 62: @romain-intel what's the idea behind this logic? Why would you want to mangle the file name?

romain-intel: To give a simple name and an unambiguous one. I could change the logic to remove the random bit though (and just through an error if I can't create a simple name). The idea of prefetching the name of the parameter is to allow, for example, to import different similarly structured data sets. It also reduces the change of colliding with reserved names.

seeravikiran: I still don't quite follow. Could you give an example of a `path` and the various transformation that `name` goes through?

seeravikiran:
@romain-intel Could you also add some examples to PR description with intended usage? Specifically I'm looking for:
* Basic usage of IncludeMultipleFiles using (*) to solve the attached github issue.
* Use of recursive during glob match.

PR #191, opened 2020-04-23
Allow files residing on s3 to be included using IncludeFile

### LINKED ISSUE #156: IncludeFile could support s3:// files
`IncludeFile` could support files in S3. It is purely a convenience feature over copying an S3 asset locally using `aws s3 cp` (which may need a separate config) and then using `IncludeFile`.

Context: https://gitter.im/metaflow_org/community?at=5e664fd5145f4d69562d7e24

romain-intel:
This will cause the file to be brought from S3 and then repackaged as a
Flow artifact.

Fixes #156

### DIFF
--- a/metaflow/includefile.py
+++ b/metaflow/includefile.py
@@ -4,6 +4,9 @@ import os
 import random
 import re
 import string
+import tempfile
+
+from shutil import move
 
 import click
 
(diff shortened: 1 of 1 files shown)


metaflow/includefile.py
```diff
@@ -17,11 +20,41 @@ def __init__(self, logger, is_text, encoding, path):
...
+            return True, None
+        try:
+            with open(path, mode='r') as _:
+                pass
+        except OSError:
+            return False, "Could not open file '%s'" % path
+        else:
+            return True, None
+
+    def handle_s3_file(self):
```
seeravikiran, line 37: Can these be private methods? So _handle_s3_file since I don't think you need them accessible outside.

metaflow/includefile.py
```diff
@@ -17,11 +20,41 @@ def __init__(self, logger, is_text, encoding, path):
         self._is_text = is_text
         self._encoding = encoding
         self._path = path
-        self._size = os.path.getsize(self._path)
+        self._size = 0
 
-    def __call__(self):
+    @classmethod
+    def is_file_handled(cls, path):
```
seeravikiran, line 26: I find the return tuple slightly uncommon for UX. By the function name I typically expect it to return a boolean. WDYT?

If you agree, one alternative would be to accept a error_msg as an inout arg and you can set it if a valid object is passed in.
I don't know if its easy to differentiate between the following cases:
```
Assume, is_file_handled(cls, path, error_msg=None).

Case 1:
is_file_handled(path)

Case 2:
error_msg = None
is_file_handled(path, error_msg)

In C++ land a valid lval is specified in the second case.
```

metaflow/includefile.py
```diff
@@ -17,11 +20,41 @@ def __init__(self, logger, is_text, encoding, path):
...
+            return True, None
+
+    def handle_s3_file(self):
+        # We bring the file back in locally so we can then properly
+        # include it like a local file; we currently do not support
+        # S3 -> S3 directly.
+        from metaflow import S3
+
+        to_return = tempfile.NamedTemporaryFile(dir='/tmp', delete=False)
+        to_return.close()
```
seeravikiran, line 44: For my understanding - why is this close() needed right after creation?

metaflow/includefile.py
```diff
@@ -33,15 +66,25 @@ def __call__(self):
...
             # it means that read failed which happens with Python 2.7 for large files
             raise MetaflowException('Cannot read file at %s -- this is likely because it is too '
                                     'large to be properly handled by Python 2.7' % self._path)
 
+    def __call__(self):
+        local_file = None
+        if self._path.startswith("s3://"):
+            local_file = self.handle_s3_file()
+        try:
+            return self.handle_local_file(local_file)
```
seeravikiran, line 83: To improve readability slightly I think you could use kwarg here and say `override_path=local_file`; so its clearer that we want to override only for S3 case.
Otherwise the function uses `self._path` for knowing the local file path; which was non-obvious to me until I read L:54-55.

romain-intel:
This PR was superseded by #220 which includes support for S3 external files.

PR #193, opened 2020-04-28
Mute `ThrottleException` for AWS Logs API for AWS Batch

### LINKED ISSUE #184: Mask ThrottleException for GetLogEvents in @batch
More context - https://gitter.im/metaflow_org/community?at=5e9f5a5a71a34b0149fe40f1

savingoyal:
Currently, the flow logs are polluted with  occasional `ThrottleExceptions`.
They don't result in flow failures since we retry log fetching till the tasks
are alive and do a best-effort pull after the tasks have finished. This patch
swallows the `ThrottleExceptions` so that the end-user isn't thrown off track.

resolves #184

### DIFF
--- a/metaflow/plugins/aws/batch/batch_client.py
+++ b/metaflow/plugins/aws/batch/batch_client.py
@@ -166,7 +166,7 @@ class BatchJob(object):
 
 
 class Throttle(object):
-    def __init__(self, delta_in_secs=1, num_tries=200):
+    def __init__(self, delta_in_secs=1, num_tries=20):
         self.delta_in_secs = delta_in_secs
         self.num_tries = num_tries
         self._now = None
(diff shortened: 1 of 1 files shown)


metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -166,7 +166,7 @@ def attempts(self, attempts):
 
 
 class Throttle(object):
-    def __init__(self, delta_in_secs=1, num_tries=200):
```
seeravikiran: I don't see this diff when I review the entire change for some reason.

seeravikiran: Never mind - it was rolled back to 200 in second commit. Makes sense.

Issue #104, opened 2020-01-17
S3 Errors when using s3.get_all() not handled properly and cause crash


rickyshelton:
Metaflow v2.0.1
Python 3.7

When trying to load files from S3 using `s3.get_all()` (and potentially `get_recursive()` looking at the code), if you encounter an S3 error (either due to permissions or some other error) it is not handled properly, and the execution crashes without useful output.

When trying to raise a `MetaflowS3Exception` in `_read_many_files` it's trying to reference `prefixes[0]`. 
https://github.com/Netflix/metaflow/blob/master/metaflow/datatools/s3.py#L577

When called called from `get_recursive`, prefixes is a map: `map(self._url, keys)`
https://github.com/Netflix/metaflow/blob/master/metaflow/datatools/s3.py#L418

This causes 
`TypeError: 'map' object is not subscriptable` instead of the original Exception being raised and handled as usual.

This is probably a Python2 vs Python3 issue.

Logs

```
Task is starting (status SUBMITTED)...
Task is starting (status RUNNABLE)...
Task is starting (status STARTING)...
Task is starting (status RUNNING)...
Setting up task environment.
Downloading code package.
Code package downloaded.
Bootstrapping environment.
Environment bootstrapped.
Task is starting.
<flow MetaflowExampleProject step start> failed:
    Internal error
Traceback (most recent call last):
  File "/metaflow/metaflow/cli.py", line 853, in main
    start(auto_envvar_prefix='METAFLOW', obj=state)
  File "/metaflow/metaflow_MetaflowExampleProject_linux-64_cffadb7002d862de02233a30fad8e993ff54043e/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(args, kwargs)
  File "/metaflow/metaflow_MetaflowExampleProject_linux-64_cffadb7002d862de02233a30fad8e993ff54043e/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/metaflow/metaflow_MetaflowExampleProject_linux-64_cffadb7002d862de02233a30fad8e993ff54043e/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/metaflow/metaflow_MetaflowExampleProject_linux-64_cffadb7002d862de02233a30fad8e993ff54043e/lib/python3.7/site-packages/click/core.py", line 956, in invoke
[...]

PR #195, opened 2020-04-29
Fix logging of prefixes in datatools.S3._read_many_files

oliverholworthy:
When `datatools.S3._read_many_files` is unsuccessful, instead of displaying the underlying error message, we get a `TypeError: 'map' object is not subscriptable` due to trying to access the first element of `prefixes` with `[0]`. The `prefixes` passed in is a generator (map) which doesn't support this sort of access. The simplest fix is ~to remove the attempted logging of the _first prefix_~ to assign to a list inside the function. 

Better to report the underlying error correctly than a mysterious `'map' object is not subscriptable`. In my case I had an issue with a version of the [typing](https://pypi.org/project/typing/) library being incompatible with a version of [`certifi`](https://pypi.org/project/certifi/) which became obvious once the underlying error message was not being hidden.

~If it's important to log the first prefix requested, we can instead change all callers of `_read_many_files` to pass a list instead of generator.~

## Related Issues

- Should Fix https://github.com/Netflix/metaflow/issues/104

### DIFF
--- a/metaflow/datatools/s3.py
+++ b/metaflow/datatools/s3.py
@@ -564,6 +564,7 @@ class S3(object):
     # Conversion between bytes and unicode is done through url_quote
     # and url_unquote.
     def _read_many_files(self, op, prefixes, **options):
+        prefixes = list(prefixes)
         with NamedTemporaryFile(dir=self._tmpdir,
                                 mode='wb',
                                 delete=not debug.s3client,


savingoyal:
Thanks! Taking a look at the PR.

metaflow/datatools/s3.py
```diff
@@ -575,8 +575,7 @@ def _read_many_files(self, op, prefixes, **options):
                                                      **options)
             if stderr:
                 raise MetaflowS3Exception("Getting S3 files failed.\n"\
-                                          "First prefix requested: %s\n"\
-                                          "Error: %s" % (prefixes[0], stderr))
+                                          "Error: %s" % (stderr,))
```
savingoyal, line 578: You can instead do `prefixes = list(prefixes)` after L:566

oliverholworthy: Updated. 😀 

I didn't spot the same pattern in the method below (`_put_many_files`). I initially thought the use of generators might have been some kind of performance/memory optimisation.

PR #197, opened 2020-04-30
Increase retry count for AWS Batch logs streaming

savingoyal:
Modify the retry behavior for log fetching on AWS Batch by adding
jitters to exponential backoffs as well as reset the retry counter
for every successful request.

Additionally, fail the metaflow task even if AWS Batch task succeeds
but we fail to stream the task logs back to the user's terminal.

### DIFF
--- a/metaflow/plugins/aws/batch/batch_client.py
+++ b/metaflow/plugins/aws/batch/batch_client.py
@@ -198,7 +198,7 @@ class TriableException(Exception):
 
 class RunningJob(object):
 
-    NUM_RETRIES = 5
+    NUM_RETRIES = 8
 
     def __init__(self, id, client):
         self._id = id
@@ -317,7 +317,8 @@ class RunningJob(object):
             elif not self.is_done:
                 self.wait_for_running()
 
-        for i in range(self.NUM_RETRIES):
+        exception = None
+        for i in range(self.NUM_RETRIES + 1):
             try:
                 check_after_done = 0
                 for line in log_stream:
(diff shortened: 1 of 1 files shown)


metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -329,12 +330,17 @@ def get_log_stream(job):
...
+                        i = 0
                         yield line
+                return
             except Exception as ex:
+                exception = ex
                 if self.is_crashed:
                     break
-                sys.stderr.write(repr(ex))
-                time.sleep(2 ** i)
+                #sys.stderr.write(repr(ex) + '\n')
```
savingoyal, line 340: This is intentionally left out in the code so that we can easily debug issues while in dev mode. The raised exception in L:343 will convey the last generated exception to the user.

seeravikiran, line 340: Why is this commented out?

metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -329,12 +330,17 @@ def get_log_stream(job):
...
+                if i < self.NUM_RETRIES:
```
seeravikiran, line 341: Why is there no sleep on the last retry? Overall it was confusing for me to understand why you don't chose the constant to 9; instead of +1 on L:321.

metaflow/plugins/aws/batch/batch_client.py
savingoyal: https://github.com/Netflix/metaflow/pull/197#discussion_r417683154

metaflow/plugins/aws/batch/batch_client.py
savingoyal: `NUM_ATTEMPTS = NUM_RETRIES + 1`. We would like the loop to execute `NUM_ATTEMPTS` times and we would immediately like to raise the exception after the last attempt - sleeping at that point is not helpful in any way.

seeravikiran: Let me know if you think this might help readability.
```
NUM_ATTEMPTS=9

for attempt in range(0, NUM_ATTEMPTS):
     try:
     // log processing (with reset on attempt)
     except:
     if attempt == NUM_ATTEMPTS - 1: # last attempt
        raise(..)
     else: # exponential back-off with jitter
       sleep(..)
```

PR #204, opened 2020-05-20
AWS Step Functions Integration

savingoyal:
Integrates Metaflow with [AWS Step Functions](https://aws.amazon.com/step-functions/).

Introduces a new command `step-functions`:
- `python my_flow.py step-functions create` will compile and export the user-defined Metaflow flow into an AWS Step Functions state machine. This will allow users of Metaflow to move their flows into production seamlessly with AWS. 
    - An additional flow level decorator, `@schedule`, allows users to optionally schedule the execution of their flows by integrating with AWS Event Bridge. 
    - All current functionality of Metaflow - containerized job execution on top of AWS Batch through `@batch`, dependency management via `@conda`, retrying mechanisms through `@retry`, `@catch` and `@timeout`, parameters, branches, and for-eaches are now available within AWS Step Functions through this integration. 
    - Additionally, introduces a notion of `production token` to ensure flow deployments to AWS Step Functions have proper safeguards against unintended production deployments
- `python my_flow.py step-functions trigger` will trigger a deployed workflow on AWS Step Functions

Requirements:
- `METAFLOW_SFN_IAM_ROLE `: IAM role that allows AWS Step Functions to interact with AWS Batch and AWS Dynamo DB
- `METAFLOW_EVENTS_SFN_ACCESS_IAM_ROLE `: IAM role that allows AWS Event Bridge to send notifications to AWS Step Functions
- `METAFLOW_SFN_DYNAMO_DB_TABLE `: AWS Dynamo DB Table which allows AWS Step Functions to keep track of the necessary state for foreaches

Notes:
- A cloudformation template provisioning all of the above should be available soon.
- Metaflow service currently doesn't provide for string identifiers for runs and tasks. This PR currently bonks when used when `DEFAULT_METADATA` is set to `service`. A subsequent update to the service will fix that issue. For testing, please use - `DEFAULT_METADATA=local`.

Instructions and detailed docs are on their way.

Links:
- [User facing documentation](https://docs.google.com/document/d/1OT5Pk6iyCoH-QmkNfwvrEnaPiEMxkHjlqiXZ-79VvRc/edit#)

### DIFF
--- a/test/core/tests/large_artifact.py
+++ b/test/core/tests/large_artifact.py
@@ -2,7 +2,7 @@ from metaflow_test import MetaflowTest, ExpectationFailed, steps
 
 class LargeArtifactTest(MetaflowTest):
     """
-    Test that you can serialize large objects (over 4GB)
+    Test that you can serialize large objects (over 2GB)
     with Python3.
     """
     PRIORITY = 2
@@ -11,7 +11,7 @@ class LargeArtifactTest(MetaflowTest):
     def step_single(self):
         import sys
         if sys.version_info[0] > 2:
-            self.large = b'x' * int(4.1 * 1024**3)
+            self.large = b'x' * int(2.1 * 1024**3)
             self.noop = False
         else:
             self.noop = True
@@ -20,7 +20,7 @@ class LargeArtifactTest(MetaflowTest):
     def step_end(self):
         import sys
         if sys.version_info[0] > 2:
-            assert_equals(self.large, b'x' * int(4.1 * 1024**3))
+            assert_equals(self.large, b'x' * int(2.1 * 1024**3))
 
     @steps(1, ['all'])
     def step_all(self):
@@ -32,4 +32,4 @@ class LargeArtifactTest(MetaflowTest):
         if not noop and sys.version_info[0] > 2:
             checker.assert_artifact('end',
                                     'large',
-                                    b'x' * int(4.1 * 1024**3))
+                                    b'x' * int(2.1 * 1024**3))


metaflow/metadata/service.py
```diff
@@ -41,7 +41,13 @@ def new_run_id(self, tags=[], sys_tags=[]):
         return self._new_run(tags=tags, sys_tags=sys_tags)
 
     def register_run_id(self, run_id, tags=[], sys_tags=[]):
-        pass
+        try:
```
romain-intel, line 44: The comment and this code is confusing. I believe the idea is that if the run id is an integer, it was obtained from the service and doesn't need to be created whereas if the integer is a string, it was generated outside the service and needs to be registered. I would put a comment here and below around this. We should also ensure that IDs generated outside of the service are NEVER integers. Also, s/MLI/Metadata/

metaflow/metadata/service.py
```diff
@@ -102,19 +120,21 @@ def _get_object_internal(cls, obj_type, obj_order, sub_type, sub_order, filters=
                 return None
             raise
 
-    def _new_run(self, tags=[], sys_tags=[]):
+    def _new_run(self, run_id=None, tags=[], sys_tags=[]):
         # first ensure that the flow exists
         self._get_or_create('flow')
-        run = self._get_or_create('run', tags=tags, sys_tags=sys_tags)
+        run = self._get_or_create('run', run_id, tags=tags, sys_tags=sys_tags)
```
romain-intel, line 126: We should improve the _get_or_create logic. Right now it does a query, waits for a 404 and creates it (which is already somewhat ugly).

metaflow/metadata/service.py
```diff
@@ -102,19 +120,21 @@ def _get_object_internal(cls, obj_type, obj_order, sub_type, sub_order, filters=
...
     def _new_task(self,
                   run_id,
                   step_name,
+                  task_id,
                   tags=[],
                   sys_tags=[]):
+        # first ensure that the step exists
         self._get_or_create('step', run_id, step_name)
-        task = self._get_or_create('task', run_id, step_name, tags=tags, sys_tags=sys_tags)
+        task = self._get_or_create('task', run_id, step_name, task_id, tags=tags, sys_tags=sys_tags)
```
romain-intel, line 137: Same as above.

metaflow/plugins/__init__.py
```diff
@@ -38,7 +42,8 @@ def get_plugin_cli():
 # careful with the choice of name though - they become top-level
 # imports from the metaflow package.
 from .conda.conda_flow_decorator import CondaFlowDecorator
-FLOW_DECORATORS = [CondaFlowDecorator]
+from .schedule_decorator import ScheduleDecorator
```
romain-intel, line 45: Should this be in the AWS directory or is it generic?

metaflow/plugins/aws/batch/batch.py
```diff
@@ -41,11 +41,13 @@ def _command(self, code_package_url, environment, step_name, step_cli):
 
     def _search_jobs(self, flow_name, run_id, user):
         if user is None:
-            regex = '-{flow_name}-{run_id}-'.format(flow_name=flow_name, run_id=run_id)
+            regex = '-{flow_name}-'.format(flow_name=flow_name, run_id=run_id)
```
romain-intel, line 44: run_id does not seem needed anymore.

metaflow/plugins/aws/batch/batch.py
```diff
@@ -41,11 +41,13 @@ def _command(self, code_package_url, environment, step_name, step_cli):
...
         else:
-            regex = '{user}-{flow_name}-{run_id}-'.format(
+            regex = '{user}-{flow_name}-'.format(
```
romain-intel, line 46: Ditto

metaflow/plugins/aws/batch/batch.py
```diff
@@ -56,10 +58,10 @@ def _job_name(self, user, flow_name, run_id, step_name, task_id, retry_count):
         return '{user}-{flow_name}-{run_id}-{step_name}-{task_id}-{retry_count}'.format(
             user=user,
             flow_name=flow_name,
-            run_id=run_id,
+            run_id=str(run_id or ''),
```
romain-intel, line 61: Small nit: this will create an empty string for a run ID of 0. Ditto for next two cases. Might be a bit problematic for retry count which may be more frequently 0?

metaflow/plugins/aws/batch/batch_cli.py
```diff
@@ -47,10 +47,10 @@ def _execute_cmd(func, flow_name, run_id, user, my_runs, echo):
     if my_runs:
         user = util.get_username()
 
-    latest_run = False
+    latest_run = True
```
romain-intel, line 50: What is this change?

metaflow/plugins/aws/batch/batch_cli.py
```diff
@@ -87,12 +87,14 @@ def echo_none(*args, **kwargs):
         except err as e:  # noqa F841
             pass
 
-@batch.command(help="List running Batch tasks of this flow")
-@click.option(
-    "--my-runs", default=False, is_flag=True, help="Run the command over all tasks."
-)
-@click.option("--user", default=None, help="List tasks for the given user.")
-@click.option("--run-id", default=None, help="List tasks corresponding to the run id.")
+
```
romain-intel, line 90: I am the first to fall in the trap but formatting changes are difficult to parse in parallel with actual code change :(.

metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -92,6 +94,11 @@ def _register_job_definition(self, image, job_role):
         response = self._client.register_job_definition(**payload)
         return response['jobDefinitionArn']
 
+    def job_def(self, image, iam_role):
```
romain-intel, line 97: This seems to be duplicated on line 110

metaflow/plugins/aws/batch/batch_decorator.py
```diff
@@ -136,6 +137,9 @@ def step_init(self, flow, graph, step, decos, environment, datastore, logger):
                     if not (my_val is None and v is None):
                         self.attributes[k] = str(max(int(my_val or 0), int(v or 0)))
         self.run_time_limit = get_run_time_limit_for_task(decos)
+        if self.run_time_limit < 60:
```
romain-intel, line 140: What is the reason for this?

metaflow/plugins/__init__.py
savingoyal: Yes, you are right. I will move it to the aws/stepfunctions directory.

metaflow/plugins/aws/batch/batch_decorator.py
savingoyal: AWS Batch has a lower limit of 60 seconds for timeouts. The job submission gets rejected otherwise - which is needed for AWS SFN.

metaflow/plugins/aws/batch/batch_cli.py
savingoyal: Apologies. There are a few, mostly related to comments and help messages to maintain consistency.

metaflow/plugins/aws/batch/batch_cli.py
savingoyal: The previous behaviour didn't work as expected.

metaflow/plugins/aws/batch/batch.py
savingoyal: Good catch, I will modify the behaviour.

metaflow/plugins/aws/batch/batch.py
savingoyal: Ah yes, I will remove it.

metaflow/plugins/aws/batch/batch.py
savingoyal: Ditto

metaflow/metadata/service.py
savingoyal: Yes, currently we ensure that is the case for SFN by prepending SFN to the execution id.

metaflow/metadata/service.py
savingoyal: What do you propose instead?

metaflow/metadata/service.py
savingoyal: Same as above.

metaflow/plugins/aws/batch/batch_client.py
savingoyal: Ah yes I will get rid of this

metaflow/cli.py
```diff
@@ -767,9 +770,19 @@ def start(ctx,
...
+        datastore_root = \
+          ctx.obj.datastore.get_datastore_root_from_config(ctx.obj.echo)
+    ctx.obj.datastore_root = ctx.obj.datastore.datastore_root = datastore_root
+
+    if decospecs:
+        decorators._attach_decorators(ctx.obj.flow, decospecs)
+
+    # initialize current and parameter context for deploy-time parameters
     current._set_env(flow_name=ctx.obj.flow.name, is_running=False)
+    parameters.set_parameter_context(ctx.obj.flow.name)
```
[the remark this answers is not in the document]
savingoyal: It didn't make much sense that `main` was aware of the notion of `parameters`.

metaflow/metadata/service.py
romain-intel: That's fine. I would rework the comment though to make it clear. It always confused me :).

metaflow/metadata/service.py
romain-intel: I would pass a always_create parameter as an additional optional param to _get_or_create that defaults to None. In _get_or_create, if that parameter is set, use that, otherwise use the logic inside to create. Also don't create on 404 to allow basically issues to be caught (basically: if we shoudln't be creating, it should be an error that the obj doesn't exist).

savingoyal: We can make that change outside of this PR.

PR #214, opened 2020-06-08
Initial commit of open source R binding

jasonge27:
Trigger github actions on R tests.


R/DESCRIPTION
```diff
@@ -0,0 +1,46 @@
+Package: metaflow
+Type: Package
+Title: R Binding for Metaflow 
+Version: 2.1.0
```
savingoyal, line 4: Maybe just `2.0.0`? The version bumps are going to be independent between metaflow and metaflow-r.

R/DESCRIPTION
```diff
@@ -0,0 +1,46 @@
...
+Author: Jason Ge [aut], 
+  Savin Goyal [aut, cre]
+Maintainer: Jason Ge <jge@netflix.com>
```
savingoyal, line 7: Can we set the email to `help@metaflow.org`. That will make is easier to provide visibility to the team about any comms with CRAN maintainers. We can add individual emails in the Author field above.

R/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
savingoyal, line 1: Can you add a change similar to https://github.com/Netflix/metaflow/pull/210

R/NAMESPACE
```diff
@@ -0,0 +1,39 @@
...
+export(new_run)
+export(new_step)
+export(new_task)
+export(parameter)
+export(py_to_r)
+export(r_to_py)
+export(run)
+export(run_client)
+export(set_default_namespace)
+export(set_namespace)
```
savingoyal, line 29: We would need to support `set_metadata` and `get_metadata` as well.

R/R/flags.R
```diff
@@ -0,0 +1,154 @@
...
+    "package_suffixes", "no-pylint",
+    "help", "resume",
+    "max_num_splits", "max_workers",
+    "other_args", "show", "force",
+    "notify_on_error",
+    "no_notify_on_error",
+    "notify_on_success",
+    "no_notify_on_success",
+    "notify_email", "authorize",
+    "my_runs", "run_id", "user",
```
savingoyal, line 139: `notify*` CLI args don't exist currently in OSS.

R/R/flags.R
```diff
@@ -0,0 +1,154 @@
...
+    "origin_run_id", "with", "force",
```
savingoyal, line 140: `force` doesn't exist as a CLI arg

R/R/flow.R
```diff
@@ -0,0 +1,102 @@
...
+                          private$.functions
+                        } else {
+                          private$.functions[!unlist(lapply(private$.functions, is.null))]
+                        }
+                      }
+                    )
+)
+
+header <- function(flow, flow_decorators = NULL) {
+  imports <- paste0(c("FlowSpec", "step", "Parameter", "retry", "environment", "batch", "timeout", "catch", "resources"), collapse = ", ")
```
savingoyal, line 76: I am wondering if `metaflow` could expose a method which lists all top-level imports for py3 and we populate this list by making a reticulate call.

R/R/metaflow_client.R
```diff
@@ -0,0 +1,147 @@
...
+#' @importFrom R6 R6Class
+#'
+#' @return Object of \code{\link{R6Class}} with fields/methods for introspection.
+#' @format \code{\link{R6Class}} object.
+#'
+
+#'
+#' @section Usage:
+#' \preformatted{
+#' client <- mf_flow$new()
```
savingoyal, line 15: nit -> `mf_flow` vs `mf_client`. What do you think of just `client`?

R/R/namespace.R
```diff
@@ -0,0 +1,24 @@
+#' Switch to a namespace specified by the given tag.
```
savingoyal, line 1: We would need similar functionality for `metadata`

R/R/run.R
```diff
@@ -0,0 +1,166 @@
...
+  invisible(file.remove(flow_file))
+  return(status_code)
+}
+
+run_cmd <- function(flow_file, ...) {
+  flags <- flags(...)
+  run_path <- system.file("run.R", package = "metaflow")
+  if ("metadata" %in% names(flags)) {
+    metadata <- paste0("--metadata=",flags$metadata)
+  } else {
```
savingoyal, line 50: We can remove the `else` statement and just rely on what the metaflow python default it. Similar comment for `datastore`, and `package_suffixes`

R/R/run.R
```diff
@@ -0,0 +1,166 @@
...
+    }
+  } else {
+    run <- "run"
+  }
+  
+  if ("batch" %in% names(flags)) {
+    if (is.logical(flags$batch)) {
+      if (flags$batch) {
+        batch <- "--with batch"
+        metadata <- "--metadata=service"
```
savingoyal, line 82: `metadata=service` is not a strict requirement here. We can avoid setting `metadata` and `datastore` and just rely on the defaults from metaflow-py.

R/tests/contexts.json
```diff
@@ -0,0 +1,40 @@
+{
+    "contexts": [
+        {
+            "name": "python3-all-local",
```
savingoyal, line 4: Just have the name as `all-local` and `batch`

R/tutorial/client_notebook.ipynb
```diff
@@ -0,0 +1,143 @@
+{
```
savingoyal, line 1: if you intend to include this notebook, let's clear the result cells.

R/R/zzz.R
```diff
@@ -0,0 +1,51 @@
...
+metaflow_version <- function(x) {
+  # utils library usually comes with the standard installation of R
+  version <- as.character(unclass(utils::packageVersion(x))[[1]])
+  if (length(version) > 3) {
+    version[4:length(version)] <- as.character(version[4:length(version)])
+  }
+  paste0(version, collapse = ".")
+}
+
+metaflow_attach <- function() {
```
savingoyal, line 45: Currently, when the R workflow executes, it prints out the python version on the CLI. We should make sure for R flows, it prints at least the R version, otherwise a lot of users are going to be confused.

R/inst/run_batch.R
```diff
@@ -1,10 +1,17 @@
...
+
 # dependencies for metaflow
-suppressMessages(if (!suppressWarnings(require("R6", character.only = TRUE))) { 
-  system("R CMD INSTALL ./metaflow-r", ignore.stdout = TRUE, ignore.stderr = TRUE)
-  install.packages(c("R6", "reticulate", "magrittr", "cli", "lubridate"), quiet = TRUE, repos='http://cran.us.r-project.org')
-})
+lapply(c("R6", "reticulate", "magrittr", "cli", "lubridate"), install_dep)
+
+# the remote code package places the R package under the metaflow-r folder
+system("R CMD INSTALL ./metaflow-r", ignore.stdout = TRUE, ignore.stderr = TRUE)
```
savingoyal, line 13: Can't we just do `install.packages('./metaflow-r')`?

R/DESCRIPTION
jasonge27: yep will set the email to help@metaflow.org.

R/R/flags.R
jasonge27: good catching. Removing the `notify` args

R/tests/contexts.json
jasonge27: yep making the change.

R/R/zzz.R
jasonge27: Yes it should still print out python version & R version on the CLI. See the Github Actions output https://github.com/Netflix/metaflow/runs/751334495?check_suite_focus=true in `Run integration test` section.  We don't have the colored output as I removed the dependency on the crayon package.

R/inst/run_batch.R
jasonge27: yep I'll swtich to `install.packages("./metaflow-r", repos=NULL, type="source")` . Just realized that it doesn't work unless we set the `repos` and `type` field as such.

.github/workflows/test_R.yml
```diff
@@ -0,0 +1,58 @@
...
+
+    - name: Set up R 
+      uses: r-lib/actions/setup-r@v1
+      with:
+        r-version: '3.6.3'
+
+    - name: Install Metaflow R binding and its dependencies
+      run: |
+        Rscript -e 'install.packages(c("R6", "reticulate", "magrittr", "lubridate", "cli"), repos="https://cloud.r-project.org")'        
+        cd R/; R CMD INSTALL .
```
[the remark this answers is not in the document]
jasonge27: I tried to avoid using methods from `devtools` library here since `devtool` has a big dependency chain and very slow to install. But yeah it's much better to use `install_deps(.)`. Switching to `devtools::insall_deps('.')`

.github/workflows/test_R.yml
```diff
@@ -0,0 +1,58 @@
+name: test-R-ubuntu
```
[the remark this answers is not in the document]
jasonge27: Will add test for macos & merge into a single `test.yml`. Will also copy the test process to publish.yml so we make sure each release does not break the R package

R/tutorial/client_notebook.ipynb
jasonge27: sounds good. Will do!

R/R/run.R
jasonge27: Makes sense. Will make the change.

R/R/run.R
jasonge27: Makes sense. Will make the change!

R/R/namespace.R
jasonge27: Yep adding it now.

R/NAMESPACE
jasonge27: Good catch. Adding these two methods.

R/DESCRIPTION
jasonge27: mark as resolved as we discussed

R/R/flags.R
jasonge27: good catch. Removing it.

R/R/flow.R
jasonge27: Are you thinking about something like `--help`? Basically we run `Rscript flow.R --help` and pass `--help` to python so the click library can list relevant command line args? I can give it a try.

R/R/metaflow_client.R
jasonge27: Good catch. `client` is better but we still have `flow_client` / `run_client` / 'step_client` / `task_client` so it does not solve the problem too much. I'm also a bit concerned that `client` may conflict in namespace with other popular packages or user's own R code

R/R/flow.R
jasonge27: I've added support for --help. Now users can do `Rscript flow.R --help` and `Rscript flow.R run --help` and see the help messages.

R/LICENSE
jasonge27: Yes I have `Copyright 2020 Netflix, Inc.` in the LICENSE file.

R/R/utils.R
```diff
@@ -0,0 +1,305 @@
...
+list_flows <- function() {
+  mf$Metaflow()$flows %>%
+    extract_ids()
+}
+
+#' Install Metaflow python dependencies
+#' 
+#' @export
+install <- function() {
+  system("python3 -m pip install git+https://github.com/Netflix/metaflow.git@R-dev --user")
```
savingoyal, line 253: TODO: Let's make sure we switch this to the correct branch before release.

R/R/utils.R
```diff
@@ -0,0 +1,305 @@
...
+}
+
+py_version <- function() {
+  reticulate::use_python(python_3(), required = TRUE)
+  mf <- reticulate::import('metaflow', delay_load = TRUE)
+  version <- mf$metaflow_version$get_version()
+  c(python_version = substr(version, 1, 5))
+}
+
+metaflow_version <- function(x) {
```
savingoyal, line 290: This function can very simply be `r_version` without an args.

R/R/utils.R
```diff
@@ -0,0 +1,305 @@
...
+install <- function() {
+  system("python3 -m pip install git+https://github.com/Netflix/metaflow.git@R-dev --user")
+  system("python3 -m pip install numpy>=1.18.4 --user")
+
+  metaflow_load()
+  metaflow_attach()
+}
+
+metaflow_load <- function(){
+  reticulate::use_python(system("which python3", intern = TRUE), required = TRUE)
```
savingoyal, line 261: should use `python_3` here.

PR #215, opened 2020-06-08
Set namespace to None in example notebook

savingoyal:
Setting namespace to None. A common scenario is where the user
copies to contents of the notebook into a sagemaker notebook and executes
the code snippets. That would fail, because the sagemaker notebook runs
with user:ec2-user. Setting the namespace to None allows us to bypass this
issue and provide a better user experience.

### DIFF
--- a/metaflow/tutorials/05-helloaws/helloaws.ipynb
+++ b/metaflow/tutorials/05-helloaws/helloaws.ipynb
@@ -22,7 +22,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from metaflow import Flow, get_metadata\n",
+    "from metaflow import Flow, get_metadata, namespace\n",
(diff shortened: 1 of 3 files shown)


seeravikiran:
Don't you need this change for tutorials 6 and upwards as well?
https://github.com/Netflix/metaflow/blob/e3541844805b747e55adc24a04e85f6f55d35b28/metaflow/tutorials/06-statistics-redux/stats.ipynb

savingoyal:
@seeravikiran Yes indeed, if you look at the commits, I am adding those currently.

GregHilston:
@savingoyal Looks perfect! I should have submitted this PR ha

I think this will help anyone on their first iteration, when deploying to a recently deployed Cloud Formation Stack.

PR #220, opened 2020-06-18
Improve IncludeFile support to support Step-Functions

romain-intel:
IncludeFiles are now compatible with Step Functions and files are properly uploaded/persistend when needed. As part of this change, we also include direct support for s3 files (and therefore include the functionality in #191). The difference here though is that any file already in s3 is not persisted/snapshotted (it will be directly lazily loaded on use).


romain-intel:
Tests are forthcoming. Modifications are welcome.

metaflow/includefile.py
```diff
@@ -70,7 +248,107 @@ def __repr__(self):
...
+                        if value_info['is_text']:
+                            return io.TextIOWrapper(f, encoding=value_info.get('encoding')).read()
+                        return f.read()
+                else:
+                    # We open this file according to the is_text and encoding information
+                    if value_info['is_text']:
+                        return io.open(obj.path, mode='rt', encoding=value_info.get('encoding')).read()
+                    else:
+                        return io.open(obj.path, mode='rb').read()
+            raise FileNotFoundError("File at %s does not exist" % self._url)
```
romain-intel, line 354: Oops, reading here, this is an error; it should be value_info['url'] and not self._url. I will fix this.

PR #230, opened 2020-06-24
Optimise `finished` and `successful` calls for task in client

savingoyal:
`finished` and `successful` calls for task currently go through the datastore, while they can be easily addressed through the much faster metadata service.

### DIFF
--- a/metaflow/client/core.py
+++ b/metaflow/client/core.py
@@ -877,7 +877,7 @@ class Task(MetaflowObject):
             True if the task completed successfully and False otherwise
         """
         try:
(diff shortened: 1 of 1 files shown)


metaflow/client/core.py
```diff
@@ -877,7 +877,7 @@ def successful(self):
             True if the task completed successfully and False otherwise
         """
         try:
-            return self['_success'].data
```
tuulos, line 880: how does this work if `_success` or `_task_ok` is False? This PR seems to change the semantics

PR #233, opened 2020-06-25
Properly handle strings passed to debug

romain-intel:


metaflow/debug.py
```diff
@@ -30,7 +30,11 @@ def __init__(self):
             setattr(self, typ, op != self.noop)
 
     def log(self, typ, args):
-        print('debug[%s]: %s' % (typ, ' '.join(args)), file=sys.stderr)
+        if isinstance(args, str):
```
tuulos, line 33: this doesn't work too well for py2 where `args` maybe a unicode object. Use `util.is_stringish` instead.

romain-intel:
Superseded by #289

PR #236, opened 2020-06-25
Moved Metadata to plugins. Also moved more defaults to metaflow_config

romain-intel:
Metadata providers are now plugins. We include the two that were already there
(local and service) but other metadata providers can now be easily added.


metaflow/metaflow_config.py
```diff
@@ -41,9 +41,15 @@ def from_conf(name, default=None):
 ###
 # Default configuration
 ###
+DEFAULT_AUTH = from_conf('METAFLOW_DEFAULT_AUTH', 'aws')
 DEFAULT_DATASTORE = from_conf('METAFLOW_DEFAULT_DATASTORE', 'local')
+DEFAULT_DATA_PROVIDER = from_conf('METAFLOW_DEFAULT_DATAPROVIDER', '')
```
seeravikiran, line 46: Nit: DEFAULT_DATAPROVIDER so its also consistent with the `from_conf` invocation across other values.

metaflow/package.py
```diff
@@ -16,13 +17,13 @@
     import io
     BytesIO = io.BytesIO
 
-DEFAULT_SUFFIXES = ['.py']
+LIST_DEFAULT_SUFFIXES = DEFAULT_SUFFIXES.split(',')
```
seeravikiran, line 20: DEFAULT_SUFFIXES_LIST is probably better, no?

metaflow/metaflow_config.py
romain-intel: Done.

metaflow/package.py
```diff
@@ -16,13 +17,13 @@
     import io
     BytesIO = io.BytesIO
 
-DEFAULT_SUFFIXES = ['.py']
```
tuulos, line 19: how is this change related to this PR?

metaflow/datastore/s3.py
```diff
@@ -105,7 +105,7 @@ def get_latest_tasks(cls,
                 task_urls = [task.url for task in task_objs]
             urls = []
             for task_url in task_urls:
-                for attempt in range(5):
```
tuulos, line 108: `MAX_ATTEMPTS` seems to be 6 but we used to have `range(5)`. Let's just double check that the change is intentional

metaflow/package.py
romain-intel: I moved this to the configuration file.

romain-intel:
Replaced by #331.

PR #237, opened 2020-06-25
Add 'include' parameter to merge_artifacts for easier control of arti…

romain-intel:
…facts to merge


metaflow/exception.py
```diff
@@ -105,9 +105,16 @@ class CommandException(MetaflowException):
...
-class MergeArtifactsException(MetaflowException):
+class UnhandledInMergeArtifactsException(MetaflowException):
     headline = "Unhandled artifacts in merge"
 
     def __init__(self, msg, unhandled):
-        super(MergeArtifactsException, self).__init__(msg)
+        super(UnhandledInMergeArtifactsException, self).__init__(msg)
+        self.artifact_names = unhandled
+
+class MissingInMergeArtifactsException(MetaflowException):
```
seeravikiran, line 115: Unrelated to the code move itself, I'm curious why you chose this over `MergeArtifactsMissingException` or `MissingMergeArtifactsException`. Specifically I felt the use of the preposition (`in`) is less-commonplace.

Let me know if this was inspired from somewhere else, since I'm also not super familiar with Python Readability recommendations; and learning myself.

Related - why did you want the explicit exception types instead of just passing a string message to `MergeArtifactsException` similar to `MetaflowException`?

metaflow/flowspec.py
```diff
@@ -288,40 +289,68 @@ def merge_artifacts(self, inputs, exclude=[]):
         inputs : List[Steps]
             Incoming steps to the join point
         exclude : List[str], optional
+            If specified, do not consider merging artifacts with a name in `exclude`.
+            Cannot specify if include is also specified
```
seeravikiran, line 293: Could you add `include` and `exclude` (using tilde) in L:293 and L:295?

metaflow/exception.py
seeravikiran: EDIT: Is it just for the sake of testing [which could be handled by peeking at the contents of the error msg], or did you have other user-facing reasons?

romain-intel: No particular reason for the In. Not sure why I did that. For the error, I think it did make it easier for error parsing. I am generally not a fan of message level parsing to determine what the error was (it's a bit brittle) but this could be simplified. Maybe in another PR?

metaflow/flowspec.py
romain-intel: Done.

metaflow/exception.py
seeravikiran: Yep of course. Most of my comments of this form are non-blocking. Just wanted to facilitate discussion while reviewing those lines of code.

tuulos:
lgtm

romain-intel:
Superseded by #287

PR #239, opened 2020-06-25
Auth plugins

romain-intel:


metaflow/datastore/util/s3util.py
```diff
@@ -4,13 +4,26 @@
...
-    return get_authenticated_boto3_client('s3', { 'endpoint_url': S3_ENDPOINT_URL, 'verify': S3_VERIFY_CERTIFICATE }), ClientError
+    from metaflow.plugins import AUTH_PROVIDERS
+    try:
+        import boto3
+        from botocore.exceptions import ClientError
+    except (NameError, ImportError):
+        raise MetaflowException("Could not import module 'boto3' which "
+                                "is required by S3 datastore. Install boto "
+                                "first.")
+    get_client = AUTH_PROVIDERS[DEFAULT_AUTH]
```
seeravikiran, line 21: Can we be explicit about requesting `aws` client instead of `DEFAULT` since `s3` only makes sense with `aws`?
I understand both are the same now, but I want to be future-proof if the default changes.
Alternatively, at least rename `get_client` to `get_aws_client` if you feel the default won't change.

metaflow/datastore/util/s3util.py
```diff
@@ -4,13 +4,26 @@
...
+    # If DEFAULT_AUTH does not support s3, this should error. Not checking for
+    # a specific value for DEFAULT_AUTH as multiple values may provide support for S3
```
seeravikiran, line 23: Could you expand on what the latter part of your comment implies? `multiples values might provide support for S3`?

romain-intel: Will do. The idea was that I wasn't going to check if DEFAULT_AUTH == 's3' for example because there could be other auth mechanisms that are backed by S3.

metaflow/datastore/util/s3util.py
```diff
@@ -4,13 +4,27 @@
...
-    return get_authenticated_boto3_client('s3', { 'endpoint_url': S3_ENDPOINT_URL, 'verify': S3_VERIFY_CERTIFICATE }), ClientError
+    from metaflow.plugins import AUTH_PROVIDERS
+    try:
+        import boto3
+        from botocore.exceptions import ClientError
+    except (NameError, ImportError):
+        raise MetaflowException("Could not import module 'boto3' which "
+                                "is required by S3 datastore. Install boto "
+                                "first.")
+    get_client = AUTH_PROVIDERS[DEFAULT_AUTH]
```
tuulos, line 21: why does this go through `AUTH_PROVIDERS` but `batch_client` doesn't?

romain-intel: Fixed. Good catch!

romain-intel:
Replaced by #332.

PR #240, opened 2020-06-25
Provide support for Metaflow customization

romain-intel:
Metaflow can be easily customized and extended through the use of plugins.
This patch allows Metaflow to look for an additional package called 'metaflow_custom'
to load additional plugins or configuration options.

More details on this feature will be forthcoming.


metaflow/metaflow_config.py
```diff
@@ -199,3 +199,14 @@ def get_authenticated_boto3_client(module, params={}):
                 raise MetaflowException(repr(e))
         return boto3.session.Session(**cached_aws_sandbox_creds).client(module, **params)
     return boto3.client(module, **params)
+
+# Check if there is a an extension to Metaflow to load and override everything
+try:
+   import metaflow_custom.config.metaflow_config as extension_module
```
seeravikiran, line 205: Indentation seems slightly off in this section of code. Could you double check?

metaflow/plugins/__init__.py
```diff
@@ -20,38 +45,40 @@ def get_plugin_cli():
...
 # Add an entry here if you need a new flow-level annotation. Be
 # careful with the choice of name though - they become top-level
 # imports from the metaflow package.
 from .conda.conda_flow_decorator import CondaFlowDecorator
-FLOW_DECORATORS = [CondaFlowDecorator]
+FLOW_DECORATORS = _merge_lists([CondaFlowDecorator], ext_plugins.FLOW_DECORATORS, 'name')
 
 # Sidecars
-SIDECAR = {}
+SIDECAR = ext_plugins.SIDECAR
```
seeravikiran, line 69: Nit: This should've ideally been `SIDECARS` since its a dictionary (so a container of a sidecars).
If you agree - feel free to handle this in a separate PR. But just want to field for your opinion via this comment.

metaflow/plugins/__init__.py
```diff
@@ -9,8 +27,15 @@ def get_plugin_cli():
     from . import package_cli
     from .aws.batch import batch_cli
 
-    return [package_cli.cli,
-            batch_cli.cli]
+    return ext_plugins.get_plugin_cli() + [package_cli.cli, batch_cli.cli]
+
+
+def _merge_lists(decos, overrides, attr):
```
seeravikiran, line 33: Maybe you can leave a short comment on usage of `merge_lists`. One thing to bring out would be to prefer entities from `metaflow_custom` as overrides - which is implicit from your usage throughout this file. WDYT?

seeravikiran, line 33: Could you rename `decos` to something less-decorator specific. It seems you use this function for `environments` as well.
Maybe base and overrides?

metaflow/metaflow_config.py
romain-intel: Fixed. Good catch.

metaflow/plugins/__init__.py
romain-intel: Done.

metaflow/plugins/__init__.py
romain-intel: Done

romain-intel: Done

tuulos:
lgtm

romain-intel:
Superseded by #295

PR #242, opened 2020-07-01
Add instructions to tutorial 6 for execution inside a conda environment

savingoyal:
Some users might not have access to pandas within the docker container
executing their AWS Batch job. The situation is different for sandbox
accounts since the image bundles in pandas by default.

Added a command line flag to enable conda and pandas for all executions

### DIFF
--- a/metaflow/tutorials/04-playlist-plus/README.md
+++ b/metaflow/tutorials/04-playlist-plus/README.md
@@ -11,11 +11,10 @@ isolated and reproducible environments for individual steps.**
 - Metaflow's conda based dependency management.
 
 #### Before playing this episode:
-This tutorial requires the 'conda' package manager to be installed with the
-conda-forge channel added.
-
-1. Download Miniconda at https://docs.conda.io/en/latest/miniconda.html
-2. ```conda config --add channels conda-forge```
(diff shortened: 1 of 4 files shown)


oonisim:
@savingoyal allow me to ask if there is a plan to be able to install metaflow via conda. The reason is as in [Using Pip in a Conda Environment](https://www.anaconda.com/blog/using-pip-in-a-conda-environment), pip and conda are separate package management systems and mixing them would lead issues. Metaflow installation is pip, but metaflow requires conda, which may mislead the users into mixing both pip and conda without being aware of the issues of mixing pip and conda.

It looks there are efforts to make pip compatible within conda but looks still work in progress.
* [PyPy builds on conda-forge](https://conda-forge.org/blog/posts/2020-03-10-pypy/)
* [mproving interoperability with pip](https://docs.conda.io/projects/conda/en/latest/user-guide/configuration/pip-interoperability.html)

If this has been already considered and addressed within metaflow, please advise.

savingoyal:
@oonisim You can use metaflow within a conda environment. Metaflow is available as a conda package on the conda-forge channel - https://anaconda.org/conda-forge/metaflow. 
Yes, we test metaflow within virtual environments, so you should be able to create one using say conda and run your code within that environment.

PR #249, opened 2020-07-07
AST-only version of graph.FlowGraph

tuulos:
this avoids importing the flow module and execution of any user-defined code

### DIFF
--- a/metaflow/graph.py
+++ b/metaflow/graph.py
@@ -2,49 +2,12 @@ import inspect
 import ast
 import re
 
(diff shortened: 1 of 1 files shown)


romain-intel:
Prior to merging, could you rebase to not remove the deindent_docstring code?

tuulos:
> Prior to merging, could you rebase to not remove the deindent_docstring code?

@romain-intel did you test that indenting is still a problem with this PR. Hopefully it shouldn't be.

PR #255, opened 2020-07-24
Remove from s3 the credentials part of conda private channels: https:…

benbemoh:
…//github.com/Netflix/metaflow/issues/252

### DIFF
--- a/metaflow/plugins/conda/conda_step_decorator.py
+++ b/metaflow/plugins/conda/conda_step_decorator.py
@@ -6,6 +6,9 @@ import platform
 import requests
 import shutil
 import tempfile
+
+from .utils import PluginHelper
+
 try:
     from urlparse import urlparse
 except:
@@ -116,7 +119,7 @@ class CondaStepDecorator(StepDecorator):
                 payload = {
                     'explicit': exact_deps,
                     'deps': [d.decode('ascii') for d in deps],
-                    'urls': urls,
+                    'urls': [PluginHelper.sanitize_conda_package_url(u) for u in urls],
                     'order': order
                 }
             else:
@@ -138,8 +141,9 @@ class CondaStepDecorator(StepDecorator):
         files = []
         to_download = []
         for package_info in self.conda.package_info(env_id):
-            url = urlparse(package_info['url'])
-            path = os.path.join(CONDA_PACKAGE_S3ROOT, 
+            url_sanitized = PluginHelper.sanitize_conda_package_url(package_info["url"])
+            url = urlparse(url_sanitized)
+            path = os.path.join(CONDA_PACKAGE_S3ROOT,
                                 url.netloc,
                                 url.path.lstrip('/'),
                                 package_info['md5'],
--- a/metaflow/plugins/conda/utils.py
+++ b/metaflow/plugins/conda/utils.py
@@ -1,0 +1,57 @@
+"""This modules provides a helper function for the conda plugin."""
+
+
+try:
+    from urlparse import urlparse
+except:  # noqa E722
+    from urllib.parse import urlparse
+
+
+class PluginHelper:
+    """A plugin helper class."""
+
+    @staticmethod
+    def sanitize_conda_package_url(url, token_ref="t"):  # noqa S107
+        """Sanitize conda private channels.
+
+        This function removes `username`, `password` and `token`
+        if they exist from the url.
+
+        Possible conda private channels configuration:
+            - https://user:password@private.com:port/<channel>
+            - https://private.com:port/t/<token>/<channel>
+            - https://private.com:port/t/<token>/<channel>/label/<labelname>
+
+        :param url: channel url
+        :type url: Union[str, ParseResult]
+        :param token_ref: token name in the  url path
+        :type token_ref: str
+        :return: the sanitized url
+        :type : Union[str, ParseResult]
+        """
+        sanitized_url = url
+        if url:
+            # parse the url (code should be py2, py3 compatible)
+            c_url = urlparse(url) if isinstance(url, str) else url
+            # remove the username/password from the url 'netloc' field
+            # use the _replace() method as described here:
+            # https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse
+            if c_url.password:
+                c_url = c_url._replace(
+                    netloc=c_url.netloc.replace(
+                        "{}:{}@".format(c_url.username, c_url.password), "", 1
+                    )
+                )
+            # remove the token, if it exists, from the url 'path' field
+            if (
+                token_ref
+                and c_url.path.startswith("/{}/".format(token_ref))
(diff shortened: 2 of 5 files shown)


savingoyal:
@jasonge27 Can you shepherd this PR?

jasonge27:
@benbemoh Thanks a lot for the PR! A quick workaround is to use [.condarc](https://docs.conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html#expansion-of-environment-variables) to store the credentials:
```
channels:
  - https://${USERNAME}:${PASSWORD}@my.private.conda.channel
```

Really love it that you've carefully documented your PR. Introducing a new class/module for the URL parsing seems to be an overkill. Also the dependency on `urlparse`and `ParseResult` is a bit concerning. We're leaning towards the idea of hashing the entire URL internally and will fix it in a future release.

parviste:
> 
> 
> @benbemoh Thanks a lot for the PR! A quick workaround is to use [.condarc](https://docs.conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html#expansion-of-environment-variables) to store the credentials:
> 
> ```
> channels:
>   - https://${USERNAME}:${PASSWORD}@my.private.conda.channel
> ```
> 

We already have `.condarc` like that, but presumably the variables are substituted there _before_ the S3 keys are determined, so the actual credentials end up there, rather than just the variable names. Or maybe I am misunderstanding what you are suggesting.

parviste:
> We're leaning towards the idea of hashing the entire URL. Will try to fix it in a future release.

So the idea is to simply hash the entire URL? I guess that would make navigating the S3 bucket a bit more difficult, but maybe that is not so bad. Is someone actively working on that already?

Otherwise we could try to implement that instead, because we've found no workaround so far, and having the credentials visible in S3 is a no-go for us.

parviste:
@jasonge27 Any news on this? As mentioned, the proposed workaround does not work, and credentials are still (after several months) being stored in plain text in S3. We might be willing to fix this, so could you confirm that simply hashing the URL is the desired way of fixing this?

jasonge27:
@parviste sorry I missed your previous comment in my github notification inbox. Please feel free to implement workarounds in your own fork so at least it unblocks you for now. Apologies for having the problem in the first place.

We had some discussions within the team on this topic. We think hashing should be better than explicitly sanitizing the URL since the conda channel URL format may change in the future and break the logic. However, there are some concerns that surfaced during our discussion regarding the hashing method: 

- the old S3 path for our packages will be invalidated. We will end up with duplicated pkg tarballs on S3. Our users would also need to re-cache their environments on S3.

- With the hashing solution, every credential change would result in a different hashed path and thus duplicated packages and env re-caching overhead for users.

- We still want to keep some readability in the S3 URLs for ease of debugging when things go wrong. One option is to only hash the credential part. The other option is to use hashing only when a config or environment variable is set, otherwise we use the current explicit S3 URL path by default.

We are still evaluating various options and happy to continue chatting on the issue if you have other ideas/suggestions.

parviste:
I'm actually not convinced that hashing is the way to go. It may be quite easy to figure out the domain part, leaving the credentials that could potentially be brute forced if no salting is used. Salting in turn would make the whole thing more complicated. The easiest and secure solution really seems to be to just remove the credentials from the bucket key, like the original PR.

As you point out, credentials changes would result in duplicate packages, which would discourage credentials rotation. Simply not including the credentials solves that problem as well.

An environment variable could be added to  enable the original behaviour, but the default should IMHO definitely be to not show credentials in plaintext S3 keys.

jasonge27:
I agree that removing the credential part from the s3 URL is better than hashing the cred part. The problem is I'm not sure what's a reliable way to remove the credential URL. The parsing & removing logic seems to depend on the URL format, an implementation detail of the private conda hosting server. The URL format may change in the future and could differ in different setup.

parviste:
> 
> 
> I agree that removing the credential part from the s3 URL is better than hashing the cred part. The problem is I'm not sure what's a reliable way to remove the credential in the URL. The parsing & removing logic seems to depend on the URL format, an implementation detail of the private conda hosting server. The URL format may change in the future and could differ in different setup.

I'm not sure what you mean here. The URI RFC specifies how credentials are to be added to a URL. Are you suggesting that conda might not use URIs for channels in the future (or is this already the case?), or that there are conda servers that break the RFC or provide credentials in a non-standard way?

If we want to be paranoid, we could remove credentials where we know how (URIs) and if we find a channel that uses some other format then we hash it.

jasonge27:
We're not sure if anaconda enterprise will continue to support this format for authentication. Looks like the community is raising concerns on the current design https://stackoverflow.com/questions/60672233/securely-pass-credentials-to-conda-to-access-corporate-proxy-channels. Besides, there are customized conda servers that run authentication in non-standard ways. 

Backward compatibility is important for us and that's the main reason for us being conservative on this before we're clear what is the best long-term solution. Sorry for the inconvenience and hope the workarounds in your fork solve your problems for now.

parviste:
I'm honestly a bit confused here. Nothing in the linked StackOverflow link suggests _not_ continuing to use URLs in Conda, but rather the question is how to indicate which credentials to use, but still using an URL.

My proposal above would also fully work with customized Conda installs, if instead of hashing non-URLs we simply leave them be. It would just mean that these customized non-standard installs would have poor security, but at least it would not be worse than now, and those who use non-customized versions would get vastly improved security.

We planned on using a fork until this is fixed upstream, but now it seems it will actually never happen because conda might possibly change at some time in the future. This means that we now have to reconsider using Metaflow, but we neither have the resources to maintain a fork indefinitely, nor can we break all kinds of security guidelines.

Again, if you have a proposal for an acceptable solution we would be willing to implement and merge upstream.

jasonge27:
My interpretation of the GitHub issue is that the community is raising concerns on how Anaconda is storing & passing credentials in clear text, quote "...I feel very uncomfortable about having to store my password in plain sight...", which includes the .condarc file and also the credentials URL format. 

The team has been discussing what's the best solution and don't have an answer yet. All I can do right now is sharing contexts and continue discussions. Your input is also super valuable to us. I'm re-surfacing this issue within the team.

jasonge27:
My interpretation of the GitHub issue is that the community is raising concerns on how Anaconda is storing & passing credentials in clear text, quote "...I feel very uncomfortable about having to store my password in plain sight...", which includes the .condarc file and also the package URLs stored in `conda-meta` folder in conda environment.

The team has been discussing what's the best solution and don't have an answer yet. All I can do right now is sharing contexts and continue discussions. Your input is also super valuable to us. I'm re-surfacing this issue within the team.

pikulmar:
Any updates on this issue?

parviste:
> The team has been discussing what's the best solution and don't have an answer yet. All I can do right now is sharing contexts and continue discussions. Your input is also super valuable to us. I'm re-surfacing this issue within the team.

@jasonge27 Any clarity on how we could proceed with this?

PR #275, opened 2020-08-07
Fix timedate conversion from service to client

romain-intel:


metaflow/client/core.py
```diff
@@ -337,8 +338,7 @@ def __init__(self,
         else:
             raise MetaflowInternalError(msg="Unknown type: %s" % self._NAME)
 
-        self._created_at = time.strftime(
-            '%Y-%m-%dT%H:%M:%SZ', time.gmtime(self._object['ts_epoch']//1000))
+        self._created_at = datetime.fromtimestamp(self._object['ts_epoch']/1000.0)
```
savingoyal, line 341: The type of `_created_at` is now a datetime object rather than the string previously.

romain-intel:
Replaced by #692

PR #282, opened 2020-08-13
Use the first environments directory instead of conda_prefix

### LINKED ISSUE #179: Permission error when running tutorial episode 04
Hi, I am new to `metaflow` and currently trying to run through the tutorials in my work environment. As a non-root user of `conda`, I don't have write permission for where `conda` is installed: `/opt/miniconda`, so I ran into this error when executing episode 04 in which `@conda` was used:
```
PermissionError: [Errno 13] Permission denied: '/opt/miniconda/mf_env-creation.lock'
```
Is there some configuration I can do to assign a new location where this `mf_env-creation.lock` will be written?

`conda` version = 4.8.3
`metaflow` version = 2.0.1

sodre:
The conda_prefix folder is not always writable, so creating a lock there can fail.
On the other hand, conda's first entry in "envs_dirs" should always be writable.

Fix #179

### DIFF
--- a/metaflow/plugins/conda/conda.py
+++ b/metaflow/plugins/conda/conda.py
@@ -141,7 +141,7 @@ class Conda(object):
         return None
 
     def _env_lock_file(self, env_id):
-        return os.path.join(self._info()['conda_prefix'], 'mf_env-creation.lock')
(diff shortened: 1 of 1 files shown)


romain-intel:
Perfect! This was the fix I had in mind. We can investigate if we need the lock but this is perfect for now :).

Issue #283, opened 2020-08-14
(R) Cannot define steps with anonymous functions


mdneuzerling:
I am unable to define a step which uses an anonymous function:

``` r
library(metaflow)
#> Metaflow (R) 2.2.0 loaded
#> Metaflow (Python) 2.2.0 loaded
library(dplyr)
metaflow("TestFlow") %>%
  step(
    step = "anonymous_function_step",
    r_function = function(step) step$x <- 3 
  )
#> Error in names(fun) <- name: 'names' attribute [4] must be the same length as the vector [1]
```

<sup>Created on 2020-08-14 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

I've traced the issue to [this line of code in step.R](https://github.com/Netflix/metaflow/blob/128d9e780c23b5282dee1be09d6c2511b7a9081f/R/R/step.R#L33):

```r
  if (!is.null(r_function)) {
    function_name <- as.character(substitute(r_function))
    ...)
  }
```

<details>

<summary>Session info</summary>

``` r
devtools::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value                       
#>  version  R version 4.0.0 (2020-04-24)
#>  os       Ubuntu 20.04.1 LTS          
#>  system   x86_64, linux-gnu           
#>  ui       X11                         
#>  language en_AU:en                    
#>  collate  en_AU.UTF-8                 
#>  ctype    en_AU.UTF-8                 
#>  tz       Australia/Melbourne         
#>  date     2020-08-14                  
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version    date       lib source                            
#>  assertthat    0.2.1      2019-03-21 [1] CRAN (R 4.0.0)                    
#>  backports     1.1.8      2020-06-17 [1] CRAN (R 4.0.0)                    
#>  callr         3.4.3      2020-03-28 [1] CRAN (R 4.0.0)                    
#>  cli           2.0.2      2020-02-28 [1] CRAN (R 4.0.0)                    
#>  crayon        1.3.4      2017-09-16 [1] CRAN (R 4.0.0)                    
#>  desc          1.2.0      2018-05-01 [1] CRAN (R 4.0.0)                    
#>  devtools      2.3.0      2020-04-10 [1] CRAN (R 4.0.0)                    
#>  digest        0.6.25     2020-02-23 [1] CRAN (R 4.0.0)                    
[...]

PR #284, opened 2020-08-14
(R) Support for anonymous functions in steps

mdneuzerling:
Fixes Issue #283


R/R/step.R
```diff
@@ -31,6 +31,9 @@ step <- function(flow, ..., step, r_function = NULL, foreach = NULL, join = FALS
   }
   if (!is.null(r_function)) {
     function_name <- as.character(substitute(r_function))
+    if (length(function_name) > 1) { # likely an anonymous function
+      function_name <- paste0(step, "_function")
```
jasonge27, line 35: Thanks for the improvement! Can we add hash of the function to this function name as well? Just in case the user may have other non-anonymous function with the same name @mdneuzerling

mdneuzerling: Sorry for the delay, @jasonge27. Base R doesn't have any hashing capabilities, so I would need to add a dependency to the `digest` package. Would this be okay? I also considered appending the location of the function in memory instead, but even this would require an additional dependency (`pryr` or `lobstr`).

bcgalvin: thanks for the PR @mdneuzerling!

Would using the python builtin hash function save the dependency here? 

```
library(reticulate)
builtins <- import_builtins()
py_hash <- function(fun) {
  builtins$hash(paste(deparse(fun), collapse = ""))
}


> sumfun <- function(x, y) { x + y}
> py_hash(sumfun)
[1] -61687620
>
> sumfun1 <- function(x, y) {x + y + y}
> py_hash(sumfun1)
[1] -504669628
```

mdneuzerling: Thanks @bcgalvin. I've got that working, although [the inbuilt Python hash function salts the inputs and so the hash will change between sessions](https://docs.python.org/3/reference/datamodel.html#object.__hash__). If this behaviour is undesirable, I think a dependency on either R's `digest` package or Python's `hashlib` module is needed.

jasonge27: Hey thank you David for the updated PR! @mdneuzerling (not sure why ..Github is not letting me mention you). I discussed with the team. We believe it's better to use `digest` since :
1.  as you've mentioned the native python hash is not very stable (overflow & changing between sessions) 
2. The `digest` pkg is maintained by Dirk and does not introduce additional dependency (`utils` pkg is already required by `reticulate`) . 

I saw that you've gone via the python path. Sorry for the late reply. Do you have time to update the PR & use `digest`?

mdneuzerling: Not a problem. I"m not an expert on hashing algorithms, but I've heard good things about spookyhash for non-cryptographic purposes, so I went with that.

``` r
#> flow <- metaflow("TestFlow") %>%
#     step(
#       step = "anonymous",
#       r_function = function(step) step$x <- 3 
#     )
#> names(flow$get_functions())
"anonymous_function_7542958bbcd6fa74bdf17e888b956426"
```

jasonge27: Thank you @mdneuzerling for the quick turnaround. Merging the PR.

PR #235, opened 2020-06-25
Add possibility for flow level decorators to have options exposed at …

romain-intel:
…the top level.

Also, minor other cleanups regarding PyLint

PR #294, opened 2020-08-18
Add possibility for flow level decorators to have options exposed at the top level

romain-intel:
Also, minor other cleanups regarding PyLint

Supersedes #235


metaflow/plugins/conda/conda_flow_decorator.py
```diff
@@ -37,7 +37,7 @@ class MyFlow(FlowSpec):
                 'python': None,
                 'disabled': None}
 
-    def flow_init(self, flow, graph, environment, datastore, logger):
+    def flow_init(self, flow, graph,  environment, datastore, logger, echo, options):
```
savingoyal, line 40: Need to make the same change **[here](https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/aws/step_functions/schedule_decorator.py)**

metaflow/cli.py
```diff
@@ -747,9 +756,6 @@ def start(ctx,
     echo(" executing *%s*" % ctx.obj.flow.name, fg='magenta', nl=False)
     echo(" for *%s*" % resolve_identity(), fg='magenta')
 
-    if decospecs:
-        decorators._attach_decorators(ctx.obj.flow, decospecs)
-
```
savingoyal, line 752: What is the rationale for this deletion?

savingoyal (changes requested):

metaflow/cli.py
romain-intel: It was duplicated (for some reason). If you look at the code path, it is also present on line 792. I can keep this one and remove that one if you want. In fairness, there is still duplication because we invoke it here, on line 792 and then either on line 804 or in before_run so this bring is down from 3 to 2 :). I could fix it more to bring it to one (not sure why we call it 3 times.

metaflow/plugins/conda/conda_flow_decorator.py
romain-intel: Good catch. Missed it in all the rebasing and intro of step functions. I'll check if I forgot another one too.

metaflow/cli.py
```diff
@@ -553,6 +553,14 @@ def resume(obj,
     if step_to_rerun is None:
         clone_steps = set()
     else:
+        # validate step name
```
tuulos, line 556: just to confirm that this chunk has nothing to do with flow-level decorators per se, right?

metaflow/cli.py
tuulos: we have to call it in multiple places to handle both `python myflow.py --with deco command` and `python myflow.py command --with decospec` formats. There are three valid places to call it (the fourth one seems to be redundant).

savingoyal:
@romain-intel Looks good to merge, pending the fix to `@schedule.`

metaflow/cli.py
romain-intel: Ya, not sure why this snuck in here. I can make a separate PR for it too or can just leave it here if that is OK.

metaflow/cli.py
romain-intel: Thanks for the clarification @tuulos. I think that even with this removal, there is one more place it could be removed. Currently it is called in the `start` function twice (once protected with `if decospecs` and another time with `if ctx.invoked_subcommand not in` a few lines later. Based on what you mention, I believe we could also remove the call to `_attach_decorator` in that location and just keep the `_init_step_decorators` there as well as in `before_run`. After that change, there would be two calls to `_attach_decorators` (one in start for top level `--with` decorators, and one in `before_run` for any run/resume level `--with` decorators). There would also be a similar `_attach_decorators_to_step` for the step level `--with` decorators. Would that be correct?

PR #295, opened 2020-08-18
Provide support for Metaflow customization

romain-intel:
Metaflow can be easily customized and extended through the use of plugins.
This patch allows Metaflow to look for an additional package called 'metaflow_custom'
to load additional plugins or configuration options.

More details on this feature will be forthcoming.

Supersedes #240.


tuulos:
the code LGTM. Pending testing

savingoyal (approved):

PR #308, opened 2020-08-26
pin conda libs version depending on different python version

jasonge27:
Pin default libs version in conda environment based on different python versions. 

Supported python version: 2.7.x, 3.5.x, 3.6.x, 3.7.x, and 3.8.x

### DIFF
--- a/metaflow/metaflow_config.py
+++ b/metaflow/metaflow_config.py
@@ -188,30 +188,21 @@ def get_version(pkg):
 # PINNED_CONDA_LIBS are the libraries that metaflow depends on for execution
 # and are needed within a conda environment
 def get_pinned_conda_libs(python_version):
-    if python_version.startswith(("2.7", "3.6", "3.7", "3.8")):
-        return {
-            'click': '7.1.2',
-            'requests': '2.24.0',
-            'boto3': '1.14.47',
-            'coverage': '4.5.4'
-        }
-    elif python_version.startswith("3.5"):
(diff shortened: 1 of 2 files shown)


metaflow/metaflow_config.py
```diff
@@ -187,13 +187,29 @@ def get_version(pkg):
...
+            'requests': '2.24.0',
+            'boto3': '1.9.88',
+            'coverage': '4.5.1'
+        }
+    elif python_version in ["2", "3"]:
+        raise MetaflowException(
+            "Please specify complete python version number, " +
+            "for example 2.7, 3.7, 3.7.8")
+    else:
+        raise MetaflowException(
```
savingoyal, line 210: @jasonge27 How about [Python 3.9](https://docs.python.org/3.9/whatsnew/3.9.html)?

jasonge27: @savingoyal Python 3.9 has not yet landed in conda-forge or conda main channel. We can ignore it for now?

savingoyal: Yes, okay to ignore but Metaflow will throw an error whenever `3.9` makes its way to conda and I am not sure we want that. Given that only `3.5` has different dependencies compared to other versions, how about a simple clause - 
```
if py_version.startswith('3.5'):
    ...
else:
    ...
```

jasonge27: @savingoyal good idea! updated the PR according your suggestion

Issue #281, opened 2020-08-13
R Install Issues


realauggieheschmeyer:
Forgive my Pythonic ignorance but I, for the life of me, cannot get {metaflow} installed on my computer. I've been following the tutorial on your website and keep getting `Error: Python shared library not found`. I have tried the steps listed on your website for this exact error, but I get the same error message no matter what I do. I was hoping you might be able to provide some guidance.

I've included a reprex but I don't know how helpful it will be as it doesn't include my system specs. Let me know what info you might need and I will happily track it down. 

Thank you.

``` r
reticulate::install_miniconda()
#> * Downloading 'https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh' ...
#> * Installing Miniconda -- please wait a moment ...
#> * Miniconda has been successfully installed at '/Users/auggieheschmeyer/Library/r-miniconda'.
#> [1] "/Users/auggieheschmeyer/Library/r-miniconda"
reticulate::use_miniconda(required = TRUE)
reticulate::conda_create(envname = "metaflow-env", packages = "python")
#> [1] "/Users/auggieheschmeyer/Library/r-miniconda/envs/metaflow-env/bin/python"
reticulate::use_condaenv(condaenv="metaflow-env", required = TRUE)
metaflow::install()
#> Metaflow (R) 2.2.0 loaded
#> Error: Python shared library not found, Python bindings not loaded.
#> Use reticulate::install_miniconda() if you'd like to install a Miniconda Python environment.
```

<sup>Created on 2020-08-13 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

bcgalvin:
The issue here is that `metaflow:install()` is using system python3 (https://github.com/Netflix/metaflow/blob/master/R/R/utils.R#L286) and is ignoring your conda environment.

To get it going with conda you can use:
```
reticulate::conda_create(
    envname = "metaflow-env", 
    packages = c("metaflow", "pandas", "numpy")
)
reticulate::use_condaenv("metaflow-test", required = TRUE)
```
Note the `pandas` and `numpy` dependencies:  reticulate won't respect the `required = TRUE` if it doesn't detect `numpy` (and `pandas` is required for the R bindings)

[...]

PR #312, opened 2020-08-27
(R) add conda + virtualenv options to install function 

bcgalvin:
Addresses https://github.com/Netflix/metaflow/issues/281

Implements a more flexible approach for handling installation similar to the RStudio tensorflow packages and MLflow:
- creates a new conda/virtualenv if one doesn't currently exist, installs deps into it
- alternative python install can be set with `METAFLOW_PYTHON` environment variable
- correct python version forced by either env variable or `use_*` functions


R/R/install.R
```diff
@@ -0,0 +1,84 @@
...
+#' @seealso
+#' [reticulate: Using reticulate in an R Package](https://rstudio.github.io/reticulate/articles/package.html),
+#' [reticulate: Installing Python Packages](https://rstudio.github.io/reticulate/articles/python_packages.html)
+#' @examples
+#' \dontrun{
+#' # not run because it requires Python
+#' install_metaflow()
+#' }
+#' @export
+install_metaflow <- function(method = c("conda", "virtualenv"),
```
savingoyal, line 26: Can we rename this to `install` to ensure backwards compatibility?

R/R/install.R
```diff
@@ -0,0 +1,84 @@
...
+                             envname = "metaflow",
```
savingoyal, line 27: Can we set the default env name to `metaflow-r`?

R/R/zzz.R
```diff
@@ -1,22 +1,47 @@
...
-    error = function(e) {
-    }
-  )
+  metaflow_python <- Sys.getenv("METAFLOW_PYTHON", unset = NA)
+  if (!is.na(metaflow_python))
+    Sys.setenv(RETICULATE_PYTHON = metaflow_python)
+  
+  if (!ensure_metaflow()) {
+    packageStartupMessage("* Metaflow Python dependencies not found *\n",
+                          "    - Call `install_metaflow()` to install into a new conda or virtualenv\n",
```
savingoyal, line 25: Modify it to `metaflow::install()`?

R/DESCRIPTION
```diff
@@ -33,8 +33,10 @@ Collate:
     'flags.R'
     'flow.R'
     'metaflow_client.R'
-    'metaflow.R'
+    'metaflow-package.R'
```
savingoyal, line 36: Any specific reason to change the name? Also, for consistency - `metaflow_package` could be a better choice.

R/NAMESPACE
```diff
@@ -12,7 +12,7 @@ export(flow_client)
 export(gather_inputs)
 export(get_metadata)
 export(get_namespace)
-export(install)
+export(install_metaflow)
```
savingoyal, line 15: For backwards compatibility, let's keep `install`

R/NAMESPACE
```diff
@@ -36,7 +36,8 @@ export(step_client)
 export(task_client)
 export(test)
 export(version_info)
+import(reticulate)
 importFrom(R6,R6Class)
 importFrom(magrittr,"%>%")
-importFrom(reticulate,import_builtins)
-importFrom(reticulate,py_get_attr)
+importFrom(rstudioapi,hasFun)
+importFrom(rstudioapi,restartSession)
```
savingoyal, line 43: Can we avoid the dependency on `rstudioapi` for now? or place a soft dependency instead? Right now the code fails because `rstudioapi` is not listed as a metaflow dependency.

R/R/zzz.R
```diff
@@ -1,22 +1,47 @@
...
+                          "    - Set `METAFLOW_PYTHON` environment variable to use a custom metaflow installation.")
```
savingoyal, line 26: Can you add more documentation around what should be the value of `METAFLOW_PYTHON`?

R/R/install.R
bcgalvin: I think it may be worth it to switch since it's so early in the package's development but can you definitely overrule me on this!

`install` has a namespace issue with the popular devtools package and there is a strong consensus on the `install_` naming convention for wrapping python projects (I can't think of one that doesn't).

R/DESCRIPTION
bcgalvin: Nope not necessary, other than it's convention to either use `____-package.R` or `package.R`.

- https://cran.r-project.org/web/packages/roxygen2/vignettes/rd.html
- `usethis::use_package_doc` creates a file with the `____-package`

Sorry about these opinionated changes in this PR, was on autopilot for this one. It's totally out of scope of the business logic being changed here so I can leave these modifications out.

R/R/flow_client.R
```diff
@@ -3,8 +3,7 @@
 #' Instances of this class contain all runs related to a flow.
 #'
 #' @docType class
-#' @importFrom R6 R6Class
-#' @include metaflow.R
+#' @include metaflow-package.R
```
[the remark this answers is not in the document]
bcgalvin: https://github.com/Netflix/metaflow/pull/312#discussion_r478639874

R/R/install.R
savingoyal: Okay, let's go with `install_`

R/DESCRIPTION
savingoyal: Can we change it to `package.R`?

R/R/install.R
```diff
@@ -0,0 +1,90 @@
...
+#'
+#' This function wraps installation functions from [reticulate][reticulate::reticulate] to install the Python packages
+#' **metaflow** and it's Python dependencies.
+#'
+#' This package uses the [reticulate][reticulate::reticulate] package
+#' to make an interface with the [Metaflow](https://metaflow.org/)
+#' Python package.
+#'
+#' @param method `character`, indicates to use `"conda"` or `"virtualenv"`.
+#' @param envname `character`, name of environment into which to install.
```
bcgalvin, line 11: one remaining issue here is `envname`: if a user specifies a custom envname, onLoad() won't pick it up. There are a few ways to go about supporting this but not sure if adding the custom environment is really necessary since that can be handled via the `METAFLOW_PYTHON` variable. 

@savingoyal @jasonge27

R/NAMESPACE
```diff
@@ -36,7 +36,8 @@ export(step_client)
 export(task_client)
 export(test)
 export(version_info)
+import(reticulate)
 importFrom(R6,R6Class)
 importFrom(magrittr,"%>%")
-importFrom(reticulate,import_builtins)
-importFrom(reticulate,py_get_attr)
+importFrom(rstudioapi,hasFun)
```
jasonge27, line 42: Did we use `rstudioapi::hasFun` in this PR? Looks like I couldn't find the usage when searching across the repo.

R/NAMESPACE
jasonge27, line 43: same as above

savingoyal:
@bcgalvin Can you also fix the tests?

R/R/install.R
```diff
@@ -0,0 +1,89 @@
...
+  packages <- c(metaflow_pkg_version, "numpy", "pandas")
+
+  # create environment if not present
+  if (method == "conda" && !envname %in% conda_list()$name) {
+    conda_create(envname)
+  } else if (method == "virtualenv" && !envname %in% virtualenv_list()) {
+    virtualenv_create(envname)
+  }
+
+  py_install(
```
savingoyal, line 55: Does this function automatically infer whether to use `conda` or `pip` to install dependencies?

bcgalvin: Currently it uses the defaults for `conda_create`, which is `pip=FALSE`. Should probably pass `...`s here or (add pip as an argument) to make it more flexible.

savingoyal: Great!

savingoyal:
@bcgalvin Is the PR ready for final review or do you intend to make any more changes? I will go ahead and test the package on AWS if you don't plan on any more changes.

bcgalvin:
> @bcgalvin Is the PR ready for final review or do you intend to make any more changes? I will go ahead and test the package on AWS if you don't plan on any more changes.

I'm not planning any substantive changes, so all clear to review @savingoyal

R/R/install.R
```diff
@@ -0,0 +1,150 @@
...
+#' }
+#' @export
+install_metaflow <- function(method = c("conda", "virtualenv"),
+                             version = NULL,
+                             ...) {
+  envname <- pkg.env$envname
+
+  env_set <- check_environment(envname)
+  if (method == "conda" && env_set[["virtualenv"]]){
+    message("Installing Metaflow python backend using conda but it has been installed in virtualenv.")
```
savingoyal, line 32: Can we simplify the messaging here - 
`Can't install within a virtualenv using conda.`
`To install using virtualenv instead, use: install_metaflow(method='virtualenv').`

R/R/install.R
```diff
@@ -0,0 +1,150 @@
...
+    message("We do not allow python backend in both conda and virtualenv.")
+    message("1. To reinstall python backend in virtualenv: install_metaflow(method='virtualenv')")
+    message("2. To remove python backend: remove_metaflow()")
+    stop("Installing python backend in both conda and virtualenv is not allowed.")
+  }
+
+  if (method == "virtualenv" && env_set[["conda"]]){
+    message("Installing Metaflow python backend using virtualenv but it has been installed in conda.")
```
savingoyal, line 40: Can we simplify the messaging here - 
`Can't install within a conda environment using virtualenv.`
`To install using conda instead, use: install_metaflow(method='conda').`

R/R/install.R
```diff
@@ -0,0 +1,150 @@
...
+                             version = NULL,
+                             ...) {
+  envname <- pkg.env$envname
+
+  env_set <- check_environment(envname)
+  if (method == "conda" && env_set[["virtualenv"]]){
+    message("Installing Metaflow python backend using conda but it has been installed in virtualenv.")
+    message("We do not allow python backend in both conda and virtualenv.")
+    message("1. To reinstall python backend in virtualenv: install_metaflow(method='virtualenv')")
+    message("2. To remove python backend: remove_metaflow()")
```
savingoyal, line 35: I am not sure it follows immediately, in what scenario they are supposed to invoke `remove_metaflow()`. If it is an edge case, then we should just mention that in our docs instead and remove it from here.

R/R/install.R
```diff
@@ -0,0 +1,150 @@
...
+
+  packages <- c(metaflow_pkg_version, "numpy", "pandas")
+
+  # create environment if not present
+  if (method == "conda"){
+      conda <- tryCatch(reticulate::conda_binary(), 
+                        error = function(e) NULL)
+      have_conda <- !is.null(conda)
+      if (!have_conda) {
+        message("No conda was found in the system.")
```
savingoyal, line 71: `No conda installation was detected in the system.`

R/R/install.R
```diff
@@ -0,0 +1,150 @@
...
+        message("Would you like to download and install Miniconda?")
```
savingoyal, line 72: Delete L:72 and L:73

R/R/install.R
```diff
@@ -0,0 +1,150 @@
...
+#' [reticulate: Installing Python Packages](https://rstudio.github.io/reticulate/articles/python_packages.html)
+#' @examples
+#' \dontrun{
+#' # not run because it requires Python
+#' install_metaflow()
+#' }
+#' @export
+install_metaflow <- function(method = c("conda", "virtualenv"),
+                             version = NULL,
+                             ...) {
```
savingoyal, line 27: Can we have a `-y` equivalent option here so that metaflow can be installed programmatically?

R/R/install.R
```diff
@@ -0,0 +1,150 @@
...
+  envname <- pkg.env$envname
+
+  env_set <- check_environment(envname)
+  if (method == "conda" && env_set[["virtualenv"]]){
+    message("Installing Metaflow python backend using conda but it has been installed in virtualenv.")
+    message("We do not allow python backend in both conda and virtualenv.")
+    message("1. To reinstall python backend in virtualenv: install_metaflow(method='virtualenv')")
+    message("2. To remove python backend: remove_metaflow()")
+    stop("Installing python backend in both conda and virtualenv is not allowed.")
```
savingoyal, line 36: Delete L:36

R/R/install.R
```diff
@@ -0,0 +1,150 @@
...
+install_metaflow <- function(method = c("conda", "virtualenv"),
+                             version = NULL,
+                             ...) {
+  envname <- pkg.env$envname
+
+  env_set <- check_environment(envname)
+  if (method == "conda" && env_set[["virtualenv"]]){
+    message("Installing Metaflow python backend using conda but it has been installed in virtualenv.")
+    message("We do not allow python backend in both conda and virtualenv.")
+    message("1. To reinstall python backend in virtualenv: install_metaflow(method='virtualenv')")
```
savingoyal, line 34: L:33 and L:34 can be deleted in favor of the above comment.

R/R/install.R
```diff
@@ -0,0 +1,150 @@
...
+        message("Would you like to download and install Miniconda?")
+        message("Miniconda is an open source environment management system for Python.")
+        message("See https://docs.conda.io/en/latest/miniconda.html for more details.")
+        ans <- utils::menu(c("Yes", "No"), 
+                    title = "Would you like to install Miniconda?")
+        if (ans == 1) {
+          reticulate::install_miniconda()
+          conda <- tryCatch(reticulate::conda_binary("auto"), error = function(e) NULL)
+        } else {
+          stop("Conda environment installation failed (no conda binary found)\n",
```
savingoyal, line 81: `Metaflow installation failed (no conda binary found).`

R/R/install.R
savingoyal: Can be a follow up to this PR.

R/R/install.R
```diff
@@ -0,0 +1,150 @@
...
+  envname = pkg.env$envname
+
+  env_set <- check_environment(envname)
+
+  if (env_set[["conda"]]){
+    conda <- tryCatch(reticulate::conda_binary(), 
+                        error = function(e) NULL)
+    have_conda <- !is.null(conda)
+    if (!have_conda) {
+      message("Conda binary is not found. Skip metaflow removal from conda.")
```
savingoyal, line 128: `No conda installation found. Aborting Metaflow removal.`

R/R/install.R
```diff
@@ -0,0 +1,150 @@
...
+    } else {
+      message("A conda environment ", envname, " will be removed\n")
```
savingoyal, line 130: `Conda environment <> will be deleted.`

R/R/install.R
```diff
@@ -0,0 +1,150 @@
...
+      ans <- ifelse(prompt, utils::menu(c("No", "Yes"), title = "Proceed?"), 2)
+      if (ans == 1) stop("conda env removal is cancelled by user", call. = FALSE)
+      python <- reticulate::conda_remove(envname = envname)
+      message("\nRemoval complete. Please restart the current R session.\n\n")
+    }
+  } 
+  
+  if (env_set[["virtualenv"]]) {
+    message("A virtualenv environment ", envname, " will be removed\n")
```
savingoyal, line 139: `Virtualenv environment <> will be deleted.`

R/R/install.R
```diff
@@ -0,0 +1,150 @@
...
+  if (env_set[["conda"]]){
+    conda <- tryCatch(reticulate::conda_binary(), 
+                        error = function(e) NULL)
+    have_conda <- !is.null(conda)
+    if (!have_conda) {
+      message("Conda binary is not found. Skip metaflow removal from conda.")
+    } else {
+      message("A conda environment ", envname, " will be removed\n")
+      ans <- ifelse(prompt, utils::menu(c("No", "Yes"), title = "Proceed?"), 2)
+      if (ans == 1) stop("conda env removal is cancelled by user", call. = FALSE)
```
savingoyal, line 132: `Cancelled. Aborting Metaflow removal.`

R/R/install.R
```diff
@@ -0,0 +1,150 @@
...
+      python <- reticulate::conda_remove(envname = envname)
+      message("\nRemoval complete. Please restart the current R session.\n\n")
+    }
+  } 
+  
+  if (env_set[["virtualenv"]]) {
+    message("A virtualenv environment ", envname, " will be removed\n")
+    ans <- ifelse(prompt, utils::menu(c("No", "Yes"), title = "Proceed?"), 2)
+    if (ans == 1) stop("virtualenv removal is cancelled by user", call. = FALSE)
```
savingoyal, line 141: `Cancelled. Aborting Metaflow removal.`

R/R/install.R
```diff
@@ -0,0 +1,150 @@
...
+    python <- reticulate::virtualenv_remove(envname = envname, confirm = FALSE)
+    message("\nRemoval complete. Please restart the current R session.\n\n")
+  }
+
+  if (!env_set[["conda"]] && !env_set[["virtualenv"]]){
+    stop("Metaflow python environment ", envname,
```
savingoyal, line 147: `Nothing to remove.`

R/R/metaflow_client.R
```diff
@@ -93,7 +91,7 @@ metaflow_object <- R6::R6Class(
       private$created_at_ <- obj$created_at
       private$parent_ <- obj$parent$id
       private$pathspec_ <- obj$pathspec
-      private$tags_ <- import_builtins()$list(obj$tags)
+      private$tags_ <- reticulate::import_builtins()$list(obj$tags)
```
savingoyal, line 94: What's the rationale for introducing `reticulate::` here and elsewhere in lieu of NAMESPACE import?

R/R/run_client.R
```diff
@@ -16,7 +14,7 @@
 #' @section Usage:
 #' \preformatted{
 #' r <- run_client$new(flow, run_id)
-#' r <- new_client$new("HelloFlow/12")
+#' r <- run_client$new("HelloFlow/12")
```
savingoyal, line 17: Good catch!

R/R/zzz.R
```diff
@@ -1,22 +1,100 @@
...
+  }
+}
+
+print_metaflow_install_options <- function(){
+  packageStartupMessage(
+      "* Metaflow Python dependencies not found *\n",
+      "  Available options:\n",
+      "    - Call `install_metaflow()` to install into a new conda or virtualenv\n",
+      "    - Set `METAFLOW_PYTHON` environment variable to the path of your python executable.\n",
+      "      Note: Metaflow needs to be available in the environment specified by `METAFLOW_PYTHON`"
```
savingoyal, line 35: Not just Metaflow, we need pandas and numpy as well. An alternate messaging could be `- Set 'METAFLOW_PYTHON' env variable to the python executable that has metaflow, numpy, and pandas available as dependencies.`

savingoyal (changes requested):

R/R/install.R
```diff
@@ -0,0 +1,150 @@
...
+#' @seealso
+#' [reticulate: Using reticulate in an R Package](https://rstudio.github.io/reticulate/articles/package.html),
+#' [reticulate: Installing Python Packages](https://rstudio.github.io/reticulate/articles/python_packages.html)
+#' @examples
+#' \dontrun{
+#' # not run because it requires Python
+#' install_metaflow()
+#' }
+#' @export
+install_metaflow <- function(method = c("conda", "virtualenv"),
```
savingoyal, line 25: What is the expected behavior if `install_metaflow` is called multiple times?

savingoyal (changes requested):

R/R/install.R
jasonge27: we don't allow `install_metaflow` with both conda & virtualenv. If the users runs `install_metaflow` with either conda/virtualenv multiple times, we simply re-install numpy & pandas & metaflow in the same environment.

R/R/zzz.R
jasonge27: good catch. Will do

R/R/metaflow_client.R
jasonge27: We don't have import reticulate in NAMESPACE anymore. Only import we have is `importFrom(magrittr,"%>%")` in NAMESPACE. This is presumably better practice since it does not expose all the methods in dependency libraries. 

`reticualte::import_builtins` is also better for readability as it's clear we're using a method from `reticulate`

R/R/install.R
jasonge27: We can delete L:72, but L:73 should be there? L:73 and L:74 kind of should go together to provide info about Miniconda which we are asking to user to install

R/R/install.R
jasonge27: They can not install both conda and virtualenv. So we are asking them to remove virtualenv first if they want to install conda. I'll simplify the message.

R/R/install.R
jasonge27: Yes made a change to 
```
install_metaflow <- function(method = c("conda", "virtualenv"),
                             prompt = TRUE,
                             version = NULL,
                             ...) {
```
for `prompt = FALSE` we will just install without prompting for confirmation

R/R/install.R
```diff
@@ -0,0 +1,154 @@
...
+  packages <- c(metaflow_pkg_version, "numpy", "pandas")
+
+  # create environment if not present
+  if (method == "conda") {
+    conda <- tryCatch(reticulate::conda_binary(),
+      error = function(e) NULL
+    )
+    have_conda <- !is.null(conda)
+    if (!have_conda) {
+      message("No conda was found in the system.")
```
savingoyal, line 70: Minor nit - `No conda installation found.`

R/R/install.R
```diff
@@ -0,0 +1,154 @@
...
+      message("Miniconda is an open source environment management system for Python.")
```
savingoyal, line 71: For the sake of correctness - `Miniconda is an open-source package management and environment management system.` Conda is not limited to Python.

R/R/install.R
```diff
@@ -0,0 +1,154 @@
...
+                             prompt = TRUE,
+                             version = NULL,
+                             ...) {
+  envname <- pkg.env$envname
+
+  env_set <- check_environment(envname)
+  if (method == "conda" && env_set[["virtualenv"]]) {
+    message("A virtualenv environment is detected for Metaflow.")
+    message("Can't install within a virtualenv using conda.")
+    stop("Please remove existing virtualenv by `remove_metaflow()` and run the install again.")
```
savingoyal, line 36: I am trying to understand the rationale here for recommending the user to remove the virtualenv. My assumption is that the user is already inside a virtualenv because they might need access to some of the packages in that virtualenv through reticulate. But now through our messaging, we are indicating that the user should remove that virtualenv before they can proceed to install metaflow-r (while just executing `install_metaflow(method='virtualenv')` will get them the desired end state). 

Also, `remove_metaflow()` signals that executing this command will somehow remove metaflow from my system, but it instead removes the environment that might contain much more than just Metaflow. It might be a good idea to change the name of the function to say, `remove_metaflow_environment()` or something similar. If we go down that path, then let's modify the messages in that function to read - `No conda installation found. Aborting...` and `Cancelled...` instead of the current messages.

savingoyal (changes requested):

savingoyal (approved):

Issue #313, opened 2020-08-31
Error in match.arg(algo): 'arg' should be one of “md5”, “sha1”, “crc32”, “sha256”, “sha512”, “xxhash32”, “xxhash64”, “murmur32”


hatemr:
Running `metaflow::test()` using Metaflow in R gives this error. Any idea what's going on?

`> metaflow::test()`
Error in match.arg(algo) : 
  'arg' should be one of “md5”, “sha1”, “crc32”, “sha256”, “sha512”, “xxhash32”, “xxhash64”, “murmur32”

jasonge27:
Metaflow depends on `digest` library for some hashing functions -- can you update `digest` and give it a try? We will make it more robust in the next release. Thanks for reporting this problem!

savingoyal:
@hatemr If you force reinstall the package again, the issue should be fixed.

PR #314, opened 2020-08-31
Set minimum version of digest dependency to 0.6.19

mdneuzerling:
Addresses issue #313.

Version 0.6.19 (2019-05-20) of the `digest` package was the earliest to support the `spookyhash` algorithm.

### DIFF
--- a/R/R/step.R
+++ b/R/R/step.R
@@ -31,9 +31,12 @@ step <- function(flow, ..., step, r_function = NULL, foreach = NULL, join = FALS
   }
   if (!is.null(r_function)) {
     function_name <- as.character(substitute(r_function))
-    function_hash <- digest::digest(deparse(r_function), algo = "sha256")
-    trunc_function_hash <- substr(function_hash, 1, 16)
-    if (length(function_name) > 1) { # likely an anonymous function
+    # If r_function is anonymous then function_name will be a vector of its
+    # components. In this case we give the function a pseudonym prefixed by the
+    # step name and suffixed with a hash of the function.
+    if (length(function_name) > 1) { 
+      function_hash <- digest::digest(deparse(r_function), algo = "sha256")
+      trunc_function_hash <- substr(function_hash, 1, 16)
       function_name <- paste(step, "function", trunc_function_hash, sep = "_")
     }
     body(r_function) <- wrap_function(r_function)
--- a/R/R/step.R
+++ b/R/R/step.R
@@ -31,9 +31,10 @@ step <- function(flow, ..., step, r_function = NULL, foreach = NULL, join = FALS
   }
   if (!is.null(r_function)) {
     function_name <- as.character(substitute(r_function))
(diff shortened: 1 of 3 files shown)


jasonge27:
@mdneuzerling Thanks for the quick fix!! We want to introduce as less dependency requirements as possible for our users. Can you change the hashing algorithm to `sha-256` and force `digest` version to be >= 0.4.0 (2009-09-24)?

mdneuzerling:
Done, @jasonge27. I also truncated the hash to make things a bit easier to read, since I don't think we need to worry too much about collisions here. Sorry for the trouble these anonymous functions have caused!

jasonge27:
Looks good to me! It's definitely nicer UX. Dependency management is always hard :-)  @savingoyal Can you take another look? Thank you both!

wlandau:
> I also truncated the hash to make things a bit easier to read, since I don't think we need to worry too much about collisions here.

What about a 64-bit hash without truncation?

mdneuzerling:
> > I also truncated the hash to make things a bit easier to read, since I don't think we need to worry too much about collisions here.
> 
> What about a 64-bit hash without truncation?

If there's a small hash that older versions of `digest` support, then that would be fine to use.

A collision at all is very unlikely here because the function name is prefixed with the step name and suffixed with the truncated hash. So the user would have to define two steps with the same name **and** R functions that whose hashes collide.

R/R/step.R
```diff
@@ -31,9 +31,10 @@ step <- function(flow, ..., step, r_function = NULL, foreach = NULL, join = FALS
   }
   if (!is.null(r_function)) {
     function_name <- as.character(substitute(r_function))
-    function_hash <- digest::digest(deparse(r_function), algo = "spookyhash")
+    function_hash <- digest::digest(deparse(r_function), algo = "sha256")
+    trunc_function_hash <- substr(function_hash, 1, 16)
     if (length(function_name) > 1) { # likely an anonymous function
-      function_name <- paste(step, "function", function_hash, sep = "_")
+      function_name <- paste(step, "function", trunc_function_hash, sep = "_")
```
savingoyal, line 37: @mdneuzerling Can we move L34:35 within the `if` clause?

wlandau:
Looks like `digest` has supported xxhash64 [since 2014](https://github.com/eddelbuettel/digest/blob/35ad0b7441010bec1dcc1a26ea0699f982634274/ChangeLog#L807-L821). It's also a lot faster, and people have studied its collision properties.

R/R/step.R
jasonge27: also can we clarify the comments a bit? Something like -- `an anonymous function will be parsed into a list of function arguments and function body, while non-anonymous function is only parsed into a function name as a length-one list`

mdneuzerling:
Done. Please let me know if there are any other concerns.

Thanks for all of this, by the way. I can't speak for other R users, but when I wrote my first flow I defaulted to using anonymous functions for everything. It seemed pretty natural to me.

PR #316, opened 2020-09-02
Metaflow Cache [WIP]

tuulos:
A new cache backend to replace the old `client.s3cache`. Work in progress!

### DIFF
--- a/metaflow/client/cache/cache_client.py
+++ b/metaflow/client/cache/cache_client.py
@@ -1,18 +1,30 @@
 import os
+import math
 import time
+import json
 
 from .cache_server import server_request, subprocess_cmd_and_env
-from .cache_store import object_path, is_safely_readable
+from .cache_store import object_path, stream_path, is_safely_readable
+
+FOREVER = 60 * 60 * 24 * 3650
 
 PING_FREQUENCY = 1
+BUSY_LOOP_FREQUENCY = 0.2
 
 class CacheServerUnreachable(Exception):
+    pass
+
+class CacheStreamMissing(Exception):
+    pass
+
+class CacheStreamTimeout(Exception):
     pass
 
 class CacheFuture(object):
 
     def __init__(self, keys, stream_key, client, action_cls, root):
         self.stream_key = stream_key
+        self.stream_path = stream_path(root, stream_key) if stream_key else None
         self.action = action_cls
         self.client = client
         self.keys = keys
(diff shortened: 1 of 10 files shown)


metaflow/client/cache/cache_store.py
```diff
@@ -0,0 +1,260 @@
...
+        dirr = os.path.dirname(path)
+        if not os.path.isdir(dirr):
+            try:
+                makedirs(dirr)
+            except Exception as ex:
+                self.warn(ex, "Could not create dir: %s" % path)
+
+    def open_tempdir(self, token, action_name, stream_key):
+        self._gc_objects()
+        if self.total_size > self.max_size:
```
romain-intel, line 166: So to be clear, the cache can go higher than that size because the objects are marked for deletion but not deleted, correct? Further, an action can just fill up this space easily before commit and it wouldn't count.

metaflow/client/cache/cache_store.py
```diff
@@ -0,0 +1,260 @@
...
+        # will invalidate the corresponding symlink that points at
+        # a file in the temp directory. The dangling symlink will get
+        # removed in close_tempdir.
+        #
+        # The cache client is expected to fall back to the object_path
+        # when it detects an invalid symlink.
+        for key in keys + ([stream_key] if stream_key else []):
+            src = os.path.join(tempdir, key_filename(key))
+            if os.path.exists(src):
+                dst = object_path(self.root, key)
```
romain-intel, line 238: Is there no risk of collision here?

metaflow/client/cache/cache_worker.py
```diff
@@ -0,0 +1,63 @@
...
+        if req['stream_key']:
+            stream = open(req['stream_key'], 'a', buffering=1)
+            def stream_output(obj):
+                stream.write(json.dumps(obj) + '\n')
+        else:
+            stream_output = None
+
+        # prepare keys
+        keys = list(req['keys'])
+        ex_keys = dict(best_effort_read(req['existing_keys'].items()))
```
romain-intel, line 41: This seems potentially costly since you assume that all keys will need to be read by the action (you read them already). Also, the keys could be garbage collected so is the contract that this is the value at the time but you shouldn't assume that they will be around when you are done?

talebzeghmi:
What will this cache?
Will this design be useful to cache across steps run on different containers (eg: on Batch or Kubernetes)?

thanks!

tuulos:
@talebzeghmi probably not. This cache is meant for executing S3 operations asynchronously in the background. The main use case is the upcoming GUI for Metaflow.

metaflow/client/cache/cache_server.py
```diff
@@ -0,0 +1,307 @@
...
+        'keys': keys,
+        'stream_key': stream_key,
+        'message': message,
+        'idempotency_token': token,
+        'disposable_keys': disposable_keys
+    }
+
+def echo(msg):
+    now = datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%S.%f')
+    sys.stderr.write('CACHE [%s] %s\n' % (now, msg))
```
saikonen, line 61: is this a deliberate choice to always write output to stderr? f.ex. the CacheStore uses this to push out initialization information, so stderr seems like an odd place.

On a separate note: when running the async cache client, any of the messages produced by the subprocess echoes are not visible on-time, but rather only after an exception is encountered. For example the cache store initialization messages are only printed after a worker fails with an action. This might either be due to the use of stderr, or some asyncio shenanigans.

metaflow/client/cache/cache_client.py
```diff
@@ -0,0 +1,230 @@
...
+            msg, keys, stream_key, disposable_keys =\
+                 cls.format_request(*args, **kwargs)
+            future = CacheFuture(keys, stream_key, self, cls, self._root)
+            if future.is_ready():
+                # cache hit
+                req = None
+            else:
+                # cache miss
+                action_spec = '%s.%s' % (cls.__module__, cls.__name__)
+                req = self._send('action',
```
saikonen, line 173: the fire-and-forget nature of executing actions is problematic for error catching at the caller level. It would be preferable that something like this could be performed when calling actions:

```
try:
  await cache.SomeAction(params)
except:
  # any exception worth catching, CacheFullException, CacheServerUnreachable, etc.
  # Including the ability to raise an exception from within an actions execute()
```

metaflow/client/cache/cache_server.py
```diff
@@ -0,0 +1,307 @@
...
+        self.request = request
+        self.prio = request['priority']
+        self.filestore = filestore
+        self.proc = None
+        self.tempdir = self.filestore.open_tempdir(request['idempotency_token'],
+                                                   request['action'],
+                                                   request['stream_key'])
+        if self.tempdir is None:
+            self.echo("Store couldn't create a temp directory. "\
+                      "WORKER NOT STARTED.")
```
saikonen, line 123: Referring to my other comment about worker functionality depending on a tempdir; shouldn't this case be an explicit exception?

PR #331, opened 2020-09-12
Moved Metadata to plugins. Also moved more defaults to metaflow_config

romain-intel:
Metadata providers are now plugins. We include the two that were already there
(local and service) but other metadata providers can now be easily added.

This replaces PR #236.


metaflow/plugins/__init__.py
```diff
@@ -12,6 +12,8 @@ def get_plugin_cli(self):
         FLOW_DECORATORS=[],
         STEP_DECORATORS=[],
         ENVIRONMENTS=[],
+        DATA_PROVIDERS={},
```
seeravikiran, line 15: What are `DATA_PROVIDERS`? It seems unused to me but if that's incorrect - could you give a pointer where this is used?

metaflow/cli.py
```diff
@@ -718,12 +717,12 @@ def version(obj):
               show_default=True,
               help='Measure code coverage using coverage.py.')
 @click.option('--event-logger',
-              default='nullSidecarLogger',
+              default=DEFAULT_EVENTLOGGER,
```
seeravikiran, line 720: Should this be `DEFAULT_EVENT_LOGGER` considering there is a hyphen in the name?
Similar comment for `DEFAULT_PACKAGE_SUFFIXES`? I'm assuming the consistency could help but I feel weakly about this comment, so let me know if you disagree.

metaflow/plugins/__init__.py
savingoyal: Yes, we don't have the notion of `DATA_PROVIDERS` anymore. Let's chuck it.

savingoyal (approved):
LGTM! Let me verify the PR manually, before merging it in.

tuulos:
the code LGTM - pending testing

PR #332, opened 2020-09-12
Move the authentication (Boto S3 here) to plugins

romain-intel:
This move will allow other authentication methods to be defined more
easily.

Replaces #239.


metaflow/datastore/util/s3util.py
```diff
@@ -4,13 +4,18 @@
 import sys
 
 from metaflow.exception import MetaflowException
-from metaflow.metaflow_config import get_authenticated_boto3_client, S3_ENDPOINT_URL, S3_VERIFY_CERTIFICATE
+from metaflow.metaflow_config import DEFAULT_AWS_CLIENT_PROVIDER, S3_ENDPOINT_URL, S3_VERIFY_CERTIFICATE
 from botocore.exceptions import ClientError
```
seeravikiran, line 8: You could remove this import here. Yes?

PR #333, opened 2020-09-15
add heartbeat at task and run level

ferras:
add sidecar for heartbeat on task and run

testing still required

Note the corresponding change needs to be made and deployed in the metadata-service (see https://github.com/Netflix/metaflow-service/pull/27) and the initialization of the sidecar should be gated based of what version of the metadata-service is deployed (old versions will not have the new heartbeat endpoints)


metaflow/plugins/metadata/service.py
```diff
@@ -19,9 +28,15 @@ def __init__(self, msg, http_code=None, body=None):
 
 class ServiceMetadataProvider(MetadataProvider):
     TYPE = 'service'
+    HB_URL_KEY = 'hb_url'
 
     def __init__(self, environment, flow, event_logger, monitor):
         super(ServiceMetadataProvider, self).__init__(environment, flow, event_logger, monitor)
+        self.mli_url_task_template = METADATA_SERVICE_URL + \
```
savingoyal, line 35: Can we use `urlparse` to construct the URL so that trailing slashes in `METADATA_SERVICE_URL` are appropriately taken care of?

metaflow/plugins/metadata/service.py
```diff
@@ -73,7 +88,55 @@ def register_task_id(self,
                            sys_tags=sys_tags)
         finally:
             self._register_code_package_metadata(run_id, step_name, task_id)
-        
+
+    def _start_heartbeat(self, heartbeat_type, flow_id, run_id, step_name=None, task_id=None):
+        if self._already_started():
+            raise Exception("heartbeat already started")
+        # start sidecar
+        if self.version() == 'local' or self.version() is None or \
```
savingoyal, line 96: Do we expect the version to ever be 'local' here?

metaflow/sidecar.py
```diff
@@ -56,13 +56,14 @@ def start(self):
 
         if (self.__worker_type is not None and \
                 self.__worker_type.startswith(NULL_SIDECAR_PREFIX)) or \
-                platform.system() == 'Darwin':
-            self.__poller = NullPoller()
+                (platform.system() == 'Darwin' and sys.version_info < (3, 0)):
+                # if on darwin and running python 2 disable sidecars
+                # there is a bug with importing poll from select in some cases
+                self.__poller = NullPoller()
```
savingoyal, line 62: Is it feasible to describe the bug - SO link or something here?

savingoyal (changes requested):
Some minor changes @ferras. I am yet to execute this PR but overall looks great!

metaflow/sidecar.py
ferras: don't have a good stackoverflow or github issues link but the error is:
```    
from select import poll
ImportError: cannot import name poll
```
I feel the comment is clear enough about where the issue is and since it only effects python2 on darwin i don't see a need to add anything here

metaflow/plugins/metadata/service.py
ferras: used urljoin from urlparse
note that it requires the user specify the protocol (http vs https) otherwise it won't work
```
>>> urljoin('http://www.example.com:8080//', '/flows/{flow_id}/runs/{run_number}/steps/{step_name}/tasks/{task_id}/heartbeat')
'http://www.example.com:8080/flows/{flow_id}/runs/{run_number}/steps/{step_name}/tasks/{task_id}/heartbeat'
>>> urljoin('www.example.com:8080//', '/flows/{flow_id}/runs/{run_number}/steps/{step_name}/tasks/{task_id}/heartbeat')
'/flows/{flow_id}/runs/{run_number}/steps/{step_name}/tasks/{task_id}/heartbeat'
```

metaflow/plugins/metadata/service.py
ferras: i guess it never should go to this provide if running in local mode

metaflow/plugins/metadata/service.py
```diff
@@ -73,7 +94,55 @@ def register_task_id(self,
...
+                              run_id,
+                              step_name,
+                              task_id)
+
+    def _already_started(self):
+        return self.sidecar_process is not None
+
+    def stop_heartbeat(self):
+        msg = Message(MessageTypes.SHUTDOWN, None)
+        self.sidecar_process.msg_handler(msg)
```
romain-intel, line 144: Should never be restarted but should we set sidecar_process to None at this time?

metaflow/sidecar_messages.py
```diff
@@ -22,7 +22,11 @@ def serialize(self):
 
 
 def deserialize(json_msg):
+    dict = {
+        1: MessageTypes.SHUTDOWN,
+        2: MessageTypes.LOG_EVENT
+    }
     parsed_json_msg = json.loads(json_msg)
-    return Message(MessageTypes(parsed_json_msg['msg_type']),
+    return Message(dict[parsed_json_msg['msg_type']],
```
romain-intel, line 30: dict is not a great name :).

metaflow/task.py
```diff
@@ -318,6 +318,8 @@ def run_step(self,
         })
         logger = self.event_logger
         start = time.time()
+        self.metadata.start_task_heartbeat(self.flow.name, run_id, step_name,
```
romain-intel, line 321: Nit: why here and not with the other side cars (logger below)?

romain-intel (commented):
A few changes/comments but LGTM overall. Once these are addressed, I think we can merge. It should be applicable to both convergence and master branch which is great :).

metaflow/plugins/metadata/service.py
```diff
@@ -2,10 +2,24 @@
 import requests
 import time
 
+from distutils.version import LooseVersion
+
 from metaflow.exception import MetaflowException
 from metaflow.metaflow_config import METADATA_SERVICE_NUM_RETRIES, METADATA_SERVICE_HEADERS, \
     METADATA_SERVICE_URL
+<<<<<<< HEAD:metaflow/metadata/service.py
```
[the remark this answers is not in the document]
ferras: i think you commented on an outdated version of the code. this was from 3 commits ago and has since been resolved

savingoyal:
@ferras This PR doesn't handle attempt-level heartbeats. Do we want to consider introducing attempt-level heartbeats - it might simplify the UI implementation significantly.

metaflow/metadata/heartbeat.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def heartbeat(self):
+        if self.hb_url is not None:
+            response = \
+                requests.post(url=self.hb_url, data="{}", headers=self.headers)
+            return json.loads(response.json()).get('wait_time_in_seconds')
+
+    def shutdown(self):
+        # attempts sending on last heartbeat
+        self.heartbeat()
```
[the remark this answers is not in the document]
ferras: what would the purpose of try/except be? This is something that is running in a separate process (sidecar) that we are trying to shutdown. In either case we would just want the process to terminate

metaflow/plugins/metadata/service.py
```diff
@@ -73,7 +94,55 @@ def register_task_id(self,
                            sys_tags=sys_tags)
         finally:
             self._register_code_package_metadata(run_id, step_name, task_id)
-        
+
+    def _start_heartbeat(self, heartbeat_type, flow_id, run_id, step_name=None, task_id=None):
+        if self._already_started():
+            raise Exception("heartbeat already started")
```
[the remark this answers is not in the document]
ferras: With the way metaflow is designed it's not really possible for the same ServiceMetadataProvider object to be shared by both a run and a task since they already execute in different sub-processes. We can avoid raising an exception but that would be a departure from the internal implementation

metaflow/task.py
ferras: it's how we do it internally. IIRC we didn't want the heartbeat sidecar within the try/except/finally block because in case of failure we wanted different behavior. But given your previous comment here:
https://github.com/Netflix/metaflow/pull/333#discussion_r625990515

If we decide to start treating all heartbeat failures as soft failures then that's really no longer a concern and we can move it 2 lines down right above the logger

metaflow/plugins/metadata/service.py
ferras: I don't think setting it to None would be a good idea. Calling shutdown will send a message that will terminate the sidecar subprocess which will prevent it from sending more heartbeats. If we set sidecar_process to know then it looks like the sidecar never started and may cause confusion

metaflow/sidecar_messages.py
ferras: haha goodpoint. That was actually just meant to be temporary. I replaced MessageTypes with IntEnum so the dict is no longer needed

metaflow/plugins/metadata/service.py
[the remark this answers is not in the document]
romain-intel: Yes, I agree that it is is not possible (and shouldn't be). The point I was making is that this is not really documented anywhere and if you just look at the code as a casual observer, you would be wondering why there are two places to call start and an exception if you call it twice :). For the exception part, my point was that this is an optional feature (sidecar) and can kill your main process for something that you really can't do anything about.

metaflow/plugins/metadata/service.py
romain-intel: so once it is shutdown, if by mistake another message is sent to it, it will just be ignored?

metaflow/metadata/heartbeat.py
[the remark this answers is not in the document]
romain-intel: Ah, nvm, I was under the (wrong) impression that this code was called from the user process but it is only called from process_message which is only called in the last sub-process. As a related point though, calling this has no effect because the worker loop doesn't call `process_message` on a shutdown. I don't know if that is what your TODO points to.

metaflow/plugins/metadata/service.py
ferras: the message will just get blackholed since that PIPE is now broken, essentially the message will have no effect

metaflow/plugins/metadata/service.py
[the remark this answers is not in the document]
ferras: I added a few comments to hopefully clarify things. We don't have any developer documentation as far as I know in OSS so in code comments will have to do.

Re: the exception: I totally get your point and we went back and forth on this in the original implementation a couple years ago. The reasoning behind throwing a hard error here is because logically in the code this situation should never happen and if it did it means data in the metadata service could become corrupted and we want to terminate the run and fail loudly. If the heartbeat fails to start or dies part way through the execution of the main process it's considered a soft failure and we opt to fail silently.

There are arguments to be made in either direction but that is the call we took back in 2019. We can revisit that now but we probably need to think through the failure scenarios again.

ferras:
> @ferras This PR doesn't handle attempt-level heartbeats. Do we want to consider introducing attempt-level heartbeats - it might simplify the UI implementation significantly.

as the name implies this is only meant to handle task and run level heartbeats :) 

Regarding the concept of attempt-level heartbeats I don't think it really makes sense to have them on an attempt level. The purpose of the heartbeat is to know if a task is currently running. If an attempt of a task failed and we were able to start a new attempt then we should use 'attempt-done' that was written to the metadata table for the purposes of rendering the UI. I don't think adding different heartbeats at the "attempt" level will simplify the UI implementation in that regard. Much of the complexity there actually comes from the need of handling legacy runs which are missing the 'attempt-done' metadata that was added in November of 2020.

If you do decide to implement a heartbeat at the attempt level, that will require changes to the data model in the underlying DB as well as the metaflow-service api which would need to happen before any changes within metaflow core

savingoyal (approved):
LGTM! Ship it! 🚢

PR #341, opened 2020-09-30
fixing a bug for parsing --authorize in R SFN

jasonge27:
Fixing a bug where we treated `--authorize` as an option without argument and failed to parse the token argument.


savingoyal (approved):
LGTM!

PR #343, opened 2020-10-05
fix an issue with aws region check in dynamodb client

jasonge27:
Replace the `curl` function call with requests in case `curl` does not exist in a custom image.

### DIFF
--- a/metaflow/plugins/aws/step_functions/dynamo_db_client.py
+++ b/metaflow/plugins/aws/step_functions/dynamo_db_client.py
@@ -1,5 +1,6 @@
 import json
 import os
+import requests
 
 from metaflow.metaflow_config import SFN_DYNAMO_DB_TABLE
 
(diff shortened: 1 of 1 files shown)


metaflow/plugins/aws/step_functions/dynamo_db_client.py
```diff
@@ -1,5 +1,6 @@
 import json
 import os
```
savingoyal, line 2: this import can be done away with.

metaflow/plugins/aws/step_functions/dynamo_db_client.py
```diff
@@ -65,6 +66,16 @@ def get_parent_task_ids_for_foreach_join(self,
         return response['Item']['parent_task_ids_for_foreach_join']['SS']
 
     def _get_instance_region(self):
-        return os.popen(
-            'curl -s http://169.254.169.254/latest/meta-data/placement/availability-zone/'
-            ).read()[:-1]
\ No newline at end of file
+        try:
+            r = requests.post(
```
savingoyal, line 70: We can do a simple `requests.get(...)` as well.

metaflow/plugins/aws/step_functions/dynamo_db_client.py
```diff
@@ -65,6 +66,16 @@ def get_parent_task_ids_for_foreach_join(self,
...
+                url = "http://169.254.169.254/latest/meta-data/placement/availability-zone/"
+            )
+            if r.status_code != 200:
+                raise RuntimeError("Failed to query AWS region from " +
+                        "http://169.254.169.254/latest/meta-data/placement/availability-zone/\n" +
```
savingoyal, line 75: `("Failed to query instance metadata. Url [%s], Error code [%s]" % (url, r.status_code)` would be cleaner.

metaflow/plugins/aws/step_functions/dynamo_db_client.py
```diff
@@ -65,6 +66,16 @@ def get_parent_task_ids_for_foreach_join(self,
...
+                        "Error code: " + str(r.status_code))
+            return r.text[:-1]
+        except Exception as e:
```
savingoyal, line 78: We might not need this block. All uncaught exceptions are caught in the task runtime, printed and re-raised appropriately.

savingoyal (changes requested):

metaflow/plugins/aws/step_functions/dynamo_db_client.py
savingoyal: @jasonge27 Let's address this comment and then you should be good to merge.

savingoyal (approved):

PR #346, opened 2020-10-08
Fix issue with IncludeFile caused in 2.1.0

romain-intel:
In some cases, overriding the default IncludeFile value would cause a crash.
This fixes this issue and adds tests to test this condition

### DIFF
--- a/test/core/metaflow_test/formatter.py
+++ b/test/core/metaflow_test/formatter.py
@@ -78,7 +78,7 @@ class FlowFormatter(object):
             tags.extend(tag.split('(')[0] for tag in step.tags)
 
         yield 0, '# -*- coding: utf-8 -*-'
-        yield 0, 'from metaflow import FlowSpec, step, Parameter, JSONType'
+        yield 0, 'from metaflow import FlowSpec, step, Parameter, IncludeFile, JSONType'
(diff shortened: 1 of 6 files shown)


test/core/metaflow_test/formatter.py
```diff
@@ -95,6 +95,10 @@ def _flow_lines(self):
             kwargs = ['%s=%s' % (k, v) for k, v in parameter.items()]
             yield 1, '%s = Parameter("%s", %s)' % (var, var, ','.join(kwargs))
 
+        for var, include in self.test.INCLUDE_FILES.items():
```
savingoyal, line 98: I don't think this will work. Not every test has `INCLUDE_FILES`

savingoyal (changes requested):

test/core/metaflow_test/formatter.py
romain-intel: gah, forgot the defaults. Adding

romain-intel:
LMK if this works for you now. All tests seem to pass with the small defaults fix so hopefully this solves the issue.

savingoyal (approved):
Looks great! Thanks for making the change!

PR #351, opened 2020-10-15
New features for the S3 datatools

romain-intel:
- support for range queries (in get and get_many)
  - support for content-type and user metadata in put, put_many and put_files
    (metadata can also be retrieved using any of the get calls)
  - support for info and info_many to retrieve information about a file without
    fetching it.


metaflow/datatools/s3.py
```diff
@@ -562,17 +836,32 @@ def put_files(self, key_paths, overwrite=True):
...
             overwrite: (optional) overwrites the key with obj, if it exists
 
         Returns:
             a list of (key, S3 URL) tuples corresponding to the files sent.
         """
         def _check():
-            for key, path in key_paths:
+            for key_path in key_paths:
+                key = getattr(key_path, 'key', key_path[0])
+                path = getattr(key_path, 'value', key_path[1])
```
hajapy, line 853: I think using value instead of path here may be slightly confusing, how about: 

```suggestion
                path = getattr(key_path, 'path', key_path[1])
```

metaflow/datatools/s3.py
```diff
@@ -161,6 +190,39 @@ def size(self):
         """
         return self._size
 
+    @property
+    def has_info(self):
```
tuulos, line 194: do we need this? Isn't it sufficient for `content_type` and `range_info` to return `None` if the info is not available

metaflow/datatools/s3.py
```diff
@@ -382,100 +467,261 @@ def get(self, key=None, return_missing=False):
...
         Args:
-            keys: (required) a list of suffixes identifying the objects.
+            keys: (required) a list of suffixes identifying the objects. Each
+                  item in the list can also be an object containing the properties
+                  `key`, `offset` and `length to specify a range query.
+                  `S3GetObject` is such an object.
             return_missing: (optional, default False) if set to True, do
                             not raise an exception for a missing key but
                             return it as an S3Object with .exists == False.
+            return_info: (optional, default False) if set to True, fetch the
```
tuulos, line 627: don't we get `content-type` and headers (i.e. user metadata) for free on the `s3op` side? If so, can't we just return info always by default?

metaflow/datatools/s3.py
romain-intel: Yes, could remove. I was trying to provide an easy way to know if the info was loaded (as opposed to being loaded but empty). It does the check you mention; I wanted to avoid users doing things like `if self.metadata` which can't really distinguish between (a) there is metadata but it wasn't loaded and (b) metadata was loaded and there is nothing to see.

metaflow/datatools/s3.py
romain-intel: I agree this is a bit confusing. I was trying to reuse the S3PutObject in all put calls. Let me see if I can come up with something a bit nicer.

metaflow/datatools/s3.py
romain-intel: I was trying to be consistent between the calls (get vs get_many). It is also not *entirely* free (close to free but not totally) due to the additional file mumbo jumbo to get the data back out. I am fine making it true by default but would prefer to keep consistency (so making it true by default for the get too which does imply an additional call).

metaflow/datatools/s3.py
```diff
@@ -527,14 +783,28 @@ def put_many(self, key_objs, overwrite=True):
...
+                       'value' are required but others are optional.
             overwrite: (optional) overwrites the key with obj, if it exists
 
         Returns:
             a list of (key, S3 URL) tuples corresponding to the files sent.
         """
         def _store():
-            for key, obj in key_objs:
+            for key_obj in key_objs:
+                key = getattr(key_obj, 'key', key_obj[0])
```
tuulos, line 798: you can't do `key_obj[0]` on an arbitrary object. Instead of `getattr`, just check `isinstance(key_obj, tuple)` and access `.key` and `value` directly in non-tuple objects

metaflow/datatools/s3.py
```diff
@@ -562,17 +832,32 @@ def put_files(self, key_paths, overwrite=True):
...
+                       'value' should point to the file path
             overwrite: (optional) overwrites the key with obj, if it exists
 
         Returns:
             a list of (key, S3 URL) tuples corresponding to the files sent.
         """
         def _check():
-            for key, path in key_paths:
+            for key_path in key_paths:
+                key = getattr(key_path, 'key', key_path[0])
```
tuulos, line 848: see my comment above about `key_obj[0]` - just check tupleness here.

metaflow/datatools/s3.py
tuulos: yeah, @romain-intel we could add `.path` in `S3PutObject`

metaflow/datatools/s3.py
```diff
@@ -37,6 +39,19 @@
...
     boto_found = False
 
+def ensure_unicode(x):
+    return None if x is None else to_unicode(x)
+
+S3GetObject = namedtuple_with_defaults(
+    'S3GetObject', 'key offset length')
+
+S3PutObject = namedtuple_with_defaults(
+    'S3PutObject', 'key value content_type metadata')
```
tuulos, line 49: let's add `path` here too so the fields make more sense for `put_files`

metaflow/datatools/s3.py
```diff
@@ -609,35 +894,43 @@ def _one_boto_op(self, op, url):
...
+                                          "Error: %s" % (prefixes_and_ranges[0], stderr))
             else:
                 for line in stdout.splitlines():
                     yield tuple(map(url_unquote, line.strip(b'\n').split(b' ')))
 
-    def _put_many_files(self, url_files, overwrite):
-        url_files = list(url_files)
+    def _put_many_files(self, url_info, overwrite):
+        url_info = list(url_info)
+        def _update_and_return(d, d1):
```
tuulos, line 922: `_update_and_return` is a bit funky construct - making stateful updates etc. Instead, you can replace the line above, `url_info = list(url_info)` with
```
url_info = [dict(chain(('local', os.path.realpath(local)), ('url': url)), info.items()))) for local, url, info in url_info]
```

romain-intel: ooh, I like that better. I'll change it.

metaflow/datatools/s3.py
tuulos: could `.metadata` return `{}` for loaded-but-empty and `None` for not loaded?

metaflow/datatools/s3op.py
```diff
@@ -73,17 +92,73 @@ def worker(queue, mode):
...
+                        # needed
+                        resp = s3.get_object(
+                            Bucket=url.bucket,
+                            Key=url.path,
+                            Range=url.range)
+                        code = str(resp['ResponseMetadata']['HTTPStatusCode'])
+                        if code[0] == '2':
+                            tmp.write(resp['Body'].read())
+                        else:
+                            tmp.close()
```
tuulos, line 132: aren't `tmp.close()` and `os.unlink` redundant given `except` below?

metaflow/datatools/s3op.py
```diff
@@ -73,17 +92,73 @@ def worker(queue, mode):
             if url is None:
                 break
 
-            if mode == 'download':
+            if mode == 'info':
```
tuulos, line 95: remind me why we need this `mode == 'info'` in the worker since we have `get_info` below?

metaflow/datatools/s3.py
romain-intel: Absolutely. It actually does that (I hope anyways). My point was just that both {} and None evaluate to False so the if expression is a bit of a trap but happy to remove it.

metaflow/datatools/s3op.py
romain-intel: Absolutely; will remove.

metaflow/datatools/s3op.py
romain-intel: Apologies, that was supposed to be in the other PR (the one about moving the get_info to the worker). I'll remove it from this one (it's not used here at all). I think I had started working on it and forgot to remove it from this PR.

metaflow/datatools/s3.py
romain-intel: Done

metaflow/datatools/s3.py
romain-intel: Done

metaflow/datatools/s3.py
romain-intel: Done

metaflow/datatools/s3.py
romain-intel: Done

metaflow/datatools/s3.py
romain-intel: Done

metaflow/datatools/s3.py
romain-intel: Done

metaflow/datatools/s3op.py
romain-intel: Done

metaflow/datatools/s3op.py
romain-intel: Done

metaflow/datatools/s3op.py
```diff
@@ -35,7 +36,25 @@
...
+else:
+    from collections import namedtuple
+    def namedtuple_with_defaults(typename, field_names, defaults=()):
+        T = namedtuple(typename, field_names)
+        T.__new__.__defaults__ = (None,) * len(T._fields)
+        prototype = T(*defaults)
+        T.__new__.__defaults__ = tuple(prototype)
+        return T
+
+S3Url = namedtuple_with_defaults(
```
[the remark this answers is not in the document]
romain-intel: Done.

metaflow/datatools/s3.py
romain-intel: No further comment here so leaving as is and resolving. Re-open if you would like to change.

PR #356, opened 2020-10-20
Dereference symlinks when packaging

jens-hummelshoej-tri:
As a user of metaflow, I would like to share external libraries among flows in different folders without code duplication. 

symlinks provide a solution but not when packaging and running --with batch, because the TarFile are keeping the symlinks as such instead of dereferencing when packaging.

### DIFF
--- a/metaflow/package.py
+++ b/metaflow/package.py
@@ -96,7 +96,7 @@ class MetaflowPackage(object):
             return tarinfo
 
         buf = BytesIO()
-        with tarfile.TarFile(fileobj=buf, mode='w') as tar:
+        with tarfile.TarFile(fileobj=buf, mode='w', dereference=True) as tar:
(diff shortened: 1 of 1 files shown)


savingoyal:
@jens-hummelshoej-tri Any reasons for 356 and 357 to be different PRs?

jens-hummelshoej-tri:
@savingoyal no, I guess they could be the same PR. Two different approaches to the same issue. PR356 works for linked files, PR357 works for linked folders. Honestly, I thought when some tests failed on PR356 that it would be ignored, so I found a different solution.

tuulos:
@jens-hummelshoej-tri I am happy to let you know that symlinks are finally being followed in the latest release. We just added a bit extra logic to avoid cyclical links which could make it hang otherwise

PR #357, opened 2020-10-20
Follow symlinks when packaging

jens-hummelshoej-tri:
As a user of metaflow, I would like to share external libraries among flows in different folders without code duplication.

symlinks to package folders provide a solution but not when packaging and running --with batch, because the os.walk is not following symlinks when packaging.


tuulos:
@jens-hummelshoej-tri I am happy to let you know that symlinks are finally being followed in the latest release. We just added a bit extra logic to avoid cyclical links which could make it hang otherwise

PR #364, opened 2020-10-26
AWS GovCloud compliance

savingoyal:
Changes to S3 access, Batch configuration and State Language specification to support AWS GovCloud partition. Should be applicable to aws-cn region as well, but it hasn't been tested yet.


joe153:
LGTM, thank you @savingoyal!

PR #366, opened 2020-10-27
New datastore

romain-intel:


metaflow/client/core.py
```diff
@@ -25,13 +26,37 @@
     # python3
     import pickle
 
+# IMPORTANT NOTE ON THE CLIENT
+#
+# The client is not fully consistent particularly when you use the client to
+# query current runs. The previous version of the client accessed data directly
```
savingoyal, line 32: Before we merge, it would be good to specify which "previous versions".

romain-intel: Will do.

talebzeghmi:
Is there a design document to reference that would help code review this? thanks!

romain-intel:
Hello @talebzeghmi. Unfortunately, there is no design document for this (I should have written one) but I can give you some of the higher level goals of the change. The biggest driver was the fact that there was not one way to access data stored in the datastore and the abstraction provided by the datastore was therefore broken. As an example, you can see that `pickle` is used in the client to access artifacts whereas that should be handled more at the datastore level. There are other examples of this as well (assumption that things are gzipped fore example). The change does a better job splitting up responsibilities (the TaskDatastore is responsible for converting a Python object to bytes (basically the pickle part but this could change if needed down the line and it would be contained in one place), the backend implements the low-level byte reading/writing operations making it easier to port this to GCS (for example) without having to change much (the old code had a lot of intermingled parts making it harder to tease out what needed to be reimplemented). Along the way, we fixed a few other things but that was the major motivation. To provide a bit more context, we are also working on a way to access artifacts in a more "sane" way particularly if they were produced in an environment that is not the current running environment. This change will make that easier. Let me know if you have more questions, happy to provide more information.

romain-intel:
Replaced by #413

PR #367, opened 2020-10-27
Change the new datastore to use Metadata to store file information

romain-intel:
The new datastore would initially store some extra information in the file itself (in a header). This uses the new metadata support in S3 to store this information.


metaflow/datastore/content_addressed_store.py
```diff
@@ -145,38 +138,49 @@ def load_blobs(self, keys, force_raw=False):
...
+                            info_format,
+                            r.read(struct.calcsize(info_format)))
+                        if file_magic == magic:
+                            unpack_code = getattr(
+                                self, '_unpack_v%d' % version, None)
                         else:
-                            # This is the backward compatible mode
                             r.seek(0)
+                            # This is the backward compatible mode
                             unpack_code = getattr(
```
tuulos, line 170: why not just ```
unpack_code = self._unpack_backward_compatible
```

File: metaflow/datastore/content_addressed_store.py [hunk 2]
```diff
@@ -145,38 +138,49 @@ def load_blobs(self, keys, force_raw=False):
...
+                        version = meta.get('cas_version', -1)
+                        if version == -1:
+                            raise DataException(
+                                "Could not extract encoding version for %s" % path)
+                        unpack_code = getattr(self, '_unpack_v%d' % version, None)
+                    if unpack_code is None:
                         raise DataException(
                             "Unknown encoding version %d for %s -- the artifact "
                             "is either corrupt or you need to update Metaflow"
                             % (version, path))
```
tuulos, line 182: `version` may be undefined here. Just print the `path`.

metaflow/datastore/content_addressed_store.py
romain-intel: Yes, dumb copy paste without thinking...

metaflow/datastore/content_addressed_store.py
romain-intel: Fair point. As discussed. We can hopefully remove this as well (or at least simplify some of it).

romain-intel: Done: I moved this code to the else part (only relevant part after refactoring to remove the version stored directly in the file.

metaflow/datastore/content_addressed_store.py
romain-intel: Done and simplified.

PR #368, opened 2020-10-27
Explore further performance improvements in S3 datatools

romain-intel:


metaflow/datatools/s3op.py
```diff
@@ -106,66 +157,125 @@ def worker(queue, mode):
...
 
     except:
         traceback.print_exc()
         sys.exit(ERROR_WORKER_EXCEPTION)
 
 def start_workers(mode, urls, num_workers):
-
+    # We start the minimum of len(urls) or num_workers to avoid starting
+    # workers that will definitely do nothing
+    num_workers = num_workers if len(urls) > num_workers else len(urls)
```
tuulos, line 216: nit: `num_workers = min(num_workers, len(urls))`

metaflow/datatools/s3op.py
```diff
@@ -106,66 +157,125 @@ def worker(queue, mode):
...
     queue = Queue(len(urls) + num_workers)
     procs = {}
 
     # 1. push sources and destinations to the queue
-    for url, _ in urls:
-        queue.put(url)
+    idx = 0
+    for elt in urls:
```
tuulos, line 222: no need to increment `idx` manually:
```
for idx, elt in enumerate(urls):
    queue.put((elt, idx))
```

metaflow/datatools/s3op.py
```diff
@@ -106,66 +157,125 @@ def worker(queue, mode):
...
+                        args = {'size': result_info['size']}
+                        if result_info['content_type']:
+                            args['content_type'] = result_info['content_type']
+                        if result_info['metadata'] is not None:
+                            args['metadata'] = result_info['metadata']
                         json.dump(args, f)
+                    # Finally, we push out the size to the result_pipe since
+                    # the size is used for verification and other purposes and
+                    # we want to avoid file operations for this simple process
+                    result_pipe.send((idx, result_info['size']))
```
tuulos, line 184: this call will block if the pipe buffer is full. It can get full looking at the receiving side below.

Since you open the `_meta` file above anyways, why not write these results in the `_meta` file, which is guaranteed to be non_blocking and get rid of the pipe?

metaflow/datatools/s3op.py
```diff
@@ -106,66 +157,125 @@ def worker(queue, mode):
...
+        # they close properly
+        recv = None
+        send = None
+
+    # 5. wait for the processes to finish; we continuously update procs
+    # to remove all processes that have finished already
+    while procs:
+        new_procs = {}
+        for proc, recv in procs.items():
+            proc.join(timeout=1)
```
tuulos, line 249: this call will block for a second while workers are executing. If means that if you have, say, 32 workers (the default), it will take 31 seconds before we get to the last worker. It is possible that a worker will have filled its pipe buffer during this long wait time and it starts blocking, which hurts performance.

metaflow/datatools/s3op.py
```diff
@@ -106,66 +157,125 @@ def worker(queue, mode):
...
+                except EOFError:
+                    pass
+                if proc.exitcode != 0:
+                    msg = 'Worker process failed (exit code %d)'\
+                            % proc.exitcode
+                    exit(msg, proc.exitcode)
+            else:
+                # Do some limited receive to avoid too much buffering on the
+                # pipes.
+                count = 10
```
tuulos, line 266: why 10?

metaflow/datatools/s3op.py
romain-intel: Will fix.

metaflow/datatools/s3op.py
romain-intel: Duh. That was stupid. Will fix.

metaflow/datatools/s3op.py
romain-intel: So yes the pipe could block. It's more OS dependent but on Linux, typically, that would take a few thousand files to block (pipes are typically 64K it seems) so possible but not highly likely. To get to the second question, I could indeed write the information in the file (it actually is contained there). The reason I was also pushing it to a pipe is because of the fact that the size is used quickly for lots of things (verifying existence being one of them). I thought that using files for this might be a bit more expensive). For a few files, it wouldn't be a big problem (but then in that case, pipes would not block either) and for many files where pipes could be problematic, opening and closing thousands of small files might also not be ideal (but I admittedly did not benchmark it). Ideally it would also be a separate file (to keep the index -- yes, we could map but that also seems more costly). We can discuss a bit more though because this is by no means the ideal solution.

metaflow/datatools/s3op.py
romain-intel: See above but yes, it would wait. I could set a quicker timeout and have this be more of a busy loop.

metaflow/datatools/s3op.py
romain-intel: No particular reason for the number. I did want to set a limit to avoid possible starvation.

metaflow/datatools/s3op.py
tuulos: yeah, a concern about many small files is a valid one. I assume you are mainly worried about the read side, since it seems you are writing a separate file for each object above already.

An interesting alternative might be to *append* all results in a single file. Small appends are atomic https://stackoverflow.com/questions/1154446/is-file-append-atomic-in-unix/24270790#24270790

metaflow/datatools/s3op.py
tuulos: A downside of heuristics like this (1 second or less?) and the 10 below is that they can cause performance degradations that are nightmarish to identify and debug.

The "correct" way of handling the read side would be to make pipes non-blocking and use e.g. epoll to multiplex input from pipes. It gets tedious. Hence my suggestion above to consider if there's a way to get rid of pipes altogether.

romain-intel: See above for the elimination of pipes. For the 1 second poll, the timeout was also present in the initial code. I am perfectly fine removing it and putting a tighter loop but it will always be a heuristical approach. Another way we could do it is have the process signal that it is done (but we are back to multiplexing) but that seems a bit overkill.

metaflow/datatools/s3op.py
romain-intel: Done

metaflow/datatools/s3op.py
romain-intel: Done

metaflow/datatools/s3op.py
romain-intel: Addressed in the refactoring using a single file per worker.

metaflow/datatools/s3op.py
romain-intel: Addressed in refactoring. This is no longer present.

metaflow/datatools/s3op.py
romain-intel: Kept 1 second poll but pipe removed.

romain-intel:
I haven't tested the latest code. Will do that tomorrow but wanted to push it out to get eyes on it.

PR #371, opened 2020-11-01
allow tagging of metadata and add attempt status to metadata table

ferras:
write attempt status at the end of every task attempt


savingoyal (approved):
lgtm!

ferras:
> can we set the type to `attempt_ok` as well

yep, i think i pushed the updated name the same time you made that comment

msavela (approved):

Issue #372, opened 2020-11-01
Allow State Machines to be prefixed outside of Metaflow AWS Sandbox


rchui:
This feature would be useful for allowing Metaflow users to group Metaflow state machines together. It would also help Metaflow admins to distinguish Metaflow state machines from non-Metaflow state machines and help with scoping permissions in IAM roles used to access the state machines.

Currently `SFN_STATE_MACHINE_PREFIX` is set to `None` by default and is gated by `AWS_SANDBOX_ENABLED` in this block:

```python
# Finalize configuration
if AWS_SANDBOX_ENABLED:
    os.environ['AWS_DEFAULT_REGION'] = AWS_SANDBOX_REGION
    BATCH_METADATA_SERVICE_URL = AWS_SANDBOX_INTERNAL_SERVICE_URL
    METADATA_SERVICE_HEADERS['x-api-key'] = AWS_SANDBOX_API_KEY
    SFN_STATE_MACHINE_PREFIX = from_conf('METAFLOW_AWS_SANDBOX_STACK_NAME')
```

This should be a relatively simple change in `metaflow/metaflow/metaflow_config.py`. Changing this:

```python
SFN_STATE_MACHINE_PREFIX = None
```

to

```python
SFN_STATE_MACHINE_PREFIX = from_conf('METAFLOW_SFN_STATE_MACHINE_PREFIX')
```

savingoyal:
@rchui Thanks for the issue. Would you like to contribute a PR with this change?

rchui:
I'd be more than happy to submit a PR for this.

PR #373, opened 2020-11-01
Allow users to prefix SFN state machines outside Metaflow sandbox.

rchui:
This PR addresses #372 and modifies the Metaflow environment variable config to allow users to prefix Step Function state machines outside of the Metaflow AWS sandbox. It adds a new environment variable `METAFLOW_SFN_STATE_MACHINE_PREFIX` that controls this behavior but does not modify existing behavior when running in the Metaflow AWS sandbox.


savingoyal (approved):
lgtm!

PR #375, opened 2020-11-05
Set METAFLOW_RUNTIME_NAME for executions by AWS Step Functions

savingoyal:


metaflow/plugins/aws/batch/batch.py
```diff
@@ -159,8 +159,6 @@ def create_job(
             # is NOT set (see get_datastore_root_from_config in datastore/local.py).
         for name, value in env.items():
             job.environment_variable(name, value)
-        for name, value in self.metadata.get_runtime_environment('batch').items():
```
romain-intel, line 162: Why are you removing this? The commit message says that you are adding a new environment variable but this seems to be removing the setting of said env var. Is it done somewhere else?

savingoyal: This was overwriting the env values set by step-functions. That's why I yanked it out. Currently, we don't have any `batch` specific runtime environment anyways.

savingoyal: See this - https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/aws/step_functions/step_functions.py#L521

romain-intel: Got it.

romain-intel (approved):
LGTM

PR #376, opened 2020-11-05
Log metaflow_version as a tag for all packaged executions

savingoyal:


metaflow/metaflow_version.py
```diff
@@ -14,7 +14,7 @@
 __all__ = ("get_version",)
 
 CURRENT_DIRECTORY = path.dirname(path.abspath(__file__))
-INFO_FILE = path.join(CURRENT_DIRECTORY, "INFO")
+INFO_FILE = path.join(path.dirname(CURRENT_DIRECTORY), "INFO")
```
romain-intel, line 17: So just to confirm, the file named 'INFO' is actually two directories up correct?

savingoyal: correct.

romain-intel (approved):
LGTM

PR #377, opened 2020-11-06
Guard against inconsistently cased filesystems for @conda on macOS with @batch

savingoyal:
Conda fails to correctly set up environments for linux-64 packages on macOS which
is needed to collect the necessary metadata for correctly setting up the conda
environment on AWS Batch. This patch simply ignores the error-checks that conda
throws while setting up the environments on macOS when the intended destination
is AWS Batch.

### DIFF
--- a/metaflow/plugins/conda/conda_step_decorator.py
+++ b/metaflow/plugins/conda/conda_step_decorator.py
@@ -108,12 +108,17 @@ class CondaStepDecorator(StepDecorator):
         cached_deps = read_conda_manifest(ds_root, self.flow.name)
         if CondaStepDecorator.conda is None:
             CondaStepDecorator.conda = Conda()
-            CondaStepDecorator.environments = CondaStepDecorator.conda.environments(self.flow.name)
+            CondaStepDecorator.environments =\
+                CondaStepDecorator.conda.environments(self.flow.name)
         if force or env_id not in cached_deps or 'cache_urls' not in cached_deps[env_id]:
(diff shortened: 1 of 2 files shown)


metaflow/plugins/conda/conda_step_decorator.py
```diff
@@ -170,10 +176,21 @@ def _prepare_step_environment(self, step_name, ds_root):
...
+            CondaStepDecorator.environments =\
+                CondaStepDecorator.conda.environments(self.flow.name)
         return env_id
 
+    def _disable_safety_checks(self, decos):
+        # Disable conda safety checks when creating linux-64 environments on
+        # a macOS. This is needed because of gotchas around inconsistently 
+        # case-(in)sensitive filesystems for macOS and linux.
+        for deco in decos:
+            if deco.name == 'batch' and platform.system() == 'Darwin':
```
jasonge27, line 190: Do we want to disable safety checks for all steps including the non-batch step? Users can create a metaflow conda environment running locally and then run --with batch. Would it be a problem if the env created in non-batch mode is then directly used for batch run?

jasonge27 (commented):
LGTM except for one minor comment.

metaflow/plugins/conda/conda_step_decorator.py
savingoyal: This only happens when we resolve linux-64 packages on mac. Currently we create a different environment for batch and non-batch mode when launched from mac. I hope this clarifies.

PR #401, opened 2021-01-05
Add Flow.tags_of_runs property

dpoznik:
It is convenient to filter runs by supplying `tags` to `Flow.runs`, but I didn't see a simple way to determine which tags are available, like one can do in `git` with the `git tag` command. Toward this end, this PR adds a simple property, `Flow.tags_of_runs`, that returns all tags associated with runs of a given flow.


romain-intel:
I believe this should now be addressed by #547 . Please let me know if that would provide equivalent functionality (the tagging CLI would now allow you to do `myflow.py tag list --all` to get the tags for all runs (you can also get for a given run or for your runs).

dpoznik:
> I believe this should now be addressed by #547. Please let me know if that would provide equivalent functionality

Cool, the new `tag` subcommand looks super useful and pretty much obviates the need for the `Flow.tags_of_runs` property I'd proposed here. Thanks for the heads-up! I'll go ahead and close this PR.

It looks like I'd also included a few tangentially related docstring fixes and clarifications (e.g., [incorrect return type on line 483](https://github.com/Netflix/metaflow/pull/401/files#diff-53de601a19b3392fea7c75048a447ff9b8bff3418573ff3ae741d5c987f8a799L483)). I'd be happy to pull those minor tweaks out into their own PR if that would be helpful. Let me know?

PR #412, opened 2021-01-09
Improve top level import

romain-intel:


seeravikiran:
Could you add a bit more description to what this PR helps with?

romain-intel:
The module import mechanism wasn't working well and could cause circular dependencies in some cases; this now only imports non modules to avoid that issue.

PR #413, opened 2021-01-09
New datastore

romain-intel:
Replaces #366 (targeting a different branch)


metaflow/client/core.py
```diff
@@ -25,13 +26,37 @@
...
+# use the client, you could get different values for the same data artifact).
+# This could happen because the client used the Metadata Service to retrieve the
+# artifacts to be read from the datastore and this information is only updated
+# at the *end* of an attempt.
+#
+# This current client goes through the datastore to access data artifacts
+# (well, it goes through the FileCache which uses the datastore). This means,
+# in particular, that it can render some things eventually consistent like the
+# attempt file and the DONE file (since it may try to read them before they
+# are written). This has *no* impact on Metaflow's execution since Metaflow
```
seeravikiran, line 45: Is this still valid after the latest update to S3 consistency?
https://aws.amazon.com/s3/consistency/

metaflow/cli.py
```diff
@@ -807,12 +752,36 @@ def start(ctx,
...
 
     if datastore_root is None:
         datastore_root = \
-          ctx.obj.datastore.get_datastore_root_from_config(ctx.obj.echo)
-    ctx.obj.datastore_root = ctx.obj.datastore.datastore_root = datastore_root
+          ctx.obj.datastore_impl.get_datastore_root_from_config(ctx.obj.echo)
+    if datastore_root is None:
+        raise CommandException(
+            "Could not find the location of the datastore -- did you correctly set the "
+            "METAFLOW_DATASTORE_SYSROOT_%s environment variable" % datastore.upper())
```
seeravikiran, line 764: Nit: Leave a "?" at the end of the mesage.

metaflow/cli.py
```diff
@@ -807,12 +752,36 @@ def start(ctx,
...
+
+    # It is important to initialize flow decorators early as some of the
+    # things they provide may be used by some of the objects initialize after.
+    decorators._init_flow_decorators(ctx.obj.flow,
+                                     ctx.obj.graph,
+                                     ctx.obj.environment,
+                                     ctx.obj.flow_datastore,
+                                     ctx.obj.logger,
+                                     echo,
+                                     deco_options)
```
seeravikiran, line 784: Looking internally (to NFLX) this call was introduced before the Metadata initialization since we wanted to register information about the project (which is a flow level decorator).

I can understand the circular dependency (newly created) since the datastore needs the metadata object - but just wanted to bring your attention to this change.

metaflow/datastore/common.py
```diff
@@ -0,0 +1,28 @@
+from ..exception import MetaflowException
```
seeravikiran, line 1: Might be worth splitting this into: `exceptions.py` and `inputs.py` since `common.py` feels too vague.

metaflow/client/core.py
romain-intel: No, I believe that we won't have as many issues. I'll update the comment.

metaflow/cli.py
romain-intel: OK, will fix.

metaflow/cli.py
romain-intel: I will check but I believe there is no use of the stuff in `current` in the intervening code but that is a good callout.

metaflow/datastore/common.py
romain-intel: Fair enough. I'll make the split.

metaflow/client/core.py
romain-intel: Done

metaflow/cli.py
romain-intel: Done

metaflow/cli.py
romain-intel: Checked; in the internal version, the project description are not used in monitor or datastore.

metaflow/datastore/common.py
romain-intel: Done

PR #417, opened 2021-01-12
Throw an exception if invalid tags are input

seeravikiran:
We will probably not be able to pass tags with special characters in them easily via CLI, so this is a safety guard for that:
E.g: '!,a,list of,#tags,with,,special:characters or "ravi's"

### DIFF
--- a/metaflow/plugins/aws/step_functions/step_functions.py
+++ b/metaflow/plugins/aws/step_functions/step_functions.py
@@ -646,6 +646,12 @@ class StepFunctions(object):
                  '--task-id %s' % task_id_params]
             # Assign tags to run objects.
             if self.tags:
+                import shlex
+                invalid_tags = [tag for tag in self.tags if shlex.quote(tag) != tag]
+                if invalid_tags:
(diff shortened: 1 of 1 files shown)


metaflow/plugins/aws/step_functions/step_functions.py
```diff
@@ -646,6 +646,12 @@ def _step_cli(self,
                  '--task-id %s' % task_id_params]
             # Assign tags to run objects.
             if self.tags:
+                import shlex
+                invalid_tags = [tag for tag in self.tags if shlex.quote(tag) != tag]
+                if invalid_tags:
+                    raise MetaflowException("Step functions doesn't support "\
+                                            "tags with special characters "\
+                                            "currently: %s" % invalid_tags[0])
```
saikonen, line 654: Same tagging logic is present in airflow and argo-workflows as well. Tested on argo and managed to get a syntax error with a tag with special characters, as expected.

`--with kubernetes` has some issues handling the tags as well, but a bit different due to the way the command is built.

tested everything with `--tag "#a,tag:'"`

PR #429, opened 2021-01-31
Added step-function-name as a parameter to the create command

ahmadhori:
this pull request is to give the user the ability to define step function name on AWS to be different from the flow name when creating it
this could be helpful when updating the step function using the pipeline then you can define which step function to update based on the current branch

### DIFF
--- a/metaflow/plugins/aws/step_functions/step_functions_cli.py
+++ b/metaflow/plugins/aws/step_functions/step_functions_cli.py
@@ -78,9 +78,15 @@ def step_functions(obj):
               help="Add a decorator to all steps. You can specify this "
                    "option multiple times to attach multiple decorators "
                    "in steps.")
+@click.option('--step-function-name',
+              default=None,
+              type=str,
+              help="Step Function name to use on AWS - will use flow name as a "
+                   "default value if not specifid.")
 @click.pass_obj
 def create(obj,
            tags=None,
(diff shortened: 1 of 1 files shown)


metaflow/plugins/aws/step_functions/step_functions_cli.py
```diff
@@ -78,9 +78,15 @@ def step_functions(obj):
               help="Add a decorator to all steps. You can specify this "
                    "option multiple times to attach multiple decorators "
                    "in steps.")
+@click.option('--step-function-name',
```
savingoyal, line 81: We can rename it to `--name` to keep it simple. Do you have any thoughts around how `step-functions trigger` command would work in this scenario, where the user overrides the name of the workflow?

corleyma:
So we already accomplish something similar by setting `METAFLOW_SFN_STATE_MACHINE_PREFIX` env var when deploying/triggering.  Perhaps it makes more sense to expose this option at the CLI-level, so that the name still has some bearing to the actual flow name?  

@savingoyal, re: your question of how such a prefix/name option would work when triggering, I think the implication is that users are responsible for setting the same prefix/name at trigger time if they wish to target the same SFN.

savingoyal:
PRs #540 , #532 and #541 have implemented this feature.

PR #434, opened 2021-02-05
DRAFT - Schedule Flows on Argo Workflows

sappier:
Metaflow Argo plugin
====================

Our proposed Argo plugin for Metaflow facilitates scheduling and execution of workflows in Kubernetes clusters comparable to the `batch` and `step-functions` with [Argo Workflows](https://argoproj.github.io/argo/).

Introduction
------------

The plugin works similarly to the `step-functions` approach and generates
[WorkflowTemplates](https://argoproj.github.io/argo/workflow-templates/) based on Metaflow flows which are then directly orchestrable on Argo.

The user interacts with Argo through similar commands:

* Deploy a WorkflowTemplate to the cluster
  ```
  python flow.py argo create
  ```
* Trigger a workflow execution
  ```
  python flow.py argo trigger
  ```
* List workflow runs
  ```
  python flow.py argo list-runs
  ```

The user can similarly pass parameters to the flow which are then mapped to workflow parameters:
```
python flow.py argo trigger --alpha=0.1
```
It is possible to generate a workflow specification with the `--only-json` flag and then create a WorkflowTemplate in an Argo Dashboard or with the `argo` command-line utility.

Argo Decorators
---------------

The plugin supports optional argo/k8s-specific parameters via
`@kubernetes`, `@argo` and `@argo_base` decorators, e.g.:

```python
  @argo_base(image='continuumio/miniconda',
	     annotations={
	         'sidecar.istio.io/inject': 'false'
             }
  )
  class MyFlow(FlowSpec):
      ...

      @kubernetes(image='amancevice/pandas')
      @step
      def training(self):
          ...
```
As it works with other decorators the base parameters can be overridden by definitions on step level.

The currently supported list of decorator parameters is in `argo_decorator.py`.
Most of them are mapped directly to the corresponding k8s/Argo specs.
Argo decorators are ignored when the flow is executed locally.

Configuration
-------------

Argo *create*, *trigger* and *list-runs* commands send HTTP requests to the [Argo Server](https://argoproj.github.io/argo/argo-server/).
`METAFLOW_ARGO_SERVER` address should be defined in the configuration to make them work.

To make pods access S3 storage, S3 credentials can be stored in k8s secrets and applied to the running pods through `env` or `envFrom` parameters, e.g.:

```python
@argo_base(envFrom = [{'secretRef': {'name': 's3-secret'}}])
class Flow(FlowSpec):
    ....
```

See https://github.com/sappier/metaflow/wiki/Metaflow-Argo-Plugin for more details.

Implementation details
----------------------

Most of the AWS state machine constructs were mapped correspondingly to the generated argo workflow specs.
Our plugin tries to follow the step-functions implementation as close as possible.

It generates a [DAG template](https://argoproj.github.io/argo/examples/#dag) where every flow's `@step` is performed by a separate DAG task.  Every task in turn executes a `container template` which is very similar to a k8s pod.

Internal parameters, e.g. `input-paths`, `split-index`, are passed through argo template [parameters](https://argoproj.github.io/argo/examples/#parameters).

Flow parameters and IncludeFiles are mapped into WorkflowTemplate input parameters.

Gaps in Functionality
--------------------

There are major Metaflow features we did not implement yet:

* [x] multi-level foreach
* [x] resume workflows
* [x] `@schedule` decorator
* [x] `@timeout`, `@catch`, `@retry` decorators
* [x] --max-workers
* [ ] `@project` decorator


metaflow/plugins/argo/argo_cli.py
```diff
@@ -0,0 +1,185 @@
...
+              help="Environment variables to be set for the workflow.")
+@click.option("--env-from",
+              default=None,
+              help="Environment variables to be set for the workflow.")
+@click.option("--token",
+              default=None,
+              help="Authentication token to call Argo Server.")
+@click.option('--namespace',
+              default=None,
+              help="Deploy into the specified kubernetes namespace.")
```
oavdeev, line 40: This may get confusing: `step-functions create` also supports `--namespace`, but in that case this is Metaflow namespace

metaflow/plugins/argo/argo_workflow.py
```diff
@@ -0,0 +1,586 @@
...
+
+    def _prepare_environment(self):
+        default = {
+            'METAFLOW_USER': get_username(),
+            'METAFLOW_DATASTORE_SYSROOT_S3': DATASTORE_SYSROOT_S3,
+        }
+        if METADATA_SERVICE_URL:
+            default['METAFLOW_SERVICE_URL'] = METADATA_SERVICE_URL
+        if DEFAULT_METADATA:
+            default['METAFLOW_DEFAULT_METADATA'] = DEFAULT_METADATA
```
oavdeev, line 462: `METAFLOW_SERVICE_HEADERS` is another one we'd need to pass through

metaflow/plugins/argo/argo_decorator.py
```diff
@@ -0,0 +1,131 @@
...
+        Follows the k8s spec.
+    envFrom: list
+        Additional environment variables to be set in all steps.
+        Follows the k8s spec.
+    imagePullSecrets: list
+        Credentials for pulling docker images.
+    """
+    name = 'argo_base'
+    defaults = {
+        'image': None,
```
oavdeev, line 41: This is a broader q beyond this PR, so just thinking aloud: there probably should be a more consistent way to specify docker images between `@batch`, `@argo` and potential future execution envs based on Docker. For example, right now for `@batch` default image is a configuration setting, not flow-level decorator parameter. There is also a docker image param introduced in #463 for AWS Lambda.

metaflow/plugins/argo/argo_cli.py
sappier: yes, the semantics differs a little.  But the name "namespace" is so common in k8s and using k8s namespaces is exactly how we separate workloads of different users in k8s cluster ...  hard to come up with a better name.
Integration with metaflow's namespaces isn't completed though.

metaflow/plugins/argo/argo_decorator.py
sappier: yes, accepting docker images from the config/envar is on our table.
There is indeed a broader topic.  Recently we had a discussion internally whether we should re-use Metaflow's @batch config names or introduce @argo's alternatives, e.g. `SFN_STATE_MACHINE_PREFIX`.  We opted to have a separate `ARGO_WORKFLOW_PREFIX` for similar functionality to avoid users' confusion.  The same applies to the `BATCH_CONTAINER_IMAGE`.
Ideally, we'd have neutral names for common config variables like `METAFLOW_CONTAINER_IMAGE` which could be re-used in different Metaflow plugins.  Even by @batch at some point though @batch would need to support the old name for backward compatibility.

metaflow/plugins/argo/argo_workflow.py
sappier: sure

metaflow/plugins/argo/argo_cli.py
oavdeev: Yeah it is unfortunate, maybe something like `--k8s-namespace`? It is cumbersome but unambiguous

metaflow/plugins/argo/argo_decorator.py
oavdeev: Started a discussion here https://github.com/Netflix/metaflow/issues/489, please chime in

oavdeev:
One idea would be to have a more generic `@kubernetes` decorator a-la #488 that would contain all k8s-specific settings (like `nodeSelector`, `envFrom`, etc.), and it would both work in "standalone" mode via #488 or with Argo via this PR. In this case, there would be no Argo-specific decorator, and argo UI would be even more similar to how step functions + batch API looks today.

sappier:
> One idea would be to have a more generic `@kubernetes` decorator

yes.  The only reason why `@argo` has such parts today is because there was no official `@kubernetes` decorator.  We'd rebase it as soon as `@kubernetes` will stabilize.

> In this case, there would be no Argo-specific decorator

we would likely still need it because Argo's Template contains more stuff on top of k8s's pod/job/container: https://argoproj.github.io/argo-workflows/fields/#template
E.g. we allow users to define Argo's output [artifacts](https://argoproj.github.io/argo-workflows/examples/#artifacts) through the `@argo(output_artifacts=...)`.
But the rest of its current parameters should be taken from the `@kubernetes`.

metaflow/plugins/argo/argo_cli.py
sappier: done

metaflow/plugins/argo/argo_workflow.py
sappier: done

talebzeghmi:
You could reuse this suite of tests to validate the Argo orchestrator 
https://github.com/zillow/metaflow/tree/feature/kfp/metaflow/plugins/kfp/tests/flows

sappier:
> You could reuse this suite of tests to validate the Argo orchestrator

Wow, a great collection of test flows.  @talebzeghmi , thanks a lot for pointing to it.

PR #449, opened 2021-03-09
New datastore based logging system

savingoyal:
This PR introduces a new logging functionality for Metaflow where all Metaflow runtime/task logs are published via a sidecar process to the datastore. The user-visible logs on the console are streamed directly from the datastore. 

For Metaflow's integrations with the cloud (AWS at the moment), the compute tasks logs (AWS Batch) are directly written by Metaflow into the datastore (Amazon S3) independent of where the flow is launched from (User's laptop or AWS Step Functions). This has multiple benefits 
- Metaflow no longer relies on AWS Cloud Watch for fetching the user logs to the console - AWS Cloud Watch has rather low global API limits which have caused multiple issues in the past for our users
- Logs for AWS Step Functions executions are now also available in Amazon S3 and can be easily fetched by simply doing `python flow.py logs 42/start`

The two downsides of this PR are - 
- We will not be able to capture final log lines during hard crashes since AWS Batch doesn't provide any such capability natively. Hopefully, other compute platforms will not have this gap in their offering. Netflix's Titus offers a feature where it ships all logs to a location in Amazon S3. We will introduce functionality to ship these final log lines in a best-effort manner from AWS CloudWatch to Amazon S3 when the container suffers a hard crash in cases when the user launches the workflow from their laptop.
- We will be issuing more Amazon S3 PUT requests which will increase the cost of a metaflow execution by a bit. We will provide knobs to control how often logs are written to Amazon S3 so that the user is in control.


savingoyal:
@romain-intel @talebzeghmi @sappier @corleyma @oavdeev the PR is ready for consumption.

oavdeev (commented):
Not sure what's the best place for this, but it would be nice to document the overall approach somewhere in more detail. For example, it took me some time to figure out that there is in fact only **one** log file in S3 that is being re-uploaded over and over, and it is not using some kind of chunked multi-file storage scheme.

Also documenting the overall process hierarchy would be helpful. I made this diagram for myself fwiw, let me know if that looks right:
mflog

romain-intel (commented):
Minor nits but LGTM o/w

metaflow/runtime.py
```diff
@@ -636,20 +643,11 @@ def persist(self, flow):
...
-                               'location': location,
-                               'attempt': self.retries}),
-                           type='log_path',
-                           tags=[])]
-        self.metadata.register_metadata(self.run_id,
-                                        self.step,
-                                        self.task_id,
-                                        datum)
-        return location
+    def save_logs(self, stdout_buffer, stderr_buffer):
```
romain-intel, line 646: As a comment here: we should probably not remove the metadata information. For the location, we can do like we do for artifacts in the convergence branch which stores the location as ':root:datastore_root'. This does not store the URL (because there would be many in part) but does allow the client to get an instance of the datastore used to stave the logs (and therefore retrieve the logs).

romain-intel (changes requested):
Getting rid of the metadata for log location is not great IMHO. This would make it hard to reconstruct the object (datastore) needed to properly fetch the log.

metaflow/runtime.py
savingoyal: Good point. Will do!

metaflow/cli.py
```diff
@@ -304,31 +322,70 @@ def logs(obj, input_path, stdout=None, stderr=None, both=None):
...
+                            monitor=obj.monitor,
+                            event_logger=obj.event_logger)
+        # get all successful tasks
+        ds_list = list(datastore_set)
+
+    if ds_list:
+        def echo_unicode(line, **kwargs):
+            click.secho(line.decode('UTF-8', errors='replace'), **kwargs)
+
+        maybe_old_style = True
```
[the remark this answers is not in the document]
savingoyal: done!

metaflow/mflog/mflog.py
```diff
@@ -0,0 +1,146 @@
...
+            return utc_dt.replace(tzinfo=timezone.utc).astimezone(tz=None)
+    except ImportError:
+        # python2
+        import calendar
+        def utc_to_local(utc_dt):
+            timestamp = calendar.timegm(utc_dt.timetuple())
+            local_dt = datetime.fromtimestamp(timestamp)
+            return local_dt.replace(microsecond=utc_dt.microsecond)
+
+def decorate(source, line, version=VERSION, now=None, lineid=None):
```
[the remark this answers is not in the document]
savingoyal: It will be used for reconciling the logs fetched from s3 and cloudwatch. That piece of code is currently listed under a TODO.

metaflow/plugins/aws/batch/batch_decorator.py
```diff
@@ -209,6 +210,7 @@ def task_pre_step(self,
         entries = [MetaDatum(field=k, value=v, type=k, tags=[]) for k, v in meta.items()]
         # Register book-keeping metadata for debugging.
         metadata.register_metadata(run_id, step_name, task_id, entries)
+        self._save_logs = SidecarSubProcess('save_logs_periodically')
```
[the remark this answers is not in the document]
savingoyal: done

metaflow/runtime.py
```diff
@@ -893,17 +921,22 @@ def terminate(self):
         # Return early if the task is cloned since we don't want to
         # perform any log collection.
         if not self.task.is_cloned:
-            self.task.save_logs('stdout', self._stdout.get_bytes())
-            self.task.save_logs('stderr', self._stderr.get_bytes())
+            self.task.save_logs(self._stdout.get_bytes(),
```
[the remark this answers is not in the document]
savingoyal: I didn't quite understand this comment. Can you elaborate?

metaflow/runtime.py
savingoyal: Can you point me to your change in the convergence branch? I don't think `save_logs` is the place to store the `datastore_root` in the metadata.

savingoyal:
@romain-intel I created #478 to keep track of the AWS CloudWatch logs within Metaflow metadata. With the release of the V4 of task metadata API for ECS, it is finally possible to get hold of the log stream name from within the container. Given that there is a current possibility to lose out on critical log lines during hard crashes, this provides complete lineage to full logs in cloudwatch.

metaflow/runtime.py
romain-intel: I just pushed stuff in #479, I took your (offline) suggestion to put the ds-root and ds-type as a task level metadata.

metaflow/runtime.py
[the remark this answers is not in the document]
romain-intel: I think I was trying to say something that this may be too soon if there is more log coming with the emit_log and task.log but looking at the code I see I am mistaken and that all log should be already there in the stdout/stderr.

metaflow/runtime.py
savingoyal: I have added those elements in this PR as well.

metaflow/client/core.py
```diff
@@ -1024,16 +1024,62 @@ def environment_info(self):
...
+    def loglines(self, stream, as_unicode=True):
+        """
+        Return an iterator over (utc_timestamp, logline) tuples.
+
+        If as_unicode=False, logline is returned as a byte object. Otherwise,
+        it is returned as a (unicode) string.
+        """
+        from metaflow.mflog.mflog import merge_logs
+        from metaflow.mflog import LOG_SOURCES
+        from metaflow.datastore import DATASTORES
```
oavdeev, line 1043: why are those imports here? recursive dep? may be worth adding a comment

romain-intel:
Thanks for adding ds-type and ds-root. LGTM now. As a minor nit, maybe change the title from 'New datastore based...' since this is not based on the new datastore :). Approving now though.

romain-intel (approved):

savingoyal:
> Thanks for adding ds-type and ds-root. LGTM now. As a minor nit, maybe change the title from 'New datastore based...' since this is not based on the new datastore :). Approving now though.

Thanks! Haha, I meant `New 'Datastore based logging system'` :P I will change the title.

metaflow/client/core.py
savingoyal: No recursive dep. They are only used in this function. I can also move them at the top.

oavdeev: Yep I think it would be better to have them at the top then. Otherwise makes one think there is something special going on here :)

PR #455, opened 2021-03-15
Fix typos

dpoznik:
This PR removes an extra word from a docstring, adds a missing indentation space, and removes a couple unused imports from the `s3` modules.


savingoyal:
Good catch with the indentation. LGTM!

PR #456, opened 2021-03-18
Fix ordering of custom parameters in help text

dpoznik:
With this change, the help text output by:
```
python my_flow.py run --help
```
will list custom options in the intended order.

Custom parameters are iteratively inserted into the front of the command parameters list. Doing so in reverse preserves the order in which they are defined in the `FlowSpec`.

### DIFF
--- a/metaflow/parameters.py
+++ b/metaflow/parameters.py
@@ -227,7 +227,7 @@ def add_custom_parameters(deploy_mode=False):
     # deploy_mode determines whether deploy-time functions should or should
     # not be evaluated for this command
     def wrapper(cmd):
-        for arg in parameters:
+        for arg in parameters[::-1]:
             kwargs = arg.option_kwargs(deploy_mode)
             cmd.params.insert(0, click.Option(('--' + arg.name,), **kwargs))
         return cmd
--- a/metaflow/parameters.py
+++ b/metaflow/parameters.py
@@ -227,6 +227,8 @@ def add_custom_parameters(deploy_mode=False):
     # deploy_mode determines whether deploy-time functions should or should
     # not be evaluated for this command
     def wrapper(cmd):
+        # Iterate over parameters in reverse order so cmd.params lists options
+        # in the order they are defined in the FlowSpec subclass
         for arg in parameters[::-1]:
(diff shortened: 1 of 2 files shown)


dpoznik:
It may be desirable to make a similar change [here](https://github.com/Netflix/metaflow/blob/master/metaflow/decorators.py#L159) in `metaflow.decorators.add_decorator_options`, as that function uses the same pattern: `cmd.params.insert(0, click.Option(('--' + <name>,), **kwargs))`.

savingoyal:
> It may be desirable to make a similar change [here](https://github.com/Netflix/metaflow/blob/master/metaflow/decorators.py#L159) in `metaflow.decorators.add_decorator_options`, as that function uses the same pattern: `cmd.params.insert(0, click.Option(('--' + <name>,), **kwargs))`.

That should be okay because we don't rely on the relative ordering on the cli-args

metaflow/parameters.py
```diff
@@ -227,7 +227,7 @@ def add_custom_parameters(deploy_mode=False):
     # deploy_mode determines whether deploy-time functions should or should
     # not be evaluated for this command
     def wrapper(cmd):
-        for arg in parameters:
+        for arg in parameters[::-1]:
```
savingoyal, line 230: Thanks for making the change! Can we just add a quick comment above this line so that a future refactor can preserve this behavior?

savingoyal (changes requested):

dpoznik:
> > It may be desirable to make a similar change [here](https://github.com/Netflix/metaflow/blob/master/metaflow/decorators.py#L159) in `metaflow.decorators.add_decorator_options`, as that function uses the same pattern: `cmd.params.insert(0, click.Option(('--' + <name>,), **kwargs))`.
> 
> That should be okay because we don't rely on the relative ordering on the cli-args

OK, thanks for confirming. I'll leave leave that one as-is.

metaflow/parameters.py
dpoznik: No prob. Done!

savingoyal: Thank you!

savingoyal (approved):

PR #461, opened 2021-03-31
Make --namespace= behave as promised

oavdeev:
The help string for `--namespace` [says](https://github.com/Netflix/metaflow/blob/de527ab8da63fa4a5895d5f6b032728769838f0a/metaflow/cli.py#L602-L607) you can pass an empty string, and it would have the same effect as `namespace(None)`.

However, it looks like this setting is never propagated to `step` command. That is, scheduler does not pass `--namespace` flag to the step process, and then namespace defaults back to username.

I have tested this in local mode and with `@batch`

### DIFF
--- a/metaflow/runtime.py
+++ b/metaflow/runtime.py
@@ -740,10 +740,18 @@ class CLIArgs(object):
 
     def get_args(self):
         def options(mapping):
+            """ Map a dictionary of options to a list of command-line arguments.
+
+            Values assigned to boolean values are assumed boolean flags. List values
+            are converted to repeated CLI arguments.
+
+            >>> list(options({"foo": True, "bar": False, "baz": "", "qux": ["a", "b"]}))
+            ["--foo", "--baz", "", "--qux", "a", "--qux", "b"]
+            """
             for k, v in mapping.items():
                 values = v if isinstance(v, list) else [v]
                 for value in values:
-                    if value:
+                    if value is not None and value is not False:
                         yield '--%s' % k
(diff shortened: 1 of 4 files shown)


metaflow/runtime.py
```diff
@@ -740,10 +740,18 @@ def __init__(self, task):
 
     def get_args(self):
         def options(mapping):
+            """ Map a dictionary of options to a list of command-line arguments.
+
+            Values assigned to boolean values are assumed boolean flags. List values
```
seeravikiran, line 745: Are we assuming boolean flags will be defaulted to false?

Wouldn't it be better to explicitly set the key (k) explicitly based on the value? Thoughts?

oavdeev: > Are we assuming boolean flags will be defaulted to false?

That bit of logic is not new, I _think_ that was the intent behind `if value:` in the previous version (but I may be wrong).

Good point on `dict_to_cli_options`, I'll do that

seeravikiran: You're right - the previous code assumed the same.

I was just pointing out since you were touching this area of code :) no worries if you want to preserve the same semantics for now.

savingoyal:
@oavdeev Thanks for the PR. I verified that it works in the local mode and with `@batch`. However, it doesn't work as expected with Step Functions. I believe `step_functions.py` needs a similar TLC.

oavdeev:
Good call, fixed for step functions

PR #462, opened 2021-03-31
Remove pylint<2.5.0 restriction

savingoyal:
context - https://gitter.im/metaflow_org/community?at=60622af8940f1d555e277c12

Pylint barfs on `self.next` call, which can be safely ignored.

### DIFF
--- a/metaflow/pylint_wrapper.py
+++ b/metaflow/pylint_wrapper.py
@@ -72,4 +72,7 @@ class PyLint(object):
             # Ditto for dynamically added properties in 'current'
             if "Instance of 'Current' has no" in line:
                 continue
+            # Ingore complaints of self.next not callable
+            if "self.next is not callable" in line:
+                continue
             yield line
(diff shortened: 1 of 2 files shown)


corleyma (approved):

corleyma:
✅  from me.  Lifting this pin would be helpful for allowing newer versions of isort to be installed alongside Metaflow as well.

metaflow/pylint_wrapper.py
```diff
@@ -72,4 +72,7 @@ def _filter_lines(self, output):
             # Ditto for dynamically added properties in 'current'
             if "Instance of 'Current' has no" in line:
                 continue
+            # Ingore complaints of self.next not callable
```
romain-intel, line 75: Nit: Ignore.

romain-intel (approved):
Ready to merge I think.

metaflow/pylint_wrapper.py
savingoyal: fixed.

savingoyal:
@romain-intel A note - when the conda bot kicks in, we would need to update the [pinned dependency](https://github.com/conda-forge/metaflow-feedstock/blob/master/recipe/meta.yaml#L29) in the resulting PR as well.

Issue #358, opened 2020-10-21
Support for AWS Lambda


joe153:
What would it take for Metaflow to support @lambda decorator? AWS Batch is powerful and good at bigger jobs but AWS Lambda would be great for smaller/fast jobs and would work together with AWS Step Functions.

savingoyal:
@joe153 For AWS Lambda support, you would have to implement [a plugin very similar to the `@batch` plugin](https://github.com/Netflix/metaflow/tree/master/metaflow/plugins/aws/batch). Job execution guarantees are going to be quite different on AWS Lambda because of the resource constraints and timing constraints on the Lambda containers. For example, `@conda` might not work because of the 3G limit on memory and 0.5G limit on disk; `Parameters` (and `foreaches`) could also be a bit problematic due to the 4K env size limit and it will be hard to gauge if the function invocation will always be within the 15 mins limit.

Zylatis:
I think it's worth looking into. I was considering doing something like this independent of Metaflow as a kind of super lazy 'parallel this loop' up-and-out onto Lambda but it would be great to have in Metaflow. 

All the cautionary points above are valid, but I think if someone is explicitly invoking @lambda not @batch it's reasonable to assume they are aware of the limitations, and if they get timeouts it's on them. If there are embarrassingly parallel tasks which the user is certain will be < 15 min (and all the other issues discussed, env size etc) then this would be a powerful tool.

calmacx:
hi all, I also agree it'd be fantastic to have this. @joe153 are you considering working implementing this plugin yourself? I might give it a go myself

joe153:
@Zylatis: I agree with you. Metaflow syntax is super simple and neat, it would be great to generate step functions using Metaflow with Lambda. 3G memory limit and 15 timeout limit are enough for many jobs and they will only increase in the future. Yes Metaflow is great at ML projects but I see broader use cases like copy data, send emails, etc.

@calmacx I haven't had a chance to do anything. Give it a go!

Zylatis:
@calmacx I'd be keen to collaborate if you're looking to take this on and it won't be too many cooks etc.

calmacx:
Great stuff. When I get round to starting a fork I'll tag you guys... also, don't let me stop you making a start yourself, might be ~1 week or so before I get started. Very happy to collab - I'm a research scientist - my excuse if my coding isn't great...

savingoyal:
Great! Keep me updated and let me know if you need any assistance grokking the codebase.

[...]

PR #463, opened 2021-04-01
@awslambda decorator

oavdeev:
This is a PR to implement https://github.com/Netflix/metaflow/issues/358 , stacked on top of #810 

### What works
You can decorate steps with `@awslambda` and they will run on Lambda. It is pretty quick, just a few seconds spin-up time, so the user experience is much better than AWS Batch. Logs are collected and streamed properly.

`@awslambda` respects `@resource`, `@retry` and `@timeout` decorators.

You can also specify a custom docker image like this:
```python
@awslambda(image="1234.dkr.ecr.us-west-2.amazonaws.com/my-image:latest")
def my_step(self):
    ...
```
The image must be pushed to a ECR registry, public images are not supported by AWS Lambda.

### Not supported in this PR
..but could be done as future improvements:

* No integration with Step Functions. It would be nice if one could write mixed Batch+Lambda workflows and deploy them to Step, but this is not implemented here (you'll get a nice error message)
* `@conda` is not supported (you'll get a nice error message). The reason is that Lambdas only get .5 GB of writable disk space. It could probably technically work for smallish conda deps, but even installing Pandas bumps into that limit, so I figured better to disable this altogether. What we could implement instead in a future PR, is to build and push a custom docker image for each `@conda` step, since Lambda supports docker images up to 10GB in size.

### How to give it a try
For now the setup is somewhat involved, hopefully we'll automate away most of this:
1. In your AWS account, create a ECR repo to hold the docker image for Lambda. 
2. Run `Makefile` from https://github.com/Netflix/metaflow-tools/pull/29 to build the image and push it to the repo. 
3. Create a Lambda IAM role that has access to Metaflow S3 bucket.
4. Set `METAFLOW_LAMBDA_ROLE_ARN` env var to lambda role ARN created in step 4
5. Set `METAFLOW_LAMBDA_CONTAINER_IMAGE` to the docker image URI from step (2)

### Other Limitations
There are a couple of limitations with Lambdas that I don't think can be addressed nicely:
1. AWS provides no way to cancel Lambda execution. If we want to support that, we'd have to invent something awkward with sending termination signal to Lambda via SQS or something. For now lambdas will always run to completion.
2. AWS provides no built-in way to invoke a function asynchronously and poll for the invocation to finish. You can either call it synchronously, but that's potentially a 15 minute API call. Or invoke it asynchronously, but you have to invent your own mechanism for it to signal completion via another service. For now this implementation sticks with the former approach but it has some downsides, for example you need to make sure the HTTP connection doesn't time out.

### DIFF
--- a/metaflow/decorators.py
+++ b/metaflow/decorators.py
@@ -234,7 +234,7 @@ class StepDecorator(Decorator):
         """
         pass
 
-    def package_init(self, flow, step_name, environment):
+    def package_init(self, flow, step_name, environment, logger):
         """
         Called to determine package components
         """
--- a/metaflow/metaflow_config.py
+++ b/metaflow/metaflow_config.py
@@ -193,6 +193,22 @@ KUBERNETES_CONTAINER_REGISTRY = (
     from_conf("METAFLOW_KUBERNETES_CONTAINER_REGISTRY") or DEFAULT_CONTAINER_REGISTRY
 )
 #
+# AWS Lambda configuration
+###
+LAMBDA_ROLE_ARN = from_conf("METAFLOW_LAMBDA_ROLE_ARN")
+
+LAMBDA_CONTAINER_IMAGE = (
+    from_conf("METAFLOW_LAMBDA_CONTAINER_IMAGE") or DEFAULT_CONTAINER_IMAGE
+)
+
+LAMBDA_CONTAINER_REGISTRY = (
+    from_conf("METAFLOW_LAMBDA_CONTAINER_REGISTRY") or DEFAULT_CONTAINER_REGISTRY
+)
+
+# This limit is used to retry lambda execution when we're throttled by AWS due to
+# being close to the parallel execution limit. This does *not* affect how much we
+# retry if the step itself has failed.
+LAMBDA_THROTTLE_RETRIES = int(from_conf("METAFLOW_LAMBDA_THROTTLE_RETRIES", 20))
 
 ###
 # Conda configuration
--- a/metaflow/plugins/aws/awslambda/lambda_cli.py
+++ b/metaflow/plugins/aws/awslambda/lambda_cli.py
@@ -1,0 +1,224 @@
+import os
+import sys
+
+import click
+import traceback
+
+from metaflow import R, util
+from metaflow.exception import METAFLOW_EXIT_DISALLOW_RETRY
+from metaflow.datastore import FlowDataStore
+from metaflow.metadata.util import (
+    sync_local_metadata_from_datastore,
+)
+from metaflow.metaflow_config import DATASTORE_LOCAL_DIR
+
+from .lambda_runner import LambdaRunner, LambdaRuntimeException
+from metaflow.mflog import TASK_LOG_SOURCE
+
+
+class CommonTaskAttrs:
+    def __init__(
+        self,
+        flow_name,
+        run_id,
+        step_name,
+        task_id,
+        attempt,
+        user,
+        version,
+    ):
+        self.flow_name = flow_name
+        self.step_name = step_name
+        self.run_id = run_id
+        self.task_id = task_id
+        self.attempt = attempt
+        self.user = user
+        self.version = version
+
+    def to_dict(self, key_prefix):
+        attrs = {
+            key_prefix + "flow_name": self.flow_name,
+            key_prefix + "step_name": self.step_name,
+            key_prefix + "run_id": self.run_id,
+            key_prefix + "task_id": self.task_id,
+            key_prefix + "retry_count": str(self.attempt),
+            key_prefix + "version": self.version,
+        }
+        if self.user is not None:
+            attrs[key_prefix + "user"] = self.user
+        return attrs
+
+
+@click.group()
+def cli():
+    pass
+
+
+@cli.group(help="Commands related to AWS Lambda.")
+def awslambda():
+    pass
+
+
+@awslambda.command(
+    help="Execute a single task using AWS Lambda. This command "
+    "calls the top-level step command inside a AWS Lambda "
+    "job with the given options. Typically you do not "
+    "call this command directly; it is used internally "
+    "by Metaflow."
+)
+@click.argument("step-name")
+@click.argument("code-package-sha")
+@click.argument("code-package-url")
+@click.option("--run-id", help="Passed to the top-level 'step'.")
+@click.option("--task-id", help="Passed to the top-level 'step'.")
+@click.option("--lambda-arn", help="Lambda ARN")
+@click.option("--lambda-name", help="Lambda name")
+@click.option("--input-paths", help="Passed to the top-level 'step'.")
+@click.option("--split-index", help="Passed to the top-level 'step'.")
+@click.option("--clone-path", help="Passed to the top-level 'step'.")
+@click.option("--clone-run-id", help="Passed to the top-level 'step'.")
+@click.option(
+    "--tag", multiple=True, default=None, help="Passed to the top-level 'step'."
+)
+@click.option("--namespace", default=None, help="Passed to the top-level 'step'.")
+@click.option("--retry-count", default=0, help="Passed to the top-level 'step'.")
+@click.option(
+    "--max-user-code-retries", default=0, help="Passed to the top-level 'step'."
+)
+@click.pass_context
+def step(
+    ctx,
+    step_name,
+    code_package_sha,
+    code_package_url,
+    lambda_arn,
+    lambda_name,
+    **kwargs
+):
+    # type: (...) -> None
+
+    def echo_without_prefix(msg, err):
+        ctx.obj.echo_always(msg, err=err)
+
+    if R.use_r():
+        entrypoint = R.entrypoint()
+    else:
+        executable = ctx.obj.environment.executable(step_name)
+        entrypoint = "%s -u %s" % (executable, os.path.basename(sys.argv[0]))
+
+    top_args = " ".join(util.dict_to_cli_options(ctx.parent.parent.params))
+
+    step_args = " ".join(util.dict_to_cli_options(kwargs))
+    step_cli = u"{entrypoint} {top_args} step {step} {step_args}".format(
+        entrypoint=entrypoint, top_args=top_args, step=step_name, step_args=step_args
+    )
+    node = ctx.obj.graph[step_name]
+
+    retry_count = kwargs.get("retry_count", 0)
+
+    common_attrs = CommonTaskAttrs(
+        flow_name=ctx.obj.flow.name,
+        step_name=step_name,
+        run_id=kwargs["run_id"],
+        task_id=kwargs["task_id"],
+        attempt=retry_count,
+        user=util.get_username(),
+        version=ctx.obj.environment.get_environment_info()["metaflow_version"],
+    )
+
+    env_deco = [deco for deco in node.decorators if deco.name == "environment"]
+    if env_deco:
+        env = env_deco[0].attributes["vars"]
+    else:
+        env = {}
+
+    def echo_with_prefix(msg, err):  # type: (str, bool) -> None
+        ctx.obj.echo_always(
+            "[%s] %s"
+            % (
+                lambda_name,
+                msg,
+            ),
+            err=err,
+        )
+
+    runner = LambdaRunner(
+        datastore=ctx.obj.flow_datastore,
+        environment=ctx.obj.environment,
+        lambda_arn=lambda_arn,
+        name=lambda_name,
+    )
+
+    ds = ctx.obj.flow_datastore.get_task_datastore(
+        mode="w",
+        run_id=kwargs["run_id"],
+        step_name=step_name,
+        task_id=kwargs["task_id"],
+        attempt=int(retry_count),
+    )
+    stdout_location = ds.get_log_location(TASK_LOG_SOURCE, "stdout")
+    stderr_location = ds.get_log_location(TASK_LOG_SOURCE, "stderr")
+
+    def _sync_metadata():
+        if ctx.obj.metadata.TYPE == "local":
+            sync_local_metadata_from_datastore(
+                DATASTORE_LOCAL_DIR,
+                ctx.obj.flow_datastore.get_task_datastore(
+                    kwargs["run_id"], step_name, kwargs["task_id"]
+                ),
+            )
+
+    try:
+        with ctx.obj.monitor.measure("metaflow.awslambda.launch"):
+            return_code = runner.run(
+                step_name,
+                step_cli,
+                code_package_sha,
+                code_package_url,
+                ctx.obj.flow_datastore.TYPE,
+                env=env,
+                attrs=common_attrs.to_dict(key_prefix="metaflow."),
+                echo=echo_without_prefix,
+                stdout_location=stdout_location,
+                stderr_location=stderr_location,
+                task_id=kwargs["task_id"],
+                run_id=kwargs["run_id"],
+                attempt=str(retry_count),
+                flow_name=ctx.obj.flow.name,
+            )
+            task_datastore = FlowDataStore(
+                ctx.obj.flow.name,
+                ctx.obj.environment,
+                ctx.obj.metadata,
+                ctx.obj.event_logger,
+                ctx.obj.monitor,
+            ).get_task_datastore(kwargs["run_id"], step_name, kwargs["task_id"])
+            _sync_metadata()
+            if return_code != 0:
+                sys.exit(1)
+    except LambdaRuntimeException as e:
+        echo_with_prefix(
+            "Lambda runtime exception (possibly an issue with the container image)",
+            True,
+        )
+        echo_with_prefix(str(e), True)
+        task_datastore = FlowDataStore(
+            ctx.obj.flow.name,
+            ctx.obj.environment,
+            ctx.obj.metadata,
+            ctx.obj.event_logger,
+            ctx.obj.monitor,
+        ).get_task_datastore(kwargs["run_id"], step_name, kwargs["task_id"])
+        _sync_metadata()
+        sys.exit(METAFLOW_EXIT_DISALLOW_RETRY)
+    except Exception as e:
+        traceback.print_exc()
+        task_datastore = FlowDataStore(
+            ctx.obj.flow.name,
+            ctx.obj.environment,
+            ctx.obj.metadata,
+            ctx.obj.event_logger,
+            ctx.obj.monitor,
+        ).get_task_datastore(kwargs["run_id"], step_name, kwargs["task_id"])
+        _sync_metadata()
+        sys.exit(METAFLOW_EXIT_DISALLOW_RETRY)
--- a/metaflow/plugins/aws/awslambda/lambda_decorator.py
+++ b/metaflow/plugins/aws/awslambda/lambda_decorator.py
@@ -1,0 +1,322 @@
+import os
+import sys
+
+from metaflow import R
+
+from metaflow.decorators import Decorator, StepDecorator
+from metaflow.exception import MetaflowException
+from metaflow.plugins.aws.awslambda.lambda_runner import LAMBDA_MAX_MEMORY_MB
+from metaflow.plugins.aws.batch.batch_decorator import (
+    BatchDecorator,
+    ResourcesDecorator,
+)
+from metaflow.metadata.util import sync_local_metadata_to_datastore
+from metaflow.plugins.conda.conda_step_decorator import CondaStepDecorator
+from metaflow.plugins.timeout_decorator import get_run_time_limit_for_task
+from .lambda_runner import ensure_lambda, lambda_name
+from metaflow.sidecar import SidecarSubProcess
+from metaflow.metadata import MetaDatum
+from ..aws_utils import get_docker_registry, compute_resource_attributes
+
+
+if sys.version_info > (3, 0):
+    from typing import TYPE_CHECKING, Dict, List, Optional, Callable
+
+    if TYPE_CHECKING:
+        from metaflow.runtime import CLIArgs
+        from metaflow.graph import FlowGraph
+        from metaflow.package import MetaflowPackage
+        from metaflow.flowspec import FlowSpec
+        from metaflow.datastore import FlowDataStore
+        from metaflow.datastore.task_datastore import TaskDataStore
+        from metaflow.metaflow_environment import MetaflowEnvironment
+        from metaflow.event_logger import EventLogger
+
+from metaflow.metaflow_config import (
+    LAMBDA_CONTAINER_REGISTRY,
+    LAMBDA_CONTAINER_IMAGE,
+    LAMBDA_ROLE_ARN,
+    DATASTORE_LOCAL_DIR,
+)
+
+DEFAULT_TIMEOUT_SECONDS = 900
+LAMBDA_DEFAULT_MEMORY_MB = 4096
+
+
+class LambdaException(MetaflowException):
+    headline = "AWS Lambda error"
+
+
+class LambdaDecorator(StepDecorator):
+    """
+    The decorator is called @awslambda to avoid confusion, since "lambda" is a reserved
+    word in Python.
+    """
+
+    name = "awslambda"
+    defaults = {
+        "image": None,
+        "memory": None,
+        "refresh_image": False,
+    }
+    resource_defaults = {
+        "memory": "4096",
+    }
+
+    package_url = None  # type: Optional[str]
+    package_sha = None  # type: Optional[str]
+    memory_mb = LAMBDA_DEFAULT_MEMORY_MB  # type: int
+    run_time_limit = None  # type: Optional[int]
+    ds_root = None  # type: Optional[str]
+    lambda_arn = None  # type: Optional[str]
+    lambda_name = None  # type: Optional[str]
+
+    def __init__(self, attributes=None, statically_defined=False):
+        super(LambdaDecorator, self).__init__(attributes, statically_defined)
+
+        if not self.attributes["image"]:
+            # If metaflow-config specifies a docker image, just use that.
+            if LAMBDA_CONTAINER_IMAGE:
+                self.attributes["image"] = LAMBDA_CONTAINER_IMAGE
+            # If metaflow-config doesn't specify a docker image, throw an error.
+            # Unlike with other platforms, AWS Lambda needs a special image
+            # and it has to be uploaded in users ECR repo, so we cannot provide
+            # a good default here.
+            else:
+                raise LambdaException(
+                    "The *@awslambda* decorator requires either image parameter or "
+                    "METAFLOW_LAMBDA_CONTAINER_IMAGE to be set to a lambda container image uri"
+                )
+        if not get_docker_registry(self.attributes["image"]):
+            if LAMBDA_CONTAINER_REGISTRY:
+                self.attributes["image"] = "%s/%s" % (
+                    LAMBDA_CONTAINER_REGISTRY.rstrip("/"),
+                    self.attributes["image"],
+                )
+
+    def step_init(
+        self,
+        flow,  # type: "FlowSpec"
+        graph,  # type : "FlowGraph"
+        step,  # type: str
+        decos,  # type: List[Decorator]
+        environment,  # type: "MetaflowEnvironment"
+        flow_datastore,  # type: "FlowDataStore"
+        logger,  # type: "Callable[[str], None]"
+    ):
+        if flow_datastore.TYPE != "s3":
+            raise LambdaException("The *@awslambda* decorator requires --datastore=s3.")
+
+        self.logger = logger
+        self.environment = environment
+        self.step = step
+        self.flow_datastore = flow_datastore
+
+        for deco in decos:
+            if isinstance(deco, CondaStepDecorator):
+                raise LambdaException(
+                    "*@awslambda* doesn't support using @conda. "
+                    "We recommend using custom container images to manage"
+                    " dependencies instead"
+                )
+
+            if isinstance(deco, BatchDecorator):
+                raise LambdaException(
+                    "You can't use @batch and *@awslambda* at the same time"
+                )
+
+        resource_attrs = compute_resource_attributes(
+            decos, self, self.resource_defaults
+        )
+        if int(resource_attrs["memory"]) > LAMBDA_MAX_MEMORY_MB:
+            raise LambdaException(
+                "{r} cannot exceed {max_v} for *@awslambda* (currently set to {v})".format(
+                    r="memory", max_v=LAMBDA_MAX_MEMORY_MB, v=resource_attrs["memory"]
+                )
+            )
+
+        # ignore all other resource attrs, we only support setting memory for lambda
+        self.memory_mb = int(resource_attrs["memory"])
+
+        # Run time limit more than 15m is effectively ignored
+        self.run_time_limit = min(
+            get_run_time_limit_for_task(decos),
+            900,
+        )
+
+        if self.run_time_limit < 10:
+            raise LambdaException(
+                "The timeout for step *{step}* should be at "
+                "least 10 seconds for execution on AWS Lambda".format(step=step)
+            )
+
+    def runtime_step_cli(
+        self,
+        cli_args,  # type: "CLIArgs"
+        retry_count,  # type: int
+        max_user_code_retries,  # type: int
+        ubf_context,
+    ):
+        # After all attempts to run the user code have failed, we don't need
+        # Lambda anymore. We can execute possible fallback code locally, so we'll
+        # leave args as is
+        if retry_count <= max_user_code_retries:
+            cli_args.commands = ["awslambda", "step"]
+            cli_args.command_args.append(self.package_sha)
+            cli_args.command_args.append(self.package_url)
+            cli_args.command_options["lambda_arn"] = self.lambda_arn
+            cli_args.command_options["lambda_name"] = self.lambda_name
+            if not R.use_r():
+                cli_args.entrypoint[0] = sys.executable
+
+    @classmethod
+    def _save_package_once(
+        cls,
+        flow_datastore,  # type: "FlowDataStore"
+        package,  # type: "MetaflowPackage"
+    ):
+        if cls.package_url is None:
+            cls.package_url, cls.package_sha = flow_datastore.save_data(
+                [package.blob], len_hint=1
+            )[0]
+
+    def package_init(
+        self,
+        flow,  # type: "FlowSpec"
+        step_name,  # type: str
+        environment,  # type: "MetaflowEnvironment"
+        echo,  # type: Callable[..., None]
+    ):
+        # self.step_init(..) should've run by now and populated this and other
+        # attributes. This assert makes mypy happy as it cannot infer that statically.
+        assert self.run_time_limit is not None
+
+        self.lambda_name = lambda_name(
+            self.memory_mb, self.run_time_limit, self.attributes["image"]
+        )
+
+        def echo_with_prefix(msg, err):  # type: (str, bool) -> None
+            echo(
+                "[awslambda decorator] %s" % (msg,),
+                err=err,
+            )
+
+        if LAMBDA_ROLE_ARN is None:
+            raise LambdaException(
+                "The *@lambda* decorator requires METAFLOW_LAMBDA_ROLE_ARN to be set to a "
+                "lambda role arn"
+            )
+
+        self.lambda_arn = ensure_lambda(
+            name=self.lambda_name,
+            memory_mb=self.memory_mb,
+            timeout_seconds=self.run_time_limit,
+            role_arn=LAMBDA_ROLE_ARN,
+            image_uri=self.attributes["image"],
+            refresh_image=self.attributes["refresh_image"],
+            echo=echo_with_prefix,
+        )
+
+    def runtime_init(
+        self,
+        flow,  # type: "FlowSpec"
+        graph,  # type: "FlowGraph"
+        package,  # type: "MetaflowPackage"
+        run_id,  # type: str
+    ):
+        self.flow = flow
+        self.graph = graph
+        self.package = package
+        self.run_id = run_id
+
+    def runtime_task_created(
+        self,
+        task_datastore,  # type: "TaskDataStore"
+        task_id,  # type: str
+        split_index,  # type: Optional[int]
+        input_paths,  # type: List[str]
+        is_cloned,  # type: bool
+        ubf_context,
+    ):
+        if not is_cloned:
+            self._save_package_once(self.flow_datastore, self.package)
+
+    def task_pre_step(
+        self,
+        step_name,
+        task_datastore,
+        metadata,
+        run_id,
+        task_id,
+        flow,
+        graph,
+        retry_count,
+        max_retries,
+        ubf_context,
+        inputs,
+    ):
+        self.metadata = metadata
+        self.task_datastore = task_datastore
+
+        meta = {}
+        if "AWS_LAMBDA_FUNCTION_NAME" in os.environ:
+            meta["aws-lambda-function-name"] = os.environ["AWS_LAMBDA_FUNCTION_NAME"]
+            meta["aws-lambda-log-group-name"] = os.environ["AWS_LAMBDA_LOG_GROUP_NAME"]
+            meta["aws-lambda-log-stream-name"] = os.environ[
+                "AWS_LAMBDA_LOG_STREAM_NAME"
+            ]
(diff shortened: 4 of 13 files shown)


romain-intel:
Thank you for this. I will try to take a look shortly.

extras/awslambda_image/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM public.ecr.aws/lambda/python:3.7
+
+RUN curl -1sLf \
+    'https://repositories.timber.io/public/vector/cfg/setup/bash.rpm.sh' \
+    | bash
+
+RUN yum install -y vector wget
+
+RUN pip install awslambdaric metaflow awscli
```
savingoyal, line 9: Do we need to install `awslambdaric` if the lambda python image is used? Curious if there is a way to somehow not require `runtime interface client` in the image - that way any arbitrary image can be used for both AWS Batch & AWS Lambda : say the entry point installs `awslambdaric` first and then invokes `python -m awslambdaric`?

metaflow/plugins/aws/awslambda/lambda_decorator.py
```diff
@@ -0,0 +1,213 @@
...
+    from metaflow.event_logger import EventLogger
+
+from metaflow.metaflow_config import (
+    LAMBDA_LOG_GROUP_NAME,
+    LAMBDA_ROLE_ARN,
+    LAMBDA_IMAGE_URI,
+)
+
+# Supported resource specs for Lambda
+LAMBDA_MAX_RESOURCES = {
```
savingoyal, line 33: I am not sure if hard coding these values (and runtime limit checks / timeouts) is a great idea since these are going to change with time as AWS releases new capabilities for AWS Lambda.

metaflow/plugins/aws/awslambda/lambda_decorator.py
```diff
@@ -0,0 +1,213 @@
...
+            cls.package_sha = package.sha
+
+    def runtime_init(
+        self,
+        flow: "FlowSpec",
+        graph: "FlowGraph",
+        package: "MetaflowPackage",
+        run_id: str,
+    ):
+        if LAMBDA_LOG_GROUP_NAME is None:
```
savingoyal, line 156: Curious, what is the reasoning behind making `LOG_GROUP_NAME` user configurable?

metaflow/plugins/aws/awslambda/lambda_decorator.py
```diff
@@ -0,0 +1,213 @@
...
+                "set to a Cloudwatch log group name"
+            )
+
+        if LAMBDA_ROLE_ARN is None:
+            raise LambdaException(
+                "The *@lambda* decorator requires METAFLOW_LAMBDA_ROLE_ARN to be set to a "
+                "lambda role arn"
+            )
+
+        if LAMBDA_IMAGE_URI is None:
```
savingoyal, line 168: If no image is specified, we can default to a `python` image.

metaflow/plugins/aws/awslambda/lambda_decorator.py
```diff
@@ -0,0 +1,213 @@
...
+        if datastore.TYPE != "s3":
+            raise LambdaException("The *@awslambda* decorator requires --datastore=s3.")
+
+        self.logger = logger
+        self.environment = environment
+        self.step = step
+        self.resource_attributes = {}
+        for deco in decos:
+            if isinstance(deco, CondaStepDecorator):
+                raise LambdaException(
```
savingoyal, line 81: What would it take to support `@conda` in `@awslambda`?

metaflow/plugins/aws/awslambda/lambda_decorator.py
```diff
@@ -0,0 +1,213 @@
...
+    headline = "AWS Lambda error"
+
+
+class LambdaDecorator(StepDecorator):
+    """
+    The decorator is called @awslambda to avoid confusion, since "lambda" is a reserved
+    word in Python.
+    """
+
+    name = "awslambda"
```
savingoyal, line 50: thoughts on using `@lambda` instead of `@awslambda` to keep it consistent with `@batch`?

metaflow/plugins/aws/awslambda/lambda_runner.py
```diff
@@ -0,0 +1,314 @@
...
+                        ):
+                            seen_start_token = True
+                        elif seen_start_token:
+                            self._echo(line)
+                else:
+                    if self._stop_deadline and self._stop_deadline < time.time():
+                        return
+
+
+def lambda_name(flow_name, step_name):
```
savingoyal, line 91: What if we constructed the `lambda_name` to be `flow_name-run_id-step_name-task_id-attempt_id` instead? That will also negate the need for introducing `force_fetch_image`, I believe?

metaflow/plugins/aws/awslambda/lambda_runner.py
```diff
@@ -0,0 +1,314 @@
...
+
+    # Compute Lambda memory_size.
+    #
+    # For Lambda, you can't specify CPU and mem independently, you always get
+    # 1769 (MB/vCPU) [1]. In @resources deco, we allow users to specify cpu or mem or
+    # both. So when it comes to lambda, we'll set memory_size to a value that would
+    # satisfy the max of both requirements.
+    #
+    # [1] https://docs.aws.amazon.com/lambda/latest/dg/configuration-memory.html
+    if cpu is not None and memory_mb is not None:
```
savingoyal, line 124: This calculation runs the risk of becoming stale as AWS introduces new resource instances for Lambda. Is there a need for us to do this validation? Given that launching an AWS Lambda function is quick and easy - we can simply launch the function with the user-defined resources and return a properly formatted error if the values are incorrect.

metaflow/plugins/aws/awslambda/lambda_runner.py
```diff
@@ -0,0 +1,314 @@
...
+    else:
+        memory_size = 2048
+
+    # Since multiple tasks in a foreach branch reuse the same Lambda, they all might
+    # get to this point simultaneously and try to update lambda configuration at the
+    # same time. For simplicity, we don't do any sychronization between them, and simply
+    # retry on conflict, as these updates should be idempotent anyway.
+    num_retries_on_conflict = 3
+
+    for retry in range(num_retries_on_conflict):
```
savingoyal, line 139: Will this logic work when I have two concurrent executions of a flow where the flow structure is the same but the resource specification is different for the same step?

metaflow/plugins/aws/sync_metadata.py
```diff
@@ -0,0 +1,62 @@
+import os
```
savingoyal, line 1: Minor nit: can we rename the file to `plugins/aws/utils.py`

metaflow/plugins/aws/awslambda/lambda_runner.py
```diff
@@ -0,0 +1,314 @@
+import shlex
+from threading import Thread
+import boto3
+import json
+from typing import Dict, Optional, TYPE_CHECKING, Callable
```
savingoyal, line 5: It would be great to find a way to make AWS Lambda compatible with Py2 as well if it's not too much work.

metaflow/plugins/aws/awslambda/lambda_cli.py
```diff
@@ -0,0 +1,202 @@
...
+    "job with the given options. Typically you do not "
+    "call this command directly; it is used internally "
+    "by Metaflow."
+)
+@click.argument("step-name")
+@click.argument("code-package-sha")
+@click.argument("code-package-url")
+@click.option("--cpu", type=float, help="CPU requirement for AWS Lambda.")
+@click.option("--memory", type=int, help="Memory requirement for AWS Lambda.")
+@click.option(
```
savingoyal, line 48: Can we set this option as a global environment variable in `metaflow_config` instead? I presume the value of `throttle_retries` will stay the same across all executions for a user?

savingoyal (changes requested):
@oavdeev Great work! I took a quick look at the PR and noted some comments. Please take a look and let me know how you feel about any of those. I am yet to test the PR - I will follow up with more comments when I get to that 😃

extras/awslambda_image/entrypoint.py
```diff
@@ -0,0 +1,88 @@
...
+    stream_name = event["parameters"]["metaflow.awslambda.log_stream"]
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        vector_config = os.path.join(tmp_dir, "vector.yaml")
+
+        os.mkdir(os.path.join(tmp_dir, "vector_data"))
+
+        region = os.environ["AWS_REGION"]
+
+        # Create a vector config to pipe logs to Cloudwatch
+        with open(vector_config, "wt") as f:
```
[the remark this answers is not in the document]
oavdeev: They would, but finding them there is tricky. That CW stream name is determined by Lambda instance id that you don't control (instance = concrete running container), and multiple lambda invocations from the same instance go to the same stream. AFAIK AWS exposes no way to find the instance id corresponding to your lambda invoke() call.

In other words, the logs are indeed there, you only reliably know the Log Group name, and finding the logs for the concrete invocation would require multiple API calls (basically scanning all log streams).

The idea with vector here is to instead stream the logs to a CW log stream with a known name, separate for each task.

Anyway, `vector` requirement will go away once we merge that mflog PR

metaflow/plugins/aws/awslambda/lambda_decorator.py
oavdeev: They're mostly to have a nicer error message, and be able to show that error _before_ your flow is executed. Everything still works if we remove them. But i figured if e.g. the last step in a long flow has incorrect limits, it would be not so great UX to get an error only when you get there.

metaflow/plugins/aws/awslambda/lambda_decorator.py
oavdeev: No huge reason, I imagine there could be people out there who have internal rules on naming the resources like CW group names.

metaflow/plugins/aws/awslambda/lambda_decorator.py
oavdeev: One idea (mentioned it briefly in the PR decription) is to build custom docker images from conda envs, push them to ECR and use them to create Lambdas.

metaflow/plugins/aws/awslambda/lambda_decorator.py
oavdeev: That's what i had at first but `lambda` is a reserved word in Python

metaflow/plugins/aws/awslambda/lambda_runner.py
oavdeev: Yes it would, but it would add 5-30 seconds of extra latency to create the Lambda for each task. One of the goals of this PR was to add a way to run your code in the cloud with minimal latency so it feels delightful, so I was optimizing for lower startup time whenever I could here.

metaflow/plugins/aws/awslambda/lambda_runner.py
oavdeev: Re: validation, see the other comment. 

As for this math, lambda doesn't allow you to specify vCPUs directly, only memory, and you get X cpu for Y memory. I figured it would be nice if we still allowed users to specify CPU using `@resources` and did the math for them. The alternative is that we could say that `@resources` deco only supports specifying memory size for lambda, and we can remove this math. But users would have to do it anyway, and it is not trivial to dig it out of AWS docs.

metaflow/plugins/aws/awslambda/lambda_runner.py
oavdeev: Nope, you're right, that would lead to weirdness. Those two executions would concurrently try to change Lambda resources and one of them would likely win

extras/awslambda_image/Dockerfile
oavdeev: You're right, we probably don't have to install it if we use the Lambda python image.

As for using an arbitrary image and installing `awslambdaric` on the fly, it [seems like](https://docs.aws.amazon.com/lambda/latest/dg/images-create.html#images-create-2) `awslambdaric` is not a pure Python package, so you'd have to have the gcc toolchain there as well. That seems like it could become pretty fragile. (I also wonder if AWS would kill the container prematurely if the runtime takes too long to initialize).

metaflow/plugins/aws/awslambda/lambda_decorator.py
savingoyal: Maybe, we can consider using [lambda layers](https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html) rather than building images on the fly - which will presumably need more infrastructure. That can be a follow-on PR.

metaflow/plugins/aws/awslambda/lambda_decorator.py
savingoyal: It might be good to rely on a default. Also, perhaps if the more recent comment in L:24 in `extras/awslambda_image/entrypoint.py` is amenable to you, we can avoid making the `LOG_GROUP_NAME` configurable at this moment. We haven't had any user requests so far for configuring the `LOG_GROUP_NAME` for AWS Batch executions.

metaflow/plugins/aws/awslambda/lambda_runner.py
savingoyal: We would have to necessarily isolate the `lambda` to at least the `run_id` which would mean that there is no specific need for `force_fetch_image`.

metaflow/plugins/aws/awslambda/lambda_runner.py
savingoyal: I think we can simply allow our users to only specify `memory`. As AWS changes the math on its end, it _might_ be tricky for us to guarantee backward compatibility.

metaflow/plugins/aws/awslambda/lambda_runner.py
savingoyal: Yeah, we should make sure we properly handle such scenarios.

extras/awslambda_image/Dockerfile
oavdeev: Gotcha, I see, so the scenario you have in mind are users using Batch today and switching to Lambda?

One additional quirk to note here is that unlike e.g. Batch, your image for Lambda has to be a ECR image. Even [public ECR repo](https://docs.aws.amazon.com/AmazonECR/latest/public/public-repositories.html) doesn't cut it. 

So it seems like it would only benefit users that satisfy _all_ of the below conditions:
1) Use batch for these tasks today and want to move to `@awslambda`
2) ..and use custom images 
3) ..and store those images in ECR
4) ..and those images happen to have gcc toolchain to build awslambdaric

I'm not sure it is worth it, given that on-the-fly installation would be a bit hacky, and it would have to finish under undocumented deadline of what seems to be ~10s, ( [see this thread](https://stackoverflow.com/questions/55412624/does-aws-lambda-charge-for-the-time-spent-initializing-code/55426800) ) or possibly trigger more weird behaviors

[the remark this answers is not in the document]
oavdeev: done in c799b10

metaflow/plugins/aws/awslambda/lambda_runner.py
oavdeev: removed the cpu->memory_size math in c799b10

metaflow/plugins/aws/awslambda/lambda_cli.py
oavdeev: done in c799b10

metaflow/plugins/aws/awslambda/lambda_runner.py
oavdeev: fixed in the last version

metaflow/plugins/aws/awslambda/lambda_decorator.py
oavdeev: unfortunately you can't use public images with Lambda, so the user would have to set this

extras/awslambda_image/Dockerfile
oavdeev: Discussed offline, we can't install awslambdaric on the fly as it has to be the entrypoint of the image - closing

extras/awslambda_image/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM public.ecr.aws/lambda/python:3.7
```
savingoyal, line 1: We can move `extras` to Metaflow docs and metaflow-tools.

metaflow/metaflow_config.py
```diff
@@ -125,6 +125,13 @@ def from_conf(name, default=None):
 # sandbox.
 SFN_STATE_MACHINE_PREFIX = from_conf("METAFLOW_SFN_STATE_MACHINE_PREFIX")
 
+###
+# AWS Lambda configuration
+###
+LAMBDA_ROLE_ARN = from_conf("METAFLOW_LAMBDA_ROLE_ARN")
+LAMBDA_IMAGE_URI = from_conf("METAFLOW_LAMBDA_IMAGE_URI")
+LAMBDA_THROTTLE_RETRIES = int(from_conf("METAFLOW_LAMBDA_THROTTLE_RETRIES", 20))
```
savingoyal, line 133: What is the utility of this variable? I see that it's used to specify `max-attempts` for the lambda invocation. We should rely on the value obtained from `@retry` instead.

metaflow/mflog/__init__.py
```diff
@@ -1,4 +1,11 @@
+import sys
 import math
+import time
+import select
+
+
+if (sys.version_info > (3, 0)):
+    from typing import Callable
```
savingoyal, line 8: Given that we are not using types at the moment, can we dispense with this import?

metaflow/mflog/mflog.py
```diff
@@ -113,11 +113,9 @@ def refine(line, prefix=None, suffix=None):
     prefix = to_bytes(prefix) if prefix else b''
     suffix = to_bytes(suffix) if suffix else b''
     parts = line.split(b']', 1)
-    if len(parts) == 2:
-        header, body = parts
-        return b''.join((header, b']', prefix, body, suffix))
-    else:
-        return line
+    assert len(parts) == 2
```
savingoyal, line 116: I am not a big fan of asserts. We can probably leave this segment of the code as-is for this PR.

metaflow/plugins/aws/awslambda/lambda_cli.py
```diff
@@ -0,0 +1,204 @@
...
+    "by Metaflow."
+)
+@click.argument("step-name")
+@click.argument("code-package-sha")
+@click.argument("code-package-url")
+@click.option("--memory", type=int, help="Memory requirement for AWS Lambda.")
+@click.option("--run-id", help="Passed to the top-level 'step'.")
+@click.option("--task-id", help="Passed to the top-level 'step'.")
+@click.option("--image", help="Lambda image")
+@click.option(
```
savingoyal, line 49: I don't think `force_fetch_image` is needed any longer.

metaflow/plugins/aws/awslambda/lambda_cli.py
```diff
@@ -0,0 +1,204 @@
...
+    def echo(msg, err):  # type (str, bool) -> None
+        ctx.obj.echo_always("%s" % (msg,), err=err)
+
+    if ctx.obj.datastore.datastore_root is None:
+        ctx.obj.datastore.datastore_root = (
+            ctx.obj.datastore.get_datastore_root_from_config(echo)
+        )
+
+    if R.use_r():
+        entrypoint = R.entrypoint()
```
savingoyal, line 94: To support R, we would need to make corresponding changes in `metaflow/R`

metaflow/plugins/aws/awslambda/lambda_cli.py
```diff
@@ -0,0 +1,204 @@
...
+    else:
+        executable = ctx.obj.environment.executable(step_name)
+        entrypoint = "%s -u %s" % (executable, os.path.basename(sys.argv[0]))
+
+    top_args = " ".join(util.dict_to_cli_options(ctx.parent.parent.params))
+
+    input_paths = kwargs.get("input_paths")
+    split_vars = None
+    if input_paths:
+        max_size = 30 * 1024
```
savingoyal, line 104: Do we have the same limits in AWS Lambda as we have in AWS Batch?

metaflow/plugins/aws/awslambda/lambda_cli.py
```diff
@@ -0,0 +1,204 @@
...
+    # Add the environment variables related to the input-paths argument
+    if split_vars:
+        env.update(split_vars)
+
+    config = Config(
+        retries={
+            "max_attempts": LAMBDA_THROTTLE_RETRIES,
+        }
+    )
+    lambda_client = boto3.client("lambda", config=config)
```
savingoyal, line 149: We should use [`get_aws_client`](https://github.com/Netflix/metaflow/blob/002e7fd5c3be196e95222dce1972020f1f41d192/metaflow/plugins/aws/aws_client.py#L4) here

metaflow/plugins/aws/awslambda/lambda_cli.py
```diff
@@ -0,0 +1,204 @@
...
+        flow_name=ctx.obj.flow.name,
+        step_name=step_name,
+        memory_mb=memory,
+        timeout_seconds=run_time_limit,
+        role_arn=LAMBDA_ROLE_ARN,
+        image_uri=image or LAMBDA_IMAGE_URI,
+        force_fetch_image=force_fetch_image,
+        echo=echo,
+    )
+    runner = LambdaRunner(
```
savingoyal, line 161: Given that the `lambda_client` is only used by the `LambdaRunner` - would it make sense to instantiate it within the `LambdaRunner` itself?

metaflow/plugins/aws/awslambda/lambda_decorator.py
```diff
@@ -0,0 +1,208 @@
...
+        from metaflow.datastore import MetaflowDataStore
+        from metaflow.metaflow_environment import MetaflowEnvironment
+        from metaflow.event_logger import EventLogger
+
+from metaflow.metaflow_config import (
+    LAMBDA_IMAGE_URI,
+    LAMBDA_ROLE_ARN,
+)
+
+DEFAULT_TIMEOUT_SECONDS = 600
```
savingoyal, line 34: This is already available from https://github.com/Netflix/metaflow/pull/463/files#diff-cfe17a874aba5a08a3ed0ae7a9d22caa5eb63caa96bf392481edbdee323fc073R68

metaflow/plugins/aws/awslambda/lambda_decorator.py
```diff
@@ -0,0 +1,208 @@
...
+class LambdaDecorator(StepDecorator):
+    """
+    The decorator is called @awslambda to avoid confusion, since "lambda" is a reserved
+    word in Python.
+    """
+
+    name = "awslambda"
+    defaults = {
+        "image": None,
+        "force_fetch_image": False,
```
savingoyal, line 50: We can dispense with `force_fetch_image` - there is no particularly great way to handle this (I believe you want to handle the latest tag of the image here) and this issue is already present in AWS Batch/K8S etc. as well.

metaflow/plugins/aws/awslambda/lambda_decorator.py
```diff
@@ -0,0 +1,208 @@
...
+            if isinstance(deco, CondaStepDecorator):
+                raise LambdaException(
+                    "*@awslambda* doesn't support using @conda. "
+                    "We recommend using custom docker images to manage"
+                    " dependencies instead"
+                )
+
+            if isinstance(deco, BatchDecorator):
+                raise LambdaException(
+                    "You can't use @batch and *@awslambda* at the same time"
```
savingoyal, line 85: We need to move this check outside this routine and check for the validity of the flow before the flow starts executing.

metaflow/plugins/aws/awslambda/lambda_decorator.py
```diff
@@ -0,0 +1,208 @@
...
+                )
+
+            if isinstance(deco, ResourcesDecorator):
+                for k, v in deco.attributes.items():
+                    if v != deco.defaults[k]:
+                        if k != "memory":
+                            raise LambdaException(
+                                "*@awslambda* only supports provisioning memory via @resources, not {r}".format(
+                                    r=k
```
savingoyal, line 94: I think rather than throwing an error, we can just ignore `cpu` etc. Makes it easier for folks to float between batch and lambda that way. Thoughts?

metaflow/plugins/aws/awslambda/lambda_decorator.py
```diff
@@ -0,0 +1,208 @@
...
+                                )
+                            )
+                        else:
+                            if float(v) > LAMBDA_MAX_MEMORY_MB:
```
savingoyal, line 98: Given that the job will fail quickly if the user asks for more than the allowed memory limit, we can get rid of this check as well. As AWS raises the limits, we wouldn't have to upgrade our logic.

metaflow/plugins/aws/awslambda/lambda_decorator.py
```diff
@@ -0,0 +1,208 @@
...
+                        self.resource_attributes[k] = str(v)
+        self.run_time_limit = get_run_time_limit_for_task(
+            decos, default_limit=DEFAULT_TIMEOUT_SECONDS
+        )
+        if self.run_time_limit < 10:
+            raise LambdaException(
+                "The timeout for step *{step}* should be at "
+                "least 10 seconds for execution on AWS Lambda".format(step=step)
+            )
+        if self.run_time_limit > 900:
```
savingoyal, line 113: Ditto as the above comment.

metaflow/plugins/aws/awslambda/lambda_runner.py
```diff
@@ -0,0 +1,291 @@
...
+
+    # Since multiple tasks in a foreach branch reuse the same Lambda, they all might
+    # get to this point simultaneously and try to update lambda configuration at the
+    # same time. For simplicity, we don't do any sychronization between them, and simply
+    # retry on conflict, as these updates should be idempotent anyway.
+    num_retries_on_conflict = 3
+
+    for retry in range(num_retries_on_conflict):
+        try:
+            return _maybe_update_lambda(
```
savingoyal, line 121: I thought we don't need this code segment anymore?

savingoyal (changes requested):
Still reviewing - here are some partial comments.

metaflow/plugins/aws/awslambda/lambda_runner.py
```diff
@@ -0,0 +1,291 @@
...
+                timeout_seconds,
+                memory_size,
+                role_arn,
+                image_uri,
+                force_fetch_image,
+                echo=echo,
+            )
+        except client.exceptions.ResourceConflictException:
+            if retry == num_retries_on_conflict - 1:
+                print("ENOUGH CONFLICT RETRIES")
```
oavdeev, line 133: forgot to remove debug print here

metaflow/metaflow_config.py
oavdeev: This parameter controls retries in case we get throttled by AWS due to Lambda parallel execution limits. I assume users will rarely need to touch this, but it is still nice to have an option. `@retry` is still respected and controls retries in case of step failure.

savingoyal: Ah got it! Can we make a note in the code base about this behavior so that it doesn't throw off future committers?

metaflow/mflog/__init__.py
oavdeev: The types are used below in `delayed_update_while`. Even though it is using the py2-compatible comment syntax (instead of py3 typing annotations), the type checker still needs the import to work. Really helps if you e.g. use vscode and have type checking enabled.

metaflow/plugins/aws/awslambda/lambda_cli.py
oavdeev: no good point i should remove this

metaflow/mflog/mflog.py
oavdeev: That's the bit we talked about offline, it seems like this is always called with mflog-formatted line so that `else` branch doesn't get called anyway. I can remove the assert itself given that it will fail on the next line anyway tho

metaflow/plugins/aws/awslambda/lambda_runner.py
oavdeev: We do since there may be a race condition where multiple branches of foreach try to create the same lambda simultaneously. It is much simpler than the previous version though.

metaflow/plugins/aws/awslambda/lambda_decorator.py
oavdeev: Yes, this is mostly for the `:latest`. I'd have something similar for k8s and its `imagepullpolicy`, not sure how to handle AWS Batch tho.

Note that with Batch the problem is less acute, since you can hope that _eventually_ the new image will be used. With lambda, the image is copied by AWS when you create the function. So without this, there is just no way to trigger it to refetch the image, you'll be stuck with the old image forever. It was very annoying even when I was testing this on toy examples.

metaflow/plugins/aws/awslambda/lambda_decorator.py
oavdeev: I agree but what's a good place for this? At first I looked at `FlowLinter` but it doesn't support plugins today

metaflow/plugins/aws/awslambda/lambda_decorator.py
oavdeev: Hmm good point, but it may be confusing for some if CPU and (especially) GPU settings are silently ignored but memory is not.

metaflow/plugins/aws/awslambda/lambda_decorator.py
oavdeev: I didn't have it at first but then I realized that doesn't fail _that_ quickly, it is only when it gets to that step you get an error. And it is a big ugly stacktrace. I think it makes a not great UX, and AWS only updated these limits twice in history, so in this I think it is worth it to put it on us to update the limits when it happens again. I'd feel differently if it was something that changes more frequently.

I understand that we probably can't do the same for each compute platform. IIRC it is much worse on Batch or k8s where you might get the job randomly stuck due to lack of resources and not get an error at all.

metaflow/mflog/__init__.py
oavdeev: addressed in #504

metaflow/mflog/mflog.py
oavdeev: addressed in #504

metaflow/plugins/aws/awslambda/lambda_cli.py
oavdeev: Discussed offline -- will leave it in but as an advanced feature that we won't advertise. Normally users are expected to build an image under a new tag when it changes

oavdeev: ..and rename to `refresh_image`

metaflow/plugins/aws/awslambda/lambda_decorator.py
oavdeev: k will ignore it if the limit it too high to be consistent

oavdeev:
Rebased this on #504 and addressed PR comments, except two things:
* i can't find a better place to create lambdas than `task_pre_step`. Tried moving it to `step_init` but that gets called _inside_ Lambda too which is not what I want. I could probably work around this but feels like it would unnecessarily complicate logic
* prefixing logs with request id turned out to be suuper cumbersome due to the fact that request id is only available inside the lambda during its execution, so I would have to do it in a hacky way that is pretty different how prefixing is done for `@batch`. So instead I made it prefix logs with just the lambda name. Request id is not super useful anyway.

oavdeev:
@savingoyal made it create lambdas in `package_init`

metaflow/plugins/aws/awslambda/lambda_runner.py
savingoyal: Is this still needed after the move to `package_init`?

metaflow/util.py
```diff
@@ -129,7 +129,7 @@ def to_unicode(x):
         return unicode_type(x)
```
savingoyal, line 129: Minor nit: Can we remove this file from the PR?

metaflow/plugins/aws/awslambda/lambda_decorator.py
```diff
@@ -0,0 +1,264 @@
...
+                            raise LambdaException(
+                                "{r} cannot exceed {max_v} for *@awslambda* (currently set to {v})".format(
+                                    r=k, max_v=LAMBDA_MAX_MEMORY_MB, v=v
+                                )
+                            )
+                        self.memory_mb = v
+
+        # Run time limit more than 15m is effectively ignored
+        self.run_time_limit = min(
+            get_run_time_limit_for_task(decos, default_limit=DEFAULT_TIMEOUT_SECONDS),
```
savingoyal, line 110: Curious what happens if we leave the run-time limit to 5 days? Does AWS Lambda complain?

savingoyal: Also, when does the run-time-limit clock start?

metaflow/plugins/aws/awslambda/lambda_decorator.py
```diff
@@ -0,0 +1,264 @@
...
+            self.ds_root = None
+
+        meta = {}
+        meta["aws-lambda-function-name"] = os.environ["AWS_LAMBDA_FUNCTION_NAME"]
+        meta["aws-lambda-log-group-name"] = os.environ["AWS_LAMBDA_LOG_GROUP_NAME"]
+        meta["aws-lambda-log-stream-name"] = os.environ["AWS_LAMBDA_LOG_STREAM_NAME"]
+        meta["aws-lambda-execution-env"] = os.environ["AWS_EXECUTION_ENV"]
+
+        # Unlike the above vars, this is set by our own code in entrypoint.py
+        meta["aws-lambda-request-id"] = os.environ.get("LAMBDA_REQUEST_ID", "")
```
savingoyal, line 244: Since this is guaranteed to exist - should we just do `os.environ["LAMBDA_REQUEST_ID"]`

metaflow/plugins/aws/awslambda/lambda_runner.py
```diff
@@ -0,0 +1,293 @@
...
+    This does all the necessary AWS API Calls to update or create Lambda. Returns
+    Lambda ARN.
+    """
+    try:
+        # Get current function configuration
+        info = client.get_function(FunctionName=name)
+
+        # If refresh_image is set, force update function code.
+        if refresh_image:
+            echo("Updating image for %s.." % name, False)
```
savingoyal, line 167: Given that this is an internal implementation detail - should we print this? Same comment for L:178.

metaflow/plugins/aws/awslambda/lambda_decorator.py
savingoyal: Why is the default time out set to 10 mins when lambdas run for up to 15 mins?

savingoyal (commented):
I had a few pending comments that I forgot to publish.

metaflow/plugins/aws/awslambda/lambda_runner.py
oavdeev: I kinda went back and forth on this; the main reason for having this is that it can often take about 30 seconds to update or create lambda (if it doesn't exist or `refresh_image` is used). So I wanted to give the user some indication that something is happening, Metaflow isn't magically "stuck".

metaflow/plugins/aws/awslambda/lambda_decorator.py
oavdeev: No strong reason. The default timeout on AWS is 3 seconds but that's too short for us (as we need to download the code too). 

On the other hand, if I try to put some really lengthy computation into a Lambda step, it would arguably be better UX if it failed faster and made me think about my choices and Lambda limitations. One could argue that if 10 minutes is not enough for your code, it is likely that 15 wouldn't be enough either.

So 10 minutes is just a point I picked between 3s and 10m. Maybe it should be even shorter?

savingoyal: hmm.. If I am executing a flow on Lambda and Metaflow kills the job at 10m interval when I know that Lambdas run till 15 mins - that would be a confusing behavior. The only reason for having a timeout is to prevent long-running expensive jobs (AWS Batch timeout is 5 days for that very reason).

metaflow/plugins/aws/awslambda/lambda_decorator.py
savingoyal: I will commit a `PluginLinter` to the codebase to help with this.

metaflow/plugins/aws/awslambda/lambda_runner.py
savingoyal: We can have a different message to address that use case. For example - AWS Batch cycles through the batch status codes so that the user understands that the job is in progress - we print a message when the status changes or if it's been 30s since the last message. We can have a similar behavior here.

savingoyal: Can we have the same(similar) message as we have for AWS Batch?

metaflow/plugins/aws/awslambda/lambda_decorator.py
oavdeev: Yeah you can't set it above 900 seconds. Starts when Lambda is invoked.

metaflow/plugins/aws/awslambda/lambda_runner.py
oavdeev: Hmm the batch is a different beast so the message is different, there is no "job status" in the case of Lambda. And note that is not a direct equivalent of what Batch is doing, you'll only see this message on the first run when Lambda resource needs to be created.

So the copy here is the best I could come up with -- but again, open to concrete alternative suggestions

metaflow/decorators.py
```diff
@@ -208,7 +208,7 @@ def step_init(self, flow, graph, step_name, decorators, environment, datastore,
         """
         pass
 
-    def package_init(self, flow, step_name, environment):
+    def package_init(self, flow, step_name, environment, logger):
```
romain-intel, line 211: Not sure this change is needed here. You can use the logger from step_init IIRC.

romain-intel (commented):
Didn't look in detail at the AWS lambda aspect (I don't know that very well). A few general comments though:
  - the mflog related changes should ideally be extracted to a separate PR (they are not related to AWS per say and it would be nice to have those singled out)
  - you can consider also stripping out the refactor on batch if you want (though that one is less of an issue)
  - if we are to merge this pre convergence branch, please also make the same PR(s) available on the convergence branch; the datastore changes will have an impact on this (no functional difference). We are going to merge back master into convergence soon so maybe after that, it would be good to make a parallel PR chain against that one as well.
 
I will run some tests as well (just quickly looked at it to give a quick initial feedback).

oavdeev:
@romain-intel the mflog changes are already factored out to https://github.com/Netflix/metaflow/pull/504 . Unfortunately Github still doesn't make it easy to manage "stacked" PRs, so this PR branch has mflog PR merged in, and those changes appear in the diff view :/

romain-intel:
Ah ok. Yes, stacked PRs are a pain (boy do I know that). You can do it manually but not as much fun. OK. I'll look at those independently as well and make sure everything still works internally.

oavdeev:
@gatesyp we're pretty close! Just had to pause before some refactorings work for data store gets merged in, which should happen next week. Then the plan is to rebase this, fix some nits and merge

gatesyp:
Ok! Sounds great :) if I'm understanding the PR correctly, this would be pretty important to my group. If there's anything I can do let me know! More than willing to help out

oavdeev:
We're focusing our efforts more on K8S support right now so unfortunately we don't have bandwidth to drive this Lambda PR to conclusion. More importantly, we want to make sure that every feature that is merged into Metaflow core is well supported going forward, e.g. works well with all future orchestrators we may introduce. Realistically we can't commit to this for AWS Lambda even if we merge the PR as is today. 

So I'm going to close this PR since its not actively being worked on.

It is possible that we'll revisit this decision in the future, so feel free to chime in if you really want Lambda or talk to us on Slack.

alamminsalo:
Hi, I know this is closed issue, but are there any plans to implement lambda tasks anymore? It would be a perfect for usecases where computation completes in seconds, avoiding the start time for batchjobs. Atleast more attractive option when there is thousands of light fanned-out jobs

PR #474, opened 2021-04-12
Use boto3 instead of awscli to download code package

tkislan:
Updated PR from https://github.com/Netflix/metaflow/pull/389
Resolves https://github.com/Netflix/metaflow/issues/387

### DIFF
--- a/metaflow/metaflow_environment.py
+++ b/metaflow/metaflow_environment.py
@@ -113,7 +113,7 @@ class MetaflowEnvironment(object):
                 "cd metaflow",
                 "mkdir .metaflow", # mute local datastore creation log
                 "i=0; while [ $i -le 5 ]; do "
-                    "mflog \'Downloading code package.\'; "
+                    "mflog \'Downloading code package...\'; "
                     "%s >/dev/null && \
                         mflog \'Code package downloaded.\' && break; "
                     "sleep 10; i=$((i+1)); "
--- a/metaflow/metaflow_environment.py
+++ b/metaflow/metaflow_environment.py
@@ -92,14 +92,16 @@ class MetaflowEnvironment(object):
         str : Command 
         """
 
-        return (
-            "%s -c \"" % self._python()
-            + "import boto3; "
-            + "exec('try:\\n from urlparse import urlparse\\nexcept:\\n from urllib.parse import urlparse');"
-            + "parsed = urlparse('%s');" % s3_path
-            + "boto3.client('s3').download_file(parsed.netloc, parsed.path.lstrip('/'), '%s');" % local_path
-            + "\""
-        )
+        try:
+            from urlparse import urlparse
+        except ImportError:
+            from urllib.parse import urlparse
+        
+        parsed_s3_path = urlparse(s3_path)
+        bucket = parsed_s3_path.netloc
+        key = parsed_s3_path.path.lstrip('/')
+
+        return "%s -c \'import sys; import boto3; boto3.client(sys.argv[1]).download_file(sys.argv[2], sys.argv[3], sys.argv[4])\' s3 \'%s\' \'%s\' job.tar" % (self._python(), bucket, key)
 
     def get_package_commands(self, code_package_url):
         cmds = ["set -e",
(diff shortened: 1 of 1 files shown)


tkislan:
No idea why the build failed

```
Downloading https://homebrew.bintray.com/bottles/qpdf-10.3.1.catalina.bottle.tar.gz
curl: (22) The requested URL returned error: 403 Forbidden
Error: Failed to download resource "qpdf"
```

savingoyal:
I launched the tests again. Hopefully, it should resolve now. The existing test suite doesn't exercise the files that the PR touches.

savingoyal:
@tkislan Have you tested this PR? I am running into this error - `boto3;: 1: boto3;: Syntax error: end of file unexpected (expecting "done")`

metaflow/metaflow_environment.py
```diff
@@ -76,21 +76,45 @@ def get_client_info(cls, flow_name, metadata):
...
+        Returns
+        -------
+        str : Command 
+        """
+
+        return (
+            "%s -c \"" % self._python()
+            + "import boto3; "
+            + "exec('try:\\n from urlparse import urlparse\\nexcept:\\n from urllib.parse import urlparse');"
+            + "parsed = urlparse('%s');" % s3_path
```
savingoyal, line 99: @tkislan This parsing can happen while we are generating the `get_s3_download_command` locally rather than in the remote environment. That will save us precious entry point chars.

savingoyal (changes requested):

metaflow/metaflow_environment.py
```diff
@@ -76,21 +76,45 @@ def get_client_info(cls, flow_name, metadata):
...
+        local_path : str
+            Path to download S3 file to
+        
+        Returns
+        -------
+        str : Command 
+        """
+
+        return (
+            "%s -c \"" % self._python()
```
savingoyal, line 96: I doubt '\"' would work here given that we use `shlex` downstream for parsing the entry point.

savingoyal (changes requested):

savingoyal:
> I only tested the generated command that downloads the file
> 
> Can you please provide more context?
> What the whole command looks like, etc

The issue is with incorrect quoting given that we use `shlex` downstream. This is what the command looks like - 
```
"/bin/sh","-c","set -e && echo 'Setting up task environment.' && python -m pip install click requests boto3 -qqq && mkdir metaflow && cd metaflow && mkdir .metaflow && i=0; while [ $i -le 5 ]; do echo 'Downloading code package.'; python -c import","boto3;","exec(try:\\n from urlparse import urlparse\\nexcept:\\n from urllib.parse import urlparse);parsed","=","urlparse(s3://oss-dev-metaflows3bucket-13oayj2nh8n4l/metaflow/ForeachFlow2q/data/95/950443769eee3cac5d35c5fa724bf390d5093b2b);boto3.client(s3).download_file(parsed.netloc,parsed.path.lstrip(/),","job.tar); >/dev/null &&                         echo 'Code package downloaded.' && break; sleep 10;...

**Comment by tkislan** [comment]
@savingoyal Thx for the pointer to `shlex` .. Honestly I didn't delve into the coda that much, as I hoped that previous PR had it almost done

This is the new full command that it generates, and I was able to run it locally
```
['/bin/sh', '-c', "set -e && echo 'Setting up task environment.' && python -m pip install click requests boto3 -qqq && mkdir metaflow && cd metaflow && mkdir .metaflow && i=0; while [ $i -le 5 ]; do echo 'Downloading code package.'; python -c 'import sys; import boto3; boto3.client(sys.argv[1]).download_file(sys.argv[2], sys.argv[3], sys.argv[4])' s3 'def3b3-mf-test' 'ParameterFlow/data/24/243633f93d432b2431cfac4518a44f4e9eb68881' job.tar >/dev/null &&                         echo 'Code package downloaded.' && break; sleep 10; i=$((i+1)); done && if [ $i -gt 5 ]; then echo 'Failed to download code package from s3://def3b3-mf-test/ParameterFlow/data/24/243633f93d432b2431cfac4518a44f4e9eb68881 after 6 tries. Exiting...' && exit 1; fi && tar xf job.tar && echo 'Task is starting.' && sleep 10"]
```

I had to go for the `sys.argv` solution, as I wasn't able to force the correct quote escaping inside the single quoted string

tkislan:
@savingoyal is anything still blocking this?

tkislan:
Closing, as nobody cares ..

Issue #472, opened 2021-04-09
'path' is not defined in _version method


matus123:
Hi,

I've encountered this error `NameError: name 'path' is not defined` in `/metaflow/plugins/metadata/service.py`.
https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/metadata/service.py#L286

Variable `path` is not defined in _version method.

Thanks

savingoyal:
Good catch. We will fix this!

PR #476, opened 2021-04-19
Properly handle error messages in `_version` metadata call

savingoyal:
The error handling logic was missing a reference to `path`
resulting in the issue outlined in #472

### DIFF
--- a/metaflow/plugins/metadata/service.py
+++ b/metaflow/plugins/metadata/service.py
@@ -261,7 +261,7 @@ class ServiceMetadataProvider(MetadataProvider):
             raise MetaflowException('Missing Metaflow Service URL. '
                 'Specify with METAFLOW_SERVICE_URL environment variable')
         path = 'ping/'
-        url = os.path.join(cls.INFO, path.lstrip('/'))
+        url = os.path.join(cls.INFO, path)
         for i in range(METADATA_SERVICE_NUM_RETRIES):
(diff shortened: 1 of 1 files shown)


metaflow/plugins/metadata/service.py
```diff
@@ -260,7 +260,8 @@ def _version(cls, monitor):
         if cls.INFO is None:
             raise MetaflowException('Missing Metaflow Service URL. '
                 'Specify with METAFLOW_SERVICE_URL environment variable')
-        url = os.path.join(cls.INFO, 'ping')
+        path = 'ping/'
+        url = os.path.join(cls.INFO, path.lstrip('/'))
```
romain-intel, line 264: lstrip is superfluous in this case. We can just set path = 'ping' and remove all lstrip.

savingoyal: The reason I added `/` is to make sure that the error message also has a `/` suffix similar to the `_request` method above. I don't feel strongly about it, so if you don't prefer `/`, I can yank it out.

romain-intel: ah, ok, but then maybe rstrip :) (the lstrip is absolutely not doing anything).

savingoyal: oh shoot. that was a brainfreeze.

PR #481, opened 2021-04-21
Back-port from convergence branch: S3 client improvements

romain-intel:


metaflow/util.py
```diff
@@ -50,6 +50,19 @@ def unquote_bytes(x):
         return unquote(to_unicode(x))
 
 
+if sys.version_info.major >= 3 and sys.version_info.minor >= 7:
```
savingoyal, line 53: Where is `namedtuple_with_defaults` used in the PR?

romain-intel: L15 of s3.py.

metaflow/datatools/__init__.py
```diff
@@ -1,2 +1 @@
-
-from .s3 import S3
\ No newline at end of file
+from .s3 import MetaflowS3Exception, S3
```
savingoyal, line 1: Maybe not related to this PR specifically but `python -m metaflow.datatools.s3op --help` results in a warning 
```
RuntimeWarning: 'metaflow.datatools.s3op' found in sys.modules after import of package 'metaflow.datatools', but prior to execution of 'metaflow.datatools.s3op'; this may result in unpredictable behaviour
```

metaflow/datatools/s3op.py
```diff
@@ -270,21 +414,29 @@ def verify_results(urls, verbose=False):
             exit(ERROR_VERIFY_FAILED, url)
         if expected != got:
             exit(ERROR_VERIFY_FAILED, url)
-
-def generate_local_path(url):
+        if url.content_type or url.metadata:
```
savingoyal, line 417: Minor nit - L:408 we should print the actual url. On the console - I get `verifying <__main__.S3Url object at 0x7fb5688b0b38>, expected 1528`

metaflow/datatools/s3op.py
```diff
@@ -517,19 +685,85 @@ def get(prefixes,
...
+@click.option('--verbose/--no-verbose',
+              default=True,
+              show_default=True,
+              help='Print status information on stderr.')
+@click.option('--listing/--no-listing',
+              default=False,
+              show_default=True,
+              help='Print S3 URL -> local file mapping on stdout.')
+@click.argument('prefixes', nargs=-1)
+def info(prefixes,
```
savingoyal, line 741: Is there a reason for not supporting `--recursive` for `info`?

metaflow/datatools/s3op.py
```diff
@@ -128,11 +263,11 @@ def process_urls(mode, urls, verbose, num_workers):
...
-    start_workers(mode, urls, num_workers)
+    sz_results = start_workers(mode, urls, num_workers)
     end = time.time()
 
     if verbose:
-        total_size = sum(size for url, size in urls)
+        total_size = sum(sz for sz in sz_results if sz is not None and sz > 0)
         bw = total_size / (end - start)
         print('%sed %d files, %s in total, in %d seconds (%s/s).'\
               % (mode.capitalize(),
```
savingoyal, line 273: Nit that we should address later - we might not always be successful in downloading all the files (--no-allow-missing and what not)

savingoyal (approved):
Verified! LGTM! Some minor comments.

metaflow/datatools/s3op.py
romain-intel: Will fix. Good catch.

metaflow/datatools/s3op.py
romain-intel: I can add it. I didn't add any info_recursive or info_all at the TL but we can do it if you think it is useful.

metaflow/datatools/s3op.py
romain-intel: those errors should be caught later ideally. At least the behavior here should not have changed.

metaflow/datatools/__init__.py
romain-intel: Let me look at that as well. That's weird.

romain-intel: It seems to do it before this patch too but checking if there is an easy fix.

PR #488, opened 2021-04-27
@kubernetes step decorator

oavdeev:
**Note that it is PR is on top of #504, since it relies on some mflog refactorings introduced there.**

# What does this do

Introduce a new decorator, `@kubernetes` that allows you to run specific steps on kubernetes as [Jobs](https://kubernetes.io/docs/concepts/workloads/controllers/job/).

## What works

Basics work. You can decorate steps with `@kubernetes` and they'll run on k8s cluster. `--with kubernetes` works too. Logging works through mflog.

You can specify a docker image as `@kubernetes(image="python:3.7")`. If the image is not specified, it uses `METAFLOW_KUBERNETES_IMAGE_URI`. If that is missing too, it'll use default python image. This behavior is essentially the same as `@batch` today.

This also introduces a dependency on [Kubernetes Python client](https://github.com/kubernetes-client/python) (only needed if you use `@kubernetes`).

#### K8S creds
It will use the k8s creds that are configured in your environment. That is, this doesn't do anything _special_ for k8s auth; if `kubectl` works in your environment, this will too.

#### AWS creds
The code still uses S3, and assumes that AWS auth is somehow taken care of via your cluster configuration. That is, AWS creds are available either via the instance profile, or via [IRSA](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html).

## What's left

- [x] support `@timeout`
- [x] support `@resources`
- [x] support `@environment`
- [x] make sure interrupting/cancelling a step works
- [ ] more testing for edge cases and failure modes


metaflow/plugins/kubernetes/kubernetes_runner.py
```diff
@@ -0,0 +1,216 @@
...
+        ttl_seconds_after_finished=600,
+    )
+
+    # Instantiate the job object
+    job = client.V1Job(
+        api_version="batch/v1",
+        kind="Job",
+        metadata=client.V1ObjectMeta(name=job_name, labels=labels),
+        spec=spec,
+    )
```
sappier, line 70: k8s python client accepts specifications written as dictionaries, then it looks identical to the k8s json spec.  The job could be created like this:
```
job = {
    "apiVersion": "batch/v1",
    "kind": "Job",
    "metadata": {
        "name": job_name,
        "labels": labels
    },
    "spec": {
        "template": {
            "metadata": {
                "labels": {
                    "app": "pi",
                },
            },
            "spec": {
                "containers": [
                    {
                        "name": job_name,
                        "image": image,
                        "command": command,
                        "env": [{"name": k, "value": v} for k, v in env.items()],
                    },
                ],
                "restartPolicy": "Never"
            }
        },
        "backoffLimit": 1,      # retries handled by Metaflow
        "ttlSecondsAfterFinished": 600
    }
}
```

oavdeev: Nice, didn't know that. One thing I like about `client.` API is that VSCode/mypy complains when I make a typo in a parameter name. For example I always forget that `"template" {...}` bit when writing YAML/JSON.

oavdeev:
closing in favor of #644

Issue #490, opened 2021-04-28
Allow setting default max-workers and max-num-split values


displacedaussie:
The default values for `--max-workers` (16) and `--max-num-splits` (100) are fine for many use cases, but for some workloads and metaflow environments, it would be useful if it were possible to set new default values. Currently we rely on the user to override them on the command line for each run.

We have some Flows where we always want max-workers to be 32 (or 64), and max-num-splits to be 256.

Reading `METAFLOW_DEFAULT_MAX_WORKERS` and `METAFLOW_DEFAULT_MAX_NUM_SPLITS` from the environment would allow us to set these in the metaflow profile/config file and then not have to worry about them being set on the command line.

savingoyal:
@displacedaussie Thanks for opening the issue. Thanks to `click`, this functionality already exists. You can set the environment variables - `METAFLOW_RUN_MAX_WORKERS` and `METAFLOW_RUN_MAX_NUM_SPLITS` to achieve the desired outcome 😄 

```
from metaflow import FlowSpec, step

class ForeachFlow(FlowSpec):

    @step
    def start(self):
        self.titles = ['Stranger Things',
                       'House of Cards',
                       'Narcos']
        self.next(self.a, foreach='titles')

    @step
    def a(self):
        self.title = '%s processed' % self.input
        self.next(self.join)

    @step
    def join(self, inputs):
        self.results = [input.title for input in inputs]
        self.next(self.end)

    @step
    def end(self):
        print('\n'.join(self.results))

if __name__ == '__main__':
    ForeachFlow()
```

```
METAFLOW_RUN_MAX_NUM_SPLITS=1 python3 ForEach.py run
```
^ this command will fail - 
```
Step start (task-id 1) failed: Foreach in step start yielded 3 child steps which is more than the current maximum of 1 children. You can raise the maximum with the --max-num-splits option.
```

```
METAFLOW_RUN_MAX_NUM_SPLITS=1 python3 ForEach.py run --max-num-splits 3
```
^ while this will succeed

displacedaussie:
Thank you Savin, that is definitely useful. I hadn't realised click could do this.

I guess there are two downsides to this though:

1) We need separate environment variables for `run` and `resume`, which means setting the same value in multiple places:

```
MEATFLOW_RUN_MAX_WORKERS=32
MEATFLOW_RESUME_MAX_WORKERS=32
METAFLOW_RUN_MAX_NUM_SPLITS=256
[...]

PR #491, opened 2021-04-28
Look up max-workers and max-num-splits values from config

displacedaussie:
This change is for #490 and allows us to set default values for `--max-workers` and `--max-num-splits`.

Currently the default values (`--max-workers=16` and `--max-num-splits=100`) cannot be customised. We rely on the user to set them on the command line for each run if they want or need different values.

Reading them in this way allows appropriate values to be set in the metaflow config file (or as environment variables) if the current default values are not suitable.


savingoyal:
@displacedaussie This functionality already exists. I have made a comment [here](https://github.com/Netflix/metaflow/issues/490#issuecomment-828645666)

> Thanks to click, this functionality already exists. You can set the environment variables - METAFLOW_RUN_MAX_WORKERS and METAFLOW_RUN_MAX_NUM_SPLITS to achieve the desired outcome

PR #495, opened 2021-04-28
Allow for more flexible (escaped) variable names in CLI options

romain-intel:


metaflow/util.py
```diff
@@ -300,7 +300,7 @@ def dict_to_cli_options(params):
 
                     # Of the value starts with $, assume the caller wants shell variable
                     # expansion to happen, so we pass it as is.
-                    if value.startswith("$"):
+                    if value.lstrip("\\").startswith("$"):
```
savingoyal, line 303: @romain-intel Can we leave a comment in the code about this change so that a future commit doesn't end up reverting it.

savingoyal (changes requested):
Added a minor comment. Feel free to merge.

romain-intel:
Will merge after tests if no other action.

savingoyal (approved):

PR #500, opened 2021-04-30
urgent-feature: Job failure classification

jithenshriyan:
Classified job failures on AWS Batch according to status code to provide deeper insight into failures.

### DIFF
--- a/metaflow/plugins/aws/batch/batch_client.py
+++ b/metaflow/plugins/aws/batch/batch_client.py
@@ -441,8 +441,11 @@ class RunningJob(object):
 
     @property
     def is_crashed(self):
-        # TODO: Check statusmessage to find if the job crashed instead of failing
-        return self.status == 'FAILED'
+        if "crash" or "crashed" in self.status_reason:
+            return self.status == 'CRASH'
+        if "fail" or "failed" in self.status_reason:
+            return self.status == 'FAILED'
+
 
     @property
     def reason(self):
@@ -575,6 +578,12 @@ class BatchLogs(object):
         self._token = None
 
     def _get_events(self):
+        rj = RunningJob(object)
+        if rj.status_code(self):
+            if rj.is_done(self):
+                if rj.status_code(self) != 0:
+                    return rj.status_code(self), rj.status_reason(self)
+
         try:
             if self._token:
                 response = self._client.get_log_events(
--- a/metaflow/plugins/aws/batch/batch_client.py
+++ b/metaflow/plugins/aws/batch/batch_client.py
@@ -14,6 +14,7 @@ except NameError:
 from metaflow.exception import MetaflowException
 from metaflow.metaflow_config import AWS_SANDBOX_ENABLED
 
+
 class BatchClient(object):
     def __init__(self):
         from ..aws_client import get_aws_client
@@ -35,20 +36,20 @@ class BatchClient(object):
             for queue in queues
             for status in ['SUBMITTED', 'PENDING', 'RUNNABLE', 'STARTING', 'RUNNING']
             for page in self._client.get_paginator('list_jobs').paginate(
-                jobQueue=queue, jobStatus=status
-            )
+            jobQueue=queue, jobStatus=status
+        )
             for job in page['jobSummaryList']
         )
 
     def describe_jobs(self, job_ids):
-        for jobIds in [job_ids[i:i+100] for i in range(0, len(job_ids), 100)]:
+        for jobIds in [job_ids[i:i + 100] for i in range(0, len(job_ids), 100)]:
             for jobs in self._client.describe_jobs(jobs=jobIds)['jobs']:
                 yield jobs
 
     def describe_job_queue(self, job_queue):
         paginator = self._client.get_paginator('describe_job_queues').paginate(
             jobQueues=[job_queue], maxResults=1)
-        return paginator.paginate()['jobQueues'][0] 
+        return paginator.paginate()['jobQueues'][0]
 
     def job(self):
         return BatchJob(self._client)
@@ -114,12 +115,12 @@ class BatchJob(object):
                 raise BatchJobException(
                     'AWS Batch Job Queue %s not found.' % job_queue)
             compute_environment = response['jobQueues'][0] \
-                                    ['computeEnvironmentOrder'][0] \
-                                    ['computeEnvironment']
+                ['computeEnvironmentOrder'][0] \
+                ['computeEnvironment']
             response = self._client.describe_compute_environments(
                 computeEnvironments=[compute_environment])
             platform = response['computeEnvironments'][0] \
-                            ['computeResources']['type']
+                ['computeResources']['type']
 
         # compose job definition
         job_definition = {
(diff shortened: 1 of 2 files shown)


metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -575,6 +582,28 @@ def __init__(self, group, stream, pos=0, sleep_on_no_data=0):
         self._token = None
 
     def _get_events(self):
+        done = RunningJob.is_done
+        status_code = RunningJob.status_code
+        if done:
```
romain-intel, line 587: This should be done later probably as there may still be messages being pulled below. I suggest checking this at the end (before the return) and appending the message to that array.

metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -575,6 +582,28 @@ def __init__(self, group, stream, pos=0, sleep_on_no_data=0):
...
+            if status_code == 1:
+                return 'Task failed with a general error. \
```
romain-intel, line 589: Each message is a dict containing at least 'message' and 'timestamp' so this needs to be formatted like that otherwise it won't work when _get_events is called.

metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -575,6 +582,28 @@ def __init__(self, group, stream, pos=0, sleep_on_no_data=0):
         self._token = None
 
     def _get_events(self):
+        done = RunningJob.is_done
```
romain-intel, line 585: is_done and status_code are instance level properties so you can't call them on a class object.

metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -575,6 +582,28 @@ def __init__(self, group, stream, pos=0, sleep_on_no_data=0):
...
+            if status_code == 2:
+                return 'Misuse of shell builtins. You may have a missing keyword or command, or \
+                       permission problem (and diff return code on a failed binary file comparison).'
+            if status_code == 126:
+                return 'Command invoked cannot execute. May be a permission problem or command is not an executable.'
+            if status_code == 127:
+                return 'Command not Found. May be a possible problem with $PATH or a type.'
+            if status_code == 128:
+                return '	Invalid argument to exit.'
+            if status_code == 130:
```
romain-intel, line 600: 129+ are signal related exit codes (so there are a bunch of them).

metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -575,6 +582,28 @@ def __init__(self, group, stream, pos=0, sleep_on_no_data=0):
...
+                return 'Script terminated by Control-C.'
+            if status_code == 255:
+                return 'Exit status out of range.'
+            else:
+                return 'Task failed.'
+
```
romain-intel, line 606: In general, I don't think that BatchLogs is the best place to classify this. As you need access to the status (done and status_code), it may better be placed after _print_available in the the batch wait method.

romain-intel (changes requested):
Thank you for your first PR! I added some comments. I am not sure it works as is. Please do test your PRs as well using either your own AWS account or a sandbox account.

metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -35,20 +36,20 @@ def unfinished_jobs(self):
             for queue in queues
             for status in ['SUBMITTED', 'PENDING', 'RUNNABLE', 'STARTING', 'RUNNING']
             for page in self._client.get_paginator('list_jobs').paginate(
-                jobQueue=queue, jobStatus=status
-            )
+            jobQueue=queue, jobStatus=status
```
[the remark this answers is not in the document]
jithenshriyan: Sorry about this. I'm using IntelliJ and sometimes it has some weird suggestions on indentation.

- [ x] Fixed

metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -563,10 +569,11 @@ def wait_for_running(self, job_id):
             jobs=[job_id]
         )
 
+
 class BatchLogs(object):
     def __init__(self, group, stream, pos=0, sleep_on_no_data=0):
         from ..aws_client import get_aws_client
-        self._client = get_aws_client('logs')
+        self._client = get_aws_client('logs', 'batch')
```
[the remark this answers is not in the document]
jithenshriyan: I was testing something out and had forgotten to remove this before commiting.

[x] Fixed

metaflow/plugins/aws/batch/batch_client.py
jithenshriyan: I moved the variables and if statement below the try and except block under `_get_events(self)`. 

- [x] Fixed

metaflow/plugins/aws/batch/batch_client.py
jithenshriyan: Would this work for status_code?

```python
status_code = self.info['container'].get('exitCode').values()
```

metaflow/plugins/aws/batch/batch_client.py
jithenshriyan: I was able to call the status_code using:
```python
self.info['Container'].get('exitCode').values()
```

If that's sufficient, I could leave it there, or if it isn't and you believe I should move it to where you suggest, I'll do so.

savingoyal:
@jithenshriyan Thanks for the PR! The current behavior is that when we detect that a job has failed, we simply fetch the `reason` written by AWS in the job details and present it to the user. The logic to handle that is here - https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/aws/batch/batch.py#L348. The problem is that the `reason` written by AWS is always the same - `Essential container in task failed` - which is not super helpful for the end-user. What we would like to do is instead print our own error message which is a function of `job.status_code`. So in [L348](https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/aws/batch/batch.py#L348) we can check the status code and print out the appropriate error message as a series of if-else.

Do you have access to an AWS account where you can test your work?

jithenshriyan:
Thanks for the feedback, @savingoyal! I've made some changes and am committing them. 

Thanks!

metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -613,4 +613,4 @@ def _fill_buf(self):
         events = self._get_events()
         for event in events:
             self._buf.append(event['message'])
-            self._pos = event['timestamp']
\ No newline at end of file
+            self._pos = event['timestamp']
```
savingoyal, line 616: Given that this file has no changes, you can remove this file from the PR.

metaflow/plugins/aws/batch/batch.py
```diff
@@ -345,9 +345,23 @@ def _print_available(tail, stream, should_persist=False):
         # TODO: AWS CloudWatch fetch logs
 
         if self.job.is_crashed:
-            msg = next(msg for msg in 
-                [self.job.reason, self.job.status_reason, 'Task crashed.']
-                 if msg is not None)
+            if self.job.status_code == 1:
+                msg = next(msg for msg in ['Task failed with a general error. You may have a miscellaneous error(s), '
```
savingoyal, line 349: I am not sure if I follow the lambda here. The list is a singleton here and below. Also, have you looked into AWS Batch error codes? I don't think the status codes map to the descriptions that you have written.

savingoyal (changes requested):

metaflow/plugins/aws/batch/batch_client.py
jithenshriyan: Got it. There was an extra line at the bottom that made this file come up on the PR. Fixed this.

metaflow/plugins/aws/batch/batch.py
jithenshriyan: I couldn't find any documentation on AWS Batch error codes. You gave me a link to exit codes earlier and that was what I used to write the descriptions. Let me know if I should use another source.

Also, it would be great if you could give me a bit more detail on the error with my If statement. Thanks!

savingoyal:
@jithenshriyan Looks like AWS Batch shipped a bug fix fixing the `status_reason` in the `describe_jobs` response. Metaflow now accurately reports back failures like OOM, Seg Faults without any extra piece of logic. I am closing this PR. In case you would like to still contribute to Metaflow, please get in touch (chat.metaflow.org) and I can find you a starter project to work on.

PR #505, opened 2021-05-04
Adding IPython Notebooks in the package

tobias-gp:
We would like to be able to execute notebooks in Metaflow steps, it would be great to include them in the package.


savingoyal:
@tobias-gp You can set the env var `METAFLOW_PACKAGE_SUFFIXES` to `.py,.ipynb,.R,.RDS` to achieve a similar outcome or use the flag `--package_suffixes` - `python flow.py --package_suffixes=.py,.ipynb,.R,.RDS run`. Setting `METAFLOW_DEFAULT_PACKAGE_SUFFIXES` in `metaflow_config.py` would also work. 

The reason we wouldn't want to bundle `.ipynb` files by default is that many times the .ipynb files can be quite voluminous since they store the state.

tobias-gp:
Thanks @savingoyal , somehow oversaw this, should have seen in the code!

PR #517, opened 2021-05-10
include headers for heartbeat and fix url template

ferras:


metaflow/plugins/metadata/service.py
```diff
@@ -38,10 +31,10 @@ class ServiceMetadataProvider(MetadataProvider):
 
     def __init__(self, environment, flow, event_logger, monitor):
         super(ServiceMetadataProvider, self).__init__(environment, flow, event_logger, monitor)
-        self.mli_url_task_template = urljoin(METADATA_SERVICE_URL,
-                                             '/flows/{flow_id}/runs/{run_number}/steps/{step_name}/tasks/{task_id}/heartbeat')
-        self.mli_url_run_template = urljoin(METADATA_SERVICE_URL,
-                                            '/flows/{flow_id}/runs/{run_number}/heartbeat')
+        self.url_task_template = os.path.join(METADATA_SERVICE_URL,
```
ferras, line 34: os.path.join probably isn't the best function to use since im pretty sure it won't work if running on windows but this is how it is done in the rest of the service.py file

PR #528, opened 2021-05-13
Add implementation docs

savingoyal:
Add developer documentation around concurrency semantics, sidecars and lifecycle.


savingoyal:
@romain-intel Can you take care of the escape hatch docs. No urgency on that.

PR #536, opened 2021-05-20
Enable state machine execution history logging for Step Functions integration

savingoyal:
Currently, it's hard to keep track of all the execution from within AWS console
for Step Functions executions. Now, the users can pass a `--log-execution-history`
flag to `step-functions create` to pipe execution history logs to an AWS CloudWatch
Logs log group ARN denoted by the environment variable -
METAFLOW_SFN_EXECUTION_LOG_GROUP_ARN

### DIFF
--- a/metaflow/plugins/aws/step_functions/step_functions_client.py
+++ b/metaflow/plugins/aws/step_functions/step_functions_client.py
@@ -1,5 +1,5 @@
 from metaflow.metaflow_config import \
-    AWS_SANDBOX_ENABLED, AWS_SANDBOX_REGION
+    AWS_SANDBOX_ENABLED, AWS_SANDBOX_REGION, SFN_EXECUTION_LOG_GROUP_ARN
 
 
 class StepFunctionsClient(object):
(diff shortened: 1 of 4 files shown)


metaflow/plugins/aws/step_functions/step_functions_client.py
```diff
@@ -75,6 +83,24 @@ def terminate_execution(self, state_machine_arn, execution_arn):
...
 
+    def _default_logging_configuration(self, log_execution_history):
+        if log_execution_history:
+            return {
+                'level': 'ALL',
+                'includeExecutionData': True,
+                'destinations': [
+                    {
+                        'cloudWatchLogsLogGroup': {
+                        'logGroupArn': SFN_EXECUTION_LOG_GROUP_ARN
```
kldavis4, line 94: @savingoyal 
```suggestion
                        'logGroupArn': f'{SFN_EXECUTION_LOG_GROUP_ARN}:*'
```
It seems a little strange to have to specify the `:*` in the config, versus just the arn of the log group

savingoyal: Isn't the ARN for the log group actually `arn:aws:logs:us-west-2:<redacted>:log-group:/aws/foo:*`?

savingoyal: I am looking at the `return values` section here - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-logs-loggroup.html

kldavis4: ah.. you are right - never mind then - I was looking at another page which made me think otherwise

PR #541, opened 2021-05-24
Introduce @project decorator

savingoyal:
This PR introduces @project decorator to help users manage multiple Metaflow flows (especially on AWS Step Functions). 

[Documentation](https://docs.google.com/document/d/1NDFrPhi5-zuUkuayDg3K277-yry8Nup2fxhQGgWGzLI/edit?usp=sharing)

I will introduce namespace-aware triggering of flows in a separate PR.

### DIFF
--- a/metaflow/plugins/aws/step_functions/step_functions_cli.py
+++ b/metaflow/plugins/aws/step_functions/step_functions_cli.py
@@ -237,7 +237,6 @@ def resolve_token(name,
                   is_project):
 
     # 1) retrieve the previous deployment, if one exists
-    #TODO - Check existing deployment with is_project enabled
     workflow = StepFunctions.get_existing_deployment(name)
     if workflow is None:
         obj.echo("It seems this is the first time you are deploying *%s* to "
--- a/metaflow/plugins/aws/step_functions/step_functions_cli.py
+++ b/metaflow/plugins/aws/step_functions/step_functions_cli.py
@@ -1,4 +1,5 @@
 import click
+from hashlib import sha1
 import json
 import re
 from distutils.version import LooseVersion
@@ -9,7 +10,7 @@ from metaflow.exception import MetaflowException, MetaflowInternalError
 from metaflow.datastore.datastore import TransformableObject
 from metaflow.package import MetaflowPackage
 from metaflow.plugins import BatchDecorator
-from metaflow.util import get_username
+from metaflow.util import get_username, to_bytes
 
 from .step_functions import StepFunctions
 from .production_token import load_token, store_token, new_token
@@ -124,9 +125,11 @@ def create(obj,
         obj.echo_always(flow.to_json(), err=False, no_bold=True)
     else:
         flow.deploy(log_execution_history)
-        obj.echo("Workflow *{name}* pushed to "
+        obj.echo("State Machine *{state_machine}* "
+                 "for flow *{name}* pushed to "
                  "AWS Step Functions successfully.\n"
-                    .format(name=obj.state_machine_name), 
+                    .format(state_machine=obj.state_machine_name,
+                            name=current.flow_name), 
                  bold=True)
 
         flow.schedule()
--- a/metaflow/plugins/aws/step_functions/step_functions_cli.py
+++ b/metaflow/plugins/aws/step_functions/step_functions_cli.py
@@ -42,7 +42,7 @@ def step_functions(obj,
                    name=None):
     obj.check(obj.graph, obj.flow, obj.environment, pylint=obj.pylint)
     obj.state_machine_name, obj.token_prefix, obj.is_project = \
-                                      resolve_state_machine_name(name)
+                                      resolve_state_machine_name(obj, name)
 
 @step_functions.command(help="Deploy a new version of this workflow to "
                     "AWS Step Functions.")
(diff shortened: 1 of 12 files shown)


metaflow/plugins/aws/step_functions/step_functions_cli.py
```diff
@@ -158,19 +164,34 @@ def attach_prefix(name):
...
           return SFN_STATE_MACHINE_PREFIX + '_' + name
       return name
+    project = current.get('project_name')
+    if project:
+        if name:
+            raise MetaflowException("--name is not supported for @projects. "
+                                    "Use --branch instead.")
+        state_machine_name = attach_prefix(current.project_flow_name)
+        project_branch = to_bytes('.'.join((project, current.branch_name)))
+        token_prefix = 'mfprj-%s' % sha1(project_branch).hexdigest()
```
oavdeev, line 174: why use a hash here? could we just use `%project.%branch` so it is human-readable?

metaflow/plugins/project_decorator.py
```diff
@@ -0,0 +1,98 @@
...
+                                    % VALID_NAME_LEN)
+
+        branch = 'test.%s' % given_branch
+    elif deploy_prod:
+        branch = 'prod'
+    else:
+        # For AWS Step Functions, we set the branch to the value of
+        # environment variable `METAFLOW_OWNER`, since AWS Step Functions
+        # has no notion of user name.
+        branch = 'user.%s' % os.environ.get('METAFLOW_OWNER', user_name)
```
oavdeev, line 96: I assume `METAFLOW_OWNER` may have dots in it, is it going to cause issues? If not should we allow dots in branch names?

metaflow/plugins/aws/step_functions/step_functions_cli.py
savingoyal: good point - the rationale is that `project` & `branch` could each be up to 128 characters - which increases the length of the token prefix.

metaflow/plugins/project_decorator.py
savingoyal: Good point! There is actually an open issue to sanitize `METAFLOW_OWNER` since it can have whitespace as well. I will open a different PR to make sure we are sanitizing the output of `utils.get_username()`

oavdeev: Maybe validate it against `VALID_NAME_RE` here then? I worry that we'll ship this PR, then people will use it to create branches with weird names, then we ship sanitization and it will be annoying to reconcile their existing branch names with the sanitized ones.

oavdeev:
One slightly confusing thing that I ran into immediately, I've added `@project("myproject")` to my test flow and...

```
  File "./helloworld1.py", line 21, in <module>
    @project("hello")
  File "/Users/oleg/src/metaflow/metaflow/decorators.py", line 347, in _base_flow_decorator
    raise BadFlowDecoratorException(decofunc.name)
metaflow.decorators.BadFlowDecoratorException: Decorator 'project' can be applied only to FlowSpecs. Make sure the decorator is above a class definition.
```

After some digging the actual issue is that is has to be `@project(name="myproject")` not `@project("myproject")`. Not sure if that's specific to `@project` (probably not) but I wonder if the error message can be made more helpful

oavdeev:
Another edge case I ran into: State Machine name limit is 80 characters and EventBridge rule name is 64 chars.

It is pretty easy to hit either limit. Moreover, you can hit one but not the other, then the State Machine is created but the rule creation fails with an error

savingoyal:
> Another edge case I ran into: State Machine name limit is 80 characters and EventBridge rule name is 64 chars.
> 
> It is pretty easy to hit either limit. Moreover, you can hit one but not the other, then the State Machine is created but the rule creation fails with an error

This has been addressed in the most recent commit.

metaflow/plugins/project_decorator.py
savingoyal: Yeah, the check needs to exist alongside the `get_username()` method. We will ship that PR and this PR in a single release.

savingoyal: @oavdeev Just so that we are on the same page - any user name that contributes to a valid step functions state machine name is allowed - which means whitespace, wildcard characters, brackets and special characters are not allowed (and currently the PR will throw a not so pretty error) - dots, hyphens and dashes are still allowed and will work as is. So I don't see the risk of shipping this PR as is.

savingoyal: There is a similar issue that has existed with @batch for a while now where we are unable to launch the AWS Batch job if there are white spaces in the user name.

oavdeev: Ah I thought you'd replace the dots in the username. I.e. if the user is `oleg.avdeev` the branch would be `oleg_avdeev` or something.

oavdeev (approved):

metaflow/plugins/project_decorator.py
```diff
@@ -0,0 +1,97 @@
...
+                  logger,
+                  echo,
+                  options):
+        self._option_values = options
+        project_name = self.attributes.get('name')
+        project_flow_name, branch_name = format_name(flow.name,
+                                                     project_name,
+                                                     options['production'],
+                                                     options['branch'],
+                                                     get_username())
```
romain-intel, line 47: Add: 
```
echo("Project flow name: *%s*\n" % project_flow_name, 
          fg='magenta', highlight='green')
```
or something of that nature.

romain-intel (approved):
Minor comment. O/w LGTM. Please pre-publish doc as well so it can be ready when releasing.

PR #547, opened 2021-05-28
Add after-execution tagging support (both in client and CLI)

romain-intel:
You can now modify tags after a run has executed. A CLI provides
the ability to see all the tags, add or remove them.

The client also allows you to operate on tags directly for all
objects.

Metadata service will need to be updated to version 2.0.6 for this
to work with the metadata service. This currently works as is with
local metadata.


metaflow/metadata/metadata.py
```diff
@@ -506,8 +556,8 @@ def _apply_filter(elts, filters):
 
     def __init__(self, environment, flow, event_logger, monitor):
         self._task_id_seq = -1
-        self.sticky_tags = []
-        self.sticky_sys_tags = []
+        self.sticky_tags = set([])
```
tuulos, line 559: nit: just do `set()`

metaflow/plugins/metadata/local.py
```diff
@@ -173,7 +208,11 @@ def _makedirs(path):
                 raise
 
     def _ensure_meta(
-            self, obj_type, run_id, step_name, task_id, tags=[], sys_tags=[]):
+            self, obj_type, run_id, step_name, task_id, tags=None, sys_tags=None):
+        if tags is None:
+            tags = []
```
tuulos, line 213: since tags are sets, we could make them sets everywhere for clarity, i.e. `tags = set()` here. We need to convert sets to lists only for JSON encoding

metaflow/client/core.py
```diff
@@ -398,6 +406,35 @@ def _filtered_children(self, *tags):
...
+                raise TaggingException("Invalid tagging operation %s" % o)
+            if o == 'add' and t in self._system_tags:
+                # We don't throw an error but just don't send it; it's a no-op
+                continue
+            if o == 'remove':
+                if t in self._system_tags:
+                    raise TaggingException("Cannot remove system tag *%s*" % t)
+                if t not in self._tags:
+                    raise TaggingException(
+                        "Cannot remove tag *%s* as it does not exist" % t)
```
tuulos, line 425: I wonder if removing a non-existent tag should be just a no-op. For instance, currently you can't do ```.replace_tag('production', 'archived')```
without having a separate path for the case if `production` doesn't exist. This is extra tricky, since multiple clients may manipulate tags concurrently

metaflow/plugins/metadata/local.py
```diff
@@ -159,6 +160,40 @@ def _get_object_internal(cls, obj_type, obj_order, sub_type, sub_order, filters=
...
+                    obj_to_update['tags'] = [
+                        x for x in obj_to_update['tags'] if x != op.args['tag']]
+                else:
+                    raise TaggingException(msg='Invalid operation %s' % op.operation)
+            else:
+                raise TaggingException(msg='Cannot find object %s' % op.id)
+            # Here, all went well
+            file_path = os.path.join(
+                LocalMetadataProvider._get_metadir(*obj_path), '_self.json')
+            LocalMetadataProvider._dump_json_to_file(file_path, obj_to_update, True)
```
tuulos, line 193: the current way of storing tags in a JSON (`_dump_json_to_file`) is not amenable to concurrent operations at all. For instance, if you run two `tag add` operations concurrently, it is possible (likely) that the second tag disappears silently.

Two possible solutions:
1. `flock` the JSON: Flock, read tags, modify tags, write tags, funlock
2. store tags as separate files

metaflow/plugins/tagging_cli.py
```diff
@@ -0,0 +1,303 @@
+import click
+from metaflow import namespace
+from metaflow.util import resolve_identity
+from metaflow.exception import CommandException, MetaflowNotFound
+from metaflow.exception import MetaflowNamespaceMismatch
+
+# a rough emulation of itertools.zip_longest for our needs
```
tuulos, line 7: you could move this to `util` which centralizes py2/py3 compatibility hacks. It allows you to use `itertools.zip_longest` for Py3 natively. Add it here https://github.com/Netflix/metaflow/blob/master/metaflow/util.py#L16

metaflow/plugins/tagging_cli.py
```diff
@@ -0,0 +1,303 @@
...
+@click.option('--hide-system-tags',
+              required=False,
+              is_flag=True,
+              default=False,
+              help="Hide system tags.")
+@click.pass_obj
+def tag_list(obj, run_id, hide_system_tags, list_all, my_runs):
+    if (run_id is None and not list_all and not my_runs):
+        raise CommandException(
+            "Please use one of the options --run-id / --all / --my-runs.\n" +
```
tuulos, line 276: let's default to `--all` - `python tag list` should return something by default instead of showing an error

metaflow/plugins/tagging_cli.py
```diff
@@ -0,0 +1,303 @@
...
+        words_per_column = num_tags
+    else:
+        words_per_column = (num_tags + column_count - 1) // column_count
+
+    # Extend all_tags by empty tags to be able to easily fill up the lines
+    addl_tags = words_per_column * column_count - num_tags
+    if addl_tags > 0:
+        all_tags.extend([' ' for _ in range(addl_tags)])
+
+    lines = ["",
```
tuulos, line 44: Isn't this message a bit redundant to be shown every time?

metaflow/plugins/tagging_cli.py
```diff
@@ -0,0 +1,303 @@
...
+             "System tags and *user tags*; only *user tags* can be removed",
+             "--------------------------------------------------------",
+             ""]
+
+    for i in range(words_per_column):
```
tuulos, line 49: would it be controversial to suggest a three-column format

`[run-id] [user|sys ] [tag]`

this would be much more CLI-friendly: compatible with `cut`, `grep` etc.

metaflow/plugins/tagging_cli.py
```diff
@@ -0,0 +1,303 @@
...
+        for run in runs:
+            system_tags.update(run.system_tags)
+            all_tags.update(run.tags)
+    else:
+        metaflow_run_client = obtain_metaflow_run_client(obj, run_id, None)
+        system_tags = metaflow_run_client.system_tags
+        all_tags = metaflow_run_client.tags
+
+    if hide_system_tags:
+        obj.echo(format_tags([], all_tags.difference(system_tags)), err=False)
```
tuulos, line 301: we should make `list` respect the `--quiet` flag, i.e. use `obj.echo_always` for data and `obj.echo` for informational lines. Compare this to `--quiet logs` vs. `logs`.

The point of `--quiet` is that it makes the output script-friendly (safe to pipe, `grep` etc)

metaflow/plugins/tagging_cli.py
```diff
@@ -0,0 +1,303 @@
...
+                                                     run_id, user_namespace)
+
+    metaflow_run_client.replace_tag([tags[0]], [tags[1]])
+
+    obj.echo("Operation successful. New tags:")
+    obj.echo(print_tags(metaflow_run_client))
+
+@tag.command('list',
+             help='List tags of a run.')
+@click.option('--run-id',
```
tuulos, line 250: we could make `run-id` an optional argument, so the command would be symmetric with existing commands:

`logs 1623618905573323/start`
`dump 1623618905573323/start`
`tag list 1623618905573323`

metaflow/client/core.py
```diff
@@ -570,6 +622,48 @@ def traverse(cls, ids_r, lst):
             self._path_components = traverse(_CLASSES[self._NAME], ids, [])
         return self._path_components
 
+    def add_tag(self, tags):
+        """
+        Add one or more tags to the current object
+
+        Parameters
+        ----------
+        tags : string or List[string]
```
tuulos, line 631: `List[string]` is not correct - any iterable is accepted (which is desirable) except strings are handled specially

metaflow/client/core.py
```diff
@@ -570,6 +622,48 @@ def traverse(cls, ids_r, lst):
...
+        remove_tags : string or List[string]
+            Tags to remove
+        add_tags : string or List[string]
+            Tags to add
+        """
+        if is_stringish(remove_tags):
+            remove_tags = [remove_tags]
+        if is_stringish(add_tags):
+            add_tags = [add_tags]
+        operations = [('remove', tag) for tag in remove_tags]
```
tuulos, line 662: you should check that each `tag` `is_stringish` - currently `add_tag([1])` happily adds an integer as a tag

metaflow/plugins/tagging_cli.py
romain-intel: THis is actually no longer used I think so I will remove.

metaflow/plugins/tagging_cli.py
romain-intel: I think the idea was to use the client if you wanted a parseable thing and use this as a "pretty printing". But I don't mind either way as long as the info is clear. Note that there can be a *lot* of tags so a single column format as you are suggesting would make for a lot of text particularly if you dedup tags based on run ids (which is not currently the case).

romain-intel:
@savingoyal : Current state:
 - implemented in local mode only (to make sure functionality is as anticipated)
 - adding a tag is never a failure from a user's point of view; at the end of the call, the tag is added:
   - adding a system tag is like a no-op; the tag doesn't get added to the user tag list but is already present in system tags
   - adding a tag already present in user tag is like a no-op; the tag is already present
   - if tag not present, added
 - removing a tag that is a system tag is a failure because it is not removed (can't be)
 - removing a tag that is absent is a no-op (not a failure)
 - Granularity of change is per object in the backend; all operations per object fail or succeed all together (so if you "replace" a tag, either it worked or it didn't, no partial state).
 - Atomic behavior at the object granularity
 - Failure for one object does not trigger failure for everything (this is never used but it's implemented this way locally; in other words, if you try to modify multiple objects in the backend and one fails, the others may succeed just fine).
 - In the CLI, the quiet mode makes things very parseable. In non-quiet mode, it's pretty looking.
 - Added new options to list tags (check them out and see if you like them)

Hopefully that covers everything :). Let me know if you have more questions, it's been a while so I may have forgotten one or two.

metaflow/metadata/metadata.py
```diff
@@ -96,7 +99,7 @@ def version(self):
         '''
         return ''
 
-    def new_run_id(self, tags=[], sys_tags=[]):
+    def new_run_id(self, tags=None, sys_tags=None):
```
savingoyal, line 102: Why change from `[]` to `None`? It just adds additional logic in L:424 and L:425.

metaflow/plugins/tagging_cli.py
```diff
@@ -0,0 +1,424 @@
...
+    return metaflow_client.runs()
+
+
+@click.group()
+def cli():
+    pass
+
+
+@cli.group(help='Commands related to Metaflow tagging.')
+def tag():
```
savingoyal, line 226: Minor nit - Should we call it `tags` rather than `tag`? Okay, either way.

metaflow/plugins/tagging_cli.py
```diff
@@ -0,0 +1,424 @@
...
+              help="Change namespace from the default (your username) to "
+                   "the specified tag.")
+@click.argument('tags',
+                required=True,
+                type=str,
+                nargs=-1)
+@click.pass_obj
+def add(obj, run_id, user_namespace, tags):
+    user_namespace = resolve_identity() if user_namespace is None else user_namespace
+    metaflow_run_client = obtain_metaflow_run_client(obj,
```
savingoyal, line 251: Minor naming nit - We can just call it `client`.

metaflow/plugins/tagging_cli.py
```diff
@@ -0,0 +1,424 @@
...
+    system_tags = {metaflow_run_client.pathspec: metaflow_run_client.system_tags}
+    all_tags = {metaflow_run_client.pathspec: metaflow_run_client.tags}
+    return format_tags(obj, system_tags, all_tags, by_tag=False)
+
+
+def is_prod_token(token):
+    return token is not None and token.startswith("production:")
+
+
+def obtain_metaflow_run_client(obj, run_id, user_namespace):
```
savingoyal, line 140: Can be simply called `get_client`

metaflow/plugins/tagging_cli.py
```diff
@@ -0,0 +1,424 @@
...
+
+    # handle error messaging for two cases
+    # 1. our user tries to tagging a new flow before it is run
+    # 2. our user makes a typo in --namespace
+    try:
+        namespace(user_namespace)
+        Flow(pathspec=flow_name)
+    except MetaflowNotFound:
+        raise CommandException("Flow *%s* does not exist " % flow_name +
+                               "in any namespace.\n" +
```
savingoyal, line 155: Suggestion for a better exception - `No execution found for *%s*`

metaflow/plugins/tagging_cli.py
```diff
@@ -0,0 +1,424 @@
...
+                               "Please run the flow before tagging.")
+    except MetaflowNamespaceMismatch:
+        raise CommandException("Flow *%s* does not belong to " % flow_name +
```
savingoyal, line 158: Suggestion for a better exception - `No execution found for *%s* in namespace *%s*`

metaflow/plugins/tagging_cli.py
```diff
@@ -0,0 +1,424 @@
...
+                               "namespace *%s*.\n" % user_namespace +
+                               "Check typos if you used option --namespace.")
+
+    # throw an error with message to include latest run-id when run_id is None
+    if run_id is None:
+        latest_run_id = Flow(pathspec=flow_name).latest_run.id
+        msg = "Please specify a run-id using --run-id\n" + \
```
savingoyal, line 165: Minor nit - `Please specify a run id using --run-id\n %s%'s latest execution in namespace *%s* has run id *%s*`.

metaflow/plugins/tagging_cli.py
```diff
@@ -0,0 +1,424 @@
...
+              "Flow %s's latest run has run-id *%s* under namespace %s" \
+            % (flow_name, latest_run_id, user_namespace)
+        raise CommandException(msg)
+    run_id_parts = run_id.split('/')
+    if len(run_id_parts) == 1:
+        run_name = "%s/%s" % (flow_name, run_id)
+    elif len(run_id_parts) == 2:
```
savingoyal, line 172: Why do we have this check? We should expect the run-id to be a run-id. That is the behavior elsewhere with other CLIs as well in Metaflow.

metaflow/plugins/tagging_cli.py
```diff
@@ -0,0 +1,424 @@
...
+
+def obtain_metaflow_run_client(obj, run_id, user_namespace):
+    if is_prod_token(user_namespace):
+        raise CommandException(
+            "Modifying a scheduled run's tags is currently not allowed.")
+
+    flow_name = obj.flow.name
+
+    # handle error messaging for two cases
+    # 1. our user tries to tagging a new flow before it is run
```
savingoyal, line 148: We don't support Flow level tagging - maybe that should allow you to dispense with a few checks. For example - we need one of L154 and L188.

metaflow/plugins/tagging_cli.py
```diff
@@ -0,0 +1,424 @@
...
+
+    # handle error messaging for three cases
+    # 1. our user makes a typo in --run-id
+    # 2. our user tries to mutate tags on a meson/production run
+    # 3. our user's --run-id does not exist in the default/specified namespace
+    try:
+        namespace(user_namespace)
+        metaflow_client = Run(pathspec=run_name)
+    except MetaflowNotFound:
+        raise CommandException("Run *%s* does not exist " % run_name +
```
savingoyal, line 188: The exception could be simply - `No run *%s* found for *%s*`.

metaflow/plugins/tagging_cli.py
```diff
@@ -0,0 +1,424 @@
...
+                               "in any namespace.\nPlease check your --run-id.")
+    except MetaflowNamespaceMismatch:
+        namespace(None)
+        metaflow_client = Run(pathspec=run_name)
+
+        run_tags = metaflow_client.tags
+
+        if any([is_prod_token(tag) for tag in run_tags]):
```
savingoyal, line 196: Shouldn't this check be outside the except block?

metaflow/plugins/tagging_cli.py
```diff
@@ -0,0 +1,424 @@
...
+    all_tags = {metaflow_run_client.pathspec: metaflow_run_client.tags}
+    return format_tags(obj, system_tags, all_tags, by_tag=False)
+
+
+def is_prod_token(token):
+    return token is not None and token.startswith("production:")
+
+
+def obtain_metaflow_run_client(obj, run_id, user_namespace):
+    if is_prod_token(user_namespace):
```
savingoyal, line 141: I am not sure this check is correct. I can pass in a different namespace for a scheduled run. Also, can we consider allowing modifying a scheduled run's tags?

metaflow/plugins/tagging_cli.py
savingoyal: Also, we should be checking the system tags here rather than all the tags. I could add a tag that begins with `production` for a local run.

metaflow/plugins/tagging_cli.py
```diff
@@ -0,0 +1,424 @@
...
+        namespace(None)
+        metaflow_client = Run(pathspec=run_name)
+
+        run_tags = metaflow_client.tags
+
+        if any([is_prod_token(tag) for tag in run_tags]):
+            raise CommandException(
+                "Modifying a scheduled run's tags is currently not allowed.")
+
+        msg = "Run *%s* does not belong to " % run_name + \
```
savingoyal, line 200: I am not sure what's the intention here. We can simply let the user know similar to L158 - `No run *%s* found for *%s* in namespace *%s*`. The user will then be able to do `tags list` to get all the current tags if they want to explicitly override the namespace.

metaflow/plugins/tagging_cli.py
```diff
@@ -0,0 +1,424 @@
...
+    # 2. our user tries to mutate tags on a meson/production run
+    # 3. our user's --run-id does not exist in the default/specified namespace
+    try:
+        namespace(user_namespace)
+        metaflow_client = Run(pathspec=run_name)
+    except MetaflowNotFound:
+        raise CommandException("Run *%s* does not exist " % run_name +
+                               "in any namespace.\nPlease check your --run-id.")
+    except MetaflowNamespaceMismatch:
+        namespace(None)
```
savingoyal, line 191: What's the intention here?

metaflow/plugins/tagging_cli.py
```diff
@@ -0,0 +1,424 @@
...
+    metaflow_run_client.remove_tags(tags)
+
+    obj.echo("Operation successful. New tags:")
+    print_tags(obj, metaflow_run_client)
+
+
+@tag.command('replace',
+             help='Replace tags of a run.')
+@click.option('--run-id',
+              required=False,  # set False here so we can throw a better error message
```
savingoyal, line 293: Why can't we set this to required? It will simplify the logic in the first 200 lines of this file tremendously.

metaflow/plugins/tagging_cli.py
```diff
@@ -0,0 +1,424 @@
...
+
+from metaflow import namespace
+from metaflow.client import Flow, Run
+from metaflow.util import resolve_identity
+from metaflow.exception import CommandException, MetaflowNotFound
+from metaflow.exception import MetaflowNamespaceMismatch
+
+def format_tags(obj, system_tags, all_tags, by_tag):
+    hdr = ["",
+           "System tags and *user tags*; only *user tags* can be removed",
```
savingoyal, line 14: We don't have a notion of `user tags` for the end user. Also, `can be removed` should read `can be modified`.

metaflow/plugins/tagging_cli.py
```diff
@@ -0,0 +1,424 @@
...
+            ','.join(runs)]))
+
+
+def print_tags(obj, metaflow_run_client):
+    system_tags = {metaflow_run_client.pathspec: metaflow_run_client.system_tags}
+    all_tags = {metaflow_run_client.pathspec: metaflow_run_client.tags}
+    return format_tags(obj, system_tags, all_tags, by_tag=False)
+
+
+def is_prod_token(token):
```
savingoyal, line 136: I am not sure if this is the right check. The `@project` decorator introduces the notion of a `production` branch as well and it is easy for future code committers to get confused about what we mean here. I think we should consider allowing the end users to modify SFN run tags.

savingoyal (changes requested):
@romain-intel I made some comments - looks good overall. I haven't tested with the service yet - will wait for your go ahead.

metaflow/client/core.py
```diff
@@ -491,6 +533,21 @@ def tags(self):
         """
         return self._tags
 
+    @property
+    def system_tags(self):
```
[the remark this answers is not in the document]
romain-intel: I don't mind doing it. Would refactor the code a bit though (I would introduce another subclass that would inherit from MetaflowObject and that would be the parent of other objects. Also, would break compatibility (`created_at` change). Am fine doing it though if that is ok. I do agree, it is better.

I am not sure however if it makes sense to do it right now in the sense that tags at the step/task/run/artifact levels are also a bit iffy in terms of what they mean. What is a tag at a step level for example?

I don't have a strong opinion on the matter though.

metaflow/client/core.py
```diff
@@ -576,6 +633,101 @@ def traverse(cls, ids_r, lst):
...
+
+        Parameters
+        ----------
+        remove_tags : string or Iterable over string
+            Tags to remove
+        add_tags : string or Iterable over string
+            Tags to add
+        """
+        if not all([is_stringish(x) for x in chain(remove_tags, add_tags)]):
+            print("Found tags: %s and %s" % (str(remove_tags), str(add_tags)))
```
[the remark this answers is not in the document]
romain-intel: oh, sorry, that's probably left over from debugging. I'll clean it up.

metaflow/metadata/metadata.py
romain-intel: This was a long-coming cleanup. I did it at the same time given I was modifying this stuff. It's unsafe to use mutable objects as default parameters in python (yes, it didn't break anything here but since I was cleaning things up, I just did it).

metaflow/plugins/tagging_cli.py
romain-intel: How would you phrase the whole sentence?

metaflow/plugins/tagging_cli.py
romain-intel: @jasonge27 : what was the rationale for preventing this?

metaflow/plugins/tagging_cli.py
romain-intel: This is because you can do `tag list MyFlow/RunID`. Am fine not doing that if you don't want to though. I'll just have to check if it breaks something internally.

metaflow/plugins/tagging_cli.py
romain-intel: Are you talking about all the exceptions handling at the top? I think the idea was to try to be friendly to the user. @jasonge27 might have more insights (I copied a lot of the code).

metaflow/plugins/tagging_cli.py
jasonge27: The original intention was to avoid modifying tags for deployed runs since the ownership of deployed runs is not clear. Agree that this is probably not the best way to check if the run is deployed on SFN/Meson

metaflow/plugins/tagging_cli.py
jasonge27: `get_run_client` ?

metaflow/plugins/tagging_cli.py
jasonge27: Simply to generate a proper error message with useful info. There's `MetaflowNamespaceMismatch` exception when the `run id` user specified does not live in the namespace. We will then try to locate the correct namespace and suggest to user in the error message

metaflow/plugins/tagging_cli.py
jasonge27: Right, simply want to be more user-friendly. If we set it to False, and if the user does not specify run-id, `click` will throw an error with the help message

metaflow/plugins/tagging_cli.py
jasonge27: Yes `tag list` should help here. The original intention was to add some more info in the error message that could help.

metaflow/plugins/tagging_cli.py
romain-intel: Not changing to avoid breaking compatibility with internal commands.

metaflow/plugins/tagging_cli.py
romain-intel: went with get_run_client

metaflow/plugins/tagging_cli.py
romain-intel: This requires a bit more thought. I think one way we could modify it is by allowing tags to be specified in the triggers. This would make ownerhsip a bit clearer.

metaflow/client/core.py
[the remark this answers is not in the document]
romain-intel: Moved all the tag operations to run; kept `created_at` to avoid breaking changes.

metaflow/plugins/tagging_cli.py
romain-intel: removed and just kept `run_id`.

metaflow/plugins/tagging_cli.py
romain-intel: I am not sure what you mean here. I simplified the logic a bit but let me know if you have other specific asks.

metaflow/plugins/tagging_cli.py
romain-intel: It's in the except block because it is basically to figure out the error message to print. Am simplifying the logic a bit but in all cases, there will be an error basically.

metaflow/plugins/tagging_cli.py
romain-intel: what would a better check be?

romain-intel:
I have done almost *no* testing yet. I did verify that the CLI works at least somewhat as expected. There is no service support yet as I want to make sure we are all on the same page for the behavior.

LMK what you think.

savingoyal:
Linking the agreed proposal here - https://docs.google.com/document/d/1YeOqU773bAH7S5FTDzT-jWcS5SENT38afi96LrnJ3Bg/edit#

metaflow/client/core.py
```diff
@@ -1667,6 +1736,100 @@ def end_task(self):
...
+        If the tag is already a system tag, this is a no-op but no error is returned.
+
+        Parameters
+        ----------
+        tag : string
+            Tag to add
+        """
+        self.add_tags([tag])
+
+    def add_tags(self, tags):
```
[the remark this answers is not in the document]
savingoyal: Similarly for `remove_tags` and `replace_tags`

metaflow/metadata/metadata.py
```diff
@@ -285,6 +289,25 @@ def _get_object_internal(
         """
         raise NotImplementedError()
 
+    @classmethod
+    def _perform_operations_internal(cls, operations):
```
savingoyal, line 293: Minor nit - this can be simply `_perform_tag_operations`

metaflow/plugins/tagging_cli.py
```diff
@@ -0,0 +1,457 @@
...
+from metaflow.client import Flow, Run
+from metaflow.util import resolve_identity
+from metaflow.exception import CommandException, MetaflowNotFound
+from metaflow.exception import MetaflowNamespaceMismatch
+
+
+def format_tags(obj, system_tags, all_tags, by_tag):
+    hdr = [
+        "",
+        "System tags and *tags*; only *tags* can be modified",
```
savingoyal, line 16: Perhaps we can remove this specific header and in the footer mention that certain tags are immutable.

metaflow/plugins/tagging_cli.py
savingoyal: Given that we don't have any notion of `prod_token`, we should remove this check entirely and rely on user being in the correct namespace.

metaflow/plugins/tagging_cli.py
savingoyal: A better check would be to ensure you are in the right namespace.

metaflow/plugins/tagging_cli.py
savingoyal: We should ensure `tag list` returns a response for the latest run-id rather than the union of tags for all runs.

metaflow/plugins/tagging_cli.py
```diff
@@ -0,0 +1,457 @@
...
+
+
+@cli.group(help="Commands related to Metaflow tagging.")
+def tag():
+    pass
+
+
+@tag.command("add", help="Add tags to a run.")
+@click.option(
+    "--run-id",
```
savingoyal, line 259: How about simplifying the logic in `get_run_client` by making one of `--latest-successful-run`, `--latest-run` and `--run-id` required?

metaflow/plugins/tagging_cli.py
```diff
@@ -0,0 +1,457 @@
...
+    obj, run_id, hide_system_tags, list_all, my_runs, by_tag, by_run, arg_run_id
+):
+    if run_id is None and arg_run_id is None and not list_all and not my_runs:
+        # Assume list_all by default
+        list_all = True
+
+    if list_all and my_runs:
+        raise CommandException(
+            "Option --all cannot be used together with --my-runs.\n"
+            "You can run the flow with commands 'tag list --help' "
```
savingoyal, line 398: Nit - you can't `run` a flow here.

metaflow/plugins/tagging_cli.py
```diff
@@ -0,0 +1,457 @@
...
+            "to check the meaning of the two options."
```
savingoyal, line 399: Ideally we should indicate the meaning right here rather than expect the user to type another command to get the information needed.

savingoyal (changes requested):
The local implementation checks out. I have added a few UX related comments. One concern is around the behavior of `replace` - should it replace tags that don't exist?

romain-intel:
> The local implementation checks out. I have added a few UX related comments. One concern is around the behavior of `replace` - should it replace tags that don't exist?

Thanks for the comments. I will address them and update this. Regarding `replace`, IMHO, if we treat it as a "remove" followed by an "add" which it is, it would replace non existing tags because the remove would be a no-op and the add would add it. However, I could see an argument to say that replace is a "no-op" if there is nothing to replace (so it would be different than a "remove" followed by an "add", it would have to be an "effectual remove" followed by an add. I am open to both. The UX seems more intuitive in the latter scenario though. Thoughts?

metaflow/client/core.py
```diff
@@ -376,13 +382,16 @@ def __init__(
 
         self._created_at = datetime.fromtimestamp(self._object["ts_epoch"] / 1000.0)
 
-        self._tags = frozenset(
-            chain(self._object.get("system_tags") or [], self._object.get("tags") or [])
-        )
+        # Keep system tags and a set union of system tags + customized tags
+        self._system_tags = self._get_tags_set(self._object.get("system_tags"), [])
```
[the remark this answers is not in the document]
romain-intel: It makes it a frozenset. I'll get that as is for now.

metaflow/client/core.py
```diff
@@ -450,6 +459,51 @@ def _filtered_children(self, *tags):
     def _url_token(cls):
         return "%ss" % cls._NAME
 
+    def _tag_operations(self, operations):
+        if self._NAME != "run":
+            raise TaggingException("Tagging operations are only allowed on runs")
+
+        if len(operations) == 0:
+            raise TaggingException("At least one tagging operation must be provided")
```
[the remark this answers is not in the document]
romain-intel: Removed it. It should never happen as this is an internal function

metaflow/client/core.py
```diff
@@ -450,6 +459,51 @@ def _filtered_children(self, *tags):
...
+                )
+            )
+        result = self._metaflow.metadata.perform_operations(formatted_operations).get(
+            self.pathspec, None
+        )
+        if result and result["status"] == "ok":
+            self._object = result["data"]
+        else:
+            raise TaggingException(
+                "Unable to perform tagging operation on %s: %s"
```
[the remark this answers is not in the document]
romain-intel: It may get confusing since a replace is a add and a remove. We could add logic here and in the backend if you want but not sure it is useful.

metaflow/plugins/tagging_cli.py
romain-intel: The idea was to show the ones that could be. I rephrased a bit though.

PR #554, opened 2021-06-04
UBF Implementation

romain-intel:


metaflow/cli_args.py
```diff
@@ -0,0 +1,73 @@
...
+                if k == 'decospecs':
+                    k = 'with'
+                k = k.replace('_', '-')
+                v = v if isinstance(v, (list, tuple, set)) else [v]
+                for value in v:
+                    yield '--%s' % k
+                    if not isinstance(value, bool):
+                        yield to_unicode(value)
+
+cli_args = CLIArgs()
```
pjoshi30, line 73: nit: new line at EOF

metaflow/unbounded_foreach.py
```diff
@@ -0,0 +1,11 @@
...
+UBF_CONTROL = 'ubf_control'
+UBF_TASK = 'ubf_task'
+
+class UnboundedForeachInput(object):
+    """
+    Plugins that wish to support `UnboundedForeach` need their special
+    input(s) subclass this class.
+    This is used by the runtime to detect the difference between bounded
+    and unbounded foreach, based on the variable passed to `foreach`.
+    """
```
pjoshi30, line 11: nit: new line at EOF

metaflow/cli.py
```diff
@@ -473,21 +481,30 @@ def step(obj,
...
+    if ctx.obj.datastore.datastore_root is None:
+        ctx.obj.datastore.datastore_root = \
+            ctx.obj.datastore.get_datastore_root_from_config(ctx.obj.echo)
 
-    obj.metadata.add_sticky_tags(tags=tags)
+    step_kwargs = ctx.params
+    # Remove argument `step_name` from `step_kwargs`.
+    step_kwargs.pop('step_name', None)
+    # Remove `opt_*` prefix from (some) option keys.
+    step_kwargs = dict([(k[4:], v) if k.startswith('opt_') else (k, v)
```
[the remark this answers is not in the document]
romain-intel: They come from the click parameters which should be a fixed set so we should be good but let me know if you think I missed something.

metaflow/cli.py
```diff
@@ -1003,4 +1021,4 @@ def main(flow, args=None, handle_exceptions=True, entrypoint=None):
             print_unknown_exception(x)
             sys.exit(1)
         else:
-            raise
+            raise
```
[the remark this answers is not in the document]
romain-intel: will address. We want to run black on the whole code base at some point. There are a *bunch* of inconsistencies.

metaflow/plugins/unbounded_foreach_decorator.py
```diff
@@ -0,0 +1,135 @@
...
+                      ubf_context):
+        if ubf_context == UBF_CONTROL:
+            from functools import partial
+            return partial(self.control_task_step_func, flow, graph, retry_count)
+        else:
+            return step_func
+
+    def step_task_retry_count(self):
+        # UBF plugins don't want retry for the control task. We signal this
+        # intent to the runtime by returning (None, None).
```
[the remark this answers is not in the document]
romain-intel: It would require a bit more change and changing the existing decorators. I think this may be a good idea to revisit when we redo the whole API for decorators (which is hopefully going to happen at some point).

metaflow/runtime.py
```diff
@@ -283,35 +289,108 @@ def _queue_task_join(self, task, next_steps):
...
-            required_tasks = [self._finished.get((task.step, s))
-                              for s in siblings(foreach_stack)]
-            join_type = 'foreach'
+        unbounded_foreach = not task.results.is_none('_unbounded_foreach')
+
+        if unbounded_foreach:
+            # Before we queue the join, do some post-processing of runtime state
+            # (_finished, _is_cloned) for the (sibling) mapper tasks.
+            # Update state of (sibling) mapper tasks for control task.
+            if task.ubf_context == UBF_CONTROL:
```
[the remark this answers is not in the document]
romain-intel: we can. There is little repeating functionality though. If you have a specific idea though, please push an update to it. I do agree it's rather large :).

metaflow/plugins/unbounded_foreach_decorator.py
```diff
@@ -0,0 +1,135 @@
+import os
+import subprocess
+import sys
+from metaflow.cli_args import cli_args
+from metaflow.decorators import StepDecorator
+from metaflow.exception import MetaflowException
+from metaflow.unbounded_foreach import UnboundedForeachInput, UBF_CONTROL, UBF_TASK
+from metaflow.util import to_unicode
+
+class InternalUnboundedForeachInput(UnboundedForeachInput):
```
tuulos, line 10: Should we rename this module to `unbounded_foreach_test_decorator.py`? The class names in this module could include the word `test` too since these are only needed to enable tests, they have no other function.

The current name implies that this module is actually required by UBF.

test/core/run_tests.py
```diff
@@ -95,6 +95,7 @@ def run_cmd(mode):
                     'LC_ALL': 'C.UTF-8',
                     'PYTHONIOENCODING': 'utf_8',
                     'PATH': os.environ.get('PATH', '.'),
+                    'PYTHONIOENCODING': 'utf_8',
```
tuulos, line 98: `PYTHONIOENCODING` seems to be defined twice

metaflow/plugins/unbounded_foreach_decorator.py
romain-intel: Sure. It was named that way in #450. @savingoyal changed it to this. I have no preference either way.

test/core/run_tests.py
romain-intel: will fix

metaflow/plugins/aws/batch/batch_cli.py
```diff
@@ -167,6 +167,7 @@ def kill(ctx, run_id, user, my_runs):
 @click.option("--shared_memory", help="Shared Memory requirement for AWS Batch.")
 @click.option("--max_swap", help="Max Swap requirement for AWS Batch.")
 @click.option("--swappiness", help="Swappiness requirement for AWS Batch.")
+@click.option('--ubf-context', default=None, type=click.Choice([None])) #TODO: Maybe remove it altogether since it's not used here
```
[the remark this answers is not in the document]
romain-intel: I think so since batch does not make use of it. I left it for now.

metaflow/plugins/unbounded_foreach_decorator.py
romain-intel: Called in InternalTest... and changed other references.

Issue #555, opened 2021-06-04
merge_artifacts loads objects into memory


peterwilton-ttam:
I was surprised to see that `FlowSpec.merge_artifacts` loads all of the merged artifacts into memory after doing all of the consistency checks with hashes. Is there any reason it’s necessary to load all of the artifacts into memory? Would it instead be possible to copy over hashes and metadata from the input-branches' datastores into the join-step's datastore? That is, would it be possible to avoid `getattr` and `setattr` in `merge_artifacts`?

I have a flow with a join step that does nothing more than merge lots of artifacts of different names from upstream branches. I expected this to be a low-memory operation, but of course I'm finding I need to request enough memory to get all of these artifacts into memory. In my case this is a substantial request. And of course decompressing, depickling, repickling, and recompressing takes some time, too.

Here's a small patch of `merge_artifacts` that seems to accomplish this:

``` patch
-        for var, (inp, _) in to_merge.items():
-            setattr(self, var, getattr(inp, var))
+        for var, (inp, sha) in to_merge.items():
+            if self._datastore.object_exists(sha):
+                self._datastore.objects[var] = sha
+                self._datastore.info[var] = inp._datastore.info[var]
+            else:
+                setattr(self, var, getattr(inp, var))
```

where `LocalDataStore.object_exists` and `S3DataStore.object_exists` are defined in the expected way.

I'm having some difficulty getting all of the `dev-local` tests to pass on `master` (three tests are failing, and it seems it has to do with my local Python environments), but those tests that do pass on `master` also pass with these changes.

Is this something you've considered before? Am I missing something that would break with such changes?

Thanks!

romain-intel:
Hi. This is an excellent suggestion (and don't know how I didn't see this). I think we may also be able to use the `passdown_partial` method in the datastore to do this. May I ask, however, why the check on `object_exists`? If it is present in one of the inputs, it *should* be present in the datastore (or am I missing something).

peterwilton-ttam:
Great, thanks for the response.

> May I ask, however, why the check on `object_exists`?

I wasn't sure whether it was safe to assume that the input datastore is always configured like the join-step datastore. I suppose it is (it's not possible to have one step backed by a local datastore and the next by S3, say), in which case I agree it would be fine to just copy over the book-keeping without checking for existence.

peterwilton-ttam:
[...]

PR #556, opened 2021-06-04
Merge artifacts without loading them into memory

peterwilton-ttam:
I was surprised to see that `FlowSpec.merge_artifacts` loads all of the merged artifacts into memory after doing all of the consistency checks with hashes. This PR proposes a way of avoiding to avoid the memory and compute costs of decompressing, depickling, repickling, and recompressing each merged artifact.

I brought up the possibility in [this issue](https://github.com/Netflix/metaflow/issues/555), and @romain-intel suggested that some checks for existence that I initially added were probably unnecessary—that a blob that's accessible from an input branch's datastore would be accessible from the join step's datastore as well. So this PR has four commits: two implementing what I first proposed, and another two removing the checks for existence.

@romain-intel also suggested that the same effect might be achieved using `MetaflowDataStore.passdown_partial`. For example, maybe one could do something like

``` python
from collections import defaultdict
...
        # (end of merge_artifacts)
        input_merge_vars = defaultdict(list)
        for var, (inp, _) in to_merge.items():
            input_merge_vars[inp].append(var)
        for inp, vars in input_merge_vars.items():
            self._datastore.passdown_partial(inp._datastore, vars)

        # or maybe, instead
        for var, (inp, _) in to_merge.items():
            self._datastore.passdown_partial(inp._datastore, [var])
```

I'm having some difficulty getting all of the dev-local tests to pass on `master` (three tests are failing, and it seems it has to do with my local Python environments), but those tests that do pass on `master` also pass with the changes in this PR.

### DIFF
--- a/metaflow/datastore/datastore.py
+++ b/metaflow/datastore/datastore.py
@@ -154,12 +154,6 @@ class MetaflowDataStore(object):
         """
         raise NotImplementedError()
 
-    def object_exists(self, sha):
-        """
-        Return whether object exists in datastore
-        """
-        raise NotImplementedError()
-
     @classmethod
     def get_latest_tasks(cls,
                          flow_name,
--- a/metaflow/datastore/datastore.py
+++ b/metaflow/datastore/datastore.py
@@ -154,6 +154,12 @@ class MetaflowDataStore(object):
         """
         raise NotImplementedError()
 
+    def object_exists(self, sha):
+        """
+        Return whether object exists in datastore
+        """
+        raise NotImplementedError()
+
     @classmethod
     def get_latest_tasks(cls,
                          flow_name,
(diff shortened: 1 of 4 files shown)


peterwilton-ttam:
Update: I did get all of the `dev-local` tests to pass. Not sure if there is a CI system in place to run these tests automatically.

romain-intel (commented):
maybe do:
```
for var, (inp, _) in to_merge.items():
  self._datastore.passdown_partial(inp._datastore, [var])
```
I would prefer to use a function than to kind of go muck in the internals (makes it a bit harder to maintain).

romain-intel:
I launched the tests as well (updates should launch more tests).

savingoyal (approved):

romain-intel (approved):
WIll merge once tests are successful (should be). Thanks for the quick change.

peterwilton-ttam:
Thanks for the speedy review and merging!

PR #559, opened 2021-06-07
Fix lifetime issue for S3() object

romain-intel:


savingoyal (approved):
LGTM. Let's merge it!

romain-intel:
Based on our discussions. We will not use this and use #568 instead.

PR #561, opened 2021-06-08
Alternative proposal to cleaning up S3 object from s3_backend

romain-intel:


romain-intel:
Based on our discussion, we will not need this and use #568 instead.

PR #565, opened 2021-06-12
R decorators

### LINKED ISSUE #563: (R) Create individual `decorator` functions and improve documentation
Currently, R decorators are specified in a very general way:

```r
decorator(x, ...)
```

where `x` is the name of the decorator and `...` are the arguments provided to that specific decorator (similar to Python's `**kwargs`). This isn't ideal because it's difficult to:

* discover which decorators are available, 
* discover which arguments should be provided for a given decorator, 
* auto-complete arguments, and
* provide specific documentation for each decorator.

I propose that we create, for each decorator, a specific function that wraps the `decorator` function. For example,

```r
resources <- function(cpu = 1L, memory = 4096L) {
  decorator("resources", cpu = cpu, memory = memory)
}
``` 

Each of these can then be individually documented and unit tested, and we can link to available decorators from the man page for `decorator`.

There may be a slight issue with `environment`, which is a (pretty important) function in base R.

I'm happy to work on this as a PR, but I want to check that this approach is okay before I begin.

mdneuzerling:
Closes #563

There are a couple of changes here, although nothing that will break backwards compatibility:

* New functions wrapping `decorator`. These are all documented, and the documentation for `decorator` links to the relevant man pages:
  * `resources`
  * `batch`
  * `retry`
  * `catch`
  * `environment_variables` --- we can't use `environment` as that's a very important base function. I've tried to make the syntax for this one match the base R `Sys.setenv` function.
* Enabled inline markdown for docstrings. This was already being used, so this just makes it parse as markdown in the man pages.
* `wrap_argument` now parses an R `NULL` to Python `None`. This is so we can use default arguments for decorators, many of which are `None` in Python-land.
* New internal utility function `%||%`, which is like a SQL `ISNULL`/`COALESCE` function. It's just handy to have.
* Added some Roxygen documentation for some internal decorator functions. These are marked `@keywords internal` so they shouldn't appear in documentation indices.
* I've added myself as a contributor. I hope that's okay. I've also bumped the version from `2.2.0` to `2.3.0`, but I'm not sure if the R version is allowed to drift from the Python version.

I'm not convinced that's all the decorators. These are the ones that are documented in the R pages of [docs.metaflow.org](https://docs.metaflow.org/v/r/). Besides these there's also `timeout` (I believe this may be broken for R?) and `project`, and maybe some others. I'm happy to create wrappers for any other decorators as part of this PR.

I also can't get `R CMD check` to pass. A lot of tests are failing, although everything passes with `devtools::test()`. I didn't invest much time debugging this because `R CMD check` may require a bespoke setup that I don't have


savingoyal:
Thanks! I will start reviewing it!~

R/R/decorators-aws.R
```diff
@@ -0,0 +1,125 @@
...
+#'   run()
+#' }
+batch <- function(
+  cpu = 1L,
+  gpu = 0L,
+  memory = 4096L,
+  image = NULL,
+  queue = NULL,
+  iam_role = NULL,
+  execution_role = NULL,
```
jasonge27, line 86: shall we add `run_time_limit` as well here? Seems like this is the only config missing from current configs supported by batch on the python side.

R/R/decorators-aws.R
```diff
@@ -0,0 +1,125 @@
...
+  )
+}
+
+#' @rdname batch
+#' @export
+resources <- function(
+  cpu = 1L,
+  gpu = 0L,
+  memory = 4096L,
+  shared_memory = NULL
```
jasonge27, line 116: the argument list should be the same as the list for batch?

mdneuzerling: My reading of [the (Python) source code](https://github.com/Netflix/metaflow/blob/4b82c586b218efe226be1dd88a1dc253b5c24e63/metaflow/plugins/aws/batch/batch_decorator.py#L33) is that `@resources` doesn't take the same arguments. To verify, I tried a Python flow with a `@resources(max_swap = 100)` and I got:

```python
metaflow.exception.InvalidDecoratorAttribute: Decorator 'resources' does not support the attribute 'max_swap'. These attributes are supported: cpu, gpu, memory, shared_memory.
```

R/R/decorators-aws.R
mdneuzerling: I'll get to work!

jasonge27:
Thanks a lot for the PR @mdneuzerling ! Checked a few things (1) ran our integration tests on AWS (2) ran example flows for `catch`, `batch`, 'retry', and 'environment_variables`, (3) ran R CMD check --as-cran. (4) sanity check the generated markdown files on RStudio. All look good to me!

We would need to update the doc as well for example https://docs.metaflow.org/v/r/metaflow/failures#catching-exceptions-with-the-catch-decorator 
@savingoyal @mdneuzerling

savingoyal:
> Thanks a lot for the PR @mdneuzerling ! Checked a few things (1) ran our integration tests on AWS (2) ran the example flows in the docstring for `catch`, `batch`, `retry`, and `environment_variables`, (3) ran R CMD check --as-cran. (4) sanity check the generated markdown files on RStudio. All look good to me!
> 
> PS: @mdneuzerling is right on `timeout` decorator -- it does not work with R yet.
> 
> We would need to update the doc as well for example https://docs.metaflow.org/v/r/metaflow/failures#catching-exceptions-with-the-catch-decorator
> @savingoyal @mdneuzerling

I have updated the spurious reference to `timeout` in the docs.

savingoyal:
The PR looks great! @mdneuzerling Can you also make the edits for the documentation? If iterating on google docs is easier, you can start a doc, and then either I or @jasonge27 can copy them over to Gitbook. The github -> gitbook bridge is a bit unreliable, unfortunately.

savingoyal (approved):

Issue #562, opened 2021-06-08
imp DeprecationWarning


dpoznik:
I get the following warning from `pytest`:
```
==================================== warnings summary =====================================
../../../../usr/local/lib/python3.8/site-packages/metaflow/R.py:2
  /usr/local/lib/python3.8/site-packages/metaflow/R.py:2: 
DeprecationWarning: the imp module is deprecated in favour of importlib; 
see the module's documentation for alternative uses
    import imp

-- Docs: https://docs.pytest.org/en/stable/warnings.html
```
I think this change would address the warning, but only for `py3`:
https://github.com/Netflix/metaflow/compare/master...dpoznik:replace-imp

I can, of course, just silence the warning on my end if there's not a good fix.

savingoyal:
Thanks for the issue! The change would definitely address the warning, but only for py3. Flows using Python2 underneath will start breaking - maybe we can introduce a `try... except...` clause to handle both py2 and py3 properly?

dpoznik:
> maybe we can introduce a `try... except...` clause to handle both py2 and py3 properly?

OK, sounds good. I'll put in a PR in a moment...

dpoznik:
Resolved in #1396

PR #566, opened 2021-06-14
Address imp DeprecationWarning

dpoznik:
The `imp` module is deprecated in favor of `importlib`, but `importlib.machinery.SourceFileLoader` wasn't introduced until Python 3.3, so it cannot be relied upon for Python 2 flows. Therefore, we use `importlib` when possible, but drop back to `imp` when necessary.

This PR resolves the following `pytest` warning:
```
==================================== warnings summary =====================================
../../../../usr/local/lib/python3.8/site-packages/metaflow/R.py:2
  /usr/local/lib/python3.8/site-packages/metaflow/R.py:2: 
DeprecationWarning: the imp module is deprecated in favour of importlib; 
see the module's documentation for alternative uses
    import imp
```

For discussion, please see: https://github.com/Netflix/metaflow/issues/562.


metaflow/R.py
```diff
@@ -1,5 +1,8 @@
 import os
-import imp
+try:
```
savingoyal, line 2: Can we do something of the order of 
```
try:
    def source(fullname, path):
        from importlib.machinery import SourceFileLoader
        return SourceFileLoader(fullname, path).load_module()
except ImportError: #Python2
    def source(fullname, path):
         import imp
         return imp.load_source(fullname, path)
```

savingoyal:
Thanks, I will test and merge the PR for the next release.

metaflow/R.py
```diff
@@ -12,6 +11,15 @@
 R_VERSION = None
 R_VERSION_CODE = None
 
+try:
+    def source(fullname, path):
+        from importlib.machinery import SourceFileLoader
+        return SourceFileLoader(fullname, path).load_module()
```
romain-intel, line 17: Since this is to support things going forward, this seems deprecated in 3.6 (https://docs.python.org/3/library/importlib.html#importlib.machinery.SourceFileLoader.load_module) and we should now use: https://docs.python.org/3/library/importlib.html#importlib.abc.Loader.exec_module. I would try with exec_module first (already present in 3.4 and so most likely works), and if that doesn't work, fall back to imp (or maybe fallback to load_module and then imp if you want to be fancy but I don't think that is worth the trouble).

dpoznik: Good call. I'll push up the change and let the CI confirm that it works.

dpoznik: The tests are unhappy. Looks like `SourceFileLoader.exec_module` has a required argument (`module`) that I neglected to include in the most recent commit. It's not immediately clear to me what that argument should be, so I think I'll have to read the `importlib` docs more closely before I can fix.

dpoznik: Hmm, I extracted out the 45f915d version of `source`, and it seems to be doing what it's supposed to:
```
$ cat foo.py
#!/usr/bin/env python

import importlib.machinery
import importlib.util

def source(fullname, path):
    loader = importlib.machinery.SourceFileLoader(fullname, path)
    spec = importlib.util.spec_from_loader(loader.name, loader)
    module = importlib.util.module_from_spec(spec)
    loader.exec_module(module)
    return module

module = source("bar", "bar.py")
print(module)
module.hello()

$ cat bar.py 
def hello():
    print("hello")

$ ./foo.py 
<module 'bar' from 'bar.py'>
hello
```
But the tests are failing with:
```
 Error in py_call_impl(callable, dots$args, dots$keywords) : 
  ModuleNotFoundError: No module named 'metaflowR'
```

romain-intel:
I tried to fix it. Let's see if that helps.

romain-intel:
Hun, that still didn't work :(. I'll try to look a bit more in detail a little later. I'm a bit stumped right now. It's probably something simple :(.

dpoznik:
Sounds good. Curious to see what the solution ends up being :)

metaflow/graph.py
```diff
@@ -168,7 +169,8 @@ def __init__(self, flow):
         self._postprocess()
 
     def _create_nodes(self, flow):
-        module = __import__(flow.__module__)
+        importlib.invalidate_caches()
```
dpoznik, line 172: Not sure if this will work, but it seemed germane: "This function should be called if any modules are created/installed while your program is running to guarantee all finders will notice the new module’s existence."
https://docs.python.org/3/library/importlib.html#importlib.invalidate_caches

Can remove if it doesn't work.

dpoznik: This call would obviously have been more appropriate in `R.py`, but looks like it didn't work anyhow. Will revert.

PR #568, opened 2021-06-16
Add a way to pass an external client to S3()

romain-intel:


metaflow/datatools/s3.py
```diff
@@ -873,11 +888,10 @@ def _one_boto_op(self, op, url):
...
                                      prefix='metaflow.s3.one_file.',
                                      delete=False)
-            self.reset_client()
-            try:    
-                op(self._s3_client, tmp.name)
+            try:
+                op(self._s3_client.client, tmp.name)
                 return tmp.name
-            except self._s3_client_error as err:
+            except self._s3_client.error as err:
```
tuulos, line 894: it is unusual to have side-effect hidden inside an `except` clause like in this case (`.error` can trigger `get_aws_client`). We could use `_s3_client_error` as before

metaflow/datatools/s3.py
```diff
@@ -304,7 +327,7 @@ def __init__(self,
             # 3. use the client only with full URLs
             self._s3root = None
 
-        self._s3_client = None
+        self._s3_client = kwargs.get('external_client', S3Client())
```
tuulos, line 330: you could add a comment here a'la
```
# The S3 datastore of Metaflow uses this class internally. Passing in an existing `boto` connection as `external_client` allows it to reuse an existing connection, which lowers latency for small operations. Note that this argument is not a part of the public API and may disappear of change in the future
```

metaflow/datatools/s3.py
romain-intel: It no longer exists. In this case though, there is no way for the side effect to happen since the try statement accesses `.client` which would do any triggering at that point.

romain-intel:
This was already included in other commits and is no longer needed.

PR #570, opened 2021-06-16
The datastore now caches the S3 boto connection and passes it to the …

romain-intel:
…S3 client

This allows us to use S3() using with (and therefore properly clean up the temp
directory) but also benefit from the connection reuse.

As part of this change, we also had to change the interface to load_bytes. This
change is invisible to the user as it is 100% internal to the datastore.

Cleaned up the s3tail.py and s3util.py files to co-locate them with the
datatools instead of the datastore (since they were no longer used directly
in the datastore). This simplified the import story.


metaflow/datastore/s3_backend.py
```diff
@@ -220,26 +204,23 @@ def load_bytes(self, paths):
...
-        to_return = {}
         if len(paths) > 10:
-            results = self.s3_datatool.get_many(
-                paths, return_missing=True, return_info=True)
+            results = s3.get_many(paths, return_missing=True, return_info=True)
             for r in results:
                 if r.exists:
-                    to_return[r.key] = (FileWithClientRef(
-                        self.s3_datatool, r.path), r.metadata)
+                    to_return[r.key] = (LazyFile(r.path), r.metadata)
```
tuulos, line 216: why do we need `LazyFile` here instead of just passing `r.path` (no FD).

It seems you just open and read files sequentially here https://github.com/Netflix/metaflow/blob/alternative2-s3-cleanup/metaflow/datastore/task_datastore.py#L744 and here https://github.com/Netflix/metaflow/blob/alternative2-s3-cleanup/metaflow/datastore/flow_datastore.py#L158 so you shouldn't use any more FDs even without `LazyFile`

metaflow/datatools/s3.py
```diff
@@ -29,7 +29,7 @@
     # python3
     from urllib.parse import urlparse
 
-from metaflow.datastore.util.s3util import get_s3_client
+from .s3util import get_s3_client
```
tuulos, line 32: can we get rid of `datastore.util` now?

metaflow/datastore/s3_backend.py
```diff
@@ -220,26 +204,23 @@ def load_bytes(self, paths):
         if len(paths) == 0:
             return {}
```
tuulos, line 205: you should return a `BytesLoader`. The `with` statement fails otherwise

metaflow/datastore/s3_backend.py
romain-intel: I didn't want to make it so that it always had to be a file so I don't want to return a path. I could return something else though that has a read interface.

metaflow/datatools/s3.py
romain-intel: Yes (sorry, forgot to erase the __init__.py file but yes, it is no longer needed. I moved it to datatools.

metaflow/datastore/s3_backend.py
romain-intel: Yep, already fixed :).

metaflow/datastore/task_datastore.py
```diff
@@ -165,14 +162,11 @@ def pathspec_index(self):
...
 
-    @property
-    def artifact_cache(self):
-        return self._artifact_cache
-
     @property
     def parent_datastore(self):
         return self._parent
 
+    @require_mode(None)
```
tuulos, line 169: why `@require_mode(None)`? Can't we just not have the decorator if the function is mode-agnostic?

tuulos:
LGTM

romain-intel:
I'll check why the tests are failing and merge once tests pass.

metaflow/datastore/task_datastore.py
romain-intel: There was other stuff in that function to check access at some point. I think it is no longer there. We can always remove later if needed. I don't mind having some sort of gatekeeper on the functions to perform some checks though.

metaflow/datastore/s3_backend.py
romain-intel: For reference, we moved to a file path for now.

PR #572, opened 2021-06-17
Enabling Basic Auth for communicating with Metadata Service

queueburt:
**Summary**

After experimenting with some other frontends for Metadata service ingress, supporting basic auth as a client-side communication mechanism seems like a worthwhile exercise.

The implementation is just the addition of a 'METAFLOW_METADATA_SERVICE_AUTH_TYPE vairable, and a few small conditionals since the HTTP header title changes based on the auth type.

**Usage**

During `metaflow configure aws`, the CLI presents an option to override the default auth type ('apigw').  If you enter 'basic', it configures the client to use the 'Authorization' header, and prepends the key with "Basic" automatically.  The key it expects in the 'METADATA_SERVICE_AUTH_KEY' variable is the base64 hashed value typically used by Basic auth, `'<username>:<password>'`.

If no environment variable is set or the CLI default is accepted, Metaflow will continue to expect API Gateway.

### DIFF
--- a/metaflow/main_cli.py
+++ b/metaflow/main_cli.py
@@ -533,6 +533,14 @@ def aws(ctx, profile):
                              default=\
                                 existing_env.get('METAFLOW_SERVICE_AUTH_KEY', ''),
                              show_default=True)
+        # Set Auth Type for the Metadata Service.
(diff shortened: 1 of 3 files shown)


saikonen:
Going through some old pull requests, is there still a need for supporting basic auth as this has not received any discussion since opening?

PR #576, opened 2021-06-22
A couple run_tests tweaks

oavdeev:
A few tweaks that didn't seem to deserve separate PRs.

1. Exclude more stuff from coverage
2. Copy `metaflow_test` module to the temp dir so it gets included into the code package. Without this, tests fail on AWS Lambda (as `metaflow_test` is not available to the step code )
3. Add `--inherit-env` option that makes run_tests.py pass its own env to the test (off by default).

### DIFF
--- a/test/core/run_tests.py
+++ b/test/core/run_tests.py
@@ -59,7 +59,7 @@ def copy_coverage_files(dstdir):
     for fname in glob.glob('.coverage.*'):
         shutil.copy(fname, dstdir)
 
-def run_test(formatter, context, coverage_dir, debug, checks):
+def run_test(formatter, context, coverage_dir, debug, checks, env_base):
 
     def run_cmd(mode):
         cmd = [context['python'], '-B', 'test_flow.py']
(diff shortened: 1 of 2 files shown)


test/core/run_tests.py
```diff
@@ -79,16 +79,20 @@ def run_cmd(mode):
         with open('check_flow.py', 'w') as f:
             f.write(formatter.check_code)
         with open('.coveragerc', 'w') as f:
-            f.write("[run]\ndisable_warnings = module-not-measured\n")
+            f.write("[run]\ndisable_warnings = module-not-measured")
```
romain-intel, line 82: nit: a reason for removing "\n" at the end?

romain-intel (commented):
Not quite sure what the impetus for these changes was but they look fine to me. @savingoyal, fine to approve and merge on my end.

test/core/run_tests.py
oavdeev: fixed (i had more changes on this line originally but then rolled them back)

romain-intel (approved):
LGTM.

PR #573, opened 2021-06-19
Check if R step names are valid Python identifiers

mdneuzerling:
Closes #414

Currently, if an R user specifies a step with a name that isn't a valid Python identifier, they're met with an obscure error:

```r
library(metaflow)
library(magrittr)

metaflow("test_flow") %>%
  step(step = "start", next_step = "bad-step-name") %>% 
  step(step = "bad-step-name", r_function = function(self) NULL, next_step = "end") %>% 
  step(step = "end") %>% 
  run()

# Error in py_call_impl(callable, dots$args, dots$keywords) : 
#   SyntaxError: invalid syntax (metaflowR.uppq6x7d, line 12)
# 
# Detailed traceback: 
#   File "/home/mdneuzerling/anaconda3/lib/python3.8/site-packages/metaflow/R.py", line 93, in run
#     module = imp.load_source('metaflowR', tmp.name)
#   File "/home/mdneuzerling/anaconda3/lib/python3.8/imp.py", line 171, in load_source
#     module = _load(spec)
#   File "<frozen importlib._bootstrap>", line 702, in _load
#   File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
#   File "<frozen importlib._bootstrap_external>", line 779, in exec_module
#   File "<frozen importlib._bootstrap_external>", line 916, in get_code
#   File "<frozen importlib._bootstrap_external>", line 846, in source_to_code
#   File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
# Calls: <Anonymous> -> py_call_impl
# Execution halted
```

This PR introduces three internal utility functions:

* `is_valid_python_identifier_py2`
* `is_valid_python_identifier_py3`
* `is_valid_python_identifier` (which calls one of the above based on `reticulate::py_version()`)

The `step` function now validates the step name using `is_valid_python_identifier`. The result is a significantly more helpful error message:

```r
library(metaflow)
library(magrittr)

metaflow("test_flow") %>%
  step(step = "start", next_step = "bad-step-name") %>% 
  step(step = "bad-step-name", r_function = function(self) NULL, next_step = "end") %>% 
  step(step = "end") %>% 
  run()

# Error in step(., step = "bad-step-name", r_function = function(self) NULL,  : 
#   bad-step-name is not a valid step name. Step names must be valid Python
# identifiers; they can contain letters, numbers, and underscores, although they
# cannot begin with a number.
```

PR #579, opened 2021-06-22
Check if R step names are valid Python identifiers (#573)

romain-intel:
DO NOT MERGE -- testing only.


romain-intel:
No longer needed; confirmed that R tests work on convergence branch.

PR #583, opened 2021-06-23
Fix return type of task.environment_info

savingoyal:
!!Breaking Change!!
The return type of task.environment_info fluctuated between a
string and a dict depending on the actual environment.

As a follow-up we need to fix the schema of the returned object.


romain-intel:
Let's hold off on this for a bit so we can merge with any other possibly breaking change. Good catch!

PR #586, opened 2021-06-26
EXPERIMENTAL FEATURE: metaflow_custom partial overrides

romain-intel:
Allow for metaflow_custom to provide a partial override for
anything that is overriden (like a plugin).

As an example, to override just the `aws_client.py` file in the `aws`
plugin, you can provide a `aws` plugin and include it in `__mf_promote_submodules__`.

The `aws` plugin only needs to provide the `aws_client.py` file as well as a
`__init__.py`. If you want to allow the base plugin to provide anything else,
add `__mf_allow_shadowed_submodules__ = ['__all__']` in `__init__.py`. By
default, an overriden module will be completely hidden.

Note that in all cases, the overriden module is visible as a subpackage of
the module called `_orig`. For example, if you override the `aws` plugin, you
can access the original plugin using `metaflow.plugins.aws._orig`. This can allow
you even more fine-grained overrides if you want.

### DIFF
--- a/metaflow/__init__.py
+++ b/metaflow/__init__.py
@@ -46,14 +46,29 @@ import importlib
 import sys
 import types
 
+if sys.version_info[0] >= 3 and sys.version_info[1] >= 4:
+    import importlib.util
+    from importlib.machinery import ModuleSpec
 
 class _LazyLoader(object):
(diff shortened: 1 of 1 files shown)


romain-intel:
Replaced with a toned-down version in #609

PR #594, opened 2021-06-30
Remove METAFLOW_ATTEMPT

oavdeev:
It is no longer necessary to have this workaround for tests, as this info is available in the `current` object.


savingoyal (approved):
LGTM! Let's ship it!

romain-intel:
LGTM

PR #596, opened 2021-07-01
Simplify new datastore [not tested thoroughly]

tuulos:
- remove artifact_cache
- simplify blob_cache (esp. in DatastoreSet)
- remove LazyFile
- small stylistic changes

### DIFF
--- a/metaflow/client/filecache.py
+++ b/metaflow/client/filecache.py
@@ -5,6 +5,7 @@ from tempfile import NamedTemporaryFile
 from hashlib import sha1
 
 from metaflow.datastore import DATASTORES, FlowDataStore
+from metaflow.datastore.content_addressed_store import BlobCache
 from metaflow.exception import MetaflowException
 from metaflow.metaflow_config import CLIENT_CACHE_PATH, CLIENT_CACHE_MAX_SIZE
 
@@ -30,18 +31,13 @@ class FileCache(object):
         self._monitor = monitor
 
         self._objects = None
-        # We will have a separate blob_cache per flow and datastore type
-        # This is to respect the fact that the keys across datastore types
-        # and/or flows may not be unique (across flows is less likely but we are
-        # keeping in line with the fact that datsatores are flow based)
-        self._blob_caches = {}
-
         # We also keep a cache for FlowDataStore objects that we created because
         # some of them may have persistent connections in their backends;
         # this is purely a performance optimization. We do *not* keep track of
         # task datastores to refresh them as needed. Caching FlowDataStore has
         # no adverse affect in terms of having to refresh the cache.
         self._store_caches = {}
+        self.blob_cache = FileBlobCache(self, cache_dir)
 
     @property
     def cache_dir(self):
@@ -110,18 +106,6 @@ class FileCache(object):
             ds_type, ds_root, data_metadata, flow_name, run_id, step_name,
             task_id, [name])[name]
 
-    def get_all_artifacts(
-            self, ds_type, ds_root, data_metadata, flow_name, run_id, step_name,
-            task_id):
-        ds = self._get_flow_datastore(ds_type, ds_root, flow_name)
-
-        # We get the task datastore for this task
-        task_ds = ds.get_task_datastore(
-            run_id, step_name, task_id, data_metadata=data_metadata)
-        # This will reuse the blob cache if needed. We do not have an
-        # artifact cache so the unpickling happens every time here.
-        return task_ds.load_artifacts([n for n, _ in task_ds.items()])
-
     def get_artifacts(
             self, ds_type, ds_root, data_metadata, flow_name, run_id, step_name,
             task_id, names):
@@ -130,8 +114,8 @@ class FileCache(object):
         # We get the task datastore for this task
         task_ds = ds.get_task_datastore(
             run_id, step_name, task_id, data_metadata=data_metadata)
-        # This will reuse the blob cache if needed. We do not have an
-        # artifact cache so the unpickling happens every time here.
+        # note that load_artifacts uses flow_datastore.castore which goes
+        # through self.blob_cache
         return task_ds.load_artifacts(names)
 
     def create_file(self, path, value):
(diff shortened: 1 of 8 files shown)


metaflow/client/filecache.py
```diff
@@ -30,18 +31,13 @@ def __init__(self, metadata=None, event_logger=None, monitor=None,
         self._monitor = monitor
 
         self._objects = None
-        # We will have a separate blob_cache per flow and datastore type
-        # This is to respect the fact that the keys across datastore types
-        # and/or flows may not be unique (across flows is less likely but we are
-        # keeping in line with the fact that datsatores are flow based)
-        self._blob_caches = {}
```
romain-intel, line 37: Why did you remove this? This can cause errors potentially (not right now but there is no guarantee that keys are unique across datastores and there is the possibility that multiple datastores are accessed by this file cache).

metaflow/datastore/content_addressed_store.py
```diff
@@ -29,18 +29,19 @@ def __init__(self, flow_datastore, prefix):
...
-        self._backend = flow_datastore._backend
-        self._logger = flow_datastore.logger
-        self._monitor = flow_datastore.monitor
-        self._blob_cache = flow_datastore.blob_cache
-
+        self._backend = backend
         self.TYPE = self._backend.TYPE
+        self._blob_cache = None
+
+    def set_blob_cache(self, blob_cache):
```
romain-intel, line 43: Nit, let's use a property here (setter and getter).

metaflow/datastore/content_addressed_store.py
```diff
@@ -189,3 +191,10 @@ def _pack_v1(self, blob):
     def _unpack_v1(self, blob):
         with gzip.GzipFile(fileobj=blob, mode='rb') as f:
             return f.read()
+
+class BlobCache(object):
```
romain-intel, line 195: I would prefer if load and store were symmetrical or named differently. It is confusing here to have a single load and a store (load/store are symmetrical but they are not in their use).

metaflow/datastore/datastore_set.py
```diff
@@ -18,31 +19,22 @@ def __init__(self,
...
-            flow_datastore.blob_cache.set_caching(False)
-            flow_datastore.artifact_cache.set_caching(False)
+                prefetch.update(ds.keys_for_artifacts(prefetch_data_artifacts))
+            # ignore missing keys
+            prefetch.discard(None)
+            
+            # prefetch artifacts and share them with all datastores
+            # in this DatastoreSet
+            cache = ImmutableBlobCache(flow_datastore.ca_store.load_blobs(prefetch))
+            flow_datastore.ca_store.set_blob_cache(cache)
```
romain-intel, line 37: This is very nice. I like this :).

metaflow/datastore/local_backend.py
```diff
@@ -215,15 +215,10 @@ def load_bytes(self, paths):
...
-                try:
-                    file_result = LazyFile(full_path)
-                except OSError:
-                    pass
-            if file_result:
                 if os.path.exists("%s_meta" % full_path):
                     with open("%s_meta" % full_path, mode='r') as f:
                         metadata = json.load(f)
-                results[path] = (file_result, metadata)
+                results[path] = (full_path, metadata)
```
romain-intel, line 221: As I mentioned, I would rather return a BufferedIOBase (which is what was being returned). This gives more flexibility to developers of the backends to return something that is not a file. It does not really add much complexity.

metaflow/datastore/content_addressed_store.py
```diff
@@ -124,29 +125,28 @@ def load_blobs(self, keys, force_raw=False):
...
-        to_load_paths = [
-            self._backend.path_join(self._prefix, k[:2], k) for k in to_load]
-        with self._backend.load_bytes(to_load_paths) as load_results:
-            for k, path in zip(to_load, to_load_paths):
+        load_paths = []
+        for key in keys:
+            blob = None
+            if self._blob_cache:
+                blob = self._blob_cache.load(key)
+            if blob:
```
romain-intel, line 134: I would keep the `is not None` to allow for empty lists, etc.

metaflow/datastore/content_addressed_store.py
tuulos: is it a better approach? `@properties` are nice for read-only getters but I find it better to use explicit functions for mutations

metaflow/datastore/content_addressed_store.py
tuulos: `_blob_cache` needs to be of type `content_addressed_store.BlobCache`, so it can be false only if not specified

metaflow/datastore/content_addressed_store.py
tuulos: sure, we can change the names

metaflow/datastore/local_backend.py
tuulos: we can have a separate discussion but I don't think it is worth introducing a dummy object which is not needed for anything currently or in the foreseeable future. We have no idea what new datastores might need in the future (they might require other changes too), so we can easily make the necessary changes later once we know what's actually needed.

metaflow/client/filecache.py
```diff
@@ -231,45 +215,39 @@ def _get_backend_datastore_cls(ds_type):
 
     def _get_flow_datastore(self, ds_type, ds_root, flow_name):
         cache_id = self._flow_ds_type(ds_type, ds_root, flow_name)
-
         cached_flow_datastore = self._store_caches.get(cache_id)
+
         if cached_flow_datastore:
             return cached_flow_datastore
+        else:
```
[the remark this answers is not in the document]
tuulos: right, it's a stylistic thing to avoid early returns. It is nice when the return value of a function is the value of the last expression evaluated (`if-else` in this case) like in many functional languages.

Returns in the middle of the function body can make it harder to follow the control flow. Obviously not a huge deal.

metaflow/client/filecache.py
```diff
@@ -110,18 +106,6 @@ def get_artifact(
             ds_type, ds_root, data_metadata, flow_name, run_id, step_name,
             task_id, [name])[name]
 
-    def get_all_artifacts(
```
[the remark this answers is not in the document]
tuulos: ok, sorry. I grepped the OSS codebase but couldn't find any uses of this.

metaflow/datastore/content_addressed_store.py
romain-intel: OK, no huge preference.

metaflow/datastore/content_addressed_store.py
romain-intel: Sorry, I meant the `if blob` part.

metaflow/client/filecache.py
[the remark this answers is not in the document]
romain-intel: Sure, I don't have a preference. There are both behaviors in the code. Something for black to check maybe :). One of the linters warns you about it (or is it in Go, I dont' remember) -- anywyas, like you say, not a huge deal either way.

metaflow/client/filecache.py
romain-intel: Reverting

metaflow/client/filecache.py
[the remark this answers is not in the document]
romain-intel: Reverting

metaflow/datastore/content_addressed_store.py
romain-intel: Reverted

metaflow/datastore/content_addressed_store.py
romain-intel: Changed names to `load_key` and `store_keys`

romain-intel (approved):
Added changes to address the comments. Let me know what you think.

tuulos:
LGTM

PR #598, opened 2021-07-02
Pass role to s3

romain-intel:


metaflow/datatools/s3op.py
```diff
@@ -111,8 +111,8 @@ def op_info(url):
 
     with open(result_file_name, 'w') as result_file:
         try:
-            from metaflow.datastore.util.s3util import get_s3_client
-            s3, client_error = get_s3_client()
+            from metaflow.datatools.s3 import S3Client
```
tuulos, line 114: when we benchmarked `s3op` a while ago, importing Metaflow was the biggest source of latency. The old `datastore.util.s3util` might have had a smaller `import` graph than this new version using `datatools.s3`. We should keep this in mind when testing perf

metaflow/datatools/s3op.py
```diff
@@ -159,7 +159,7 @@ def op_info(url):
                                 raise RuntimeError("Could not load file")
                         tmp.close()
                         os.rename(tmp.name, url.local)
-                    except client_error as err:
+                    except s3client.error as err:
```
tuulos, line 162: I left a comment about this in another PR too: `.error` is a side-effectful property which is quite confusing in this context

metaflow/datatools/s3op.py
romain-intel: That's a good point. I doubt that s3 is bigger than everything in datastore but what we can do is move S3Client to its own file to have the smallest add (it would just add whatever is in that file -- which will add the get_s3_client and what not but it will be as small as possible).

romain-intel:
Made obsolete by #987

PR #600, opened 2021-07-02
Delete duplicate conda plugin code

smith-kyle:
An error is thrown saying that the symlink already exists, that's because these 10 lines are duplicated above

### DIFF
--- a/metaflow/plugins/conda/conda_step_decorator.py
+++ b/metaflow/plugins/conda/conda_step_decorator.py
@@ -236,23 +236,6 @@ class CondaStepDecorator(StepDecorator):
                 # for the conda interpreter
                 self.addl_paths = [os.path.split(p)[0] for p in custom_paths]
 
-        # Do the same for metaflow_custom
-        try:
-            import metaflow_custom as m
-        except ImportError:
-            # No additional check needed because if we are here, we already checked
-            # for other issues when loading at the toplevel
-            pass
-        else:
-            custom_paths = list(m.__path__)
-            if len(custom_paths) == 1:
-                # Regular package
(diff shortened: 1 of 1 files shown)


romain-intel:
Ooh, good catch. Some merge crap we didn't see. Might I ask if/how you are using the `metaflow_custom` functionality. This is not something that we have advertised yet and is not fully supported (still in flux). I'd be curious about your use case though as it can inform how we move it forward.

romain-intel (approved):
LGTM

smith-kyle:
I'm using it for a custom @pip decorator. Planning to release an internal version of metaflow that includes this plugin. I'd like to avoid forking metaflow but enable the org to install just one package that includes metaflow and our custom plugins. I'd also like to add a CLI plugin to install the latest `config.json` but I haven't looked into where the CLI plugin mechanism is at yet.

Any idea when/if this will be officially released some time soon?

savingoyal:
@smith-kyle - @romain-intel can follow up with instructions on how to use metaflow-custom.

romain-intel:
I'll clean up what I have and post it here. In the meantime, in terms of CLI, you can definitely add to them by providing a `get_plugin_cli` function (see https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/__init__.py#L22). But yes, your use case should be easily supported by providing a `pip` decorator modeled on the `conda` one (or similar) and adding it to the `plugins` directory in `metaflow_custom`. I'll follow-up shortly.

romain-intel:
I haven't forgotten; we are still cleaning up the document a little bit to share something a bit clean.

PR #607, opened 2021-07-10
IncludeFile now returns the included file in the client

romain-intel:
`IncludeFile`s were properly returned as their content when using them as artifacts but not when accessing them via the client or the CLI. This patch fixes this behavior *for the client only*. The CLI dump command will still return an object that captures the path at which the `IncludeFile` is stored.

### DIFF
--- a/metaflow/includefile.py
+++ b/metaflow/includefile.py
@@ -165,6 +165,27 @@ from .datatools import S3
 DATACLIENTS = {'local': Local,
                's3': S3}
 
+class IncludedFile(object):
+    # Thin wrapper to indicate to the MF client that this object is special
+    # and should be handled as an IncludedFile when returning it (ie: fetching
+    # the actual content)
+
+    def __init__(self, descriptor):
+        self._descriptor = json.dumps(descriptor)
+
+    @property
+    def descriptor(self):
+        return self._descriptor
+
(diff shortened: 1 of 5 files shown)


metaflow/includefile.py
```diff
@@ -173,7 +194,16 @@ def __init__(self, is_text, encoding, path):
...
     @classmethod
     def is_file_handled(cls, path):
+        # This returns a tuple:
+        #  - True/False indicating whether the file is handled
+        #  - None if we need to create a handler for the file, a LocalFile if
+        #    we already know what to do with the file or a Uploader if the file
+        #    is already present remotely (either starting with s3:// or local://)
+        #  - An error message if file is not handled
         if path:
+            if isinstance(path, IncludedFile):
```
savingoyal, line 204: isn't this introducing a circular dependency?

savingoyal (changes requested):
I will test the PR shortly.

metaflow/includefile.py
romain-intel: I am confused as to why this introduces a circular dependency. The class is defined on line 168 and returned by encode_url.

savingoyal: And the `LocalFile` class is accessed in L:181 within `IncludedFile`

PR #608, opened 2021-07-13
Parse environment variables passed from CLI

LarsDu:
Right now if one attempted to pass environment variables to the `@environment` decorator using the following type of syntax:
```
python workflow/myflow.py --with environment:vars=FOO:$(FOO),BAR:$(BAR)
```
vars would incorrectly register as a string rather than parsing the the contents as a dict. 

This PR enables parsing of environment for the environment decorator which are passed using --with

### DIFF
--- a/metaflow/plugins/environment_decorator.py
+++ b/metaflow/plugins/environment_decorator.py
@@ -27,6 +27,9 @@ class EnvironmentDecorator(StepDecorator):
     name = 'environment'
     defaults = {'vars': {}}
 
+    def step_init(self, flow, graph, step_name, decorators, environment, datastore, logger):
+        self.attributes["vars"] = self.vars_dict().items()
+
     def vars_dict(self):
         vars = self.attributes["vars"]
         if isinstance(vars, str):  # env specified using --with will be a string that we must parse
--- a/metaflow/plugins/environment_decorator.py
+++ b/metaflow/plugins/environment_decorator.py
@@ -39,10 +39,8 @@ class EnvironmentDecorator(StepDecorator):
 
             except Exception as e:
                 raise ValueError(
-                    f"Encountered a problem parsing stringified @environment vars attribute as a dict: {e}"
+                    "Encountered a problem parsing stringified @environment vars attribute as a dict: {}".format(e)
                 )
-
-            logger.debug(f"Parsed stringified @environment vars attribute to dict with keys: {vars_dict.keys()}")
             self.attributes["vars"] = vars_dict
 
 
--- a/metaflow/plugins/environment_decorator.py
+++ b/metaflow/plugins/environment_decorator.py
@@ -27,7 +27,7 @@ class EnvironmentDecorator(StepDecorator):
     name = 'environment'
     defaults = {'vars': {}}
 
-    def step_init(self, flow, graph, step_name, decorators, environment, datastore, logger):
+    def vars_dict(self):
         vars = self.attributes["vars"]
         if isinstance(vars, str):  # env specified using --with will be a string that we must parse
             try:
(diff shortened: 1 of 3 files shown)


oavdeev:
Nice catch! The only thing that gives me a bit of a pause here is that it is not uncommon to have `:` in env variable values (eg `$PATH`). On the other hand I don't have a good idea how to support those without making decospec syntax even more complicated.

LarsDu:
> Nice catch! The only thing that gives me a bit of a pause here is that it is not uncommon to have `:` in env variable values (eg `$PATH`). On the other hand I don't have a good idea how to support those without making decospec syntax even more complicated.

Perhaps I can update this to only parse the first `:` in each comma delimited block

oavdeev:
After testing this a bit more, this doesn't seem to work with `--batch`. In that case gets converted back to `--with` CLI arg [here](https://github.com/Netflix/metaflow/blob/master/metaflow/decorators.py#L112), and it turns it into `--with 'environment:vars={'\"MY_FOO\"': '\"bar\"', '\"MY_BAR\"': '\"baz\"'}'`. Let me take a deeper look, there gotta be a way around this

oavdeev:
Something like this https://github.com/Netflix/metaflow/commit/d654bc8727b18f88c27e90bb95288a0f52ff8c3c should work but may still need a bit more testing

corleyma:
> Something like this [d654bc8](https://github.com/Netflix/metaflow/commit/d654bc8727b18f88c27e90bb95288a0f52ff8c3c) should work but may still need a bit more testing

This looks good, though probably still need to limit the split on `:` to 1.  This implementation should make it so that the only disallowed character in env var values would be `,`.

LarsDu:
> > Something like this [d654bc8](https://github.com/Netflix/metaflow/commit/d654bc8727b18f88c27e90bb95288a0f52ff8c3c) should work but may still need a bit more testing
> 
> This looks good, though probably still need to limit the split on `:` to 1. This implementation should make it so that the only disallowed character in env var values would be `,`.

I've updated this to split on the first colon

LarsDu:
> Something like this [d654bc8](https://github.com/Netflix/metaflow/commit/d654bc8727b18f88c27e90bb95288a0f52ff8c3c) should work but may still need a bit more testing

I've updated the PR to incorporate these changes, but I have not thoroughly tested this.

corleyma:
@oavdeev @LarsDu still needs some testing but i've made the changes I think would make this work for both step-functions and batch plugins.

metaflow/plugins/environment_decorator.py
```diff
@@ -27,5 +27,23 @@ def myStep(self):
     name = 'environment'
     defaults = {'vars': {}}
 
+    def vars_dict(self):
+        env_vars = self.attributes["vars"]
+        if isinstance(vars, str):  # env specified using --with will be a string that we must parse
```
oavdeev, line 32:
```suggestion
        if isinstance(env_vars, str):  # env specified using --with will be a string that we must parse
```

we should really get mypy set up on this :)

oavdeev (approved):
otherwise LGTM!

LarsDu:
LGTM!

oavdeev:
@savingoyal had a suggestion that we could solve it in a slightly more elegant way by making `@environment` a "varargs" style decorator. So instead of `@environment(vars={"FOO": "BAR"})` you could support `@environment(FOO="BAR")`. Metaflow already supports passing multiple parameters to a decorator from the command line, so it allows to implement this without making `--with` syntax special for `@environment`.

I've implemented it on top of this PR here https://github.com/Netflix/metaflow/commit/31519fe -- please take a look

savingoyal:
Also, [this](https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/conda/conda_step_decorator.py#L92) is how we currently resolve a similar issue in `@conda`.

savingoyal:
Any updates on this PR?

oavdeev:
See also https://github.com/Netflix/metaflow/pull/768

LarsDu:
A bit busy at the moment. Will try to revisit this by the end of the week.

PR #610, opened 2021-07-15
Move S3Client to s3util.py from s3.py

romain-intel:
This is to make the import of S3Client from s3op (upcoming change) as cheap
as possible in terms of import time


romain-intel:
Made irrelevant by other work on the S3 client.

PR #611, opened 2021-07-15
Add an option to define your own AWS client provider

romain-intel:
You can now specify a function that returns an AWS client. This is
useful if you want to use something other than boto3 and can be used
with the metaflow_custom mechanism to provide your own authentication


metaflow/plugins/aws/aws_client.py
```diff
@@ -1,37 +1,57 @@
...
-    return boto3.client(module, **params)
+                **cached_aws_sandbox_creds).client(module, **params)
+        if with_error:
+            return boto3.client(module, **params), ClientError
+        return boto3.client(module, **params)
+
+
+def get_aws_client(module, with_error=False, params={}):
+    global cached_provider_class
+    from metaflow.metaflow_config import DEFAULT_AWS_CLIENT_PROVIDER
```
tuulos, line 47: since the point of `cached_provider_class` is perf optimization, you could move these imports too inside the `cached_provider_class is None` clause

metaflow/plugins/aws/aws_client.py
```diff
@@ -1,37 +1,57 @@
...
+    from metaflow.plugins import AWS_CLIENT_PROVIDERS
+    if cached_provider_class is None:
+        cached_provider_class = [p for p in AWS_CLIENT_PROVIDERS
```
tuulos, line 50: here's a bit simpler logic for the clause:
```
for p in AWS_CLIENT_PROVIDERS:
    if p.name == DEFAULT_AWS_CLIENT_PROVIDER:
        cached_provider_class = p
        break
else:
    raise ValueError("Cannot find AWS Client provider %s"
                         % DEFAULT_AWS_CLIENT_PROVIDER)
```

romain-intel: OK, I'll make the change. I initially had something like this and then though "nah, Ville is going to tell me to use a more "Python-y" way". No kidding :).

PR #613, opened 2021-07-16
Add S3 tests

romain-intel:


test/data/s3/s3_data.py
```diff
@@ -0,0 +1,451 @@
...
+from metaflow.datatools.s3 import S3PutObject
+
+from metaflow.util import to_fileobj, to_bytes, url_quote
+
+import numpy
+
+from .. import s3client, S3ROOT
+
+BASIC_METADATA = {
+    #'no_meta': (None, None), # No metadata at all but going through the calls
```
savingoyal, line 25: Is this supposed to remain commented out?

savingoyal (approved):
LGTM! @oavdeev can we hook these tests up with an AWS account in this repo?

test/data/s3/s3_data.py
romain-intel: That's a good point. Not sure. I can try uncommenting and seeing if it works well.

test/README.md
```diff
@@ -1,5 +1,47 @@
...
+The data tests are standard `pytest` suites. In the `s3` folder, you will
+find two files: `s3_data.py` which generates synthetic
+data and `test_s3.py` which contains the actual tests.
+
+The test data is cached in S3. If you change anything in the `s3_data.py`
+module (or you have another reason for wanting to regenerate test
+data), you can regenerate the data easily by changing the S3 test prefix
+at `test/data/__init__.py`. The `s3_data.py` detects that data
+is missing in S3 and they will upload the data in the new location
+automatically.
```
oavdeev, line 26: One thing worth clarifying here: do I need to run `s3_data.py` manually once before I can run tests via pytest?

romain-intel: You should just be able to run pytest. Feel free to edit.

test/data/s3/s3_data.py
```diff
@@ -0,0 +1,451 @@
...
+        cases.append((put_prefix, blobs, expected))
+    return {'argvalues': cases, 'ids': ids}
+
+def ensure_test_data():
+    # update S3ROOT in __init__.py to get a fresh set of data
+    print('Ensuring that test data exists at %s' % S3ROOT)
+    mark = urlparse(os.path.join(S3ROOT, 'ALL_OK'))
+    try:
+        # Check if parquet data exists and has been modified in the last
+        # 29 days (default TTL - 1).
```
oavdeev, line 400: What is "default TTL" here?

test/README.md
```diff
@@ -1,5 +1,47 @@
...
+    exercising all aspects of Metaflow.
+
+You can run the tests by hand using `pytest` or `run_tests.py` as described
+below.
+
+## Data Test Suite
+
+The data tests are standard `pytest` suites. In the `s3` folder, you will
+find two files: `s3_data.py` which generates synthetic
+data and `test_s3.py` which contains the actual tests.
```
oavdeev, line 19:
```suggestion
data when tests run and `test_s3.py` which contains the actual tests.
```

test/README.md
oavdeev: cool added a wording suggestion (i can't edit your PR :)

test/data/s3/s3_data.py
```diff
@@ -0,0 +1,451 @@
...
+            _do_upload(prefix, filespecs)
+        for prefix, filespecs in BASIC_DATA:
+            _do_upload(prefix, filespecs, meta=BASIC_METADATA)
+
+        s3client.upload_fileobj(to_fileobj('ok'),
+                                Bucket=mark.netloc,
+                                Key=mark.path.lstrip('/'))
+        print('Test data uploaded ok')
+
+ensure_test_data()
```
oavdeev, line 451: Is there an easy way to call this inside a fixture or something like that? I'd generally avoid doing any non trivial work on module import. However, as this is just test code, I don't feel very strongly

test/data/s3/s3_data.py
oavdeev: Also, roughly how long does it take to generate test data? If it is fast enough I'd err on the side of having shorter TTLs

romain-intel: It seems to take under 5 min from an EC2 instance. YMMV (a lot of files and some large ones are uploaded). I don't know if there is a lot of value in re-uploading a lot though so I think 29 days is enough. I'll remove the reference to default TTL; this is probably something Netflix specific.

test/data/s3/s3_data.py
romain-intel: I'm not sure. As you say, it's a test so I didn't look to change it much (I just copied what we have). If it bugs you, definitely propose something else :).

test/data/s3/s3_data.py
romain-intel: I uncommented. Tests pass (you do have to re-upload your data if you already have it uploaded). THere are a few more tests to run but don't expect any issues.

test/data/s3/s3_data.py
oavdeev: I'd maybe go lower, like a day or two maybe? I mean, it is probably fine as is, but generally when I see a code path that gets executed only _once every X weeks_ it makes me a little uneasy, as I worry that if there is an obscure bug by some chance, it'll take a while to spot

romain-intel: I left as is for now. I did clarify the TTL requirement though. IMHO, uploading data can be expensive so I didn't think it very useful to re-upload too much. But don't have a strong opinion.

test/README.md
romain-intel: I took into account the suggestion that was below but not sure if you had another one.

oavdeev (approved):

PR #614, opened 2021-07-16
Allow Metaflow Custom packages to define additional suffixes to package

romain-intel:
This enables you to package non-python files in your metaflow_custom package
and have them included when Metaflow creates the code package.

### DIFF
--- a/metaflow/package.py
+++ b/metaflow/package.py
@@ -39,9 +39,8 @@ class MetaflowPackage(object):
         self.blob, self.sha = self._make()
 
     def _walk(self, root, exclude_hidden=True, addl_suffixes=None):
-        suffixes_to_check = self.suffixes
-        if addl_suffixes is not None:
-            suffixes_to_check = list(set().union(suffixes_to_check, addl_suffixes))
+        if addl_suffixes is None:
+            addl_suffixes = []
         root = to_unicode(root)  # handle files/folder with non ascii chars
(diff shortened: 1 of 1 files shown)


metaflow/package.py
```diff
@@ -46,7 +53,7 @@ def _walk(self, root, exclude_hidden=True):
             for fname in files:
                 if fname[0] == '.':
                     continue
-                if any(fname.endswith(suffix) for suffix in self.suffixes):
+                if any(fname.endswith(suffix) for suffix in suffixes_to_check):
```
savingoyal, line 56: minor nit - in lieu of L:42-44 we can just have `... in self.suffixes + addl_suffixes` here

savingoyal (commented):
LGTM! A minor comment.

metaflow/package.py
```diff
@@ -34,7 +38,9 @@ def __init__(self, flow, environment, echo, suffixes=DEFAULT_SUFFIXES_LIST):
                                   environment)
         self.blob, self.sha = self._make()
 
-    def _walk(self, root, exclude_hidden=True):
+    def _walk(self, root, exclude_hidden=True, addl_suffixes=None):
+        if addl_suffixes is None:
```
savingoyal, line 42: We can default to an empty list in L:41 and in L:31 to do away with this check.

savingoyal (changes requested):

savingoyal (approved):

PR #615, opened 2021-07-16
Add inputs to task_pre_step

romain-intel:


metaflow/plugins/aws/step_functions/step_functions_decorator.py
```diff
@@ -20,7 +21,11 @@ def task_pre_step(self,
                       graph,
                       retry_count,
                       max_user_code_retries,
-                      ubf_context):
+                      ubf_context,
+                      input_obj):
+        if ubf_context == UBF_TASK:
+            # Don't register any metadata for tasks that don't run within SFN.
```
oavdeev, line 27: Can you elaborate a bit more?  I'm not sure I (or future generations reading this code) would understand the link between UBF and SFN and metadata here

metaflow/task.py
```diff
@@ -439,7 +410,43 @@ def run_step(self,
...
+                # times.
+                step_func = deco.task_decorate(step_func,
+                                               self.flow,
+                                               self.flow._graph,
+                                               retry_count,
+                                               max_user_code_retries,
+                                               self.ubf_context)
+
+            if join_type:
+                self._exec_step_function(step_func, input_obj)
```
savingoyal, line 448: Any reason to move the code around? `input_obj` only exists `if join_type` so it makes sense to invoke `_exec_step_function` where `input_obj` is defined.

metaflow/plugins/aws/step_functions/step_functions_decorator.py
savingoyal: Adding to @oavdeev 's comment - I don't think we need this check here.

metaflow/decorators.py
```diff
@@ -273,7 +273,8 @@ def task_pre_step(self,
                       graph,
                       retry_count,
                       max_user_code_retries,
-                      ubf_context):
+                      ubf_context,
+                      input_obj):
```
savingoyal, line 277: Rather than `input_obj`, can we use `inputs` - that seems semantically correct looking at L433 of task.py

savingoyal (changes requested):

metaflow/plugins/aws/step_functions/step_functions_decorator.py
romain-intel: This is in the Meson implementation. Happy to strip it. I don't know why it got skipped when we ported to SFN. I figured that since the only UBF is in Meson, it's probably the correct one but I don't care either way.

metaflow/decorators.py
romain-intel: I can change that. No issues.

metaflow/task.py
romain-intel: I will check. This is how we do it internally so kept that to avoid breaking things.

romain-intel: I checked. The difference is that with this new approach, the datastore is properly set and you can access the attributes in the flow properly which is needed in one of the internal plugins. In fairness, we also will need to move this around for the new datastore since we pass the datastore down to task_pre_step.

savingoyal: pre_step and decorate don't rely on datastore at the moment. I am not sure how this change is functionally different from the tip of the master.

savingoyal (approved):

savingoyal (approved):

PR #427, opened 2021-01-30
[R] serialize all the things with qs

bcgalvin:
@savingoyal @jasonge27 

Currently passing complex R objects to and fro leads to weird results; below is a simple flow demonstrating this with bootstrap sampling from the rsample package:

```
library(metaflow)
start <- function(self){
    suppressPackageStartupMessages(library(rsample))
    self$data <- bootstraps(mtcars)
}
end <- function(self){
    suppressPackageStartupMessages(library(rsample))
    print(self$data)
}

metaflow("HelloFlow") %>%
    step(step = "start", 
         r_function = start, 
         next_step = "end") %>%
    step(step = "end", 
         r_function = end) %>% 
    run()
```

With some mangled results coming back out:

```Metaflow 2.2.0 executing HelloFlow for user:bryangalvin
Validating your flow...
    The graph looks good!
2021-01-30 00:14:56.162 Workflow starting (run-id 1611994496156093):
2021-01-30 00:14:56.168 [1611994496156093/start/1 (pid 7539)] Task is starting.
2021-01-30 00:14:59.277 [1611994496156093/start/1 (pid 7539)] Task finished successfully.
2021-01-30 00:14:59.286 [1611994496156093/end/2 (pid 7561)] Task is starting.
2021-01-30 00:15:02.258 [1611994496156093/end/2 (pid 7561)]                                                                                                                                                                                          splits
2021-01-30 00:15:02.258 [1611994496156093/end/2 (pid 7561)] 1    <environment: 0x7fb423d26d18>, 9, 28, 24, 15, 4, 24, 32, 5, 30, 29, 15, 14, 17, 4, 19, 11, 17, 18, 18, 7, 30, 5, 27, 11, 19, 25, 22, 6, 24, 28, 22, 6, TRUE, <environment: 0x7fb423d32838>
2021-01-30 00:15:02.258 [1611994496156093/end/2 (pid 7561)] 2   <environment: 0x7fb424067920>, 27, 1, 10, 10, 10, 7, 13, 3, 7, 28, 15, 31, 27, 15, 16, 16, 12, 13, 29, 13, 12, 12, 15, 13, 11, 5, 10, 21, 1, 27, 2, 26, TRUE, <environment: 0x7fb424075aa8>
[...]

PR #619, opened 2021-07-18
More aggressive serialization of R objects

### LINKED ISSUE #464: Empty collection will always become a list after storing on self in the R environment
I think it must be saving the typed collection (in this case an empty numeric vector) as [] (typeless) and then not reconstituting the type when you reload it from self.
```R
library(metaflow)

start <- function(self) {
  numlist <- numeric(0)
  self$set <- numlist
  print(class(numlist))
  print(class(self$set))
}
end <- function(self){
}

metaflow("testset") %>%
  step(
    step = "start",
    r_function = start,
    next_step = "end") %>%
  step(step = "end",
       r_function = end) %>%
  run()
```

This tripped us up because we had an optional list that was then passed into c++ modelling code by R - a preparation step created the parameter as numeric(0) but when the parameter was reloaded from self and handed to the compute job it failed. This is brittle code on our part (and has been re-enforced to work around) but I think the type changing is potentially invalid behaviour by metaflow.

### LINKED ISSUE #592: Non-invertible R objects
This expands on issue #464. Let's call an R object `x` invertible if, for a given pair of `serialise` and `deserialise` functions, the following function returns `TRUE`:

```r
is_invertible <- function(x, serialise, deserialise) {
  identical(
    x,
    x |> serialise() |> reticulate::r_to_py() |> reticulate::py_to_r() |> deserialise()
  }
}
```

Invertible objects are those that are returned _uncorrupted_ when stored to the `self` state object in a Metaflow step. Issue #464 notes that `integer(0)` becomes `list()` when it goes through the serialise and `reticulate` pipeline, so it is _not_ invertible. Let's consider a broad list of candidates:

```r
how_invertible <- function(serialise, deserialise) {
  inverted <- function(x) {
    x |> serialise() |> reticulate::r_to_py() |>
      reticulate::py_to_r() |> deserialise()
  }
  is_invertible <- function(x) {
    # ignore environments when comparing closures
    identical(x, inverted(x), ignore.environment = TRUE)
  }
  candidates <- list(5, 5.5, 5L, "a", "abc", TRUE, FALSE, NULL, NA, 
                     Sys.Date(), Sys.time(), mtcars, tibble::as_tibble(mtcars),
                     data.table::as.data.table(mtcars), c(1, 2, 3),
                     c(1.5, 2.5, 3.5), c("a", "b", "c"), c(TRUE, FALSE, TRUE),
                     integer(), numeric(), character(), logical(),
                     list(), list("red panda"), list(animal = "red panda"),
                     list(animal = "red panda", number = 5), as.raw(c(1:10)),
                     function(x) x + 1, matrix(c(1,2,3,4),nrow = 2, ncol = 2)
                     )
  
  invertible_candidates <- purrr::map_lgl(candidates, is_invertible)
  return(sum(invertible_candidates) / length(candidates))
}

how_invertible(identity, identity)
# 0.6206897

how_invertible(mf_serialize, mf_deserialize)
# 0.6551724

how_invertible(\(x) serialize(x, NULL), unserialize) # base R serialisation
# 0.9655172

how_invertible(qs::qserialize, qs::qdeserialize)
# 0.9655172
```

The tricky object above is the `data.table`, since the pointer changes during serialisation and deserialisation (without `reticulate`), but the underlying data is the same. You could argue that this isn't a big deal.

For the currently used `mf_serialize` it seems pretty bad:
* `NA` is inverted to `TRUE`
* Empty vectors become empty lists
* Tibbles become data frames
* `list("red panda")` becomes `"red panda"`

We should look into serialising all data into `raw`, which seems to survive `reticulate` unscathed. The base serialisation is an option, as is the supposedly faster serialisation offered by the `qs` package, as proposed by #427.

Once concern here is backwards-compatibility: we want to be able to load objects stored in S3 before the change in serialisation. We can do this by not deserialising anything that isn't a raw vector. If we choose to use the `qs` package, we can use the following function to detect if something was serialised with `qs`, and otherwise revert to the previous method:

```r
is_qs_serialised <- function(raw_data) {
  qs_magic_bits <- as.raw(c(0x0b, 0x0e, 0x0a, 0x0c))
  identical(raw_data[1:4], qs_magic_bits)
)
```

mdneuzerling:
Closes #464
Closes #592
Recommend we also close #427

This PR is the counterpart to the exploration I performed in #592. The core issue is that some R objects are corrupted when they are saved the to `self` object, as they undergo serialisation, `reticulate::r_to_py`, and then back again.

### Key changes:
* Serialization functions have been moved into a separate file `serialization.R`, and their tests have been moved into a separate test file. Recall that there's no concept of a module/namespace with R files, so this has zero impact on the actual package --- I just think it will be easier to maintain.
* `mf_serialize` currently leaves "simple" objects unchanged, which leaves them to the mercy of `reticulate`'s conversion. This meant that R objects without clear Python analogues are corrupted. The most egregious example is `NA` becoming `TRUE`! (See #592 for more examples). The `mf_serialize` proposed here will **convert everything to raw/bytes**, which are untouched by `reticulate`.  I've included as many tests as I can for this.
* I've put a minimum version requirement on the `testthat` dependency. It's a suggested dependency, so this only affects developers.
* Version increase 2.3.0 -> 2.3.1

`mf_deserialize` is **unchanged**. This is important, because we want users to be able to deserialize objects they've stored to S3 using previous versions of Metaflow.

### Concerns:

I've included a flow which is actually `run` (locally) in `test-serialization.R`. I couldn't find any use of the `metaflow::run` function  in other tests (although flows are run through other means). I don't think this is a problem, and it's probably a good idea to include end-to-end tests like this, but I wanted to highlight it as it slows things down a bit.

Currently, simple R objects and data frames are converted to their Python/pandas equivalents. This means that a multi-language flow would be able to use the data from either language. Under the proposed changes, all R data will appear as raw/bytes to Python. Now, I'm not even sure if a multi-language flow _exists_, or is even supported. I just want to highlight this change because it may have consequences I'm not aware of.

For example, perhaps there is a particular workflow used for monitoring that picks up on flow objects that it expects to be data frames, regardless of which language is used to generate those data frames. This PR would break that workflow.


mdneuzerling:
Ahh darn... I'll look into this:

```python
Error in py_call_impl(callable, dots$args, dots$keywords) : 
  MetaflowNotFound: Flow('BasicArtifactsTestFlow') does not exist
```

savingoyal:
Thanks for the PR! We can merge this in as soon as the tests pass again.

mdneuzerling:
I can't yet find the cause of the missing flow error, nor can I reproduce it (either locally or with `act`). I'll keep looking.

There is a second error which doesn't appear in the GA logs. The argument provided to `foreach` in an R step is passed directly through to Python. Currently, R vectors/lists are converted to Python lists, so everything is fine. With this PR, however, these are passed through as raw vectors! Python sees these as byte arrays and tries to split on them. Given that a simple `c("a", "b", "c")` vector serialises to a raw vector of 58 bytes, this usually appears as an error that the maximum number of children has been exceeded.

The argument to "foreach" _must_ be a string pointing to an instance (`self`) variable, in either language. I can think of two solutions to this but, honestly, I think both are hacky:

* if an argument is given to "foreach", save it to a non-serialized dummy variable and use that instead, or
* if an argument is given to "foreach", save it to the same variable but don't serialise it.

mdneuzerling:
I've fixed that first error. By introducing a metaflow `run` in the `testthat` tests, it set the local metadata store to the `testthat` directory. When running the integration tests, the metadata stores didn't line up, so the `flow_client` couldn't find the test flows.

I've removed my `run` test, since it was superfluous anyway. Perhaps at a later date we'll be able to bring all tests under the umbrella of `testthat`?

Of course this leaves the `foreach` error that I described above...

mdneuzerling:
I've resolved the `foreach` error by introducing special Python wrapper classes for wrapping serialised R data. The `MetaflowRObject` is initialised from R with a given length, which Python can use when splitting in a `foreach` step instead of trying to calculate the length itself. Indexing a `MetaflowRObject` (eg. `x[0]`) yields a `MetaflowRObjectIndex` object, which just delays the actual calculation until the object can be deserialised in R.

Consequently, I thought it would be easier to convert `mf_deserialize` into an S3 function that can dispatch on the above classes, as well as maintaining backwards compatibility.

Now, there's still some buggy behaviour that's preventing the tests from passing. _Something_ is going on with `merge_artifacts`. I'll need to continue digging around to find out what.

mdneuzerling:
I got it working, but I'm not sure this should be merged!

The goal of the new serialisation method is to get Python to "carry" R data without interfering with it. [I go through what this means here](https://mdneuzerling.com/post/i-tried-to-improve-how-metaflow-converts-r-to-python-and-i-failed/) but programmatically I'm serialising R data to bytes and then storing those bytes in a `MetaflowRObject` Python class. The length of object _as interpreted by R_ is also stored in this class.

I had to make some changes to the integration tests:
* In the `nested_foreach.json` graph I changed strings like "abc" into vectors like c("a", "b", "c"). Previously Metaflow would split on "abc" and create three subsequent tasks, because Python sees "abc" as an object of length 3. But R sees it as a vector of length 1! My new serialisation forces Metaflow to use the R interpretation of length.
* In the `nested_foreach.R` test file I change `stopifnot(length(self$x) == length(x[[2]]))` to `stopifnot(length(self$x) == x[[2]])`. This is probably a good idea regardless of this patch. Previously, `self$x` was a string and so it was interpreted as length 1, and `x[[2]]` is the depth of the foreach stack at that particular point, which is an integer and so also of length 1. That is, this test trivially passes under the current implementation.
* Also in `nested_foreach.R` I deserialize the input before comparing it to the expected value. Python can't index R objects, so deserialisation has to be done R-side.

I have two concerns here:

* My serialisation approach probably isn't too great for memory usage. If I have a 10GB data frame, it's serialised to raw bytes and stored in a (roughly) 10GB Python object.
* My serialisation approach does not allow Python to compare inputs in a `foreach` stack. When a `MetaflowRObject` object is indexed, it returns an object of class `MetaflowRObjectIndex`, which contains the original object and the intended index. This is to delay computation until R can deserialise and index. The only two things that Python knows about an object is serialised bytes and its length. I'm not sure what impact this would have.

There's a lot of effort here to stop Python from messing with R objects, but I'm not convinced that it's worth the effort?

mdneuzerling:
Hi @saikonen. I believe we can close this PR without merging. I don't think this is the correct approach to fixing serialisation compatibility between R and Python --- I'm not sure there is a correct approach.

PR #631, opened 2021-07-29
Add resource tags to AWS Batch jobs

savingoyal:
This PR assumes that `Batch:TagResource` is whitelisted for the role which submits the AWS Batch job.

Gated on env var - `METAFLOW_BATCH_EMIT_TAGS`

### DIFF
--- a/metaflow/plugins/aws/step_functions/step_functions.py
+++ b/metaflow/plugins/aws/step_functions/step_functions.py
@@ -813,7 +813,8 @@ class State(object):
             .parameter('RetryStrategy', 
                 to_pascalcase(job.payload['retryStrategy'])) \
             .parameter('Timeout', 
-                to_pascalcase(job.payload['timeout']))
+                to_pascalcase(job.payload['timeout'])) \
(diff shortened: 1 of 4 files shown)


metaflow/plugins/aws/step_functions/step_functions.py
```diff
@@ -813,7 +813,8 @@ def batch(self, job):
             .parameter('RetryStrategy', 
                 to_pascalcase(job.payload['retryStrategy'])) \
             .parameter('Timeout', 
-                to_pascalcase(job.payload['timeout']))
+                to_pascalcase(job.payload['timeout'])) \
+            .parameter('Tags', job.payload['tags'])
```
corleyma: might need to add `propagateTags` somewhere here as well.

savingoyal: The `job-definition` will take care of setting `propagateTags` to `True` always, so setting it here is un-neccessary.

oavdeev (approved):

PR #637, opened 2021-08-11
Revert "IncludeFile now returns the included file in the client"

savingoyal:
Reverts Netflix/metaflow#607

Netflix/metaflow#607 breaks `step-functions create` blocking any new workflow creation.


savingoyal:
We would need to find a different mechanism to address the original bug that #607 was trying to fix.

Issue #441, opened 2021-02-20
Adding volume info to batch decorator


mattmcclean:
Would be useful to add volume information into the batch decorator so that EBS, EFS or FSx volumes can be attached to the jobs to access data quickly for training, cache pre-trained weights, cache data from S3 etc. 

Simplest version would be to pass in a mount path parameter into the decorator that will add two entries: `volumes` and `mountPoints` to the AWS Batch Job Definition `containerProperties` object.

https://docs.aws.amazon.com/batch/latest/userguide/job_definition_parameters.html#containerProperties

savingoyal:
Makes sense. I will add it in the next release.

savingoyal:
also, ulimits could be a great addition - https://gitter.im/metaflow_org/community?at=60403e47d2619a4f2e1f6766

samuelthan:
Yeah, looking for to have this implement too. Our use case if more of intergration with AWS CodeArtifact to pull pip libraries, however there needs to be away of sharing the logged in mechanism :)

savingoyal:
An update - a partial fix for this issue was part of Metaflow 2.3.5 - https://github.com/Netflix/metaflow/releases/tag/2.3.5#441

cc @samuelthan @mattmcclean

NerdToMars:
hi, @savingoyal  do you have any plan to fully support this? 
I am new to metaflow. What I did is to modify the metaflow-aws terraform code to create efs with the correct subnet settings and mount efs when booting using the user data (https://aws.amazon.com/premiumsupport/knowledge-center/batch-mount-efs/). 

 Another tots is to modify the metaflow to create full volume json objects in the container definition, but will require user modify their cloud infra. Also clear documentation is required.

PR #640, opened 2021-08-11
allow to mount host volumes in AWS Batch

oavdeev:
# What does this PR do?

Adds an option to `@batch` decorator to mount host volumes inside the container. `@batch(mount_host_volumes="/mnt/")`. This way, `/mnt` from host will be mounted at `/mnt` inside the container. This (partially) addresses #441

In this release this is considered an _advanced_ feature due to less friendly UX and the fact that troubleshooting this requires a decent knowledge of AWS Batch. 

I've tested this on Batch and Step Functions and it seems to work!

## Potential improvements in future PRs, not included here

* First class support for [per-job EFS mounts](https://aws.amazon.com/blogs/hpc/introducing-support-for-per-job-amazon-efs-volumes-in-aws-batch/). Not included this because the configuration is pretty hairy, needs some extra thinking. However, you should be able to use this PR as is to expose EFS volumes mounted at host to containers, as described [here](https://aws.amazon.com/premiumsupport/knowledge-center/batch-mount-efs/).

* Support for customizing mount points inside a container. Right now `/mnt` from host always gets mounted as `/mnt` inside the container.

### DIFF
--- a/metaflow/plugins/aws/batch/batch.py
+++ b/metaflow/plugins/aws/batch/batch.py
@@ -164,7 +164,8 @@ class Batch(object):
         max_swap=None,
         swappiness=None,
         env={},
-        attrs={}
+        attrs={},
+        host_volumes=None,
     ):
         job_name = self._job_name(
             attrs.get('metaflow.user'),
@@ -186,7 +187,7 @@ class Batch(object):
             .execution_role(execution_role) \
             .job_def(image, iam_role,
                 queue, execution_role, shared_memory,
-                max_swap, swappiness) \
+                max_swap, swappiness, host_volumes=host_volumes) \
             .cpu(cpu) \
             .gpu(gpu) \
             .memory(memory) \
@@ -244,6 +245,7 @@ class Batch(object):
         shared_memory=None,
         max_swap=None,
         swappiness=None,
+        host_volumes=None,
         env={},
         attrs={},
         ):
@@ -272,8 +274,9 @@ class Batch(object):
                         shared_memory,
                         max_swap,
                         swappiness,
-                        env,
-                        attrs
+                        env=env,
+                        attrs=attrs,
+                        host_volumes=host_volumes
         )
         self.job = job.execute()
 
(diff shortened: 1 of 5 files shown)


corleyma:
While a good start, I think this implementation falls short for a couple of reasons:
- as mentioned, not being able to set the mount path inside the container seems like a strange limitation, and will likely place a burden on application developers to change their applications to fit the details of the infrastructure.
- including this in `@batch` means it can only be used in flows that are forced to run a particular step on Batch (or else in every step if using `with` syntax)
  - it seems a more natural way for this to work would be some mechanism to specify these Batch-specific resource definitions in `@resources` (either as nested configs, or else as part of an `@batch_resources` decorator).
    - Combined with the ability to set the container mount path, this should make it easy to have local runs (e.g. for integration tests) that work seamlessly with remote runs in production.
    - This desire to define Batch-specific resources arises in other contexts, e.g. the desire to set a per-step job queue without forcing the step to run on Batch.  Figuring out the idiom for this configuration in Metaflow and embracing it for all such configs would really increase the flexibility.

oavdeev:
Good points, tbh this PR was intended to solve a pretty narrow (advanced) use case of mounting EFS/FSx volumes from host into the container.

Re: local runs, wouldn't you have to fork the logic? You'd have to remember that the step may run in local mode and make sure to branch, something like:

```python
@resources(batch_mount_points=["/mnt"])
def step_foo(self):
    if os.path.exists("/mnt"):
         # running on Batch
        model = load_model("/mnt/foo.tar.gz")
    else:
        # running locally
        model = MockLocalModel()
...
```

In this sense, one benefit of keeping it in `@batch` like in this PR that the UX is more foolproof; you're guaranteed that the step will run on Batch and `/mnt` will be there. I do see the benefit of being able to run the code locally without changing the annotation though.

popsonebz:
Thanks @oavdeev. I am hoping this PR gets merged as I need this feature for similar use case. Were you able to test this on AWS metaflow deployment? If yes, Kindly share the steps so I can replicate it. Thanks

savingoyal (changes requested):
PR LGTM! Can we replace `mount_host_volumes` with `host_volumes` before shipping it?

PR #649, opened 2021-08-19
Added decorator hook for adding custom package

valayDave:
Added a hook in the`StepDecorator` to add custom packages.


metaflow/package.py
```diff
@@ -74,6 +76,19 @@ def path_tuples(self):
...
                 yield path_tuple
+             
+        # Any custom packages exposed via decorators
+        deco_module_paths = set()
+        for step in self._flow:
+            for deco in step.decorators:
+                for path_tuple in deco.add_to_package():
+                    file_path,_ = path_tuple
+                    # Check if the path is not duplicated as
+                    # many steps can have the same card
```
savingoyal, line 87: nit: update the comment to remove reference to `card`.

savingoyal (changes requested):
A minor comment, LGTM otherwise!

valayDave:
## TODOs
- [ ] Add information about hook to decorator lifecycle

metaflow/package.py
```diff
@@ -74,6 +76,20 @@ def path_tuples(self):
                     exclude_hidden=False,
                     addl_suffixes=self.metaflow_custom_addl_suffixes):
                 yield path_tuple
+             
+        # Any custom packages exposed via decorators
+        deco_module_paths = set()
+        for step in self._flow:
+            for deco in step.decorators:
+                for path_tuple in deco.add_to_package():
+                    file_path,_ = path_tuple
```
romain-intel, line 85: Out of curiosity, what is the second element of the tuple?

valayDave: `add_to_package` returns a list of (path, arcname) to be added to the job package, where `arcname` is the alternative name for the file in the package. This follows the same convention as `MetaflowPackage.path_tuples`. 

An example implementation of this decorator hook can be seen [here](https://github.com/valayDave/metaflow/blob/c6d9260ec32c7bdc13a8fb1f728197f5af135133/metaflow/plugins/cards/card_decorator.py#L26).

metaflow/package.py
```diff
@@ -74,6 +76,20 @@ def path_tuples(self):
                     exclude_hidden=False,
                     addl_suffixes=self.metaflow_custom_addl_suffixes):
                 yield path_tuple
+             
+        # Any custom packages exposed via decorators
+        deco_module_paths = set()
+        for step in self._flow:
+            for deco in step.decorators:
+                for path_tuple in deco.add_to_package():
```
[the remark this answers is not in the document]
valayDave: In my implementation of this [decorator hook](https://github.com/valayDave/metaflow/blob/c6d9260ec32c7bdc13a8fb1f728197f5af135133/metaflow/plugins/cards/card_decorator.py#L26) I return a list of tuples with the absolute path at the first item in the tuple and the `arcname` as the second item in the tuple; 

Should we just add a comment on what the decorator hook should return _or_ should we add some check to see if the path is absolute or relative?

valayDave:
I will reopen a branch for this. Resolving all of these conflicts would be too consuming compared to just plucking the diff and sending a new PR.

Issue #651, opened 2021-08-19
Foreach over list parameter results in None value for input


lsimoneau:
We have a flow with a `JSONType` parameter containing a list of values. We attempted to split into a `foreach` step across the parameter. Metaflow splits off the correct number of tasks (one per value in the list), but in each task the `input` value is `None`. 

I would expect this to either fail validation (tell us we can't use a parameter in a `foreach`) or Just Work™ and make the value available in the split tasks.

Example flow:

```python
from metaflow import FlowSpec, Parameter, step, JSONType

class ForeachFlow(FlowSpec):
    numbers_param = Parameter(
        "numbers_param",
        type=JSONType,
        default='[1,2,3]'
    )
    
    @step
    def start(self):
        # This works, and passes each number to the run_number step:
        #
        # self.numbers = self.numbers_param
        # self.next(self.run_number, foreach='numbers')
        
        # But this doesn't:
        self.next(self.run_number, foreach='numbers_param')
        
    @step
    def run_number(self):
        print(f"number is {self.input}")
        self.next(self.join)
        
    @step
    def join(self, inputs):
        self.next(self.end)
        
    @step
    def end(self):
        pass
    
if __name__ == '__main__':
    ForeachFlow()
```
    
This results in:

```
2021-08-19 21:50:37.718 Workflow starting (run-id 390):
2021-08-19 21:50:37.801 [390/start/82041 (pid 8328)] Task is starting.
2021-08-19 21:50:40.221 [390/start/82041 (pid 8328)] Foreach yields 3 child steps.
2021-08-19 21:50:40.222 [390/start/82041 (pid 8328)] Task finished successfully.
2021-08-19 21:50:40.558 [390/run_number/82042 (pid 8368)] Task is starting.
2021-08-19 21:50:40.684 [390/run_number/82043 (pid 8371)] Task is starting.
2021-08-19 21:50:40.868 [390/run_number/82044 (pid 8374)] Task is starting.
2021-08-19 21:50:42.566 [390/run_number/82043 (pid 8371)] number is None
2021-08-19 21:50:42.667 [390/run_number/82042 (pid 8368)] number is None
2021-08-19 21:50:42.748 [390/run_number/82044 (pid 8374)] number is None
2021-08-19 21:50:43.879 [390/run_number/82043 (pid 8371)] Task finished successfully.
2021-08-19 21:50:44.633 [390/run_number/82042 (pid 8368)] Task finished successfully.
[...]

PR #652, opened 2021-08-20
Handle parameters as `foreach vars`

savingoyal:
Fix for #651

### DIFF
--- a/metaflow/flowspec.py
+++ b/metaflow/flowspec.py
@@ -246,6 +246,8 @@ class FlowSpec(object):
                 #                 self.x is not available
                 self._cached_input[stack_index] = None
             else:
+                if isinstance(var, Parameter):
+                    var = var.load_parameter(self._datastore[frame.var])
                 try:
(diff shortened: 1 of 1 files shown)


metaflow/flowspec.py
```diff
@@ -43,13 +43,19 @@ class FlowSpec(object):
...
                   '_graph',
                   '_flow_decorators',
                   '_steps',
                   'index',
                   'input'}
+    # When checking for parameters, we look at dir(self) but we want to exclude
+    # attributes that are definitely not parameters and may be expensive to
+    # compute (like anything related to the `foreach_stack`). We don't need to exclude
+    # names starting with `_` as those are already excluded from `_get_parameters`.
+    _NON_PARAMETERS = {'cmd', 'foreach_stack', 'index', 'input', 'script_name', 'name'}
```
savingoyal, line 58: A note for a follow up PR - we should throw an error if we detect that a user is using any of these as parameter names.

PR #655, opened 2021-08-23
host_volumes for batch now takes an array instead of a csv

oavdeev:
You can now do this:
```python
@batch(host_volumes=['/home', '/var/log'])
```


savingoyal (approved):
LGTM!

PR #612, opened 2021-07-15
New flow CLI, SDK

ryan-williams:
This is a large change, factored into 4 sequential groups of commits, addressing pain points I hit as a new user/developer of Metaflow; [work leading up to this has been discussed in Slack](https://outerbounds-community.slack.com/archives/C020U025QJK/p1623446101135200):

- Loosen requirement that "1 file = 1 flow = 1 class = 1 Python process":
  - support multiple flows per file (new CLI: `metaflow flow <file[:flow_name]>`)
  - support multiple flows being defined in one Python process
- Reduce flow-definition boilerplate
- Define graph structure using decorators, not `self.next` calls
- Compose flows via inheritance
- add `pytest` tests under `metaflow/tests` (in addition to existing tests under `test/core`)

Everything should be backwards-compatible; flows can optionally be defined using a new API under `metaflow.api`; existing Metaflow public members work as they always have.

## Examples
See [`metaflow/api/README.md` @ `dsl`](https://github.com/celsiustx/metaflow/tree/dsl/metaflow/api#readme) for a detailed write-up and [examples](https://github.com/celsiustx/metaflow/tree/dsl/metaflow/api#examples), but here is a short overview:

### Support multiple flows per file
Using new (optional) `metaflow flow` CLI entrypoint:
```diff
-python my_flow.py <help|run|show|…> …
+metaflow flow my_flow.py[:MyFlow] <help|run|show|…> ….
```
The flow name (`[:MyFlow]` above) can be omitted if there is just one flow in the file.

### Reduce flow-definition boilerplate
`start` and `end` steps, `self.next` calls, and `__main__` handler are often superfluous:
```diff
-from metaflow import FlowSpec, step
+from metaflow.api import FlowSpec, step

 class LinearFlow(FlowSpec):
-    def start(self):
-        self.next(self.one)
-
     @step
     def one(self):
         self.a = 111
-        self.next(self.two)

     @step
     def two(self):
         self.b = self.a * 2
-        self.next(self.three)

     @step
     def three(self):
         assert (self.a, self.b) == (111, 222)
-        self.next(self.end)
-
-    @step
-    def end(self): pass

-if __name__ == '__main__':
-    LinearFlow()
```

### Define graph structure using decorators, not `self.next` calls
Mixing graph structure logic (`self.next` calls) into step implementations is clumsy; they are two different levels of abstraction and should be better delineated/separated:
```python
class SumSquares(FlowSpec):
    num = Parameter("num", required=True, type=int, default=4)

    @step
[...]

PR #662, opened 2021-08-24
continue removing python2 from CI

### LINKED ISSUE #668: Drop Python 2 support
Context:
- discussed, but not directly addressed, in #473
- blocks #612, #666
- #662 mostly achieves it (removes Python 2 from CI; perhaps there is other cleanup that can happen in the codebase once Python 2 has been dropped?)

Does it have to wait for Metaflow 3.x? If so, perhaps we can:
- start a `3.x` branch
- make it the default merge target / backport to the existing ("`2.x`") branch on a best-effort basis

ryan-williams:
**Update**: Partly superceded by #867, but removes a few more vestigial Python 2 configurations.

----

Fixes https://github.com/Netflix/metaflow/issues/668; opening partly for discussion.

This may have to wait for Metaflow 3.x; @villelaitila mentioned that long-running Metaflow services may exist where upgrading them (to a new version that removes Python 2), but not their clients, could cause breakage if those clients try to do Python 2 things.

I'm not familiar enough with what services like that may exist, or whether anyone is likely to run into that situation. Interested in others' thoughts!

This was factored out of #612.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -73,5 +73,3 @@ jobs:
       run: |
         cd R/tests
         Rscript run_tests.R
-
-
--- a/metaflow/_vendor/click/_compat.py
+++ b/metaflow/_vendor/click/_compat.py
@@ -270,7 +270,6 @@ if PY2:
             value = value.decode(get_filesystem_encoding(), "replace")
         return value
 
-
 else:
     import io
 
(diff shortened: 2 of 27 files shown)


ryan-williams:
bumping this, seems like #867 already did the main work of taking python 2 out of the CI, but there's some additional Python 2 CI configs laying around that this cleans up

Issue #677, opened 2021-08-31
MetaflowExceptionWrapper always contains None?


lsimoneau:
When calling `task.exception` the returned object is always `metaflow.exception.MetaflowExceptionWrapper(None)` regardless of the exception that was actually raised.

Looking at the code it seems like this is meant to be holding the underlying exception, so I imagine it's not supposed to just have `None`?

savingoyal:
@lsimoneau Depending on how the task exited, the value of `task.exception` may well be None. If you can expand on your scenario, we can do some digging.

cc @oavdeev

valayDave:
@lsimoneau : Can you check `task.exception.exception`. It is storing the exception but the `repr` is not perfect coming out perfect;  To see the exception better you can do `str(task.exception)`

lsimoneau:
@valayDave Thanks for that, `task.exception.exception` does have the underlying exception. I can do a PR to add a `__repr__` here https://github.com/Netflix/metaflow/blob/master/metaflow/exception.py#L11 to at least show the class of the underlying exception at least?

valayDave:
@lsimoneau: Maybe try `task.exception.type` to get the class of the underlying exception?

PR #679, opened 2021-08-31
Representation change for `task.exception`

valayDave:
- Added Repr to fix error representation problem
- Added a test case to ensure this code works as expected. 
- Based on #677

### DIFF
--- a/metaflow/exception.py
+++ b/metaflow/exception.py
@@ -30,6 +30,9 @@ class MetaflowExceptionWrapper(Exception):
 
     def __setstate__(self, state):
         self.__dict__ = state
+    
+    def __repr__(self):
+        return str(self)
 
     def __str__(self):
         if self.stacktrace:
--- a/test/core/tests/task_exception.py
+++ b/test/core/tests/task_exception.py
@@ -1,0 +1,26 @@
+from metaflow_test import MetaflowTest, ExpectationFailed, steps
+
+
+
+class TaskExceptionTest(MetaflowTest):
+    """
+    Test that an artifact defined in the first step
+    is available in all steps downstream.
+    """
+    PRIORITY = 1
+    SHOULD_FAIL = True
+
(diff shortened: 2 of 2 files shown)


test/core/tests/task_exception.py
```diff
@@ -0,0 +1,25 @@
...
+class TaskExceptionTest(MetaflowTest):
+    """
+    A test to validate if exceptions are stored and retrieved correctly
+    """
+    PRIORITY = 1
+    SHOULD_FAIL = True
+
+    @steps(0, ['singleton-end'],required=True)
+    def step_start(self):
+        raise KeyError('Something has went wrong')
```
savingoyal, line 14: Minor nit - `Something has gone wrong`

test/core/tests/task_exception.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class TaskExceptionTest(MetaflowTest):
+    """
+    A test to validate if exceptions are stored and retrieved correctly
+    """
+    PRIORITY = 1
+    SHOULD_FAIL = True
+
+    @steps(0, ['singleton-end'],required=True)
```
savingoyal, line 12: minor nit - white space `], r`

savingoyal (changes requested):

metaflow/exception.py
```diff
@@ -30,6 +30,9 @@ def __getstate__(self):
 
     def __setstate__(self, state):
         self.__dict__ = state
+    
+    def __repr__(self):
+        return str(self)
```
lsimoneau, line 35: I think `str` returns the whole backtrace, is that desirable here?

valayDave: @savingoyal What are your thoughts?

valayDave: We can also use `self.type` for `__repr__`;

PR #686, opened 2021-09-08
Leave conda dependencies floating instead of pinned.

### LINKED ISSUE #399: Leave conda dependencies floating instead of pinned
`@conda` currently pins some dependencies to specific versions (`requests`, `click` etc.). This can cause conflict with some other user-defined packages (`snowflake-client` etc.) which might pin the same packages to a different version. Also, anytime Metaflow updates the package versions, it ends up resolving the user environment again, which is undesirable. We should leave the dependencies floating.

bishax:
Closes #399 by reviving #411

### DIFF
--- a/metaflow/metaflow_config.py
+++ b/metaflow/metaflow_config.py
@@ -207,23 +207,9 @@ def get_version(pkg):
     return pkg_resources.get_distribution(pkg).version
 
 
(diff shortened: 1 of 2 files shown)


romain-intel:
Closing as this was included in #929

PR #688, opened 2021-09-09
Modify datastore availability for `task_pre_step` decorator hook

savingoyal:
The decorator hook `task_pre_step` is now executed after the flow has been assigned the datastore. This change has no impact on any of the current plugin implementations.


romain-intel (approved):
LGTM

PR #580, opened 2021-06-22
New Datastore abstraction for Metaflow

romain-intel:
Testing in progress...

Detailed notes to follow.

PR #692, opened 2021-09-09
[Breaking] Modify return type of created_at/finished_at to datetime

savingoyal:
Metaflow Client currently returns a string for `created_at` and `finished_at` properties. This PR changes the response type to a more appropriate Datetime object.

### DIFF
--- a/metaflow/client/core.py
+++ b/metaflow/client/core.py
@@ -1,4 +1,5 @@
 from __future__ import print_function
+from datetime import datetime
 import os
 import time
 import tarfile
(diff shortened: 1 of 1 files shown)


metaflow/client/core.py
```diff
@@ -1,4 +1,5 @@
 from __future__ import print_function
+from datetime import datetime
 import os
 import time
```
romain-intel, line 4: Nit: can remove since no longer used.

romain-intel (approved):
Minor nit but otherwise LGTM.

metaflow/client/core.py
savingoyal: Done

savingoyal:
I will merge this PR right when we are merging #580

PR #695, opened 2021-09-10
Emit project related tags from within project decorator

savingoyal:


romain-intel (changes requested):
The datastore will need the metadata in the convergence branch to pass it down to the task datastores.

savingoyal:
> The datastore will need the metadata in the convergence branch to pass it down to the task datastores.

Resolved offline that this change is independent of metadata being referenced in datastore.

PR #696, opened 2021-09-10
Start compressing Metaflow code packages

savingoyal:


romain-intel (changes requested):
As discussed offline, we should keep to gz for speed but we can indeed set the timestamp to zero in the convergence branch to avoid hashing differently for the same package.

romain-intel:
I believe this is no longer relevant as we merged something similar in the convergence branch.

PR #699, opened 2021-09-13
Execute Metaflow Fast Data tests on Linux

savingoyal:


.github/workflows/test_fast_data.yml
```diff
@@ -1,4 +1,4 @@
-name: pr-datastore-tests
+name: Tests - Metaflow Fast Data
```
saikonen, line 1: Unsure on the naming here, what does "Fast Data" mean in this context? The original names pointing to S3 seemed a bit more descriptive?

PR #702, opened 2021-09-15
Tag metadata with attempt_id

romain-intel:


metaflow/metadata/metadata.py
```diff
@@ -475,6 +475,8 @@ def _register_code_package_metadata(self, run_id, step_name, task_id):
                 value=json.dumps({'ds_type': code_ds, 'sha': code_sha, 'location': code_url}),
                 type='code-package',
                 tags=[]))
+        # We don't tag with attempt_id here because not readily available; this
```
savingoyal, line 478: Let's add attempt id's here as well for consistency.

savingoyal (changes requested):

PR #704, opened 2021-09-17
k8s fixes: auth and pod/job race

oavdeev:
This PR has a couple fixes as a result of running our internal test suite:

1. A workaround for https://github.com/kubernetes-client/python/issues/741 . This is properly fixed in https://github.com/kubernetes-client/python-base/pull/250 but that PR is not yet merged into the kubernetes python client. In this PR, I've modified our client retrieval code to auto-refresh the client at fixed interval.
2. A race condition where the pod object doesn't get immediately created after we submit a Job. Fixed by retrying with timeout.
3. A race condition where the Job status isn't immediately updated after container exits. Fixed by waiting for job status to update with a timeout.

I have also slightly refactored the `kubernetes_client.py` for readability (obviously this is very subjective). Mostly replacing properties with methods, in cases where the property logic was more than simple read from a private attribute. For example, `client.is_done` is now more imperatively-sounding `client.check_is_done()` because it actually calls K8S API under the hood to retrieve the job status.


metaflow/plugins/aws/eks/kubernetes_client.py
```diff
@@ -40,14 +51,21 @@ def __init__(self):
...
             config.load_kube_config()
         self._client = client
+        self._client_refresh_timestamp = time.time()
 
     def job(self, **kwargs):
-        return KubernetesJob(self._client, **kwargs)
+        return KubernetesJob(self, **kwargs)
+
+    def get(self):
+        if time.time() - self._client_refresh_timestamp < CLIENT_REFRESH_INTERVAL_SECONDS:
```
akyrola, line 60: Shouldn't the comparison be ">" ?

oavdeev: yep 🤦‍♂️

metaflow/plugins/aws/eks/kubernetes_client.py
```diff
@@ -367,31 +387,35 @@ def __repr__(self):
...
-                self._client.CoreV1Api()
-                .list_namespaced_pod(
-                    namespace=self._namespace,
-                    label_selector="job-name={}".format(self._name),
-                )
-                .to_dict()["items"]
-                or [None]
-            )[0]
-        except self._client.rest.ApiException as e:
+            # Pod objects may not get created immediately after job submission
```
savingoyal, line 405: Should we do an exponential back-off here instead of constant backoff?

metaflow/plugins/aws/eks/kubernetes_client.py
```diff
@@ -367,31 +387,35 @@ def __repr__(self):
...
+                        namespace=self._namespace,
+                        label_selector="job-name={}".format(self._name),
+                    ).to_dict()["items"]
+
+                if pods:
+                    return pods[0]
+                else:
+                    time.sleep(POD_FETCH_BACKOFF_SECONDS)
+            else:
+                raise Exception('Could not fetch pod status in %s seconds' % (POD_FETCH_RETRIES * POD_FETCH_BACKOFF_SECONDS))
```
savingoyal, line 417: When we throw an exception here, does it automatically terminate the k8s job?

metaflow/plugins/aws/eks/kubernetes.py
```diff
@@ -337,8 +337,32 @@ def _print_available(tail, stream, should_persist=False):
...
-            exit_code, reason = self._job.reason
+        try:
+            # Now the pod is no longer running, but we need to wait for the job
+            # status to update. *Usually* that happens pretty much immediately,
+            # but since it is technically done by the job controller
+            # asynchronously, there's room for a race condition where pod is
+            # done but the job is still "active". It is more likely to happen
+            # when the control plane is overloaded, e.g. on minikube/kind.
+            self._job.wait_done(timeout_seconds=20)
+        except TimeoutError:
```
savingoyal, line 348: Is it feasible to encapsulate this logic in `kubernetes_client.py`? Currently, that module takes care of all the complexity of ascertaining the correct job status.

savingoyal (commented):
Some quick comments. I haven't been able to play with this PR live.

metaflow/plugins/aws/eks/kubernetes.py
```diff
@@ -317,7 +317,7 @@ def _print_available(tail, stream, should_persist=False):
                 now = time.time()
                 log_update_delay = update_delay(now - start_time)
                 next_log_update = now + log_update_delay
-                is_running = self._job.is_running
+                is_running = self._job.check_pod_is_running()
```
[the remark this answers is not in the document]
oavdeev: To make it clear that it checks that the _pod_ is running not the k8s _job_. The entire issue was that `is_running` was checking for the pod status but `is_done` was checking for the job status, as a result `RunningJob` could be neither "running" nor "done" at the same time.

Actually, now reading your comments on this PR I realize that the "Job" in `RunningJob` wasn't intended to mean "Kubernetes Job" but a _separate logical concept_ that kinda abstracts away both k8s jobs and pods. It makes sense now but I wonder if there's a way to make it more clear for the reader; when I encounter a variable named `_job` in `kubernetes.py` it makes me think of a K8S job object.

metaflow/plugins/aws/eks/kubernetes_client.py
```diff
@@ -472,8 +496,7 @@ def id(self):
         # TODO (savin): Should we use pod id instead?
         return self._id
 
-    @property
-    def is_done(self):
+    def check_is_done(self):
```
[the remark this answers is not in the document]
oavdeev: It is def subjective but when I was debugging this issue, I found that `@property` made this code considerably harder to understand, at least for me. When I see `self.is_done` I assume that it is either a simple variable read or maybe a thin wrapper over one; where in reality it does pretty complex stuff: calls K8S API and updates the internal state.

metaflow/plugins/aws/eks/kubernetes.py
oavdeev: Let me see, I now get your intent that `RunningJob` was supposed to abstract away k8s job and pod together

metaflow/plugins/aws/eks/kubernetes_client.py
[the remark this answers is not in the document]
savingoyal: I agree with the sentiment - this is just to keep the codebase a bit consistent.

metaflow/plugins/aws/eks/kubernetes_client.py
oavdeev: I wanted to say yes but after thinking a bit, a constant backoff here should be fine. The rationale for using exponential backoff is usually about retrying service _errors_; since you may be getting the error response because the service is overloaded in the first place, you don't want to hammer it even more with constant retries.

However here we're not retrying an error response, just waiting for the pod to appear.

metaflow/plugins/aws/eks/kubernetes.py
```diff
@@ -358,9 +358,7 @@ def _print_available(tail, stream, should_persist=False):
                 "%s. This could be a transient error. "
                 "Use @retry to retry." % msg
             )
-        elif not self._job.is_done:
-            # Kill the job if it is still running by throwing an exception.
-            raise KubernetesKilledException("Task failed!")
+
```
oavdeev, line 361: Removed since it is no longer possible for `RunningJob` to be not "running" anymore and not "done" at the same time.

metaflow/plugins/aws/eks/kubernetes_client.py
```diff
@@ -534,38 +583,50 @@ def status(self):
...
+        # - the container hasn't started *yet*
+        if not self._check_is_job_done():
             # If not done, check if pod has been assigned and is in Running
             # phase
             self._pod = self._fetch_pod()
-            return self._pod.get("status", {}).get("phase") == "Running"
+            pod_phase = self._pod.get("status", {}).get("phase")
+            if pod_phase == "Running":
+                return True
+            elif pod_phase in ("Succeeded", "Failed"):
```
savingoyal, line 610: Maybe, I am missing something - why do we need to wait for 40 seconds for _us_ to mark the job as not running and not immediately?

savingoyal (commented):
A comment, but LGTM otherwise!

metaflow/plugins/aws/eks/kubernetes_client.py
oavdeev: It is about the issue 3 in the PR description, and trying to make `RunningJob` state machine simpler.

The core issue here was that we'd rely on pod status to determine "running" and V1Job status to determine "done", and it is technically possible for pod to be "Succeeded" but V1Job would still have "active= 1" for a brief period, before the job controller updates the V1Job metadata. That caused a successful "RunningJob" go through these states:

1. not-running, not-done (haven't started yet)
2. running, not-done
3. not-running, not-done (again!)
4. not-running, done

I wanted to get rid of this 3rd state because it is confusing for the caller, sometimes doesn't even happen, and in this case practically this complexity caused this bug. For the caller, it is also indistinguishable from the very first state, which feels like a trap for future us. So in this PR, I've changed it so that if the pod is done but the job isn't, `is_running` just kinda wait it out so that next call to `is_done` is guaranteed to return True. Or in other words, after the job has started, it is impossible for it to be not-running and not-done at the same time.

savingoyal: I agree, but we can return `False` for `_check_is_running` when `pod_phase in ("Succeeded", "Failed")` immediately and in `_check_is_job_done` we can add a check to for `_check_is_running` if the V1Job status is active. That should remove the need for a 40-second poll.

oavdeev: So something like
```
def _check_is_job_done(..)
    if self._job["active"] and not self._check_is_running():
       return True
```
Effectively "folding" that intermediate state into "done" state instead?

That's an option, I think what's stopped me is that it would complicate `is_succeeded`/`is_failed` since they wouldn't be able to rely on V1Job status, as it may not be yet up to date.

savingoyal: We can change the logic of `is_succeeded/is_failed` to also inspect the status of "pod_phase". We do have this issue where the "pod" itself may disappear depending on how the job was canceled.

savingoyal: This will get us out of the business of maintaining the magic number of 40

oavdeev: Hmm that may work but i think i need to draw the entire state diagram to make sure we're not missing anything. Btw remind me why don't we define `is_running()` as `V1JobStatus.active > 0` in the first place?

savingoyal: `is_running` returns `True` only when the container starts running (V1Job may be active but for one reason or the other either the pod or the container may not have come up) - otherwise, we mark the job as waiting.

savingoyal: Another point that we need to think through if we are introducing a check for `succeeded` and `failed` pod phases is what happens when a new pod phase is introduced. We can consider inspecting the PodConditions to have a greater degree of certainty, although it makes reasoning about the code a bit more convoluted. That was one of the reasons why I was only looking at the `Running` pod phase in the original implementation.

oavdeev: Soo at the end of the day,
```
RunningJob.running := pod.phase is "Running"
RunningJob.done := V1JobStatus.succeeded > 0 OR V1JobStatus.failed > 0 OR pod.phase in ("Succeded", "Failed") OR V1JobSpec.parallelism is 0
RunningJob.succeeded := done AND (pod.phase is "Succeeded" OR V1JobStatus.succeeded > 0)
RunningJob.failed := done AND (pod.phase is "Failed" OR job parallelism is 0 OR V1JobStatus.failed > 0)
RunningJob.waiting = NOT (running OR done)
```
(if pod doesn't exist, pod.phase is assumed to be "null" above)

This should ensure that the job goes through `waiting -> running -> done` without weird intermediate states, and that a "done" job is always either "succeeded" or "failed".

savingoyal: Seems reasonable to me (need to dust up my notes to find edge cases). This state diagram is pretty much the original implementation with accommodation for eventual consistency of job status? This still doesn't technically ensure that the job goes through `waiting -> running -> done` since pod phases can flip unpredictably to `unknown`. 

Also, as an aside, the implementation currently also tries to optimize for the number of pod and job status API calls, so the order in which these conditions are evaluated is important.

oavdeev: Btw I have also removed another wait-loop where I'd wait for pod object to appear after the job was created, I think it is not necessary as long as we carefully handle pod-does-not-exist in the state machine logic

PR #712, opened 2021-09-21
Rfc compliant 1123 job names

oavdeev:
This fixes a known TODO item on K8S API to make job names always-valid


metaflow/plugins/aws/eks/kubernetes.py
```diff
@@ -44,6 +45,28 @@ class KubernetesKilledException(MetaflowException):
...
+def gen_job_name(flow_name,
+    run_id,
+    step_name,
+    task_id,
+    attempt
+):
+    """ Generate RFC 1123 compatible job name """
+    long_name = "-".join(
+            [
+                flow_name,
```
savingoyal, line 57: Technically `flow_name` can start with an underscore, which will make the output of this function non-compliant. 

Also, we would need to sanitize the strings for labels as well - we can (at a later stage) make this a utility method.

savingoyal (changes requested):

metaflow/plugins/aws/eks/kubernetes.py
oavdeev: Fixed. Also I realized that the previous version could generate identical names for `HelloFlow` and `Helloflow`, fixed that too

oavdeev:
either works, #719 has this commit and one more for labels

savingoyal:
Great! Merged #719. We can sunset this PR.

PR #714, opened 2021-09-23
Final set of patches to convergence

savingoyal:


metaflow/cli.py
```diff
@@ -316,43 +316,50 @@ def logs(obj,
...
-                # a legacy run (unless we have seen new-style data already
-                # for another stream). This return an empty string if
-                # nothing is found
-                log = ds.load_log_legacy(stream)
-                if log and timestamps:
-                    raise CommandException("We can't show --timestamps for "
-                                            "old runs. Sorry!")
-                echo_unicode(log, nl=False)
-
+    if ds_list:
```
romain-intel, line 319: This is not required; both cases allow for `allow_unsuccessful` (also called `allow_not_done` except for here). In other words, it is handled automatically already.

metaflow/client/core.py
```diff
@@ -376,7 +344,8 @@ def __init__(self,
         else:
             raise MetaflowInternalError(msg="Unknown type: %s" % self._NAME)
 
-        self._created_at = datetime.fromtimestamp(self._object['ts_epoch']/1000.0)
+        self._created_at = time.strftime(
```
romain-intel, line 347: I would revert this because when we merged #692 in master and then back to here, it will conflict needlessly.

metaflow/datastore/datastore_storage.py
```diff
@@ -8,7 +8,7 @@ class CloseAfterUse(object):
     """
     Class that can be used to wrap data and a closer (cleanup code).
     This class should be used in a with statement and, when the with
-    scope exists, `close` will be called on the closer object
+    scope exits, `close` will be called on the closer object
```
romain-intel, line 11: exists

metaflow/metadata/util.py
```diff
@@ -5,16 +5,17 @@
...
 def sync_local_metadata_to_datastore(metadata_local_dir, task_ds):
     with util.TempDir() as td:
         tar_file_path = os.path.join(td, 'metadata.tgz')
-        with tarfile.open(tar_file_path, 'w:gz') as tar:
+        buf = BytesIO()
+        with tarfile.open(name=tar_file_path, mode='w:gz', fileobj=buf) as tar:
             tar.add(metadata_local_dir)
-        with open(tar_file_path, 'rb') as f:
-            _, key = task_ds.parent_datastore.save_data([f], len_hint=1)[0]
+        blob = buf.getvalue()
```
romain-intel, line 17: We should probably just do buf.seek(0) and pass buf directly; no need to load it separately in memory.

romain-intel (commented):
A few comments. I can make the changes if you are ok with them. I will also add the fix for the capture log needed for Titus.

metaflow/client/core.py
savingoyal: How will this conflict? It's a no-change from master and should flow seamlessly on rebase.

metaflow/cli.py
romain-intel: For posterity: leaving here in case the task/step does not exist but modified the error message to more accurately reflect the reality.

PR #722, opened 2021-09-26
(R) Better error message when trailing commas are used in step definitions

mdneuzerling:
When a user accidentally leaves a trailing comma in the definition of a step, they get an unhelpful error message that doesn't really describe the problem:

```r
step(
    flow = metaflow("TrailingComma"),
    step = "start",
)
```

```
Error in lapply(decorators, is.decorator) : 
  argument is missing, with no default
```

Trailing commas aren't allowed in R, but this one slips by because the function accepts `...` additional arguments (intended to be decorators). I put a check in place that looks for trailing commas, which show up in function calls as empty symbols. It still returns an error, but I'd argue it's a much better error message:

```
No decorators have been provided.
  Is there a trailing comma in your step definition?
```

### DIFF
--- a/R/R/step.R
+++ b/R/R/step.R
@@ -26,6 +26,15 @@ step <- function(flow, ..., step, r_function = NULL, foreach = NULL, join = FALS
 identifiers; they can contain letters, numbers, and underscores, although they
 cannot begin with a number.")
   }
+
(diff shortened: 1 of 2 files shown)


mdneuzerling:
Going to rethink this one. I think I can do better.

PR #725, opened 2021-09-27
Per attempt task

romain-intel:


metaflow/datastore/task_datastore.py
```diff
@@ -462,7 +462,8 @@ def done(self):
             self._metadata.register_metadata(
                 self._run_id, self._step_name, self._task_id,
                 [MetaDatum(field='attempt-done', value=str(self._attempt),
-                           type='attempt-done', tags=[])])
+                           type='attempt-done',
+                           tags=['attempt_id:{0}'.format(self._attempt)])])
```
tuulos, line 466: what's the purpose of this?

metaflow/metadata/metadata.py
```diff
@@ -364,7 +383,52 @@ def get_object(cls, obj_type, sub_type, filters, *args):
...
+            attempt_int = None
+
+        pre_filter = cls._get_object_internal(
+            obj_type, type_order, sub_type, sub_order, filters, attempt_int, *args)
+        if attempt_int is None or sub_order != 6:
+            # If no attempt or not for metadata, just return as is
+            return pre_filter
+        # Otherwise, we filter out the metadata. If we have attempt_id information
+        # (post MF 2.4.0), it's easy; if not, we do our best to reconstruct
+        # the order.
```
tuulos, line 403: you could move this special logic to handle older versions in a function of its own. Now it adds a lot of complexity in `get_object`

metaflow/datastore/task_datastore.py
romain-intel: Just that we tag all metadata with attempt id and I guess this one fell through the cracks. I should have put it separately. This is for the UI mostly.

metaflow/metadata/metadata.py
romain-intel: ok, will refactor.

romain-intel: so to clarify, just this part of the function that builds the metadata filtered out when we don't have the attempt value right?

PR #727, opened 2021-09-29
fix: Properly treat S3_RETRY_COUNT as integer

saikonen:
**Issue** 
The s3_retry_count is read from environment variables and treated as a string. This breaks when the config value is passed to `range()` further down the line. 

**Change**
Explicitly cast the value set to S3_RETRY_COUNT in configuration to avoid such issues.


savingoyal (approved):
lgtm!

PR #728, opened 2021-09-30
Fix two issues with MetaflowCode and get_data

msavela:
- `get_data` properly handling generator iteration
- `MetaflowCode` accidentally tried to init `BytesIO` with tuple instead of data blob

### DIFF
--- a/metaflow/plugins/conda/conda_environment.py
+++ b/metaflow/plugins/conda/conda_environment.py
@@ -109,12 +109,12 @@ class CondaEnvironment(MetaflowEnvironment):
         if info is None or env_id is None:
             return {'type': 'conda'}
         info = json.loads(info)
-        with cls._filecache.get_data(info['ds_type'], flow_name, info['location'], info['sha']) as f:
-            with tarfile.open(fileobj=f, mode='r:gz') as tar:
-                conda_file = tar.extractfile(CONDA_MAGIC_FILE)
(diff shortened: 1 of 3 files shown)


romain-intel:
As mentioned offline: there is another similar issue l112 on conda_environment.py. We can address that one here and merge. Changes LGTM.

msavela:
Good find, pushed additional fix at `conda_environment.py`

metaflow/plugins/conda/conda_environment.py
```diff
@@ -109,8 +109,9 @@ def get_client_info(cls, flow_name, metadata):
         if info is None or env_id is None:
             return {'type': 'conda'}
         info = json.loads(info)
-        with cls._filecache.get_data(info['ds_type'], flow_name, info['location'], info['sha']) as f:
-            with tarfile.open(fileobj=f, mode='r:gz') as tar:
+        with cls._filecache.get_data(info['ds_type'], flow_name, info['location'], info['sha']) as data:
```
romain-intel, line 112: I actually don't think we need the with here, it's just _, blob = cls._filecache and then blob = BytesIO(blob) (I think)

msavela: Made changes based on your suggestion

romain-intel (approved):

PR #729, opened 2021-09-30
fix: Removed forward slash to prevent double slashes

### LINKED ISSUE #684: Double slashes in `metaflow.S3` keys
With a config like this
```
{
}
```

vamagithub:
closes #684

### DIFF
--- a/metaflow/metaflow_config.py
+++ b/metaflow/metaflow_config.py
@@ -60,7 +60,7 @@ DATASTORE_SYSROOT_LOCAL = from_conf('METAFLOW_DATASTORE_SYSROOT_LOCAL')
 # S3 bucket and prefix to store artifacts for 's3' datastore.
 DATASTORE_SYSROOT_S3 = from_conf('METAFLOW_DATASTORE_SYSROOT_S3')
 # S3 datatools root location
-DATATOOLS_SUFFIX = from_conf('METAFLOW_DATATOOLS_SUFFIX', 'data')
+DATATOOLS_SUFFIX = from_conf('METAFLOW_DATATOOLS_SUFFIX', 'data').rstrip('/')
 DATATOOLS_S3ROOT = from_conf(
     'METAFLOW_DATATOOLS_S3ROOT', 
         '%s/%s' % (from_conf('METAFLOW_DATASTORE_SYSROOT_S3').rstrip('/'), DATATOOLS_SUFFIX)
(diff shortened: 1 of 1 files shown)


romain-intel:
I made a slight change to this. Let me know what you think but it LGTM otherwise.

romain-intel:
One note on this though: this will be a *breaking* change of sorts in the sense that it may point things to a different "bucket".

vamagithub:
> I made a slight change to this. Let me know what you think but it LGTM otherwise.

Thank You. rstrip solution is better than removing "/". Also good for readability

vamagithub:
> One note on this though: this will be a _breaking_ change of sorts in the sense that it may point things to a different "bucket".

Need to check this. Can u point me in a direction where i can test this?

metaflow/metaflow_config.py
```diff
@@ -63,12 +63,12 @@ def from_conf(name, default=None):
 DATATOOLS_SUFFIX = from_conf('METAFLOW_DATATOOLS_SUFFIX', 'data')
```
savingoyal, line 63: This needs to be sanitized as well

savingoyal (changes requested):

metaflow/metaflow_config.py
romain-intel: Fair point, I suppose with a strip in both directions to be safe there :).

romain-intel:
> > One note on this though: this will be a _breaking_ change of sorts in the sense that it may point things to a different "bucket".
> 
> Need to check this. Can u point me in a direction where i can test this?

Well, I am just saying that if your config had a `/` at the end, you will be pointing to a bucket like `something//foo` and now it would be `something/foo` which isn't the same in S3.

metaflow/metaflow_config.py
vamagithub: Done.  Please review it. Thank You :)

savingoyal (changes requested):

saikonen:
Going through older PR's, is this one still active and relevant? After a quick test I was at least unable to reproduce the original issue with trailing slashes in the config

romain-intel:
It is probably still valid but would be a breaking change since "foo//bar" is different than "foo/bar" in S3. It tries to "do the right thing by the user" if the user has a trailing slash in their config.

PR #730, opened 2021-10-01
k8s configuration wizard

oavdeev:
Note that this adds a few configuration variables that are not yet actually used by the K8S plugin.

`metaflow configure aws` logic is unchanged, but the code is refactored a bit since I needed to reuse parts of that flow in `metaflow configure eks`


savingoyal:
@oavdeev What is the expected user workflow with this change?

PR #731, opened 2021-10-01
Fix bug with current.paramter_names

savingoyal:
Without this patch, the following flow returns an empty list for `parameter_names` - 

```
from metaflow import FlowSpec,Parameter,step

class ParameterNameFlow(FlowSpec):

    x = Parameter('x',default=3,type=int)

    @step
    def start(self):
        from metaflow import current
        print(current.parameter_names)
        self.next(self.end)

    @step
    def end(self):
        print("Done Compute")

if __name__ == "__main__":
    ParameterNameFlow()
```


romain-intel (approved):
LGTM. You can add the test flow as discussed if you want and merge.

PR #744, opened 2021-10-08
Support dict and list-of-dictionary arguments for decorators

oavdeev:
There seems to be several places where we convert arguments to strings, I'm not 100% sure how they are all used. Need this for K8S as some parameters there can be somewhat more complex than just strings.

Without this, if you add a new decorator with an argument that takes a dictionary or list-of-dictionaries, Metaflow will trigger the subprocess like this:

```bash
python ./helloworld1.py ...  --myarg {'key': 'key1',  'value': 'value1'} ...
```

After this change, the argument will be passed as JSON instead, so it can be parsed by the subprocess easily:
```bash
python ./helloworld1.py ...  --myarg {"key": "key1",  "value": "value1"} ...
```

### DIFF
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -1,5 +1,6 @@
 import inspect
 import sys
+import json
 import traceback
 from datetime import datetime
 from functools import wraps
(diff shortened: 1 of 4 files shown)


romain-intel:
Did some limited testing internally; didn't see anything borked.

PR #764, opened 2021-10-14
handling exceptions for non-picklizable artifacts

### LINKED ISSUE #763: Better error message on failure of Pickle serialization
When Metaflow serializes artifacts to Pickle, It will be a good practice to fail loudly by stating what artifact couldn't be serialized. Currently, when Metaflow fails pickle serialization, it doesn't tell me which particular artifact was the cause of failure during pickle serialization.

valayDave:
This resolves  #763

### DIFF
--- a/metaflow/datastore/exceptions.py
+++ b/metaflow/datastore/exceptions.py
@@ -4,7 +4,7 @@ class DataException(MetaflowException):
     headline = "Data store error"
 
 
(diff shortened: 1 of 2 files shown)


romain-intel:
Totally in favor of this; I actually frequently add a few lines like this myself to help when I am debugging. Thanks!

PR #769, opened 2021-10-15
K8S: support setting tolerations

oavdeev:
(previously #741)

For now this is for advanced uses only, so not the nicest UX and error handling.

Note that there is a separate commit that is included in this PR but also has a PR of its own in #768

### DIFF
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -1,6 +1,7 @@
 import inspect
 import os
 import sys
+import json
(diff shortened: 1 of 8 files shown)


oavdeev:
I'm gonna make it a config variable for now instead of a decorator parameter after discussing this with @savingoyal

savingoyal:
support has been merged

PR #773, opened 2021-10-16
Update R tests to use R 4.1.1

savingoyal:
Reverts Netflix/metaflow#737

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -36,17 +36,17 @@ jobs:
         python3 -m pip install --upgrade pip
         python3 -m pip install tox numpy
     
-    - name: Set up R 3.6
+    - name: Set up R 4.1
       if: matrix.lang == 'R'
       uses: r-lib/actions/setup-r@v1
       with:
-        r-version: '3.6.3' 
+        r-version: '4.1.1' 
     
-    - name: Install R 3.6 system dependencies
+    - name: Install R 4.1 system dependencies
       if: matrix.lang == 'R' && matrix.os == 'ubuntu-latest'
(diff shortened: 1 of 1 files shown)


savingoyal:
hmmph.. R test setup is still borked. @mdneuzerling have you ever encountered this issue with R test setup failure?

mdneuzerling:
Actually, the failing Metaflow build is on _Big Sur_ (11.6) whereas the successful build I posted is on Catalina (10.15.7). Both use `macos-latest`, but we're in the middle of a rollout. "This change will be rolled out over a period of several weeks beginning on September, 15. We plan to complete the migration by November, 3. This change will be rolled out over a period of several weeks beginning on September, 15. We plan to complete the migration by November, 3."

I'm not 100% certain this is the cause but it seems the most likely culprit. See the [relevant issue](https://github.com/actions/virtual-environments/issues/4060) which states: "R was not available for Big Sur when we prepared the image for the first time." You'll notice that in my successful build R is being _upgraded_, whereas in the failing build R is being _installed_.

mdneuzerling:
Hi @savingoyal. It seems like Metaflow has been caught in the migration of the `macos-latest` environment from _Catalina_ (10.15) to _Big Sur_ (11.0). [I've raised an issue with the `r-lib` team](https://github.com/r-lib/actions/issues/412).

I can recommend either of the two below fixes:

* Force the use of `macos-10.15` instead of `macos-latest`, **or**
* Use R 4.1.1 instead of R 3.6.3 --- the latest version of R seems to work with the Big Sur environment.

savingoyal:
Great catch! I will pin to the latest version of R! Thank you 🙇🏼

PR #776, opened 2021-10-19
Fix issue where a "soft" failure would not be retried in the presence…

romain-intel:
… of a catch

A soft failure (like an exception) in a step having both catch and
retry decorators would not actually cause the step to be first retried
before catching the eventual last exception and instead the exception would
be caught right away and propagated down. This fixes this issue and adds a test.


savingoyal:
@romain-intel Can you verify this behavior on Step Functions/Meson as well?

PR #788, opened 2021-10-27
S3.get_data() to avoid creating temporary files when loading data

akyrola:
When working with large datasets, such as Imagenet, you don't want to thrash the local disk with temporary files. This PR adds function get_data() that loads the data directly into memory.

Test:

```
>>> from metaflow import S3
>>> s3 = S3()
>>> d = s3.get_data("s3://imagenet-outerbounds/LOC_val_solution.csv")
>>> d = s3.get_data("s3://imagenet-outerbounds/LOC_train_solution.csv")
>>> len(d)
23760866
```

The length matches the correct file length and the data looks correct.

### DIFF
--- a/metaflow/datatools/s3.py
+++ b/metaflow/datatools/s3.py
@@ -22,6 +22,7 @@ from ..util import namedtuple_with_defaults,\
 from ..exception import MetaflowException
 from ..debug import debug
 
+import io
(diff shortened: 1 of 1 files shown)


akyrola:
I guess this is not wanted.

PR #801, opened 2021-11-01
mflog changes for supporting AWS Lambda

savingoyal:
Ready for review - general housekeeping for #463

### DIFF
--- a/metaflow/mflog/__init__.py
+++ b/metaflow/mflog/__init__.py
@@ -43,13 +43,13 @@ BASH_SAVE_LOGS = " ".join(BASH_SAVE_LOGS_ARGS)
 def bash_capture_logs(bash_expr, var_transform=None):
     if var_transform is None:
         var_transform = lambda s: "$%s" % s
-    cmd = "python -m metaflow.mflog.tee %s %s"
-    parts = (
+
+    return "python -m metaflow.mflog.redirect_streams %s %s %s %s" % (
+        TASK_LOG_SOURCE,
+        var_transform("MFLOG_STDOUT"),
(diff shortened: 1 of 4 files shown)


metaflow/mflog/redirect_streams.py
```diff
@@ -0,0 +1,54 @@
...
+        else:
+            # Python 3
+            for line in src:
+                decorated = decorate(SOURCE, line)
+                f.write(decorated)
+                dest_stream.buffer.write(line)
+
+
+if __name__ == "__main__":
+    SOURCE = sys.argv[1].encode("ascii")
```
romain-intel, line 32: why ascii and not the more general utf8?

savingoyal:
@oavdeev this PR should unblock #463.

metaflow/mflog/__init__.py
```diff
@@ -43,13 +48,13 @@
 def bash_capture_logs(bash_expr, var_transform=None):
```
oavdeev, line 48: I think it is worth renaming at least the argument, as it cannot be an arbitrary bash expression anymore. Also.. I'm pretty sure there is a bug here (and in the original PR) related to this, in cases when `bash_expr` is not a simple cmd + args

oavdeev: See https://github.com/Netflix/metaflow/pull/805 for a tentative fix

savingoyal: I can rename the arg. Keeping the function name constant will make it easier for @romain-intel to merge this PR into the converged branch.

PR #804, opened 2021-11-02
Addressed an issue with class members

romain-intel:
Class members that were not parameters caused an error (regression
in 2.4.0). This addresses this issue.

This patch also clarifies and makes explicit that modifying class
variables is not safe; an error message is now raised.


romain-intel:
Needs test. Please look it over though and let me know what you think. In particular, the exclusions that I avoid for the class variables.

Issue #524, opened 2021-05-11
`--max-value-size` should be ignored when using `dump --file`


oavdeev:
I get why its needed when using `dump` to stdout, I don't want to get gigabytes of output written to my terminal by accident.

But when using `--file`, having the values _silently_ replaced by a placeholder is surprising, and pretty much never the behavior I'd want.

oliver-pham:
At first glance, I think this issue should be resolved if `max_value_size` (in the code below) is set to `None` when `--file` is provided. Should I consider anything else?

https://github.com/Netflix/metaflow/blob/0d3ef3387db83f804f33e7bfcbaa300bcfd9c745/metaflow/cli.py#L211-L213

If nobody has been assigned yet, may I work on this issue?

PR #806, opened 2021-11-03
Addressing `max-value-size` getting used when a `file` argument is passed 

westford14:
* adding code to alert users that when a --file argument is provided the --max-value-size gets set to None as well as updating the max_value_size var

This PR addresses issue #524 and also adds logging to let users know that the `--file` argument will override the `--max-value-size` argument when used in conjunction to prevent values getting overwritten.

### DIFF
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -226,12 +226,19 @@ def output_dot(obj):
 def dump(obj, input_path, private=None, max_value_size=None, include=None, file=None):
 
     output = {}
-    if file is not None:
+    if file is not None and max_value_size is not None:
         echo(
             "max_value_size will be set to None when a file variable "
-            "is provided so that values will not get replaced"
+            "is provided so that values will not get replaced in the file"
         )
-        max_value_size = None 
+        max_value_size = None
+    elif file is None and max_value_size is None:
+        echo(
+            "max_value_size will be set to 1000 when neither file nor "
+            "max_value_size is set to prevent all information being printed "
+            "to stdout"
+        )
+        max_value_size = 1000
     kwargs = {
         "show_private": private,
         "max_value_size": max_value_size,
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -226,6 +226,12 @@ def output_dot(obj):
 def dump(obj, input_path, private=None, max_value_size=None, include=None, file=None):
 
     output = {}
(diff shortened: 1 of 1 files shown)


westford14:
@savingoyal -- i think you had labeled this issue previously, let me know what you think about this code change

savingoyal:
@oavdeev Can you shepherd this PR?

metaflow/cli.py
```diff
@@ -226,6 +226,12 @@ def output_dot(obj):
 def dump(obj, input_path, private=None, max_value_size=None, include=None, file=None):
 
     output = {}
+    if file is not None:
+        echo(
```
savingoyal, line 230: How about throwing an error instead when both `file` and `max_value_size` are specified?

westford14: yea that's a good callout -- let me rework this to `raise` an error instead

oavdeev: I'm not a `click` expert but if you *don't* specify `--max-value-size` on the command line, would it pass the default value 1000 to `dump()` or None?

westford14: i just did a simple test with this click file:

```
import click

@click.command()
@click.option('--n', default=5, show_default=True)
def dots(n=None):
    print(f"n: {n}")
    click.echo('.' * n)

if __name__ == "__main__":
    dots()
```

and then running -> `python test.py` yielded:

```
n: 5
.....
```

so that seems to imply that if the `--max-value-size` isn't passed it still populates the default value of 1000 to the `dump()` argument instead of `None`, which i guess begs the question of should we try to change the `max_value_size` if a `--file` argument is passed or should it error out

oavdeev: I think at the end of the day, from the end user POV we want `--max-value-size` to default to unlimited if `--file` is specified, and to 1000 if not.

westford14: Sure that makes sense - let me work that up and push up another pass at this

westford14:
@oavdeev -- let me know what you think of this next pass to get to the user experience

oavdeev:
@westford14 from a cursory look it seems like it won't work as expected since `max_value_size` is never `None` due to the click default, no?

savingoyal:
@westford14 any updates on this PR? Please re-open when you have any new comments on this PR.

PR #810, opened 2021-11-04
Fix default resource math 

### LINKED ISSUE #467: Correctly compute container resources when `@resources` is overridden by `@batch`
```
from metaflow import resources, FlowSpec, step

class LinearFlow(FlowSpec):

    @resources(cpu=0.5)
    @step
    def start(self):
        self.my_var = 'hello world'
        self.next(self.end)

    @step
    def end(self):
        pass

if __name__ == '__main__':
    LinearFlow()
```

```
python flow.py run --with batch
```

This command will launch an AWS Batch container with 1 vCPU rather than 0.5 vCPU. A similar problem exists when the workflow is launched on AWS Step Functions.

oavdeev:
Fixes #467.

### DIFF
--- a/metaflow/plugins/aws/aws_utils.py
+++ b/metaflow/plugins/aws/aws_utils.py
@@ -50,3 +50,35 @@ def get_docker_registry(image_uri):
     if registry is not None:
         registry = registry.rstrip("/")
     return registry
+
+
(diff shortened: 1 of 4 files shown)


savingoyal:
@oavdeev Can you verify with SFN as well?

romain-intel:
No issues with this (haven't tested).

PR #817, opened 2021-11-08
Fix an issue with load_artifacts when several artifacts have the same…

romain-intel:
… hash

Incorrect results could be returned from load_artifacts when multiple
artifacts hash to the same value and are all requested at the same
time.

### DIFF
--- a/metaflow/datastore/task_datastore.py
+++ b/metaflow/datastore/task_datastore.py
@@ -338,7 +338,6 @@ class TaskDataStore(object):
                 "load artifacts" % self._path
             )
         to_load = []
-        sha_to_names = {}
         for name in names:
             info = self._info.get(name)
             # We use gzip+pickle-v2 as this is the oldest/most compatible.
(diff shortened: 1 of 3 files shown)


metaflow/datastore/task_datastore.py
```diff
@@ -353,15 +352,13 @@ def load_artifacts(self, names):
...
-                sha_to_names[sha] = name
-                to_load.append(sha)
+                to_load.append(self._objects[name])
         # At this point, we load what we don't have from the CAS
         # We assume that if we have one "old" style artifact, all of them are
         # like that which is an easy assumption to make since artifacts are all
         # stored by the same implementation of the datastore for a given task.
-        for sha, blob in self._ca_store.load_blobs(to_load):
-            yield sha_to_names[sha], pickle.loads(blob)
+        for name, (_, blob) in zip(names, self._ca_store.load_blobs(to_load)):
```
tuulos, line 360: this logic assumes that `load_blobs` returns the blobs in the order of `to_load`, which could break in the future and lead to very subtle bugs. You could grab the `(key, blob)` pair in the return value and use it to produce names

tuulos:
LGTM

PR #821, opened 2021-11-09
Fix a couple of typos/issues in s3 error handling

oavdeev:
...caught by `S3FailureTest`

### DIFF
--- a/metaflow/datatools/s3op.py
+++ b/metaflow/datatools/s3op.py
@@ -89,7 +89,7 @@ def normalize_client_error(err):
     try:
         return int(error_code)
     except ValueError:
-        if error_code == "AccessDenied":
+        if error_code in ("AccessDenied", "AllAccessDisabled"):
             return 403
         if error_code == "NoSuchKey":
             return 404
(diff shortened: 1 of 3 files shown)


savingoyal (approved):
lgtm! waiting for tests to pass. @oavdeev can we enable `S3FailureTest` on buildkite?

romain-intel:
Unsurprisingly, the S3FailureTest does *not* pass (nor will it -- we don't use that code path in the new datastore).

metaflow/datatools/s3op.py
```diff
@@ -89,7 +89,7 @@ def normalize_client_error(err):
     try:
         return int(error_code)
     except ValueError:
-        if error_code == "AccessDenied":
+        if error_code in ("AccessDenied", "AllAccessDisabled"):
```
romain-intel, line 92: Do we want to do this? This is a change in behavior and technically it seems AllAccessDisabled exists separately from AccessDenied and I guess doesn't mean the same thing.

oavdeev: Not sure ¯\_(ツ)_/¯ For AWS SDK it is also HTTP code 403 I believe, so it least this line of logic is consistent with that

romain-intel: I am not opposed to it per say. I actually considered it when the tests started failing but saw that the exception already existed a while ago so decided that if it was already different, there must be a reason (but then again, that's me being rational :) )

romain-intel (approved):

romain-intel:
OK to merge.

PR #824, opened 2021-11-16
`origin-task-id` in `Task` metadata and `origin_pathspec` property in `Task` object

valayDave:
## Context
We currently don't store `task-id` of the origin task for resumed workflows. 

## PR Changes
- Added `origin_pathspec` to metaflow client to retrieve the pathspec of the origin task. 
- Added `origin-task-id` to `Task` metadata

## Todos
- [x] Tests
- [x] `origin_pathspec` in all levels of Metaflow Client.

### DIFF
--- a/metaflow/client/core.py
+++ b/metaflow/client/core.py
@@ -853,13 +853,6 @@ class DataArtifact(MetaflowObject):
         return self.created_at
 
 
-def _origin_run_id_from_task(task):
-    for metadata in task.metadata:
-        if metadata.name == "origin-run-id":
-            return metadata.value
-    return None
-
-
 class Task(MetaflowObject):
     """
     A Task represents an execution of a step.
@@ -904,6 +897,8 @@ class Task(MetaflowObject):
         Code package for this task (if present)
     environment_info : Dict
         Information about the execution environment (for example Conda)
+    origin_pathspec : str
+        Pathspec of the origin task (if it is present)
     """
 
     _NAME = "task"
@@ -956,18 +951,14 @@ class Task(MetaflowObject):
             Pathspec of the origin task or None
         """
         origin_task_id = None
-        origin_run_id = None
         for metadata in self.metadata:
             if metadata.name == "origin-task-id":
                 origin_task_id = metadata.value
-            elif metadata.name == "origin-run-id":
-                origin_run_id = metadata.value
+                break
         if origin_task_id is None:
             return None
-
-        step_name = self.parent.id
-        flow_name = self.parent.parent.parent.id
-        return "%s/%s/%s/%s" % (flow_name, origin_run_id, step_name, origin_task_id)
+        step_orig = self.parent.origin_pathspec
+        return "%s/%s" % (step_orig, origin_task_id)
 
     @property
     def metadata_dict(self):
@@ -1378,6 +1369,8 @@ class Step(MetaflowObject):
         Time this step finished (time of completion of the last task)
     environment_info : Dict
         Information about the execution environment (for example Conda)
+    origin_pathspec : str
+        Pathspec of the origin step (if it is present)
     """
 
     _NAME = "step"
@@ -1512,13 +1505,11 @@ class Step(MetaflowObject):
             Pathspec of the origin step or None
         """
         task = self.task
-        origin_run_id = None
-        origin_run_id = _origin_run_id_from_task(task)
-        if origin_run_id is None:
+        runorig = self.parent.origin_pathspec
+        if runorig is None:
             return None
         step_name = self.id
-        flow_name = self.parent.parent.id
-        return "%s/%s/%s" % (flow_name, origin_run_id, step_name)
+        return "%s/%s" % (runorig, step_name)
 
 
 class Run(MetaflowObject):
@@ -1541,6 +1532,8 @@ class Run(MetaflowObject):
         Code package for this run (if present)
     end_task : Task
         Task for the end step (if it is present already)
+    origin_pathspec : str
+        Pathspec of the origin run (if it is present)
     """
 
     _NAME = "run"
@@ -1563,7 +1556,10 @@ class Run(MetaflowObject):
         """
         task = next(self.steps()).task
         origin_run_id = None
-        origin_run_id = _origin_run_id_from_task(task)
+        for metadata in task.metadata:
+            if metadata.name == "origin-run-id":
+                origin_run_id = metadata.value
+                break
         if origin_run_id is None:
             return None
         flow_name = self.parent.id
--- a/metaflow/client/core.py
+++ b/metaflow/client/core.py
@@ -1504,7 +1504,6 @@ class Step(MetaflowObject):
         str
             Pathspec of the origin step or None
         """
-        task = self.task
         runorig = self.parent.origin_pathspec
         if runorig is None:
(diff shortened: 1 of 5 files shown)


metaflow/client/core.py
```diff
@@ -938,6 +938,30 @@ def metadata(self):
             for obj in all_metadata
         ]
 
+    @property
+    def origin_pathspec(self):
```
savingoyal, line 942: For consistency, do you want to add this method to the parent objects as well? That will simplify the logic here as well to `os.path.join(self.parent.origin_pathspec, origin_task_id)`.

metaflow/datastore/task_datastore.py
```diff
@@ -553,6 +555,25 @@ def done(self):
         )
 
         if self._metadata:
+            origin_task_metadata = []
```
savingoyal, line 558: You can add more assertions in https://github.com/Netflix/metaflow/blob/master/test/core/tests/resume_*.py to test this PR.

metaflow/datastore/task_datastore.py
```diff
@@ -111,6 +111,8 @@ def __init__(
         self._attempt = attempt
         self._metadata = flow_datastore.metadata
         self._parent = flow_datastore
+        self._origin_task_id = None
```
savingoyal, line 114: Can we also expose `origin_task_id` in the current object?

metaflow/client/core.py
romain-intel: Agreed it should be on all levels. Also make it clear in the doc that if this isn't a clone, it returns None.

metaflow/datastore/task_datastore.py
```diff
@@ -553,6 +555,25 @@ def done(self):
         )
 
         if self._metadata:
+            origin_task_metadata = []
+            if self._origin_task_id is not None:
```
romain-intel, line 559: This is probably not the right place for this. Why not place it in the run_step method where we currently set origin-run-id?

romain-intel (changes requested):

metaflow/datastore/task_datastore.py
valayDave: One reason for placing it there was because [clone_only](https://github.com/Netflix/metaflow/blob/584b7e1996fa1798a1cb9d6ede03ecf145d7ddc4/metaflow/task.py#L238) method was exposed by the `MetaflowTask`. Since in the CLI we call `clone_only` [or](https://github.com/Netflix/metaflow/blob/584b7e1996fa1798a1cb9d6ede03ecf145d7ddc4/metaflow/cli.py#L535) `run_step`, if we place this in run_step then we are not getting origin-task-id as origin-task-id is available on cloning a past task-datastore and is nonexistent for a freshly run task.  

To move it out of the task_datastore I can add this to the `clone_only` method of the `MetaflowTask`. Is that a better option?

metaflow/datastore/task_datastore.py
valayDave: The `current` object is only useful during Task execution. `origin-task-id` is captured during task cloning. So having an `origin-task-id` as a part of the current object wouldn't be useful as during execution we have a fresh task that is non-cloned.

metaflow/datastore/task_datastore.py
valayDave: As a change, I have moved setting `origin-task-id` into `MetaflowTask.clone_only` instead of `TaskDatastore.done`

metaflow/datastore/task_datastore.py
valayDave: Added Test for this too.

valayDave:
I have also add the comment to the docstrings of the core objects. We also need to add this to the actual docs repo.

romain-intel:
@valayDave : OK, I think I got what you are trying to do a little better; basically, for cloned tasks, you want to preserve information about the original run id and task id. That makes sense and was indeed not there before. Putting it where you put it in the `clone_only` makes sense to me at this point.

To @savingoyal 's other point though, after thinking a little bit more, I agree with you @valayDave, that there is very little use of putting the origin task id in the current object and/or storing it in the metadata in `run_step` because we can't really easily reconstruct this information in all cases. There *could* be cases where you re-run a task that already ran (so the information could theoretically be available) but it would be hard to figure out which task specifically for your input (in the case of a foreach or branch) and the value of it is dubious.

romain-intel:
So, if you simplify the origin_pathspec and move it up a bit (a default impl can be like the one you have for step), this hsould be good to go.

valayDave:
@romain-intel : by default impl do you mean that it is a part of the `MetaflowObject`? Currently its a part of the `Step` `Run` and `Task` Objects.

metaflow/client/core.py
```diff
@@ -1501,6 +1543,27 @@ def _iter_filter(self, x):
...
+        """
+        Pathspec of the origin run that the current run was cloned from.
+
+        Returns
+        -------
+        str
+            Pathspec of the origin run or None
+        """
+        task = next(self.steps()).task
```
romain-intel, line 1556: Rare but maybe check if this returns ok because if there is no step at all (for example killed too early), I think this may fail.

romain-intel:
Yes, I mean have an implementation at the MetaflowObject level which is basically "get the one for my parent, if not none, add my id or origin id to it". You can also has a simple if on `self._NAME` to determine if you need to use origin_run_id or origin_task_id I guess so you can have one function pretty much so it's a little cleaner.

valayDave:
@romain-intel; I have made the requested changes. Do these work ?

metaflow/client/core.py
```diff
@@ -311,6 +311,8 @@ class MetaflowObject(object):
         Pathspec of this object (for example: 'FlowName/RunID' for a `Run`)
     path_components : List[string]
         Components of the pathspec
+    origin_pathspec : str
+        Pathspec of the `Run`/`Step`/`Task`, the current MetaflowObject has been cloned from
```
romain-intel, line 315: "or None if not a cloned object"

test/core/tests/resume_originpath.py
```diff
@@ -0,0 +1,45 @@
...
+
+    @steps(2, ["all"])
+    def step_all(self):
+        pass
+
+    def check_results(self, flow, checker):
+        run = checker.get_run()
+        if run is not None:
+            for step in run.steps():
+                if step.id == "start":
```
savingoyal, line 38: Minor nit - we can also add a test for the end step.

savingoyal (approved):
lgtm

romain-intel:
I updated with a quick simplification of the function to get the pathspec. If this looks good. I am fine to merge. I'll approve but please make sure you are ok with the change (it seems I didn't black format when editing on github).

romain-intel (approved):

test/core/tests/resume_originpath.py
valayDave: Do u mean that we call `ResumeFromHere()` in the start step ?

valayDave:
@romain-intel: Let me have a look properly and set reformatted PR. If tests pass then we can merge :)

valayDave:
@romain-intel : The changes fail tests.  I have added a fix to solve the bug introduced in the refactor.

PR #826, opened 2021-11-17
Multinode/Parallel Steps Part 1: with self.next(num_parallel=X), AWS Batch

akyrola:
Implemented multinode support for Metaflow with following structure:
```
@step
def start(self):
   self.next(self.train, num_parallel=4)

@batch
@parallel
@step
def train(self):
   ...
   self.next(train_end)

@step
def train_end(self, inputs):
   ...
```

Currently works only with AWS Batch, working to enable local mode in part 2.

In addition:
* linting related to the new API
* new parallel property to current

Part 2 contains some follow-up fixes.


metaflow/flowspec.py
```diff
@@ -504,6 +517,14 @@ def next(self, *dsts, **kwargs):
                 raise InvalidNextException(msg)
             funcs.append(name)
 
+        if num_parallel:
```
romain-intel, line 520: nit: this seems to allow num_parallel=0 which may cause issues?

akyrola:
Let's look at only #827

PR #827, opened 2021-11-17
Multinode/Parallel Steps Part 1 & 2:  Local support, environment type

akyrola:
Implemented multinode support for Metaflow with following structure:
```
@step
def start(self):
   self.next(self.train, num_parallel=4)

@batch
@parallel
@step
def train(self):
   ...
   self.next(train_end)

@step
def train_end(self, inputs):
   ...
```

Currently works only with AWS Batch, working to enable local mode in part 2.

In addition:

* linting related to the new API
* new parallel property to current
* Fix the issue of client not including control task in tasks of a parallel step by adding a new tag CONTROL_AND_MAPPER. Did not really figure any other way to do this, perhaps it is fine?

--
### How the UBF based implementation works

Parallel execution leverages the Unbounded Foreach (UBF) model of Metaflow, but this is not really visible to the user. How this works:

* If `self.next()` has `num_parallel=N`,  then  (flowspec.py line 520->) a synthetic UBF iterator object is created and assigned to field `self. _parallel_ubf_iter`. (With UBF, user would write `self.next(.., foreach= ParallelUBF(N)`).
* Then, the parallel step needs to have a decorator that takes care of the UBF execution, i.e of spinning the tasks.  This decorator is implemented in `parallel_decorator.py`, i.e `@parallel`. It will simply inject a CLI argument `--num_parallel=N` (parallel_decorator.py line 22), if step is run with AWS Batch or some other execution environment.
* For AWS Batch, the batch decorator launches the N workers and does the UBF business of setting `flow._control_mapper_tasks` (line 263, batch_decorator.py).
* In local mode, the parallel decorator will actually wrap the step function and launch locally the workers. This enables local testing and unit tests.
* Change to the UBF mechanism was made to allow the control task to be also a "mapper", i.e a real task. There is property `flow. _control_task_is_mapper_zero` to enable this.  (Reason is that we don't want to dedicate, say,  a 8 gpu machine to be a dummy control task).  Small changes were done to the core to enable this behavior.

---

### Local testing
To run local test:
```
```

The flow will test the multinode execution happens correctly.

Parallel decorator in local mode will launch subprocesses to handle the different nodes concurrently. To make it possible to detect we are indeed in local mode, batch and eks environments set a new environment variable METAFLOW_RUNTIME_ENVIRONMENT.


akyrola:
I am not sure how to make a stacked PR? I could not compare this to #826

akyrola:
Just noted some tests are failing. The bug is not in the implementation but I need to extend the test system to understand parallel steps.

romain-intel:
@crk-codaio: this is the PR I was talking to you about if you have time to look.

crk-codaio:
> @crk-codaio: this is the PR I was talking to you about if you have time to look.

Yep can take a look during the weekend or upcoming holidays. Excited to see a use-case of UBF (unbounded foreach). 💯

metaflow/plugins/aws/batch/batch_cli.py
```diff
@@ -141,8 +141,9 @@ def kill(ctx, run_id, user, my_runs):
 @click.option("--max-swap", help="Max Swap requirement for AWS Batch.")
 @click.option("--swappiness", help="Swappiness requirement for AWS Batch.")
 # TODO: Maybe remove it altogether since it's not used here
-@click.option("--ubf-context", default=None, type=click.Choice([None]))
+@click.option("--ubf-context", default=None, type=click.Choice([None, "ubf_control"]))
 @click.option("--host-volumes", multiple=True)
+@click.option("--num-parallel", type=int)
```
romain-intel, line 146: Should we add a default?

metaflow/plugins/aws/batch/batch_cli.py
```diff
@@ -141,8 +141,9 @@ def kill(ctx, run_id, user, my_runs):
 @click.option("--max-swap", help="Max Swap requirement for AWS Batch.")
 @click.option("--swappiness", help="Swappiness requirement for AWS Batch.")
 # TODO: Maybe remove it altogether since it's not used here
-@click.option("--ubf-context", default=None, type=click.Choice([None]))
+@click.option("--ubf-context", default=None, type=click.Choice([None, "ubf_control"]))
```
romain-intel, line 144: We have other "--ubf-context" options but they are not always a choice between None and a string. In the main CLI the default is the string "none" for example. Maybe we could make them all consistent to avoid any potential issues/pitfalls?

metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -94,6 +95,49 @@ def execute(self):
                 self._max_swap,
                 self._swappiness,
             )
+
```
romain-intel, line 98: Did not review this section but I assume it was tested and works.

metaflow/plugins/aws/batch/batch_decorator.py
```diff
@@ -292,9 +312,64 @@ def task_finished(
...
+        print("Waiting for batch secondary tasks to finish")
+        while t + TIMEOUT > time.time():
+            time.sleep(2)
+            try:
+                step_path = "%s/%s/%s" % (flow.name, current.run_id, step_name)
+                tasks = [task for task in Step(step_path)]
+                if len(tasks) == len(flow._control_mapper_tasks):
+                    if all(
+                        task.finished_at is not None for task in tasks
+                    ):  # for some reason task.finished fails
```
romain-intel, line 338: That should not be the case. What is the error for that?

metaflow/plugins/aws/batch/batch_decorator.py
```diff
@@ -292,9 +312,64 @@ def task_finished(
...
+        tasks to finish cleanly and produce their output before exiting the
+        main task. Otherwise main task finishing will cause secondary nodes
+        to terminate immediately, and possibly prematurely.
+        """
+        from metaflow import Step  # avoid circular dependency
+
+        t = time.time()
+        TIMEOUT = 600
+        print("Waiting for batch secondary tasks to finish")
+        while t + TIMEOUT > time.time():
```
romain-intel, line 330: nit: maybe put t + TIMEOUT outside the loop and give it a more meaningful name.

metaflow/plugins/catch_decorator.py
```diff
@@ -84,6 +84,12 @@ def task_exception(
 
         # pretend that self.next() was called as usual
         flow._transition = (graph[step].out_funcs, None, None)
+
+        # If this task is a UBF control task, it will return itself as the singleton
+        # list of tasks.
+        flow._control_mapper_tasks = [
+            "/".join((current.run_id, current.step_name, current.task_id))
```
romain-intel, line 91: is there a reason we are not also storing the full pathspec?

metaflow/plugins/catch_decorator.py
```diff
@@ -105,14 +111,12 @@ def task_decorate(
...
-        # running the flow downsteam, as we have a proper entry for this task.
+        # running the flow downstream, as we have a proper entry for this task.
 
         def fallback_step(inputs=None):
             raise FailureHandledByCatch(retry_count)
 
-        # We don't run fallback for `ubf_control` since it doesn't support
-        # any retries.
-        if ubf_context != UBF_CONTROL and retry_count > max_user_code_retries:
+        if retry_count > max_user_code_retries:
```
romain-intel, line 119: Why was this stripped? I get that here the control task is also a mapper but, unless there is a reason to, can we keep the control-and-not-a-mapper mode as well?

metaflow/plugins/conda/conda_step_decorator.py
```diff
@@ -81,9 +81,6 @@ def _python_version(self):
         )
 
     def is_enabled(self, ubf_context=None):
```
romain-intel, line 83: Ditto here.

metaflow/task.py
```diff
@@ -366,14 +366,6 @@ def run_step(
         output.init_task()
 
         if input_paths:
```
romain-intel, line 368: Ditto in terms of stripping out just control tasks. I am curious to know if we can support both modes.

romain-intel (commented):
Preliminary review. Hopefully @pjoshi30 can also look at it. Looks mostly ok (haven't tested yet) and my main concern is around the pure control versus control-and-mapper task-0; I am not sure this PR fully supporst both modes which I think may be useful to do in the long run.

metaflow/cli.py
```diff
@@ -476,6 +476,12 @@ def echo_unicode(line, **kwargs):
     type=click.Choice(["none", UBF_CONTROL, UBF_TASK]),
     help="Provides additional context if this task is of type " "unbounded foreach.",
 )
+@click.option(
+    "--num-parallel",
+    default=0,
+    type=int,
+    help="Num of parallel instances of a step. Ignored in local mode.",
```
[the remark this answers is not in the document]
akyrola: In local mode, it is the parallel decorator which does the spinning of the jobs.

metaflow/client/core.py
```diff
@@ -1366,7 +1366,7 @@ def task(self):
             A task in the step
         """
         for t in self:
-            if CONTROL_TASK_TAG not in t.tags:
+            if CONTROL_TASK_TAG not in t.tags or CONTROL_AND_MAPPER_TAG in t.tags:
```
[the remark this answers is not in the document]
akyrola: That's right.

metaflow/graph.py
```diff
@@ -93,15 +94,22 @@ def _parse(self, func_ast):
             self.invalid_tail_next = True
             self.tail_next_lineno = tail.lineno
             self.out_funcs = [e.attr for e in tail.value.args]
-            keywords = dict((k.arg, k.value.s) for k in tail.value.keywords)
 
+            keywords = dict(
+                (k.arg, getattr(k.value, "s", None)) for k in tail.value.keywords
```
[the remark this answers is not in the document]
akyrola: This is related how the python ast is parsed. For normal self.next(.. foreach="xyz") the foreach has a string value. However, for num_parallel, we accept integers, variable names, or even function calls.

metaflow/plugins/aws/batch/batch_cli.py
akyrola: I think None is the proper default here.

metaflow/plugins/catch_decorator.py
akyrola: This is not related to control as mapper, but just enabling catch semantics for the control task. Before, @catch was not supported for UBF control tasks.

metaflow/task.py
akyrola: Yeah, so it was just arbitrary that conda was not supported for control tasks. Now it is.

metaflow/plugins/catch_decorator.py
romain-intel: right. That was by design in a sense since in the control task is not a worker initial paradigm, the control task wasn't written by the user so we wouldn't catch any errors from it. Here, the control is also a mapper so it makes sense for it to follow the catch semantics but in the previous version (which I agree is no longer used but unless there is a strong reason to remove, we could keep it as is) the control task was not user code.

metaflow/client/core.py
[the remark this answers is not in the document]
romain-intel: Apologies, I don't know how I missed this in the first review but this condition makes it seem like the task is yielded if CONTROL_TASK_TAG is not in tags or CONTROL_AND_MAPPER_TAG is in tags but not in the default case (I suppose) where there is nothing in tags. Ditto for below.

metaflow/plugins/aws/batch/batch_cli.py
romain-intel: Sorry, I should have been clearer: I was suggesting making these options consistent across CLIs (both for this and the ubf-context above). The `step` command in the main CLI has the same arguments but different types/values which, although it should work fine, it may be preferable to make them consistent.

metaflow/plugins/conda/conda_step_decorator.py
romain-intel: I am again not sure we want to re-enable conda for pure control tasks. The logic there was that conda was defined by the user and may be irrelevant for the non-user control task.

romain-intel (commented):
A few unaddressed things but otherwise I think this should be good to merge soon.

metaflow/plugins/conda/conda_step_decorator.py
akyrola: @tuulos suggested we enable conda for control tasks as well. @tuulos  do you want to chime in? Personally I feel it would be weird to have frankenstein situation where the control task runs with different libraries than the normal tasks.

metaflow/plugins/catch_decorator.py
akyrola: @tuulos  do you want to chime in? I think you also suggested that catch should work with UBF as expected.

metaflow/plugins/aws/batch/batch_cli.py
akyrola: Ah got it, makes sense.

metaflow/plugins/catch_decorator.py
akyrola: Sorry I don't understand this question?

metaflow/plugins/catch_decorator.py
tuulos: @romain-intel the old code path didn't support `@catch` for UBF which was a deliberate choice at the time for the use cases we considered back then. Now that we are considering different use cases, we need to enable `@catch` for control.

In case we get some other UBF use cases in the future that require different behavior, we can make it configurable by the decorator. This is a reversible change in that sense

metaflow/plugins/conda/conda_step_decorator.py
tuulos: @romain-intel same deal as in the above comment - the new use cases require `@conda` for control. We can change the behavior for new decorators if needed

metaflow/plugins/catch_decorator.py
romain-intel: It starts with run_id instead of flow_name. Just wondering if there is a reason it is not the full pathspec.

romain-intel (approved):
My last remaining small issue is the following: https://github.com/Netflix/metaflow/pull/827/files#r764346481. I do remember running into issues with this at some point so maybe something to clean up but otherwise looks fine to me. I haven't tested it internally but I don't anticipate any issues.

metaflow/plugins/aws/batch/batch_decorator.py
```diff
@@ -323,7 +323,7 @@ def _wait_for_mapper_tasks(self, flow, step_name):
         to terminate immediately, and possibly prematurely.
         """
         from metaflow import Step  # avoid circular dependency
-
+        TIMEOUT = 600
         last_completion_timeout = time.time() + 600  # wait for 10 minutes
```
romain-intel, line 327: I would use TIMEOUT here (instead of 600).

akyrola: ah sorry, fixed now.

metaflow/plugins/aws/batch/batch_cli.py
akyrola: The solution in the main cli is weird, but it then has:

```
    if ubf_context == "none":
        ubf_context = None
```
Since this CLI is always called by the metaflow itself (never by the user), I don't see why we should support "none" string input. And if that gets passed from somewhere, this will fail gracefully.

And as this PR did not actually change this (the None was the sole argument option before), I suggest leaving this as is. Follow-up PR to clean up the main CLI can be done if wanted.

romain-intel: Sounds ok. I think it was due to something click related which wasn't happy with multiple types (or something). As long as the result is consistent, should be fine.

PR #828, opened 2021-11-18
Allow CLI args that are falsy but not False or None

### LINKED ISSUE #748: Batch decorator fails when `gpu=0` (as opposed to `gpu='0'`)
**I thought this was an R bug but I've been able to reproduce it in Python. Python example to follow.**

The below flow raises `invalid gpu value: (None) (should be 0 or greater)`:

```r
library(metaflow)

metaflow("testflow") %>%
  step(step = "start", 
       r_function = function(self) NULL, 
       next_step = "end") %>%
  step(step = "end", 
       batch(cpu = 1),
       r_function = function(self) NULL)

testflow %>% run()
```

Curiously when I print the (Python) form of this flow I see `gpu=0`. It's hard to imagine where this would be converted to a `NULL`/`None`.

What fixes this:
* changing `batch(cpu = 1)` to `decorator("batch", cpu = 1)`
* changing `batch(cpu = 1)` to `resources(cpu = 1)`

What doesn't fix this:
* removing the `cpu` argument entirely
* specifying `batch(cpu = 1, gpu = 0)`

This suggest that the problem sits with the new batch decorator shortcut functions I implemented a few months ago 😬

mdneuzerling:
closes #748

Currently, Metaflow CLI argument values that are "falsy" are ignored. The intention --- I believe --- is to omit CLI arguments that are `False`. This has the undesired consequence of also ignoring CLI arguments like `--gpu 0`, which arises from `@batch(gpu=0)`. The proposed solution explicitly ignores values in which `v is None or v is False`, allowing falsy values like `0` to go through.

Thank you to @romain-intel for help with this PR.

I couldn't find any existing tests around this functionality. I'd be happy to add a unit test if required, although I would be grateful for where this would fit into the Metaflow test suite.

### DIFF
--- a/metaflow/runtime.py
+++ b/metaflow/runtime.py
@@ -910,17 +910,22 @@ class CLIArgs(object):
         # TODO: Make one with dict_to_cli_options; see cli_args.py for more detail
         def _options(mapping):
             for k, v in mapping.items():
-                if v:
-                    # we need special handling for 'with' since it is a reserved
-                    # keyword in Python, so we call it 'decospecs' in click args
-                    if k == "decospecs":
-                        k = "with"
-                    k = k.replace("_", "-")
-                    v = v if isinstance(v, (list, tuple, set)) else [v]
-                    for value in v:
-                        yield "--%s" % k
-                        if not isinstance(value, bool):
-                            yield to_unicode(value)
+
+                # None or False arguments are ignored
+                # v needs to be explicitly False, not falsy, eg. 0 is an acceptable value
+                if v is None or v is False:
+                    continue
+
+                # we need special handling for 'with' since it is a reserved
+                # keyword in Python, so we call it 'decospecs' in click args
+                if k == "decospecs":
+                    k = "with"
+                k = k.replace("_", "-")
+                v = v if isinstance(v, (list, tuple, set)) else [v]
+                for value in v:
+                    yield "--%s" % k
+                    if not isinstance(value, bool):
+                        yield to_unicode(value)
 
         args = list(self.entrypoint)
         args.extend(_options(self.top_level_options))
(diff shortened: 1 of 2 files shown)


metaflow/runtime.py
```diff
@@ -910,17 +910,22 @@ def get_args(self):
...
+
+                # None or False arguments are ignored
+                # v needs to be explicitly False, not falsy, eg. 0 is an acceptable value
+                if v is None or v is False:
+                    continue
+
+                # we need special handling for 'with' since it is a reserved
+                # keyword in Python, so we call it 'decospecs' in click args
+                if k == "decospecs":
+                    k = "with"
```
romain-intel, line 922: Separately, there is actually no need for this clause here but this whole thing needs to be reworked with dict_to_cli_options in cli_args.py. For consistency, could you change that one as well?

romain-intel (changes requested):
Make same type of change in the other instance of this in cli_args.py. Otherwise LGTM

metaflow/runtime.py
mdneuzerling: I'm not sure I understand. Should we be unifying everything to use the `dict_to_cli_options` function in `utils.py`?

romain-intel: Longer term yes, for now, there is a function that looks almost identical in `cli_args.py` and I was suggesting to make the same change there.

romain-intel:
LGTM. Please reformat properly and we can ship :).

mdneuzerling:
Ahh I forgot that y'all use `black`! Sorry about that. I've formatted `runtime.py` and `cli_args.py`, but I haven't checked any unchanged file.

romain-intel:
It's new (our use of black :)). Seems like the validator is happy now :).

romain-intel (approved):
LGTM

PR #830, opened 2021-11-20
Fixes #589: validation logic to capture reserved keywords

### LINKED ISSUE #589: Add validation logic to capture reserved names for parameters
Some parameter names are disallowed in Metaflow - `with`, `max-num-splits`, `max-workers` etc. We should throw a nice error if we detect any of these in the user's flow.

rushic24:
Screenshot from 2021-11-20 11-28-18
Screenshot from 2021-11-20 11-27-45


metaflow/parameters.py
```diff
@@ -165,12 +165,13 @@ def __init__(self, name, **kwargs):
         self.kwargs = kwargs
         # TODO: check that the type is one of the supported types
         param_type = self.kwargs["type"] = self._get_type(kwargs)
+        reserved_params = ['params', 'with', 'max-num-splits', 'max-workers']
```
savingoyal, line 168: Other values would be `tag`, `run-id-file` and `namespace`.

savingoyal (approved):

PR #832, opened 2021-11-22
Remove default use of repo.anaconda.com

romain-intel:
Switch to miniforge instead of the conda executable from anaconda
to avoid any possible licensing issues given the new TOS.


metaflow/plugins/conda/batch_bootstrap.py
```diff
@@ -49,11 +49,8 @@ def download_conda_packages(flow_name, env_id):
 def install_conda_environment(env_id, packages):
     args = [
         "if ! type conda  >/dev/null 2>&1; \
-            then wget --no-check-certificate https://repo.anaconda.com/pkgs/misc/conda-execs/conda-latest-linux-64.exe -O conda >/dev/null 2>&1; \
-            chmod +x conda; \
-            export PATH=$PATH:{0}; fi".format(
-            os.getcwd()
-        ),
+            then wget --no-check-certificate https://github.com/conda-forge/miniforge/releases/latest/download/Mambaforge-pypy3-Linux-x86_64.sh -O Miniforge3.sh >/dev/null 2>&1; \
```
oavdeev, line 52: Why use pypy and mamba instead of a more "vanilla" `Miniforge3-Linux-x86_64` distribution flavor here?

romain-intel: Whoops. Sorry, that's a mistake. It should just be the more generic one. I think I had scrolled down too fast and didn't see the first part. You can change it to the more generic one.

metaflow/plugins/conda/batch_bootstrap.py
```diff
@@ -49,11 +49,8 @@ def download_conda_packages(flow_name, env_id):
 def install_conda_environment(env_id, packages):
     args = [
         "if ! type conda  >/dev/null 2>&1; \
-            then wget --no-check-certificate https://repo.anaconda.com/pkgs/misc/conda-execs/conda-latest-linux-64.exe -O conda >/dev/null 2>&1; \
-            chmod +x conda; \
-            export PATH=$PATH:{0}; fi".format(
-            os.getcwd()
-        ),
+            then wget --no-check-certificate https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-Linux-x86_64.sh -O Miniforge3.sh >/dev/null 2>&1; \
+            bash ./Miniforge3.sh -b; ln -sf /root/miniforge3/bin/conda /usr/local/bin/conda ; fi",
```
savingoyal, line 53: @oavdeev Would we be able to always execute `ln -sf /root/miniforge3/bin/conda /usr/local/bin/conda`?

savingoyal: Also, it seems that Miniforge3 brings in Python3.9 with it as well, which is pure bloat for us. Should we look into instead using micromamba `wget -qO- https://micromamba.snakepit.net/api/micromamba/linux-64/latest | tar -xvj bin/conda`

oavdeev: For `ln -sf`, it prob won't work in a non-root container, but the previous version wouldn't work either.

No opinion on micromamba, I'd just pick whatever distribution seems to be the most mainstream so we don't have to yank it out again in a year.

oavdeev: I gave micromamba a try, unfortunately it looks like it wouldn't work as a drop in replacement. For one thing, it doesn't seem to support `CONDA_JSON` flag used [here](https://github.com/Netflix/metaflow/blob/switch-to-miniforge/metaflow/plugins/conda/conda.py#L174:L174) (i didn't look further)

oavdeev: And I did a little more investigating on non-root, it is a bit finicky. It seems to generate conda entrypoint differently when installer is run as non-root which may cause it to use the wrong python interpreter. Someone would need to look further into that; i'd propose we do this in as a separate ticket and PR.

oavdeev (approved):

romain-intel:
I guess we need to specify MacOS for Darwin @oavdeev ?

romain-intel:
The following test fails: `2021-12-22T08:22:13.1086300Z [pid 2956] ### test 'ResumeForeachInnerTest' graph 'nested-foreach' in context python3-all-local ###`. Not sure why. I'll try re-triggering to see if it was a fluke.

PR #833, opened 2021-11-22
Address another issue with load_artifacts

### LINKED ISSUE #819: Datastore save/load blobs/data relies on order
In #817, an issue was fixed that causes another reliance on the order in which items are returned from underlying functions. This order is also present in the `save_blobs` method in the `content_addressed_store.py` file. It may be worth investigating if it is possible to make this more robust and generally not rely on order from underlying functions.

romain-intel:
When an artifact is present in the blob cache, load_blobs would return
things out of order and load_artifacts would therefore return the wrong
blob for the artifact name.

This patch hopefully addresses this issue fully and also adds a small
optimization whereas blobs are not requested multiple times if they are
identical.

Fixes #819 .

### DIFF
--- a/metaflow/datastore/content_addressed_store.py
+++ b/metaflow/datastore/content_addressed_store.py
@@ -118,8 +118,8 @@ class ContentAddressedStore(object):
 
         Returns
         -------
-        Returns an iterator of (string, bytes) tuples; the iterator will return the keys
-        in the same order as the input argument.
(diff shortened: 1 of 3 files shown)


akyrola (approved):
LGTM.

savingoyal (approved):

savingoyal (approved):

Issue #497, opened 2021-04-28
Support for alternate conda package dependency resolver


jamesbsilva:
I wanted to open this as a possible feature/enhancement. Solving for the environment can take some time in running flows and was wondering if support for alternate dependency resolvers like https://github.com/mamba-org/mamba is planned

The Proposal
Support for alternate dependancy resolver like  https://github.com/mamba-org/mamba to be able to speed up the part of running flows that requires resolving dependencies

Why
Iterate through flows faster

savingoyal:
@jamesbsilva Yes, indeed a very timely issue. I have been wanting to carve out some time and test swapping out conda with mamba.

savingoyal:
Also, linking #24 here.

savingoyal:
Mamba as a conda package resolver is now available - you can set the environment variable (or add it into your `~/.metaflowconfig`) - `METAFLOW_CONDA_DEPENDENCY_RESOLVER=mamba` to use mamba instead of conda. The mamba executable needs to be in your `$PATH`.

PR #836, opened 2021-11-23
Use mamba where possible

jimgoo:
#497 - uses mamba in place of conda where possible to speed things up


savingoyal:
@jimgoo Thanks for the PR!~ We have been *very interested* in a potential mamba integration. Have you tested the PR on AWS Batch and Step Functions?

savingoyal:
@romain-intel FYI

jimgoo:
@savingoyal Currently mamba runs locally and does the heavy lifting for conda decorator environment solving. It's working on AWS batch even though mamba isn't getting installed on the remote instance. Presumably this is because the python libraries uploaded to S3 are the same no matter if they were downloaded with conda or mamba. I'm still learning my way around metaflow and haven't tried the AWS step functions yet. Mamba is really great though, it's fast and just works.

PR #840, opened 2021-11-25
Implementation for triggering a SFN execution from the result of a previous SFN execution

### LINKED ISSUE #837: Scheduling Flows to run after successful execution of a previous Flow
Hi All,

One feature I've been experimenting with is having a Flow trigger (via Step Functions) after the successful execution of another Flow. This has allowed me to abstract out data cleaning code into the parent Flow, then have child flows consume this data.

I think this is a common pattern within most workflows, and I've got a base implementation to show off the feature. All I do is use Eventbridge to instead of creating a Schedule Rule, create a Rule which listens for the SUCCEEDED message from SFN for a specific SFN ARN.

How can I push my local branch so it can be tried out?

Thanks,
Robbie

iiAnderson:
An implementation to allow SFN Executions to be triggered from the finishing state of another SFN execution. I have been using this to trigger multiple consumer Flows when a specific Producer flow finishes. For example:

Hourly -> DataCleaningFlow 
DataCleaningFlow -> Analysis1
DataCleaningFlow -> Analysis2

This ensures that when Analysis1/2 run, they will be able to obtain the data from the DataCleaningFlow, no need to guess at when it might end.

To implement this, I've modified the @schedule annotation to take a flow_name. This could be extended to add multiple flows and wait for their state (currently only support SUCCEEDED from SFN).

Closes #837

### DIFF
--- a/metaflow/plugins/aws/step_functions/schedule_decorator.py
+++ b/metaflow/plugins/aws/step_functions/schedule_decorator.py
@@ -1,22 +1,73 @@
 from metaflow.decorators import FlowDecorator
+from .event_bridge_client import EventBridgeClient
+from .step_functions_client import StepFunctionsClient
+class ScheduleDefinition():
 
+    def __init__(self, schedule_obj):
+        pass
+
+    def generate_eventbridge_rule(self):
+        pass
+
+class CronScheduleDefinition(ScheduleDefinition):
+
+    def __init__(self, schedule_obj):
+        self.cron = schedule_obj
+
+    def generate_eventbridge_rule(self, name, state_machine_arn, sfn_access_role):
+        return (
+            EventBridgeClient(name)
+                .cron(self.cron)
+                .role_arn(sfn_access_role)
+                .state_machine_arn(state_machine_arn)
+                .schedule()
+        )
+
+class FlowTriggerScheduleDefinition(ScheduleDefinition):
+
+    def __init__(self, schedule_obj):
+        self.flow_name = schedule_obj
+
+    def _get_state_machine_arn(self):
+
+        flow = StepFunctionsClient().search(self.flow_name)
+
+        if flow:
+            return flow.get('stateMachineArn', None)
+        else:
+            raise Exception("Dependant State Machine Found")
+
+    def generate_eventbridge_rule(self, name, state_machine_arn, sfn_access_role):
+
+        dependant_state_machine_arn = self._get_state_machine_arn()
+
+        return (
+            EventBridgeClient(name)
+                .dependent_state_machine_arn(dependant_state_machine_arn)
+                .role_arn(sfn_access_role)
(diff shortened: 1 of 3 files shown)


iiAnderson:
Syntax still requires formatting.

savingoyal:
@iiAnderson I am taking the next couple of weeks off. I can take a look at this once I am back. Would that work?

iiAnderson:
No rush @savingoyal, just a feature I've found useful!

npow:
Hi @savingoyal, any updates on this PR? This would be a really useful feature to have!

metaflow/plugins/aws/step_functions/schedule_decorator.py
```diff
@@ -1,22 +1,73 @@
...
+                .dependent_state_machine_arn(dependant_state_machine_arn)
+                .role_arn(sfn_access_role)
+                .state_machine_arn(state_machine_arn)
+                .schedule()
+        )
 
 class ScheduleDecorator(FlowDecorator):
     name = "schedule"
-    defaults = {"cron": None, "weekly": False, "daily": True, "hourly": False}
+    defaults = {"cron": None, "weekly": False, "daily": True, "hourly": False, "flow_name": None}
```
savingoyal, line 54: @iiAnderson - What's your opinion on instead of overriding `@schedule` - we introduce `@TriggerOnFinish`? We can make sure that `@TriggerOnFinish` obeys Metaflow namespace.

savingoyal: @npow - this could be potentially interesting for you as well.

iiAnderson: @savingoyal An interesting idea, I've been playing around with how we could tie all of the Events into EventBridge - so you could essentially implement a `@TriggerOn`, then select an event to trigger on.

When a flow finishes/fails/succeeds you could then push events to Eventbridge, which then triggers dependent flows.

I think that makes it more of a complete framework for event handling, rather than just a single "Succeeded" as it currently is. I don't know how it plays into other Compute environments though (as it's certainly possible for Batch).

npow: Probably out of scope for this PR, but we were thinking about triggering arbitrary resources on Step Function completion (eg. Glue Crawlers, Lambda functions, etc). Any thoughts on how this would fit into the design?

iiAnderson: If you built the system using Eventbridge this would be easy, as they all tie into it. We would just need to implement a definition language to define these events within a decorator in Metaflow. Designing this to be user-friendly would be challenging though.

savingoyal:
closing this PR since ideally it would be preferrable to have [an equivalent support for event triggering](https://docs.metaflow.org/production/event-triggering) instead

PR #845, opened 2021-11-30
support namespace packages in MetaflowPackage

pikulmar:


pikulmar:
Without this change, PEP-420 namespace packages cannot be used for `metaflow_extensions`.

benbemoh (approved):

metaflow/package.py
```diff
@@ -21,14 +21,9 @@ def __init__(self, flow, environment, echo, suffixes=DEFAULT_SUFFIXES_LIST):
...
-            self.metaflow_extensions_root = None
+            self.metaflow_extensions_path = []
         else:
-            self.metaflow_extensions_root = os.path.dirname(
-                metaflow_extensions.__file__
-            )
-            self.metaflow_extensions_addl_suffixes = getattr(
-                metaflow_extensions, "METAFLOW_EXTENSIONS_PACKAGE_SUFFIXES", None
-            )
+            self.metaflow_extensions_path = list(metaflow_extensions.__path__)
```
romain-intel, line 26: This seems fine but please re-add the PACKAGE_SUFFIXES portion to the else clause.

pikulmar: `METAFLOW_EXTENSIONS_PACKAGE_SUFFIXES` seems to be referenced at only this one location in the Metaflow code. Also, Google and GitHub searches for `METAFLOW_EXTENSIONS_PACKAGE_SUFFIXES` only return the removed code.

Is the feature associated with `METAFLOW_EXTENSIONS_PACKAGE_SUFFIXES` documented anywhere? Is this used / needed by anything? If yes, I could add a corresponding comment.

romain-intel: Good point but yes, it is used in the extensions mechanism. In our case, we actually package a `.so` with metaflow (for a not-as-yet released feature of Metaflow) and so we set this variable in our extension package.

romain-intel (commented):
Since I initially reviewed this, I am actually interested in giving full namespace package support to the metaflow_extensions package (so that you can have multiple contributions to plugins, etc). If it is OK with you, can we hold off on this for a bit (hopefully not too long) while I work on full support. If this is something that interests you as well and you would be willing to contribute, I am happy to provide a bit more context and detail how I am thinking of going about it.

metaflow/package.py
pikulmar: Ok, thanks for explaining this point. I will add re-introduce the corresponding lines of code.

pikulmar: Done in add3c40

pikulmar:
> Since I initially reviewed this, I am actually interested in giving full namespace package support to the metaflow_extensions package (so that you can have multiple contributions to plugins, etc). If it is OK with you, can we hold off on this for a bit (hopefully not too long) while I work on full support. If this is something that interests you as well and you would be willing to contribute, I am happy to provide a bit more context and detail how I am thinking of going about it.

I fully agree that complete namespace package support would be a valuable enhancement. In fact, one item on my to-do list was to open a corresponding Metaflow issue. However, I believe it might still be useful to merge this small fix first, because the problem addressed in this PR precludes the use of namespace packages, whereas other issues related to namespace packages can be worked around.

BTW, other parts of the Metaflow source code already appear to support namespace packages, see, e.g., https://github.com/Netflix/metaflow/blob/5186c6c5bba36d9e77077413ee2495dc79da3dca/metaflow/plugins/conda/conda_step_decorator.py#L269

So far, the only real (non-blocking) issue encountered with namespace packages, is that the code around https://github.com/Netflix/metaflow/blob/5186c6c5bba36d9e77077413ee2495dc79da3dca/metaflow/plugins/__init__.py#L55 does not seem to permit using a namespace package to provide `metaflow_extensions.plugins.STEP_DECORATORS`. Clearly, 3rd-party extension packages could ship their own [plugin discovery logic](https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/#using-namespace-packages), where one "root" package would populate `metaflow_extensions.plugins.STEP_DECORATORS`  based on the discovered 3rd-party extension plugins. However, this approach only works so long as all extension packages are provided by the same party / adhere to the same discovery protocol, which is why it would be strongly preferable for Metaflow to provide a standardized framework for this.

romain-intel:
Fair enough. It is true that this does at least allows the inclusion of the namespace packages. And yes, there was a partial support there for conda (I don't know why I just added it there and nowhere else -- probably coded on a different day).

In terms of full support, there are actually several areas where things won't work as easily, namely because I anticipate that there will be for example: `metaflow_extensions.my_company.X` and `metaflow_extensions.my_team.X` where `X` can be the several extension points for metaflow_extensions (and yes, I realize that all this isn't (yet) documented). Internally, we have `X` be one of `config`, `plugins`, `exceptions`, `datatools`, `toplevel` among other things so there are a bunch of these configurations that need to be "merged". You point out the one for plugins which is definitely one of the big ones and the current logic isn't setup to properly merge things. I did consider have sub namespace packages as well (ie: `metaflow_extensions.plugins.my_plugin` and `metaflow_extensions.plugins.my_other_plugin`) but this feels a little clunkier in a way particularly since you may need/want to have an extensions plugin that contributes to both the `plugins` portion and the `config` portion (it's do-able but feels harder).

I'm open to thoughts/ideas though. I would also be curious as to your current use case for `metaflow_extensions`. As I mentioned, it's not yet a fully supported feature but we do use it internally at Netflix so some version of it will survive :). I will also see about sharing the current template we have for metaflow_extensions to maybe make some of this a bit more concrete.

romain-intel:
Here is a repository that shows the possibilities for metaflow_extensions. I opened an issue there for namespace packages as well: https://github.com/Netflix/metaflow-extensions-template/issues/1

As a caveat (again), this is currently not a fully supported feature and is subject to change. Definitely happy to get feedback on it and on how it can be made more useful. This was instrumental in helping us internally at Netflix provide a metaflow package that contained integrations to our internal tools without having the user "notice" that they are using effectively two packages.

pikulmar:
@romain-intel Thanks for sharing your thoughts. I think the outlined design (`metaflow_extensions.my_scope` as a common root for `metaflow_extensions.my_scope.plugins` etc.) is indeed simpler.

I was not aware of the [template repository](https://github.com/Netflix/metaflow-extensions-template). Although I was able to get things to work by reading the source code, it might still be a good idea to add a link to the template repository in https://docs.metaflow.org/internals-of-metaflow/technical-overview#decorators. Our use-case is actually similar to https://github.com/Netflix/metaflow-extensions-template#providing-another-flow-or-step-level-decorator, hence my initial focus on `.plugins.STEP_DECORATORS`.

In terms of changes required, I have the feeling that supporting namespace packages in the way you suggested might not be too complicated to implement: Based on https://www.python.org/dev/peps/pep-0420/#differences-between-namespace-packages-and-regular-packages, namespace packages should be identifiable by checking for the absence of the `__file__` attribute. So, whenever `metaflow_extensions` has no `__file__` attribute, one could loop over its sub-packages and check for `metaflow_extensions.some_sub_package.plugins` etc. Of course, the various contents would need to be aggregated somehow (e.g., by concatenating the `STEP_DECORATORS` from the different sub-packages).

The above should work fine as long as users stick to _either_ no `metaflow_extensions` package, one non-namespace `metaflow_extensions` package, or one or more `metaflow_extensions` namespace packages with contents as outlined above; to avoid the caveats discussed in <https://discuss.python.org/t/mixing-namespace-and-non-namespace-packages-bad-practice/7536/9>.

romain-intel:
This is included in #892.

Issue #683, opened 2021-09-07
Confusing error from `metaflow.S3` if Metaflow is not configured properly


tuulos:
If Metaflow is not configured properly (e.g. no config file at all), The `S3DemoFlow` at  https://docs.metaflow.org/metaflow/data#store-and-load-objects-in-a-metaflow-flow fails with a confusing error. We could show a friendlier error

```2021-09-07 01:25:02.077 Workflow starting (run-id 1630977902074982):
2021-09-07 01:25:02.089 [1630977902074982/start/1 (pid 31904)] Task is starting.
2021-09-07 01:25:02.339 [1630977902074982/start/1 (pid 31904)] <flow S3DemoFlow step start> failed:
2021-09-07 01:25:02.343 [1630977902074982/start/1 (pid 31904)] Internal error
2021-09-07 01:25:02.344 [1630977902074982/start/1 (pid 31904)] Traceback (most recent call last):
2021-09-07 01:25:02.344 [1630977902074982/start/1 (pid 31904)] File "/home/ec2-user/metaflow/metaflow/cli.py", line 1007, in main
2021-09-07 01:25:02.344 [1630977902074982/start/1 (pid 31904)] start(auto_envvar_prefix='METAFLOW', obj=state)
2021-09-07 01:25:02.344 [1630977902074982/start/1 (pid 31904)] File "/home/ec2-user/.local/lib/python3.7/site-packages/click/core.py", line 1137, in __call__
2021-09-07 01:25:02.344 [1630977902074982/start/1 (pid 31904)] return self.main(args, kwargs)
2021-09-07 01:25:02.345 [1630977902074982/start/1 (pid 31904)] File "/home/ec2-user/.local/lib/python3.7/site-packages/click/core.py", line 1062, in main
2021-09-07 01:25:02.345 [1630977902074982/start/1 (pid 31904)] rv = self.invoke(ctx)
2021-09-07 01:25:02.345 [1630977902074982/start/1 (pid 31904)] File "/home/ec2-user/.local/lib/python3.7/site-packages/click/core.py", line 1668, in invoke
2021-09-07 01:25:02.345 [1630977902074982/start/1 (pid 31904)] return _process_result(sub_ctx.command.invoke(sub_ctx))
2021-09-07 01:25:02.345 [1630977902074982/start/1 (pid 31904)] File "/home/ec2-user/.local/lib/python3.7/site-packages/click/core.py", line 1404, in invoke
[...]

PR #853, opened 2021-12-07
Adding friendlier exception if no S3 config provided

exotikh3:
Adding code that tries to resolve this issue #683

Looking forward for any kind of feedback as this is my first contribution to Metaflow

### DIFF
--- a/metaflow/datatools/s3.py
+++ b/metaflow/datatools/s3.py
@@ -316,6 +316,12 @@ class S3(object):
                 bucket = parsed.netloc
             if not prefix:
                 prefix = parsed.path
+
+                if DATATOOLS_S3ROOT is None:
+                    raise MetaflowS3URLException(
+                        "No config or correct S3 root path provided when trying to use S3 storage"
+                    )
+
             if isinstance(run, FlowSpec):
                 if current.is_running_flow:
                     prefix = os.path.join(prefix, current.flow_name, current.run_id)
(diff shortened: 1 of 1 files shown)


metaflow/datatools/s3.py
```diff
@@ -316,6 +316,12 @@ def __init__(
                 bucket = parsed.netloc
             if not prefix:
                 prefix = parsed.path
+
+                if DATATOOLS_S3ROOT is None:
+                    raise MetaflowS3URLException(
+                        "No config or correct S3 root path provided when trying to use S3 storage"
+                    )
+
```
oavdeev, line 324: why only check this inside the `if run.. if not prefix..` ?

romain-intel: It is only used in the `if run` portion so IMHO, it could be checked there. I would also check it in `get_root_from_config` (used for `IncludeFile`).

savingoyal: @exotikh3 Can you address this comment so that we can proceed with the merge?

exotikh3:
@romain-intel please review latest changes

roofurmston:
I recently stumbled across this issue, so it seems like it still requires a fix. 

Given the PR is now pretty old (over 18 months) is there any objection to me starting a new one? @savingoyal That is ok with you?

saikonen:
closing as this got superseded by another PR which has been merged.

PR #861, opened 2021-12-11
Stacked @card PR : Base (Card Decorator, Card Datastore, Card CLI + Metaflow Plumbing) 

valayDave:
# Metaflow Cards ( Base Plumbing ) 

This PR introduces the base plumbing needed to create HTML visualization of metaflow `Task`'s via the `@card` decorator. 

## PR Merging Order 

This feature will be fully integrated when the Following stacked PRs get merged in the following order;
`master` <-merges into-- #861 (base PR ) <-merges into-- #862  <-merges into-- #863 <-merges into-- #864 <-merges into-- #866

Order of reviewing should be   #861 -> #862 -> #863 -> #864 -> #866

### Card Decorator
- Introduces the `@card` decorator that can be placed over a step 
- The decorator supports packaging custom modules named `metaflow_cards` into the code package. `metaflow_cards` are namespace packages that need to have a `CARD` attribute exposed by every submodule in `metaflow_cards`. The import logic can be found in at `metaflow/plugins/cards/card_modules/__init__.py`
- The Decorator calls the CLI module in a separate subprocess to create cards

### Card CLI 
- The cli exposes `create`,`view`, and `get` commands 
- Card creation only happens through the cli 
- To create a card a `MetaflowCard` class needs to be implemented. The returned string from this class's `render` function is stored as HTML. 
- The cli also sets the storage root of the card.

### Card Datastore. 
- The `CardDatastore` object helps store the actual cards to the underlying `S3` or `local` datastore. 
- The datastore stores `Task` cards in the following directory format : `$card_root/$flow_id/runs/$run_id/tasks/$task_id/cards/$card_name-$hash.html`

### Config Changes
- Adding `DATASTORE_CARD_S3ROOT` and `DATASTORE_CARD_LOCALROOT` environment variables


metaflow/__init__.py
```diff
@@ -212,6 +212,9 @@ def _can_handle_orig_module():
 # this auto-generates decorator functions from Decorator objects
 # in the top-level metaflow namespace
 _import_plugin_decorators(globals())
+# Setting card import for only python 3.4
```
romain-intel, line 215: This feels a little out of place; could it not go into `_import_plugin_decorators` or something like that? The mechanism for minimum version is an interesting one and we don't have to solve it now but it doesn't feel like it should belong here. If it is too complicated, we can leave it here but should open an issue to fix it.

metaflow/plugins/__init__.py
```diff
@@ -178,6 +184,10 @@ def _merge_lists(base, overrides, attr):
     "name",
 )
 
+# Cards
```
romain-intel, line 187: Strip and just put the single card for now. The import mechanism needs to be thought out a bit more IMHO so that it plays nice with metaflow_extensions. Since right now we just want to support a single card, I would just put `CARDS = [<the default card>]`

metaflow/plugins/__init__.py
```diff
@@ -11,6 +11,7 @@
     "MONITOR_SIDECARS": {},
     "AWS_CLIENT_PROVIDERS": [],
     "get_plugin_cli": lambda: [],
+    "CARDS": [],
```
romain-intel, line 14: Remove; unused.

metaflow/plugins/cards/card_cli.py
```diff
@@ -0,0 +1,388 @@
...
+    CardClassFoundException,
+    IncorrectCardArgsException,
+    UnrenderableCardException,
+    CardNotPresentException,
+)
+
+from .card_resolver import resolve_paths_from_task, resumed_info
+
+
+def serialize_flowgraph(flowgraph):
```
romain-intel, line 24: Add TODO to move to the artifact; I think that was the idea (but it hasn't been reviewed yet).

metaflow/plugins/cards/card_cli.py
```diff
@@ -0,0 +1,388 @@
...
+        }
+    return graph_dict
+
+
+def open_in_browser(card_path):
+    url = "file://" + os.path.abspath(card_path)
+    webbrowser.open(url)
+
+
+def resolve_card(ctx, pathspec, hash=None, type=None, follow_resumed=True):
```
romain-intel, line 42: Since we only have one card per task, remove hash and type. I would keep just pathspec and follow_resumed. Or, if you really want to keep them, put them last because I am not sure if that is how we will end up specifying them.

metaflow/plugins/cards/card_cli.py
```diff
@@ -0,0 +1,388 @@
...
+    """Resolves the card path based on the arguments provided. We allow identifier to be a pathspec or a id of card.
+
+    Args:
+        ctx : click context object
```
romain-intel, line 46: nit: space before : is extraneous.

metaflow/plugins/cards/card_cli.py
```diff
@@ -0,0 +1,388 @@
...
+        pathspec : pathspec
+        hash (optional): This is to specifically resolve the card via the hash. This is useful when there may be many card with same id or type for a pathspec.
+        type : type of card
+    Raises:
+        CardNotPresentException: No card could be found for the pathspec
+
+    Returns:
+        (card_paths, card_datastore, taskpathspec) : Tuple[List[str],CardDatastore,str]
```
romain-intel, line 54: Nit: space after comma is not :).

metaflow/plugins/cards/card_cli.py
```diff
@@ -0,0 +1,388 @@
...
+    flow_name = ctx.obj.flow.name
+    run_id, step_name, task_id = None, None, None
+    # what should be the args we expose
+    run_id, step_name, task_id = pathspec.split("/")
+    pathspec = "/".join([flow_name, run_id, step_name, task_id])
+    try:
+        task = Task(pathspec)
+    except MetaflowNamespaceMismatch as e:
+        # todo : What to do in such cases where there is a namespace mismatch
+        namespace(None)
```
romain-intel, line 70: why not just do namespace(None) initially? It seems like we do that anyways so no point doing it twice.

metaflow/plugins/cards/card_cli.py
```diff
@@ -0,0 +1,388 @@
...
+    if len(pathspec.split("/")) != 3:
+        raise CommandException(
+            msg="Expecting pathspec of form <runid>/<stepname>/<taskid>"
+        )
+
+    flow_name = ctx.obj.flow.name
+    run_id, step_name, task_id = None, None, None
+    # what should be the args we expose
+    run_id, step_name, task_id = pathspec.split("/")
+    pathspec = "/".join([flow_name, run_id, step_name, task_id])
```
romain-intel, line 65: You can probably just do: `pathspec = "/".join([flow_name, pathspec])`

metaflow/plugins/cards/card_datastore.py
```diff
@@ -0,0 +1,200 @@
...
+                        break  # We are no longer making upward progress
+                    current_path = new_path
+                    check_dir = os.path.join(
+                        current_path, DATASTORE_LOCAL_DIR, DATASTORE_CARD_SUFFIX
+                    )
+                result = orig_path
+
+            return result
+
+    def __init__(self, flow_datastore, run_id, step_name, task_id, path_spec=None):
```
romain-intel, line 68: nit: inconsistent path_spec vs pathspec. And why do we need both the run_id, step_name, task_id and pathspec. Isn't the info duplicated?

metaflow/plugins/cards/card_datastore.py
```diff
@@ -0,0 +1,200 @@
...
+from .exception import CardNotPresentException
+
+TEMP_DIR_NAME = "metaflow_card_cache"
+NUM_SHORT_HASH_CHARS = 5
+
+CardInfo = namedtuple("CardInfo", ["type", "hash"])
+
+
+def path_spec_resolver(pathspec):
+    run_id, step_name, task_id = None, None, None
```
romain-intel, line 28: nit: consider something like `*splits, *[None]*(4-len(splits))` for this whole function.

metaflow/plugins/cards/card_cli.py
```diff
@@ -0,0 +1,388 @@
...
+    )
+    @click.option(
+        "--type",
+        default=None,
+        show_default=True,
+        type=str,
+        help="Type of card being created",
+    )
+    @click.option(
+        "--dont-follow-resumed",
```
romain-intel, line 219: I would use the usual thing like `--no-pylint` and `--pylint` so something like `--no-follow-resume` and `--follow-resume`.

metaflow/plugins/cards/card_cli.py
```diff
@@ -0,0 +1,388 @@
...
+    )
+    metadata(setting_metadata)
+    # set the card root to the datastore according to the configuration.
+    root_pth = CardDatastore.get_storage_root(ctx.obj.flow_datastore._storage_impl.TYPE)
+    if root_pth is not None:
+        ctx.obj.flow_datastore._storage_impl.datastore_root = root_pth
+
+
+def card_read_options_and_arguments(func):
+    @click.option(
```
romain-intel, line 204: We can expose this but since there is only one card, maybe consider removing for now.

metaflow/plugins/cards/card_cli.py
```diff
@@ -0,0 +1,388 @@
...
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        return func(*args, **kwargs)
+
+    return wrapper
+
+
+def render_card(mf_card, task, timeout_value=None):
+    rendered_info = None
+    if timeout_value is None or timeout_value < 0:
```
romain-intel, line 233: do we want a default timeout of None then?

metaflow/plugins/cards/card_cli.py
```diff
@@ -0,0 +1,388 @@
...
+        ctx.obj.flow_datastore, runid, step_name, task_id, path_spec=full_pathspec
+    )
+
+    if len(filtered_cards) == 0 or type is None:
+        raise CardClassFoundException(type)
+
+    if len(filtered_cards) > 0:
+        filtered_card = filtered_cards[0]
+        ctx.obj.echo(
+            "Creating new card of type %s With timeout %s"
```
romain-intel, line 301: nit: "with"

metaflow/plugins/cards/card_decorator.py
```diff
@@ -0,0 +1,192 @@
...
+        self._environment = None
+        self._metadata = None
+        # todo : first allow multiple decorators with a step
+
+    def add_to_package(self):
+        return list(self._load_card_package())
+
+    def _load_card_package(self):
+        try:
+            import metaflow_cards
```
romain-intel, line 40: do we need this for now? If not, I would strip so we can come up with a unified story about adding cards/extensions and what not. I think there can be a good unified story here.

metaflow/plugins/cards/card_decorator.py
```diff
@@ -0,0 +1,192 @@
...
+        self, flow, graph, step_name, decorators, environment, flow_datastore, logger
+    ):
+        # We do this because of Py3 support JSONDecodeError and
+        # Py2 raises ValueError
+        # https://stackoverflow.com/questions/53355389/python-2-3-compatibility-issue-with-exception
+        try:
+            import json
+
+            RaisingError = json.decoder.JSONDecodeError
+        except AttributeError:  # Python 2
```
romain-intel, line 80: Is this necessary since cards are py 3.4+?

metaflow/plugins/cards/card_decorator.py
```diff
@@ -0,0 +1,192 @@
...
+            # the context of the main processs
+        }
+        # FlowDecorators can define their own top-level options. They are
+        # responsible for adding their own top-level options and values through
+        # the get_top_level_options() hook.
+        for deco in flow_decorators():
+            top_level_options.update(deco.get_top_level_options())
+        return list(self._options(top_level_options))
+
+    def _run_cards_subprocess(self, runspec):
```
romain-intel, line 156: Quick note here; I don't think it matters much but in a local mode this will effectively double the number of processes which may not be desirable. I don't think it is a huge issue but pointing it out.

metaflow/plugins/cards/card_modules/__init__.py
```diff
@@ -0,0 +1,97 @@
+import sys
```
romain-intel, line 1: As mentioned, I would leave this out entirely for this first version.

romain-intel (commented):
Left a few comments to further simplify particularly around the story of multiple cards.

metaflow/__init__.py
valayDave: One reason I did this is that cards are not python 2 compatible and GH started shouting test failures for py-2 because of card imports. Should I remove the module import here?

metaflow/plugins/cards/card_cli.py
valayDave: We can have many cards given that users can create an arbitrary number of cards for a task using the `create` cli command.  Keeping these args is necessary because type and hash are necessary filters when many cards are created (Even in future). I can change the order

metaflow/plugins/cards/card_cli.py
valayDave: Fixed this

metaflow/plugins/cards/card_cli.py
valayDave: Fixed.

metaflow/plugins/cards/card_cli.py
valayDave: can you elaborate ?

metaflow/plugins/cards/card_cli.py
valayDave: default is about 45 seconds. This is when the user explicitly gives `timeout=None` in the decorator.

metaflow/plugins/cards/card_datastore.py
valayDave: Fixing this.

metaflow/plugins/cards/card_cli.py
valayDave: Added FIXME tag

metaflow/plugins/__init__.py
valayDave: Done,

metaflow/__init__.py
romain-intel: I guess we can leave it but if we no longer support Py2, we can just avoid this but can clean this up later.

metaflow/plugins/cards/card_cli.py
romain-intel: If there is only one type of card, and assuming that generating the card for a step is stable, there should only be one hash ever available. It is fine to leave though since it won't be much advertised I guess.

metaflow/plugins/cards/card_cli.py
romain-intel: Ah, so the timeout is somewhere else? This function makes it look like the timeout is None by default.

romain-intel (approved):
I think we should be good to go.

metaflow/plugins/cards/card_cli.py
valayDave: Cards can be created from the cli using `card create` command. Cards also support `options` that can account for creating cards of a different hash altogether ( when created from CLI ). Due to this, we have the hash.

metaflow/plugins/cards/card_cli.py
valayDave: Yes timeout comes from the subprocess that invokes the cli create command.

PR #862, opened 2021-12-11
Stacked @card PR : Card Client 

valayDave:
# Metaflow Cards (Card Client) 

This PR is built on top of #861 . This PR introduces a metaflow client to `get_cards`; The client helps retrieve the already present cards in the datastore.  

### `Card`
- A container object that holds an individual card that has been stored. 

### `CardContainer`
- A `list` like object that holds all many `Card`s. 

## Usage Examples

```python
from metaflow import Task
# the place from where `get_cards` is imported will change later on 
from metaflow.cards import get_cards
taskspec = '<MY_TASK_SPEC>'
task = Task(taskspec)
card_iterator = get_cards(task)

# view card in browser
card_iterator[0].view()

# Get HTML of card
html =  card_iterator[0].get()
```


romain-intel (approved):
This looks fine to me.

PR #863, opened 2021-12-11
Stacked @card PR  : Supporting `DefaultCard` and `ErrorCard`

valayDave:
# Metaflow Cards ( `DefaultCard` and `ErrorCard`)

This PR is built on top of #862 . It introduces a default card that visualizes all artifacts and provides the metadata about a `Task`. 

### DefaultCard
- Card that holds the information of a task in a dictionary form; Artfiacts are `repr`'d before passing them down to HTML. The `TaskToDict` object helps convert Task object to JSON so that it can be passed to the default card HTML. 

## MetaflowCardComponents
- A class which helps build many subcomponents in the final HTML like lego blocks. 
- Introduces many components like `ArtifactComponent`, `DagComponent` , `TableComponent` etc. 

### CLI Changes
- Added an argument called`with_error_card` in `card_cli.create` that creates a `ErrorCard` if rendering of a card fails for any reason.


metaflow/plugins/__init__.py
```diff
@@ -186,8 +186,11 @@ def _merge_lists(base, overrides, attr):
 
 # Cards
 from .cards.card_modules import MF_CARDS_EXTERNAL_MODULES
+from .cards.card_modules.basic import DefaultCard, TaskSpecCard, ErrorCard
 
-CARDS = MF_CARDS_EXTERNAL_MODULES
+CARDS = _merge_lists(
```
romain-intel, line 191: Just put the list of cards you have for now.

metaflow/plugins/cards/card_cli.py
```diff
@@ -261,16 +261,24 @@ def render_card(mf_card, task, timeout_value=None):
     type=int,
     help="Maximum amount of time allowed to create card.",
 )
+@click.option(
+    "--with-error-card",
```
romain-intel, line 265: I would avoid `with` as it has other connotations. Maybe something like `--render-error-card` and `--no-render-error-card`?

metaflow/plugins/cards/card_decorator.py
```diff
@@ -44,16 +44,23 @@ def _load_card_package(self):
...
             # What if a file is too large and
             # gets tagged along the metaflow_cards
             # path; In such cases we can have huge tarballs
             # that get created;
             # Should we have package suffixes added over here?
             for path_tuple in self._walk(metaflow_cards_root):
-                # print(path_tuple)
                 yield path_tuple
 
+        for path_tuple in self._walk(card_modules_root):
```
romain-intel, line 60: Unsure this is needed. Isn't the whole metaflow code base included anyways? Is this because there are more suffixes? If that's the case, you can just add those as suffixes. There is an var that does that already.

romain-intel (commented):
Minor comments. I also didn't review the actual cards themselves.

metaflow/plugins/cards/card_decorator.py
valayDave: Can you point to that variable?

metaflow/plugins/cards/card_cli.py
valayDave: `--render-error-card` is a flag.

metaflow/plugins/cards/card_decorator.py
romain-intel: METAFLOW_EXTENSIONS_PACKAGE_SUFFIXES

valayDave: But this variable requires stuff to be present in the `self.metaflow_extensions_root` from which it will allow the `METAFLOW_EXTENSIONS_PACKAGE_SUFFIXES`; Currently we don't have that.

PR #864, opened 2021-12-11
Stacked @card PR : Added Tests for `@card`

valayDave:
# Metaflow Cards (Test Cases) 

This is built on top of #863. This PR adds test cases to the metaflow test suite for the `@card` decorator. 

## Changes to Test Suite 
- As Metaflow Cards introduce a client as well as CLI methods to retrieve card, This PR makes change the `CliCheck` and `MetadataCheck` objects with the `assert_card` and `get_card` functions. 
- The  test suite though doesn't have easy ways of retrieving taskids so the tests use the `artifact_dict` function to get the taskids needed to retrieve the cards.

### CardBasicTest
- Test to validate cards are created with different types of graph topologies and also ensure the client and cli work as intended

### CardTimeoutTest
- Test to validate the timeout argument in the card decorator works as intended. 
- 
### CardErrorTest
- Test to validate if cards error out then they don't affect the execution of a flow as a whole.


romain-intel (approved):
Didn't review or run tests but I suppose you did :).

PR #871, opened 2021-12-20
Pytorch Parallel Decorator and test

akyrola:
Follow-up for @parallel, with @pytorch_parallel that sets the environment for Pytorch's DDP (distributed data parallel).

Making this work requires small change to the parallel decorator so that the initilaiization code is run at the right time.

Added a test flow:

```
python  test/parallel/pytorch_parallel_test_flow.py --no-pylint  run
```

(--no-pylint is needed for torch)


savingoyal (approved):

metaflow/plugins/frameworks/pytorch.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    # Choose port depending on run id to reduce probability of collisions, unless
+    # provided by the user.
+    master_port = master_port or (51000 + abs(int(current.run_id)) % 10000)
+    os.environ["MASTER_PORT"] = str(master_port)
+    os.environ["MASTER_ADDR"] = current.parallel.main_ip
+    os.environ["NODE_RANK"] = str(current.parallel.node_index)
+    os.environ["WORLD_SIZE"] = str(current.parallel.num_nodes)
+    os.environ["NUM_NODES"] = str(current.parallel.num_nodes)
+    os.environ["PL_TORCH_DISTRIBUTED_BACKEND"] = "gloo"  # NCCL crashes on aws batch!
```
savingoyal, line 39: Just leaving a note here that we may need to provide the capability to the user to override this at a later point (for k8s etc.).

metaflow/plugins/frameworks/pytorch.py
```diff
@@ -0,0 +1,39 @@
...
+        setup_torch_distributed(self.attributes["master_port"])
+
+
+def setup_torch_distributed(master_port=None):
+    """
+    Set up environment variables for PyTorch's distributed (DDP).
+    """
+    # Choose port depending on run id to reduce probability of collisions, unless
+    # provided by the user.
+    master_port = master_port or (51000 + abs(int(current.run_id)) % 10000)
```
savingoyal, line 33: There is no requirement that run_id will be an int. It can be a string as well.

savingoyal (commented):
Tiny fix needed. Let's merge this right after the fix.

savingoyal:
To fix the test - can you add an `__init__.py` file to `plugins/frameworks/pytorch` folder?

test/core/metaflow_extensions/frameworks/pytorch_parallel_test_flow.py
```diff
@@ -0,0 +1,78 @@
+from metaflow import FlowSpec, step, batch, current, pytorch_parallel, Parameter
```
savingoyal, line 1: Can we move this test to `test/parallel/`?

savingoyal:
Thanks, let's merge this in as soon as the tests pass.

romain-intel:
The failures have to do with cards for some reason and that was for the 212 commit. It totally bombed for the later one (can't even download/build MF -- I will need to debug this, it has been happening for a bit but I don't think it is your issue). Most likely something on my end.

romain-intel:
ok, nvm, the total bomb was for the 212 commit because frameworks didn't have an init file (gah, that took me a while to figure out). Looking at why the cards failures now.

romain-intel:
OK, and the card timeout is due to that fix in the formatter that I didn't have yet internally. OK. Phew, I get it now. Let me fix all that and retrigger.

romain-intel:
well this is getting infuriating.

PR #872, opened 2021-12-20
Tensorflow parallel decorator

akyrola:
@tensorflow_parallel

Note that this is stacked on top of #871 and thus includes also its commits.

Tensorflow has more involved handshaking than pytorch. We use S3 files to exchange information of each node's address and port. In the future, this could be rewritten to use a simple tcp server to exchange the info.

Simple test flow:

```
 python test/core/metaflow_extensions/frameworks/tensorflow_parallel_test_flow.py --no-pylint run
```


akyrola:
Closing as leaving the project. Can be resurrected.

PR #876, opened 2021-12-20
`aws_retry`'s `S3_RETRY_COUNT` kept min at 1

valayDave:
When`METAFLOW_S3_RETRY_COUNT=0`  the `get_cards` functions fails because `CardDatastore._list_card_paths` calls `S3Storage.list_content`; This class uses `s3op.py` (which uses `s3util.py` ). These files use the `S3_RETRY_COUNT` property of `metaflow_config`. 

In all other files except for `s3util.py` `S3_RETRY_COUNT` is iterated over in the following way: `range(S3_RETRY_COUNT +1)`; Here is it is iterated over `range(S3_RETRY_COUNT)`. 

This seems to be a bug and this PR fixes it.

### DIFF
--- a/metaflow/datatools/s3util.py
+++ b/metaflow/datatools/s3util.py
@@ -30,7 +30,7 @@ def get_s3_client():
 def aws_retry(f):
     def retry_wrapper(self, *args, **kwargs):
         last_exc = None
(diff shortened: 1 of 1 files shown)


saikonen (approved):

romain-intel:
LGTM

PR #877, opened 2021-12-21
Checks embed URL param and adds it as a class

obgibson:
## To test

* `cd metaflow/plugins/cards/ui`
* `yarn dev`
* Go to `http://localhost:5000`
* Add an `embed` parameter to the URL

Screen Shot 2021-12-21 at 2 54 24 PM

Screen Shot 2021-12-21 at 2 54 34 PM


seethroughdev (changes requested):
can you wrap this in an if `window?.location` and default to false?

if we run headless tests, I'm not sure we can always assume window exists.

seethroughdev (approved):
looks good!

seethroughdev (approved):

savingoyal (approved):

PR #879, opened 2021-12-22
Multiple `@card` decorators for a step

valayDave:
## Changes in the PR 

This PR allows multiple `@card` decorators for a step. It modifies the `StepDecorator` with an `allow_multiple` attribute. This attribute is set to `False` by default. If set to `True` then multiple decorators of the same type are allowed for a step.


romain-intel:
Superseded by #893 ? If so, please close.

PR #883, opened 2022-01-06
Remove unused code in BatchJob.execute()

oavdeev:
`execute()` is only called [here](https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/aws/batch/batch.py#L321:L321), but that code calls `create_job()` right before that, which in turn calls `job_def()`, which populates `jobDefinition` in `self.payload`.

The unused code inside `if` has been broken for a while anyway: over the last few months, two more required parameters were added to `_register_job_definition()`, namely `num_parallel` and `host_volumes`. That's how I noticed it is not used -- pylance type checker highlighted it for me as an error.


savingoyal (approved):
LGTM! Let's ship it

PR #661, opened 2021-08-24
nits: sort imports, doc typos, minor cleanups

ryan-williams:
factored out of #612

PR #885, opened 2022-01-06
fix `black .` issue

ryan-williams:
Rebasing #612 and friends, `black .` is showing this unrelated change. I folded it into #661 but [the CI is failing there for a reason I'm not sure is due to my changes](https://github.com/Netflix/metaflow/pull/661#issuecomment-1006693763), so I'm submitting this separately partly to see if I see the same failure here.

### DIFF
--- a/test/core/tests/card_timeout.py
+++ b/test/core/tests/card_timeout.py
@@ -9,8 +9,7 @@ class CardTimeoutTest(MetaflowTest):
     """
 
     PRIORITY = 2
-
-    @tag('card(type="test_timeout_card",timeout=10,options=dict(timeout=20),save_errors=False)')
+    @tag('card(type="test_timeout_card",timeout=10,options=dict(timeout=20),save_errors=False)')  # fmt: skip
     @steps(0, ["start"])
     def step_start(self):
         from metaflow import current


ryan-williams:
[This GHA also failed, and says](https://github.com/Netflix/metaflow/runs/4728927281?check_suite_focus=true#step:11:72959):
```
This step has been truncated due to its large size. Download the full logs from the  menu once the workflow run has completed.
```
I downloaded the logs and do not see a clear message about the failure yet.

ryan-williams:
It seems like the change by `black .` messes up the generation of the `test_flow.py`:

```python
# -*- coding: utf-8 -*-
from metaflow import FlowSpec, step, Parameter, project, IncludeFile, JSONType, current, parallel
from metaflow_test import assert_equals, assert_exception, ExpectationFailed, is_resumed, ResumeFromHere, TestRetry
from metaflow import card

class CardTimeoutTestFlow(FlowSpec):
    @card(type="test_timeout_card",timeout=10,options=dict(timeout=20),save_errors=False)
    @step
    def start(self):
        'card(type="test_timeout_card",timeout=10,options=dict(timeout=20),save_errors=False)'

        ps(0, ["start"])
        step_start(self):
        from metaflow import current

        self.task = current.pathspec
        self.next(self.foreach_split)
    @step
    def foreach_split(self):
        self.arr = [1, 2, 3]
        pass
        self.next(self.foreach_inner_first, foreach="arr")
    @step
    def foreach_inner_first(self):
        pass
        self.next(self.foreach_inner_second)
    @step
    def foreach_inner_second(self):
        pass
        self.next(self.foreach_join)
    @step
    def foreach_join(self, inputs):
        pass
        self.next(self.end)
    @step
    def end(self):
        pass
if __name__ == '__main__':
    CardTimeoutTestFlow()
```

No idea what is going on there, changed this PR to just have `black` ignore this line

test/core/tests/card_timeout.py
```diff
@@ -10,6 +10,7 @@ class CardTimeoutTest(MetaflowTest):
 
     PRIORITY = 2
 
+    # fmt: off
```
savingoyal, line 13: Can you add the equivalent `# fmt: on` after L:14?

ryan-williams: ah, didn't realize that was necessary, sry

looks like `# fmt: skip` at end of line now works (https://github.com/psf/black/pull/1800) so I did that, lmk what you think

savingoyal: thanks!

savingoyal (approved):
LGTM

PR #891, opened 2022-01-13
Make playlist.py easier to understand by using the built-in csvreader

hamelsmu:
Using python's built in csv reader to make reading the csv file less verbose and confusing.  

cc:/ @savingoyal @tuulos


hamelsmu:
On further thought, I am going to close this PR, because I think this example could distract from the larger goal of explaining Metaflow by inadvertently nerd sniping data scientists to focus instead on this csv to dataframe construction.  

I'll bring this topic up in a different forum and see if there is a better way to go about this.  

Thank you 🙇🏽

PR #892, opened 2022-01-13
Ns package support

romain-intel:


metaflow/extension_support.py
```diff
@@ -0,0 +1,441 @@
...
+        )
+
+    # Figure out the per extension point order
+    for k, v in extension_points_to_pkg.items():
+        l = [v[pkg] for pkg in mf_pkg_list if pkg in v]
+        extension_points_to_pkg[k] = l
+    return mf_pkg_list, extension_points_to_pkg
+
+
+_all_packages, _pkgs_per_extension_point = _get_extension_packages()
```
valayDave, line 316: `_pkgs_per_extension_point` is currently a dict of the form : 
```python
{
    "plugins.cards":[
        MFExtPackage(package_name='metaflow-ext-testing', tl_package='somedevorg', config_module=None), 
    ]
}
```

Can we change this to : 

```python
{
    "plugins.cards":{
        "somedevorg":[ # "somedevorg" is a top level module
            MFExtPackage(package_name='metaflow-ext-testing', tl_package='somedevorg', config_module=None), 
            MFExtPackage(package_name='metaflow-ext-testing', tl_package='somedevorg2', config_module=None)
        ]
    }
}
```

### Why make this change? 

Conside the following directory structure : 


```
your_package/ # the name of this dir doesn't matter
├ setup.py
├ metaflow_ext/ # namespace package name (This will change to `metaflow_extensions`
│  └ organizationA/      # package to keep org level extensions seperate ; # dir name must match the package name in `setup.py`
│      ├__init__.py    # This will have an __init__.py file
│      └ plugins/ # NO __init__.py file
│        └ cards/ # NO __init__.py file # This is a namespace package. 
│           └ my_card_module/  # Name of card_module
│               └ __init__.py. # This is the __init__.py is required to recoginize `my_card_module` as a package
│               └ somerandomfile.py. # Some file as a part of the package. 
│  └ organizationB/      
│      ├__init__.py    
│      └ plugins/ 
│        └ cards/ 
│           └ my_card_module_2/  # Name of card_module
│               └ __init__.py. # This is the __init__.py is required to recoginize `my_card_module_2` as a package
│               └ somerandomfile_2.py. # Some file as a part of the package. 
.
```

When I install the `setup.py` here, the current method will only recognize the top level module `organizationA` and  not `organizationB`

metaflow/extension_support.py
```diff
@@ -0,0 +1,441 @@
...
+    # At this point, we have all the packages that contribute to EXT_PKG,
+    # we now check to see if there is an order to respect based on dependencies. We will
+    # return an ordered list that respects that order and is ordered alphabetically in
+    # case of ties.
+    pkg_to_reqs_count = {}
+    req_to_dep = {}
+    mf_ext_packages_set = set(mf_ext_packages)
+    for pkg_name in mf_ext_packages:
+        req_count = 0
+        req_pkgs = [x.split()[0] for x in metadata.requires(pkg_name)]
```
valayDave, line 270: `metadata.requires(pkg_name)` returns a `Distribution(pkg_name).requires` which can be None. Can we have : 
```python
distmd = metadata.requires(pkg_name)
if distmd:
    req_pkgs = [x.split()[0] for x in metadata.requires(pkg_name)]
    ....
    ...
```

metaflow/extension_support.py
```diff
@@ -0,0 +1,441 @@
...
+                                parts[-1] = parts[-1][:-3]  # Remove the .py
+                                config_module = ".".join(parts)
+
+                            cur_pkg = extension_points_to_pkg[
+                                _extension_points[idx]
+                            ].get(dist.metadata["Name"], None)
+                            if cur_pkg is not None:
+                                if (
+                                    config_module is not None
+                                    and cur_pkg.config_module is not None
```
valayDave, line 223: Can you elaborate about why we do this?

valayDave (commented):
I was testing this PR with the #897 and made the following changes to the `extension_support.py` because I was facing issues importing cards for a usecase. The suggestion left in the review followed such a usecase where multiple top level modules get installed via one `setup.py`.  You can see the diff [here](https://github.com/valayDave/metaflow/commit/70caf9218f74e6e3035af0ec9d48882b467e1859)

Few questions : 
1. Can we extend `extension_support.py` to also allow `metaflow_extensions` in PYTHONPATH ?
2. Can we change `metaflow_extensions` to `mfx` ?

metaflow/extension_support.py
```diff
@@ -0,0 +1,441 @@
...
+    # the package. One way would be to rely on the description and have that info there.
+    # Not sure of the use though so maybe we can skip for now.
+    mf_ext_packages = []
+    extension_points_to_pkg = defaultdict(dict)
+    for dist in metadata.distributions():
+        if any(
+            [pkg == EXT_PKG for pkg in (dist.read_text("top_level.txt") or "").split()]
+        ):
+            _ext_debug("Found extension package '%s'..." % dist.metadata["Name"])
+            mf_ext_packages.append(dist.metadata["Name"])
```
valayDave, line 193: There can be more than one top-level package in `dist`. See the example in the comments left earlier about `_pkgs_per_extension_point`

metaflow/extension_support.py
romain-intel: Fair point. I had not considered this setup. I will integrate your suggestion.

metaflow/extension_support.py
romain-intel: Good catch. Will fix.

metaflow/extension_support.py
romain-intel: This is just to catch the case of multiple config files.

romain-intel:
@valayDave : yes, PYTHONPATH support has been pushed. And no, I would prefer not to rename things to `mf_ext` (is there a reason for this? it was `metaflow_custom` before and we then settled on `metaflow_extensions` -- `mf_ext` feels really short).

savingoyal (approved):

PR #895, opened 2022-01-13
Stacked @card v1.2: Read cli changes for Supporting Multiple`@card`s

valayDave:
Stacked on #894 

## Supported CLI Variations for `card get`/`card view` 

```sh
# If there are no slashes detected then we assume it to be a stepname for the `latest` runid. Extracts card in the `Step.task` of the lastest run
python myflow.py card view stepname

# If there are slashes detected then we assume it to be of the following form
python myflow.py card view runid/stepname
python myflow.py card view runid/stepname/taskid

# Extracts card with id abc in the `Step.task` of the lastest run
python myflow.py card view stepname --id abc

# filters cards by id. 
python myflow.py card view runid/stepname --id abc
python myflow.py card view runid/stepname/taskid --id abc
```

### Why is this needed  
Because we are now supporting multiple cards we need list functionality in the CLI to ensure that the test suite can work as it should. Since the `get` /`view` commands are bound to resolve a single card, we need an unambiguous command that resolves all the metadata about all cards for a taskspec. 

### Usage
```sh
# Command Signature is similar to `get`/`view` commands. Provides the list response from `get`/`view`. 
python myflow.py card list stepname

# Prints the pathspec along with the metadata about the cards as a JSON dictionary
python myflow.py card list stepname --as-json
```
## TODO : 
1. [x] Added the `--id` option to `card view`/`card get`
2. [x] Allow CLI to have new variations of the `get`/`view` supported. 
3. [x]  changes in the `get_cards` function. 
4. [x] `card list` function


metaflow/plugins/cards/card_cli.py
```diff
@@ -42,12 +43,68 @@ def open_in_browser(card_path):
...
 def resolve_card(
     ctx,
     pathspec,
     follow_resumed=True,
     hash=None,
     type=None,
+    card_id=None,
+    no_echo=False,
 ):
     """Resolves the card path based on the arguments provided. We allow identifier to be a pathspec or a id of card.
```
valayDave, line 109: TODO : Need to remove the comment : `We allow identifier to be a pathspec or a id of card.`

metaflow/plugins/cards/card_cli.py
```diff
@@ -56,58 +113,50 @@ def resolve_card(
...
-            pathspec = origin_taskpathspec
-            ctx.obj.echo(
-                "Resolving card resumed from: %s" % origin_taskpathspec,
-                fg="green",
-            )
-        else:
-            ctx.obj.echo(print_str, fg="green")
-    else:
+            card_pathspec = origin_taskpathspec
+            print_str = ("Resolving card resumed from: %s" % origin_taskpathspec,)
```
valayDave, line 139: TODO : remove `,)` in `print_str`

metaflow/plugins/cards/card_cli.py
```diff
@@ -42,12 +43,68 @@ def open_in_browser(card_path):
...
+    # since pathspec can have many variations.
+    pthsplits = pathspec.split("/")
+    namespace(None)
+    task = None
+    run_id = None
+    resolving_from = "task_pathspec"
+    if len(pthsplits) == 1:
+        # This means stepname
+        resolving_from = "stepname"
+        latest_run = Flow(flow_name).latest_run
```
romain-intel, line 68: Do we want the latest run or latest_successful_run?

metaflow/plugins/cards/card_cli.py
```diff
@@ -42,12 +43,68 @@ def open_in_browser(card_path):
...
+        if latest_run is not None:
+            run_id = latest_run.pathspec
+            try:
+                task = latest_run[pathspec].task
```
romain-intel, line 72: Do we want to warn if there are multiple tasks?

metaflow/plugins/cards/card_cli.py
```diff
@@ -56,58 +113,50 @@ def resolve_card(
...
-    pathspec = "/".join([flow_name, pathspec])
-    # we set namespace to be none to avoid namespace mismatch error.
-    namespace(None)
-    task = Task(pathspec)
-    print_str = "Resolving card: %s" % pathspec
+    (
+        task,
+        run_id,
+        resolved_from,
+    ) = resolve_task_from_pathspec(flow_name, pathspec)
```
romain-intel, line 129: `run_id` and `resolved_from` do not seem to be used except to raise an exception. The logic in `resolve_task_from_pathspec` is fairly long and could probably be simplified if the exception could be simplified a bit. Further, based on the logic in that function, there are cases where when task is none, run_id is also none and doesn't contain a lot of useful information. Perhaps we can simplify and just say that the string given as pathspec does not allow us to resolve to a task. You could, if you wanted to, also give a warning about resolving to multiple tasks.

metaflow/plugins/cards/card_cli.py
```diff
@@ -131,31 +180,67 @@ def raise_timeout(signum, frame):
...
+            "Found %d card matching for your query..." % len(path_tuples), fg="green"
+        )
+    task_pathspec = "/".join(pathspec.split("/")[1:])
     card_list = []
-    for path_tuple in path_tuples:
-        card_list.append(path_tuple.filename)
+    for path_tuple, file_path in zip(path_tuples, card_paths):
+        full_pth = card_datastore.create_full_path(file_path)
+        cpr = """
+        Card Id : %s
```
romain-intel, line 216: Nit: no need for " " before ":"

metaflow/plugins/cards/card_cli.py
```diff
@@ -432,23 +589,82 @@ def get(
...
+@click.argument("pathspec", required=False)
+@card_read_options_and_arguments
+@click.option(
+    "--as-json",
+    default=False,
+    is_flag=True,
+    help="Print all available cards as a JSON object",
+)
+@click.pass_context
+def list(
```
romain-intel, line 633: It may make sense in these commands to respect the "--quiet" flag which should only print things that are easily machine parseable (at least that's the intent IIRC). This would impact the listing functions above to have a quiet mode in essence and maybe list things simply as opposed to prettily.

metaflow/plugins/cards/card_datastore.py
```diff
@@ -171,17 +170,19 @@ def _list_card_paths(self, card_type=None, card_hash=None):
...
                     continue
+            elif card_id is not None and card_info.id != card_id:
+                continue
 
             if task_card_path.is_file:
                 cards_found.append(card_path)
 
         return cards_found
 
+    def create_full_path(self, card_path):
```
romain-intel, line 183: This may be a stupid comment so ignore it if it is but I thought we had a separate value for card root (`METAFLOW_CARD_S3ROOT`). Granted, it defaults to the one for the datastore but shouldn't we use the actual value in case it isn't like the one for the datastore?

romain-intel: Given your reliance on the flow_datastore, it may be better to not leave this as an option to users (and maybe only allow them to configure the card sub-directory -- with restrictions that it should probably not be a valid python name to avoid conflicts with a flow).

metaflow/plugins/cards/card_cli.py
valayDave: I can refactor the function by raising the error from inside `resolve_task_from_pathspec`and also returning only a `pathspec`.

metaflow/plugins/cards/card_cli.py
valayDave: Currently, we respect `--quiet` in `list` unless it is called with `--as-json`; the same goes for `get`. If I call `card get stepname` then it will print the card in the `stdout`. But calling `card get stepname filename.html` will save that card to the file. The test uses `--quiet` so I need this functionality to print/pass information when calling CLI commands in CliCheck.

metaflow/plugins/cards/card_datastore.py
valayDave: I actually set the storage root [here](https://github.com/Netflix/metaflow/blob/b38c9aa25ef1a189dd9ba7a993533f1fafeaf4eb/metaflow/plugins/cards/card_cli.py#L340). That's why the path generally lands up correct even when `METAFLOW_CARD_S3ROOT` is set.

PR #897, opened 2022-01-13
Stacked @card v1.2 : Namespace Packages with `@card`

valayDave:
# Importing External Cards
Requires Merging #892 

Stacked on : #911 

## Summary of changes

- Using `metaflow_extensions` to import external cards
- Added import modules in `card_modules`
- Added hook in card decorators to add custom packages

## How to implement a custom module 
Create custom cards with the following directory structure. 
```
your_package/ # the name of this dir doesn't matter
├ setup.py
├ metaflow_extensions/ # namespace package name
│  └ organizationA/ # NO __init__.py file, This is a namespace package.
│      └ plugins/ # NO __init__.py file, This is a namespace package.
│        └ cards/ # NO __init__.py file, This is a namespace package. 
│           └ my_card_module/  # Name of card_module
│               └ __init__.py. # This is the __init__.py is required to recoginize `my_card_module` as a package
│               └ somerandomfile.py. # Some file as a part of the package. 
.
```

The `__init__.py` of the `metaflow_extensions.organizationA.plugins.cards.my_card_module`, requires a `CARDS` attribute which needs to be a `list` of objects inheriting `MetaflowCard` class. For Example, in the below `__init__.py` file exposes a `MetaflowCard` of `type` "y_card2". 

```python
from metaflow.cards import MetaflowCard

class YCard(MetaflowCard):
    type = "y_card2"

    ALLOW_USER_COMPONENTS = True

    def __init__(self, options={}, components=[], graph=None):
        self._components = components

    def render(self, task):
        return "I am Y card %s" % '\n'.join([comp for comp in self._components])

CARDS = [YCard]
```
The above card can be used in the step code as follows : 

```python
@card(type='y_card2')
@step
def train(self):
.....
```

Having this `metaflow_extensions` module present in the PYTHONPATH can also work the same way.


romain-intel (commented):
Will wait for this to be refactored using #892.

romain-intel:
This does not seem updated.

valayDave:
@romain-intel : Sorry had forgotten to sync the code. Can you check now ?

PR #898, opened 2022-01-13
Stacked @card v1.2 : Test cases for Multiple `@card`s

valayDave:
Stacked on #897 

# Test Suite Changes 
The `MetadataCheck` and the `CliCheck` objects contain `assert_card` and `get_card` functions. This PR leverages the `card list` command introduced in #895 . `card list` cli command is used by `list_cards` function in the`CliCheck`.  A similar function is added to `MetadataCheck`. Both these functions return `dict`s with the same structure. 

The `assert_card` function now also takes `card_id` and `card_hash` as arguments. This makes filtering a single card easily possible. 

# Test Cases
The test cases are derived from the changes we introduce in #894  . #894 introduces using components from userland i.e. `current.card`. The below test cases also test and validate the functioning of this user interface. 
1.  [x] `current.card.append`/`current.card.extend` Should work for:  
	```python
	@card(type='x',id='a')
	@card(type='x')
	```

	```python
	@card(type='x')
	```

	```python
	@card(type='x',id='a')
	```

	```python
	@card(type='x',id='a')
	@card(type='x')
	@card(type='x',customize=True)
	```

2. [x] `current.card` should not fail : 
	1. [x] `current.card.append` shouldn't fail under ambiguity of resolution
	2. [x] `current.card['myid']` shouldn't fail if `myid` doesn't exist
	3. [x] `current.card.get(type='nonexistingtype')` shouldn't fail. 
	4. [x] `current.card.append(SomeRandomType())` shouldn't fail. 
		
3.  [x] `current.card['myid']` should work for:
	```python
	@card(type='x',id='a')
	@card(type='x')
	@step
	def mystep(self):
		current.card['a'].append
	```
		
	```python
	@card(type='x',id='b') # b is non-editable
	@step
	def mystep(self):
		current.card['b'].append
	```

4. [x]  `current.card` should not fail on serialization when passing to subprocess. We should have graceful serialization failures which don't intrude in user space. 
5. [x] Tests for `view`/`get` input argument variations. 
6. [x] multiple cards should work`@card`.


romain-intel (commented):
Did not review

PR #900, opened 2022-01-14
Fix issue with R executions specifically due to capture_output_to_mflog

romain-intel:
The new capture_output_to_mflog was not friendly to env variables at the
head of the command line. This solves this issue by passing environment variables
separately.


romain-intel:
I haven't tested this on batch yet. I plan to test it when I get my env back up and running.

romain-intel:
Are we good to merge this one @oavdeev or @savingoyal ? It's been a while :).

savingoyal:
this may no longer be needed due to #979

PR #901, opened 2022-01-14
Fix @local @parallel and refactor task ids

akyrola:
1) Pass parallel_task_prefix from decorator to client so the task id logic is in one place only.
2) Now works with local metadata. The problem was _wait_for_mapper_tasks. There is now special version for local metadata, which uses S3 files to communicate finishing of the tasks. I could not make it work with the metadata object itself.
3) some small logging message changes for clarity.


akyrola:
Closing as leaving the project. Can be resurrected.

PR #905, opened 2022-01-14
Fail early step function create with @parallel.

akyrola:
As title. Now `step-functions create` fails.


savingoyal (approved):
LGTM! Let's merge as soon as tests pass.

PR #907, opened 2022-01-17
PyTorch Lightning support

akyrola:
This PR might be a bit controversial as it has quite a lot of plumbing to make PyTorch Lightning multiprocessing act nicely with Metaflow.

This PR adds support for multinode & multigpu/cpu PyTorch Lightning training. Lightning's way of launching subprocesses is problematic for Metaflow as it sniffs the current command line and just re-executes that and then runs that with different environment variables to distinguish the worker. This would lead to same task being relaunched multiple times.

Instread, we must ask users to call a PytorchHelper to spawn a trainer function in their separate module. That is then dispatched via a separate spawner script. 

PyTorch Lightning also has handy ways to specify checkpointing and TensorBoard logging. For that, a new module "plugins.frameworks.paths" is created. A naming convention to the trainer functions arguments allows user to easily pass dthese paths to the trainer function. See the test script:

```
python  test/parallel/pytorch_parallel_test_flow_lightning.py run
```


test/parallel/pytorch_parallel_test_flow_lightning.py
```diff
@@ -0,0 +1,60 @@
...
+
+    @pytorch_parallel
+    @step
+    def parallel_step(self):
+        """
+        Run a simple torch parallel program which learns to sum a vector and also
+        includes a simple parallel reduction to validate the parallel environment setup.
+        """
+
+        self.node_index = current.parallel.node_index
```
hamelsmu, line 32: Why is this line necessary?  Is this some kind of boilerplate you need to make this work?

test/parallel/pytorch_simple_reduce.py
```diff
@@ -0,0 +1,60 @@
...
+    print("result")
+    print(
+        model(torch.arange(10, dtype=torch.float32)), "expect close to", sum(range(10))
+    )
+    assert model.rank_reduction == sum(range(1, 1 + num_local_processes * num_nodes))
+    return model.rank_reduction
+
+
+if __name__ == "__main__":
+    train(3)
```
hamelsmu, line 60: Why is this argument to `train` hardcoded here?  How does `num_local_processes` in the `train` function here relate to the `num_local_processes` in `PyTorchHelper.run_trainer` ?

test/parallel/pytorch_parallel_test_flow_lightning.py
akyrola: Good catch, this is stray code, needs to be removed.

test/parallel/pytorch_simple_reduce.py
akyrola: This is just to enable running the script outside metaflow. Can remove this line as well to reduce confusion.

PR #916, opened 2022-01-24
K8s tolerations

hermesdt:
finishes k8s tolerations support + allow requesting nvidia gpus

### DIFF
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -1,6 +1,7 @@
 import inspect
 import os
 import sys
+import json
 import traceback
 from datetime import datetime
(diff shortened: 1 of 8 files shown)


savingoyal:
@hermesdt You should be able to now simply specify GPUs using `@kubernetes(gpu=1)` for running on top of NVIDIA GPUs. For AMD, you can use `@kubernetes(gpu=1, gpu_vendor='amd')` or override the default vendor in the config - `METAFLOW_KUBERNETES_GPU_VENDOR ` to `amd`. I will close out this PR.

PR #917, opened 2022-01-24
Store system tags as task metadata

savingoyal:
The PR in action -
https://olegui.outerbounds.xyz/RetryFlow/7394/start/135356?attempt=0&direction=asc&group=true&order=startTime&section=taskinfo&status=failed

### DIFF
--- a/metaflow/metadata/metadata.py
+++ b/metaflow/metadata/metadata.py
@@ -513,6 +513,46 @@ class MetadataProvider(object):
             tags.append("r_version:" + env["r_version_code"])
         return tags
 
+    def _register_system_metadata(self, run_id, step_name, task_id, attempt):
+        env = self._environment.get_environment_info()
+        entries = {}
+        entries["runtime"] = env["runtime"]
+        entries["python_version"] = env["python_version"]
+        entries["date"] = datetime.utcnow().strftime("%Y-%m-%d")
+        if env["metaflow_version"]:
+            entries["metaflow_version"] = env["metaflow_version"]
+        if "metaflow_r_version" in env:
+            entries["metaflow_r_version"] = env["metaflow_r_version"]
+        if "r_version_code" in env:
+            entries["r_version"] = env["r_version_code"]
+        
+        metadata = [
+                MetaDatum(
+                    field=k,
+                    value=v,
+                    type=k,
+                    tags=["attempt_id:{0}".format(attempt)],
+                )
+                for k, v in entries.items()
+            ]
+
+        code_sha = os.environ.get("METAFLOW_CODE_SHA")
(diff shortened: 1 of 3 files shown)


metaflow/metadata/metadata.py
```diff
@@ -513,8 +513,29 @@ def _tags(self):
...
         return tags
 
-    def _register_code_package_metadata(self, run_id, step_name, task_id, attempt):
-        metadata = []
+    def _register_system_metadata(self, run_id, step_name, task_id, attempt):
+        env = self._environment.get_environment_info()
+        entries = {}
+        entries["runtime"] = env["runtime"]
+        entries["python_version"] = env["python_version"]
+        entries["date"] = datetime.utcnow().strftime("%Y-%m-%d")
```
oavdeev, line 521: What is `date` used for? Since we're changing things anyway, maybe we could use a more descriptive name here, `task_start_utc_date` or something (or just drop it altogether). To me everything else here is self explanatory but date stands out 🤔

oavdeev: (otherwise lgtm)

savingoyal: Technically, we don't need the date here. It's available through the `created_at` attribute. I am in favor of dropping it altogether.

oavdeev: 👍 lets drop it then and merge this PR

savingoyal: done

metaflow/metadata/metadata.py
```diff
@@ -513,8 +513,28 @@ def _tags(self):
             tags.append("r_version:" + env["r_version_code"])
         return tags
 
-    def _register_code_package_metadata(self, run_id, step_name, task_id, attempt):
-        metadata = []
+    def _register_system_metadata(self, run_id, step_name, task_id, attempt):
+        env = self._environment.get_environment_info()
```
romain-intel, line 517: Could we, instead of doing this, rely on the information coming from `self._tags`. The reason for this is that internally I extend `_tags` and that would allow me to reuse this code as is (ie: not do much else than call this function) to get all the proper tags in place as metadata. As a transition (if we keep both for a while), we can just strip out the date here. The identity can also be stripped out though I suppose it would be nice to be added the run.

I am also curious as to why we want to keep things that are non-varying in the metadata (like the r information). My thought was that we could:
- keep all current system tags that are non-varying as system tags at the run level
- keep things that can vary across tasks/steps as metadata (like the python version for example)

romain-intel (commented):
A quick comment that will make it easier for me to integrate with the MLI service as well as a general comment on what to store in metadata and what not to store in metadata.

savingoyal:
#1039 supersedes this PR

PR #919, opened 2022-01-24
Fix compute_resources to take into account string values

romain-intel:


metaflow/plugins/aws/aws_utils.py
```diff
@@ -62,23 +64,41 @@ def compute_resource_attributes(decos, compute_deco, resource_defaults):
...
             return result
 
     # If there is no resources decorator, values from compute_deco override
     # the defaults.
-    for k, v in resource_defaults.items():
-        if compute_deco.attributes.get(k):
-            result[k] = str(compute_deco.attributes[k])
+    for k in resource_defaults:
+        if compute_deco.attributes.get(k) is not None:
+            result[k] = compute_deco.attributes[k] or "0"
```
oavdeev, line 102: Why did you drop `str(..)` here? Without it this function is no longer guaranteed to return a `Dict[str]`

PR #920, opened 2022-01-25
Fix issue in Step Functions integration with CLI defined decorators

savingoyal:


metaflow/plugins/aws/step_functions/step_functions.py
```diff
@@ -799,9 +797,6 @@ def _step_cli(self, node, paths, code_package_url, user_code_retries):
             "--retry-count $((AWS_BATCH_JOB_ATTEMPT-1))",
             "--max-user-code-retries %d" % user_code_retries,
             "--input-paths %s" % paths,
-            # Set decorator to batch to execute `task_*` hooks for batch
-            # decorator.
-            "--with=batch",
```
oavdeev, line 804: can you explain why `--batch` is no longer needed here?

savingoyal: It's added in step_functions_cli.py as well

PR #929, opened 2022-01-25
Vendor click as dependency for Metaflow

savingoyal:
Additionally - 
1. Removes coverage and click from Metaflow's pinned dependency requirements for `@conda` and floats other dependencies.
2. This PR will result in dependency re-resolution for all workflows that rely on `@conda`. We can handle it as a special case but the effort/complexity may not be worth it.
As a side effect this PR now ensures Metaflow @conda works with Py3.9


romain-intel:
This LGTM with the following comments:
- LICENSES need to be included (it's a requirement of the license) for both click and importlib_metadata
- As discussed, we should roll back importlib_metadata so it supports Py 3.6
- I wonder if the coverage test in formatter.py should exclude _vendor (is it present)?

savingoyal:
Added LICENSES for click, importlib_metadata and their dependencies. `vendor.py` automatically takes care of that - we would need to add more edge cases for LICENSE detection in case we add more dependencies into the mix. Rolled back both click and importlib_metadata.

PR #933, opened 2022-01-28
Default card JSON Bug Fixes. 

valayDave:
## Reason for the PR
- If an artifact has a large HTML string then the resultant JSON in the HTML file breaks the HTML interpreter in the browser. 

## Changes
- JSON object is given to JS framework in base64 format. The JS framework converts the base64 back to JSON during page load. 
- Built JS + CSS
- Added a `default_json` type to quickly retrieve the JSON object for a default card.


obgibson (approved):
Tested locally and it works

PR #937, opened 2022-02-01
Save logs of local parallel tasks to the task datastore.

akyrola:
When using @parallel in local mode (which is intended only for testing), the non-main tasks logs were not see in Metaflow UI as the logs were not store in the task datastore. Since we lauch the tasks manually, without the runtime, we need to do this slightly manually. 

Test:

```

 ...
 # in python repl
>>> 
>>> from metaflow import Task
>>> t = Task('ParallelTest/1643707565805761/parallel_step/1643707565805761-start-1_node_2')
>>> print(t.stdout)
parallel_step: node 2 finishing.

```

### DIFF
--- a/metaflow/plugins/parallel_decorator.py
+++ b/metaflow/plugins/parallel_decorator.py
@@ -109,8 +109,8 @@ def _local_multinode_control_task_step_func(flow, env_to_use, step_func, retry_c
         kwargs["retry_count"] = str(retry_count)
 
         cmd = cli_args.step_command(executable, script, step_name, step_kwargs=kwargs)
-        p = subprocess.Popen(cmd)
-        subprocesses.append(p)
+        p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        subprocesses.append((task_id, p))
 
     flow._control_mapper_tasks = [
         "%s/%s/%s" % (run_id, step_name, mapper_task_id)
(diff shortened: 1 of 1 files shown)


metaflow/plugins/parallel_decorator.py
```diff
@@ -122,8 +122,24 @@ def _local_multinode_control_task_step_func(flow, env_to_use, step_func, retry_c
...
+    errors = []
+    for task_id, p in subprocesses:
+        stdout, stderr = p.communicate()
+
+        print("--- Task {} stdout: ---".format(task_id))
+        print(stdout.decode("utf8"))
+        print("--- Task {} stderr: ---".format(task_id))
+        print(stderr.decode("utf8"))
+
+        task_datastore = flow_datastore.get_task_datastore(
```
romain-intel, line 136: Suggest using the `task_pre_step` hook to get the `task_datastore` (it's one of the parameters) and using it directly. This would also set the proper attempt ID for you (which may be incorrect here since you always use 0). To be consistent with runtime.py:829, I think we should also use RUNTIME_LOG_SOURCE.

metaflow/plugins/parallel_decorator.py
```diff
@@ -122,8 +122,24 @@ def _local_multinode_control_task_step_func(flow, env_to_use, step_func, retry_c
...
     step_func()
 
-    # join the subprocesses
-    for p in subprocesses:
-        p.wait()
+    # join the subprocesses and store logs
+    flow_datastore = flow._datastore.parent_datastore
+    errors = []
+    for task_id, p in subprocesses:
+        stdout, stderr = p.communicate()
```
romain-intel, line 129: This may not be a concern but this has the potential to deadlock if too much output is produced on the stdout/stderr pipes AND the parallel processes actually need to all make forward progress (ie: they all have their own internal synchronization mechanism).

romain-intel (commented):
A few comments on this. If this is purely for debugging, the comment regarding deadlock may not be as critical but raising it just in case.

metaflow/plugins/parallel_decorator.py
akyrola: I fixed the attempt.  Using the task_pre_step won't work in this case, because it is the launcher task that needs to access the logs of the launched subtasks.

metaflow/plugins/parallel_decorator.py
romain-intel: I don't think it fixes it. Suppose two subprocesses A and B and somehow, A and B have a sync in them (for eg: A and B both need to sync at a barrier). Suppose further that A and B produce enough output to saturate the PIPEs. You would first `communicate` with A but that would never return because `communicate` waits for A to complete but that would never happen because B can't make progress because it is stuck on a full pipe and A can't terminate because it is stuck on waiting for B.

PR #938, opened 2022-02-02
Reduce Metaflow overhead by optimizing content addressable store

akyrola:
I started to profile why Metaflow runs trivial flows so slowly, at least when S3 storage is used. Turns out that metaflow stores a lot of trivial objects in S3, like None, True, and False and empty list. These arise when artifacts are stored via the "content_addressable_store" which stores the objects in paths derived from their hash. This is slow because for each object, it needs to query S3 to see if the file exits (it usually does for these trivial objects), and this takes ~250ms on my laptop.

In addition, the caching is disabled by default. I added a simple cache for small objects with bounded size. Also, we should check for the cache when putting values (they are immutable anyway). This helps when Metaflow stores small objects like step names and other metadata, which is queried twice during the task (when loading the task and when storing the task).

In addition, one small optimization to prevent loading all attempts: we shoudl stop loading attempts when attemp N does not exist anymore. N+1 won't either. This hits S3 again.

With the local parallal test:
```
```

This reduced on my laptop the running time from 50s to 30s.   So still pretty slow, but an improvement.

(We should probably actually do something different for small objects and for htem just store the value instead of the hash... I might propose solution for this in a different PR.)

### DIFF
--- a/metaflow/datastore/content_addressed_store.py
+++ b/metaflow/datastore/content_addressed_store.py
@@ -34,8 +34,7 @@ class ContentAddressedStore(object):
         self._prefix = prefix
         self._storage_impl = storage_impl
         self.TYPE = self._storage_impl.TYPE
-        self._blob_cache = None
-
+        self._blob_cache = SmallObjectCache()
         # Map of very common short values. Store both v2 and v4 pickled versions...
         constants = [
             None,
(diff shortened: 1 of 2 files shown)


romain-intel:
Given #939 seems even better, are you OK closing this one in favor of that one or do you see a benefit to having both?

PR #939, opened 2022-02-02
Directly store small objects in the key instead of storing in S3

akyrola:
**Alternative to #938**
I started to profile why Metaflow runs trivial flows so slowly, at least when S3 storage is used. Turns out that metaflow stores a lot of trivial objects in S3, like None, True, and False and empty list. In addition, many small strings like step names are stored. These arise when artifacts are stored via the "content_addressable_store" which stores the objects in paths derived from their hash. This is slow because for each object, it needs to query S3 to see if the file exits (it usually does for these trivial objects), and this takes ~250ms on my laptop.

This PR fixes this by storing small objects directly in the content key. Keys starting with "!" have the blob itself as hex string in the rest of the key. Thus, most metadata will not hit S3.

In addition, one small optimization to prevent loading all attempts: we should stop loading attempts when attemp N does not exist anymore. N+1 won't either. This hits S3 again.

With the local parallal test:
```
```
This reduced on my laptop the running time from 50s to 20s. So still pretty slow, but an improvement.

### DIFF
--- a/metaflow/datastore/task_datastore.py
+++ b/metaflow/datastore/task_datastore.py
@@ -147,6 +147,9 @@ class TaskDataStore(object):
                     )
                     if self.has_metadata(check_meta, add_attempt=False):
                         max_attempt = i
+                    else:
+                        # We can assume that if attempt i does not exist, i + 1 does not either
+                        break
                 if self._attempt is None:
                     self._attempt = max_attempt
                 elif max_attempt is None or self._attempt > max_attempt:
--- a/metaflow/datastore/content_addressed_store.py
+++ b/metaflow/datastore/content_addressed_store.py
@@ -6,6 +6,7 @@ from io import BytesIO
 
 from ..exception import MetaflowInternalError
 from .exceptions import DataException
+import binascii
 
 
 class ContentAddressedStore(object):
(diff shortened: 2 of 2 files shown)


romain-intel:
*Love* this PR. I will be benchmarking it and reducing the size to 127 (because the SHA field in the metadata service is limited to 255 characters). Thanks for this!

akyrola:
Btw, if this is released, then the metaflow-service should be updated as well, right? Otherwise the artifacts in UI would not show up properly?

romain-intel:
IIRC, we switched the UI to rely on the client (exactly for things like this) so it should just work provided the UI uses the new version of metaflow. This does break forward compatibility of course but may be worth it depending on perf gains. I still need to run the benchmarks. Going to try to do that soon.

metaflow/datastore/task_datastore.py
```diff
@@ -147,6 +147,9 @@ def __init__(
                     )
                     if self.has_metadata(check_meta, add_attempt=False):
                         max_attempt = i
+                    else:
+                        # We can assume that if attempt i does not exist, i + 1 does not either
+                        break
```
tuulos, line 152: I don't this assumption is true: It can easily happen that we miss one attempt in the middle

akyrola:
Closing as leaving project -- can be resurrected by others.

PR #945, opened 2022-02-10
Card Import Refactors

valayDave:
## Purpose of PR
We were double importing packages unnecessarily and also double packaging card-related extensions files. This PR tries to clean up those changes. 

## Changes. 
- Removes functions that resolved paths. 
- Leaving packaging of extensions up to code managing extensions. 

## todos
1. There are slight issues with packaging cards with HTML and other assets. Need to document and test that approach.

### DIFF
--- a/metaflow/plugins/cards/card_decorator.py
+++ b/metaflow/plugins/cards/card_decorator.py
@@ -7,7 +7,6 @@ from metaflow.decorators import StepDecorator, flow_decorators
 from metaflow.current import current
 from metaflow.util import to_unicode
 from .component_serializer import CardComponentCollector, get_card_class
-from .card_modules import _get_external_card_package_paths
 
 
 # from metaflow import get_metadata
@@ -65,16 +64,6 @@ class CardDecorator(StepDecorator):
         ):
             file_path, arcname = path_tuple
             yield (file_path, os.path.join("metaflow", "plugins", "cards", arcname))
-
-        external_card_pth_generator = _get_external_card_package_paths()
-        if external_card_pth_generator is None:
-            return
-        for module_pth, parent_arcname in external_card_pth_generator:
-            # `_get_card_package_paths` is a generator which yields
-            # path to the module and its relative arcname in the metaflow-extensions package.
-            for file_pth, rel_path in self._walk(module_pth, prefix_root=True):
-                arcname = os.path.join(parent_arcname, rel_path)
-                yield (file_pth, arcname)
 
     def _walk(self, root, filter_extensions=[], prefix_root=False):
         root = to_unicode(root)  # handle files/folder with non ascii chars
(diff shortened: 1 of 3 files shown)


metaflow/plugins/cards/card_modules/__init__.py
```diff
@@ -14,53 +16,39 @@ def iter_namespace(ns_pkg):
...
+    # Caching card related modules.
+    global _CARD_MODULES
+    if len(_CARD_MODULES) > 0:
+        return _CARD_MODULES
     for m in get_modules("plugins.cards"):
         # Iterate submodules of metaflow_extensions.X.plugins.cards
         # For example metaflow_extensions.X.plugins.cards.monitoring
         card_packages = []
-        for fndx, card_mod, ispkg_c in iter_namespace(m.module):
+        if getattr(m, "__path__", None):
```
romain-intel, line 35: There are three cases we want to support (I think):
- your current system where you have a namespace package and then sub packages that have `__init__.py`
- a namespace package support with `mfextinit_X.py`
- a regular package support

In the last two cases, you can just do the `append(m)` with no issues. You can check to see if you have `__file__` as an attribute of `m.module` (not `m`). I would then put an `else` and do the `for` loop.

metaflow/plugins/cards/card_decorator.py
```diff
@@ -66,16 +65,6 @@ def _load_card_package(self):
...
-        if external_card_pth_generator is None:
-            return
-        for module_pth, parent_arcname in external_card_pth_generator:
-            # `_get_card_package_paths` is a generator which yields
-            # path to the module and its relative arcname in the metaflow-extensions package.
-            for file_pth, rel_path in self._walk(module_pth, prefix_root=True):
-                arcname = os.path.join(parent_arcname, rel_path)
-                yield (file_pth, arcname)
-
     def _walk(self, root, filter_extensions=[], prefix_root=False):
```
romain-intel, line 68: You should be able to remove all of `add_to_package`, _load_card_package` and `_walk`.

romain-intel (commented):
Minor comments but lgtm.

metaflow/plugins/cards/card_decorator.py
```diff
@@ -66,16 +65,6 @@ def _load_card_package(self):
             file_path, arcname = path_tuple
             yield (file_path, os.path.join("metaflow", "plugins", "cards", arcname))
```
valayDave, line 66: @romain-intel : you mean these too right ?

romain-intel: Yep, I think all packaging should now be taken care of (the metaflow packaging includes the .html, .js, .css by default). I think going forward, we should probably evolve the `add_to_package` function to return the extensions that it wants packaged or something like that.

valayDave:
Closing PR because #973 addresses this

PR #946, opened 2022-02-10
Python Multi Version Test Suite

valayDave:
# Changes 
- Ported changes to GH action tests from #940 
- Python3.5 is not playing nice with bytes-based JSON loading. Throws an error like [TypeError: the JSON object must be str, not 'bytes'](https://stackoverflow.com/questions/42683478/typeerror-the-json-object-must-be-str-not-bytes). This error doesn't persist in py3.6+. This resulted in minor changes to the way we read task metadata in task_datastore.py and flow_datastore.py
- The tests in the `publish.yml` are refactored to come from `test.yml`. We reuse the jobs in the Test gh action. Workflows being reused need to have [on.workflow_call](https://docs.github.com/en/actions/using-workflows/reusing-workflows#example-reusable-workflow). This helps reuse [it locally](https://github.blog/changelog/2022-01-25-github-actions-reusable-workflows-can-be-referenced-locally/).

### DIFF
--- a/metaflow/datastore/flow_datastore.py
+++ b/metaflow/datastore/flow_datastore.py
@@ -1,6 +1,5 @@
 import itertools
 import json
-import sys
 
 from .. import metaflow_config
 
--- a/metaflow/datastore/flow_datastore.py
+++ b/metaflow/datastore/flow_datastore.py
@@ -159,9 +159,7 @@ class FlowDataStore(object):
                     elif fname == TaskDataStore.METADATA_DATA_SUFFIX:
                         # This somewhat breaks the abstraction since we are using
                         # load_bytes directly instead of load_metadata
-                        with open(path, "rb") as f:
-                            if sys.version_info < (3, 6):
-                                f = f.decode("utf-8")
+                        with open(path, encoding="utf-8") as f:
                             data_objs[(run, step, task, attempt)] = json.load(f)
         # We now figure out the latest attempt that started *and* finished.
         # Note that if an attempt started but didn't finish, we do *NOT* return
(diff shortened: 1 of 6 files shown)


oavdeev:
should we add Python 3.10 there too? its twice as popular as 3.5, if you go by pypi download numbers since the beginning of this year
image

savingoyal:
Yep, let's add 3.10 and 3.11. @valayDave we would have to install Rcpp from source for R3.5 and R3.6. It's not worth the hassle at the moment, so we can safely disable those tests for now.

metaflow/datastore/flow_datastore.py
```diff
@@ -1,5 +1,6 @@
 import itertools
 import json
+import sys
```
savingoyal, line 3: spurious import?

metaflow/datastore/flow_datastore.py
```diff
@@ -158,7 +159,7 @@ def get_latest_task_datastores(
                     elif fname == TaskDataStore.METADATA_DATA_SUFFIX:
                         # This somewhat breaks the abstraction since we are using
                         # load_bytes directly instead of load_metadata
-                        with open(path, "rb") as f:
+                        with open(path, encoding="utf-8") as f:
```
savingoyal, line 162: I am not familiar with this part of the code. @romain-intel is this the right thing to do here?

romain-intel: This looks ok yes. The data is stored with `save_metadata` which does a json.dumps followed by an encode("utf-8") so this should be reasonable to open as text with that encoding.

romain-intel (approved):
Minor comments. O/w LGTM. If addressed, this can go in.

valayDave:
Addressed all comments + nit fixes. Tests are passing. Seems ready to merge.

savingoyal (approved):
lgtm!

PR #955, opened 2022-02-15
Exporting graph in JSON format via `output-raw` cli command 

valayDave:
Added a flag `--json` flag in the `output-raw` command.


savingoyal (approved):
LGTM!

PR #958, opened 2022-02-16
follow symlinks in packaging

tuulos:
- files pointed by symlinks are now dereferenced by tar (no symlinks inside tar)
 - symlink cycles are detected
 - if no symlinks are present, the package should be identical to the previous versions


savingoyal (approved):
lgtm!

romain-intel (approved):

PR #959, opened 2022-02-16
Extension packaging improvements

romain-intel:
This PR includes several smaller commits to improve packaging of metaflow extensions. To ease review, you can view each commit independently (I can make a stack too but they were small enough that hopefully this is legible).

### DIFF
--- a/metaflow/extension_support.py
+++ b/metaflow/extension_support.py
@@ -231,7 +231,7 @@ def _get_extension_packages():
     # other ways of specifying "load me after this if it exists" without depending on
     # the package. One way would be to rely on the description and have that info there.
     # Not sure of the use though so maybe we can skip for now.
-    mf_ext_packages = []
+    mf_ext_packages = set([])
     # Key: distribution name/full path to package
     # Value:
     #  Key: TL package name
@@ -242,6 +242,12 @@ def _get_extension_packages():
         if any(
             [pkg == EXT_PKG for pkg in (dist.read_text("top_level.txt") or "").split()]
         ):
+            if dist.metadata["Name"] in mf_ext_packages:
+                _ext_debug(
+                    "Ignoring duplicate package '%s' (duplicate paths in sys.path? (%s))"
+                    % (dist.metadata["Name"], str(sys.path))
+                )
+                continue
             _ext_debug("Found extension package '%s'..." % dist.metadata["Name"])
 
             # Remove the path from the paths to search. This is not 100% accurate because
@@ -249,7 +255,7 @@ def _get_extension_packages():
             # package but it is exceedingly unlikely so we are going to ignore this.
             all_paths.discard(dist.locate_file(EXT_PKG).as_posix())
 
-            mf_ext_packages.append(dist.metadata["Name"])
+            mf_ext_packages.add(dist.metadata["Name"])
 
             # At this point, we check to see what extension points this package
             # contributes to. This is to enable multiple namespace packages to contribute
@@ -350,12 +356,11 @@ def _get_extension_packages():
     # case of ties. We do not do any checks because we rely on pip to have done those.
     pkg_to_reqs_count = {}
     req_to_dep = {}
-    mf_ext_packages_set = set(mf_ext_packages)
     for pkg_name in mf_ext_packages:
         req_count = 0
         req_pkgs = [x.split()[0] for x in metadata.requires(pkg_name) or []]
         for req_pkg in req_pkgs:
-            if req_pkg in mf_ext_packages_set:
+            if req_pkg in mf_ext_packages:
                 req_count += 1
                 req_to_dep.setdefault(req_pkg, []).append(pkg_name)
         pkg_to_reqs_count[pkg_name] = req_count
--- a/metaflow/extension_support.py
+++ b/metaflow/extension_support.py
@@ -23,6 +23,8 @@ __all__ = (
 
 EXT_PKG = "metaflow_extensions"
 EXT_CONFIG_REGEXP = re.compile(r"^mfextinit_[a-zA-Z0-9_-]+\.py$")
+EXT_META_REGEXP = re.compile(r"^mfextmeta_[a-zA-Z0-9_-]+\.py$")
+EXT_EXCLUDE_SUFFIXES = [".pyc"]
 
 METAFLOW_DEBUG_EXT_MECHANISM = os.environ.get("METAFLOW_DEBUG_EXT", False)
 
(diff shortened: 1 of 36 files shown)


metaflow/extension_support.py
```diff
@@ -141,22 +173,22 @@ def multiload_globals(modules, dst_globals):
...
         )
         load_globals(m.module, dst_globals)
 
 
-_py_ver = sys.version_info[0] * 10 + sys.version_info[1]
+_py_ver = sys.version_info[0] * 100 + sys.version_info[1]
 _mfext_supported = False
 
-if _py_ver >= 34:
+if _py_ver >= 304:
```
oavdeev, line 187:
```suggestion
if sys.version_info[:2] >= (3, 4):
```

metaflow/extension_support.py
```diff
@@ -430,10 +569,13 @@ def _get_extension_packages():
                                 for x in map(EXT_CONFIG_REGEXP.match, files)
                                 if x is not None
                             ]
+                            if "__init__.py" in files:
+                                init_files.append("__init__.py")
```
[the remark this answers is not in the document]
valayDave: Ignore this comment.

metaflow/extension_support.py
```diff
@@ -445,56 +587,80 @@ def _get_extension_packages():
...
+                        continue
+                    have_null_config = True
+                else:
+                    package_path = package_name_to_path.get(pkg.package_name)
+                    if package_path:
+                        package_path = "at '%s'" % package_path
+                    else:
+                        package_path = "'%s'" % pkg.package_name
+                    raise RuntimeError(
+                        "Package %s does not define a configuration file for '%s'"
```
valayDave, line 657: An import on my local looks like the following 
Screen Shot 2022-02-17 at 9 29 04 AM

It ends up failing with this error. I added additional logs that print `k, pkg`
Screen Shot 2022-02-17 at 9 29 57 AM

romain-intel: Good catch. I will fix this. I am trying to be more pedantic about well formed packages but missed the case of "only define a card and no plugins". I'll fix that. It should actually not even catch the `plugins` extension if there is no other file but I see now that for PYTHONPATH extensions, I don't do that properly. Ooh, this is going to be fun.

metaflow/extension_support.py
```diff
@@ -525,111 +691,219 @@ def _attempt_load_module(module_name):
...
+
     if extension_module:
+        # We update the path to this module; this will be helpful if/when we need
+        # to package modules so we have the base path for them.
+        # Note that we cannot always get this initially because, in the case of Conda
+        # and/or remote execution, the path of the files will change (Conda does symlinks
+        # and creates intermediate directories and remote puts everything in one place)
+        if _all_packages[distribution_name]["root_path"] is None:
+            root_path = "/".join(
+                extension_module.__file__.split("/")[: -len(module_name.split("."))]
```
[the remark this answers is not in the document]
romain-intel: Ah yes, for the cards thing that will cause issue (because that can be a ns package which I will have to handle separately). Gah. I will add a special case for that one as well. In all other cases, there should be a single file (which will either be the `__init__.py` file or the `mfextinit_X.py`.

metaflow/extension_support.py
romain-intel: I'll make that change (and others related to it).

romain-intel (commented):
Thanks for reviewing. I will address the comments today.

metaflow/extension_support.py
```diff
@@ -445,56 +619,80 @@ def _get_extension_packages():
...
+        # In all other cases, we error out if we don't have a configuration file for the
+        # package (either a __init__.py of an explicit mfextinit_*.py)
+        final_list = []
+        have_null_config = False
+        for pkg in chain(*l):
+            if pkg.config_module is None:
+                if k == "plugins.cards":
+                    # This is allowed here but we only keep one
+                    if have_null_config:
+                        continue
```
valayDave, line 680: When I was testing, I noticed that we are swallowing all other card-related modules and explicitly keeping only one. Below is the state of variable `extension_points_to_pkg` at time of for loop. 
```python
defaultdict(None, {'plugins.cards': {'metaflow-card-uplot': {'card_uplot': MFExtPackage(package_name='metaflow-card-uplot', tl_package='card_uplot', config_module=None)}, 'metaflow-card-notebook': {'card_notebook': MFExtPackage(package_name='metaflow-card-notebook', tl_package='card_notebook', config_module=None)}, 'metaflow-card-html': {'card_html': MFExtPackage(package_name='metaflow-card-html', tl_package='card_html', config_module=None)}}})
```
Due to the `continue` we only take the first package and ignore the rest since we don't add them to `final_list`.

savingoyal (commented):
I just browsed through the conda related changes and add a couple of clarifying questions.

metaflow/plugins/conda/conda_step_decorator.py
```diff
@@ -250,30 +251,48 @@ def runtime_init(self, flow, graph, package, run_id):
         self.addl_paths = None
         os.symlink(path_to_metaflow, os.path.join(self.metaflow_home, "metaflow"))
 
-        # Also symlink the INFO version to properly propagate down version information
-        # from, for example, a step-function execution
+        # Symlink the INFO file as well to properly propagate down the Metaflow version
+        # if launching Titus from a meson run for example
```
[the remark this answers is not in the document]
romain-intel: Thanks. Will remove that and put when running remotely.

metaflow/plugins/conda/conda_step_decorator.py
```diff
@@ -250,30 +251,48 @@ def runtime_init(self, flow, graph, package, run_id):
...
         if os.path.isfile(path_to_info):
             os.symlink(path_to_info, os.path.join(self.metaflow_home, "INFO"))
-
-        # Do the same for metaflow_extensions
+        else:
+            # If there is no "INFO" file, we will actually create one in this new
+            # place because we won't be able to properly resolve the EXT_PKG extension
+            # since they may all be linked
```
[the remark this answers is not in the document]
romain-intel: I'll clarify; it refers to what is happening around line 286 in this file. The logic is that without the INFO file, we look around for the metaflow_extensions packages whereas when there is an INFO file, we use the information from there to figure out thee metaflow_extensions packages. In the case of a conda environment, we don't want to "look around" because now things are all installed as part of a custom path (and not distributions) which means we wouldn't be able to do the analysis we do on them (ordering particularly) so we therefore create the INFO file so that we can use it directly in the Conda environment.

PR #960, opened 2022-02-16
Pass the paths used by the external interpreter to the one launching …

romain-intel:
…the Env Escape server

We were currently not reflecting possible values passed using PYTHONPATH or
programmatically added to the interpreter launching the environment escape server.


savingoyal (approved):
LGTM! I don't think any of the escape hatch tests are part of the current CI test suite. Would you like to add them (or at least just open a GH issue noting that so that it can be done in the future)?

PR #962, opened 2022-02-17
Don't explicitly break py2 support

romain-intel:


savingoyal:
@romain-intel can you also enable Py2 tests?

PR #967, opened 2022-02-21
Fixing some hacky plumbing in card test suite

valayDave:
## Purpose
This PR helps resolve some hacky code added to the card test suite where we checked stdout to validate if a command threw a specific error. This was a brittle solution. Patching for something that is more conventional. By explicitly piping errors from stderr. 

## Changes 
- Added `--file` arg to `card list` cli
- Using files to get the information written by cli
- changes to test to use files instead of stdout
- Piping std errors to stderr to capture card not found errors.

### DIFF
--- a/test/core/metaflow_test/cli_check.py
+++ b/test/core/metaflow_test/cli_check.py
@@ -23,7 +23,7 @@ except:
 
 
 class CliCheck(MetaflowCheck):
-    def run_cli(self, args, capture_output=False, pipe_error_to_output=False):
+    def run_cli(self, args, capture_output=False):
         cmd = [sys.executable, "test_flow.py"]
 
         # remove --quiet from top level options to capture output from echo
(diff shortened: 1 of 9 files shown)


test/core/metaflow_test/cli_check.py
```diff
@@ -148,44 +147,46 @@ def list_cards(self, step, task, card_type=None):
...
+                card_type,
+            ]
+
+            if card_hash is not None:
+                cmd.extend(["--hash", card_hash])
+            if card_id is not None:
+                cmd.extend(["--id", card_id])
+
+            self.run_cli(cmd, capture_output=True)
+            with open(f.name, "r") as jsf:
```
romain-intel, line 188: I think we may need to specify encoding here as well (read and write maybe -- some older version of python seem to prefer this)

romain-intel (approved):
whoohoo. I'll try this out internally but very nice!

Issue #969, opened 2022-02-24
Cannot seem to use Conda & Batch on MacOS since 2.5.2


bishax:
`python anyFlow.py --environment conda run --with batch`

results in the following on MacOS (M1 chip)

```
Bootstrapping conda environment...(this could take a few minutes)
    Conda ran into an error while setting up environment.:
    Step: start, Error: command '['/opt/homebrew/Caskroom/miniconda/base/condabin/conda', 'create', '--yes', '--no-default-packages', '--name', 'metaflow_BugFlow_linux-64_4f87119c38b4db696c5a3d1a0f205a449a9a8ae6', '--quiet', b'python==3.7.12', b'requests==>=2.21.0', b'boto3==>=1.14.0']' returned error (1): b'{\n  "command": "/opt/homebrew/Caskroom/miniconda/base/condabin/conda create --yes --no-default-packages --name metaflow_BugFlow_linux-64_4f87119c38b4db696c5a3d1a0f205a449a9a8ae6 --quiet python==3.7.12 requests==>=2.21.0 boto3==>=1.14.0",\n  "

...

[...]

PR #972, opened 2022-02-25
Fix "Too many symbolic links" error when using Conda + Batch on MacOS

bishax:
Addresses #969


bishax:
A workaround for this issue is for the user to set the `CONDA_SAFETY_CHECKS` env var to `disabled`.

Issue #814, opened 2021-11-05
Add a way to securely inject secrets into AWS Batch


mmiermans:
# Problem
Some of our Metaflow steps need to connect to a database using secret credentials. Currently we use boto3 to get these from AWS SecretsManager, but this requires [a little bit of code](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/secretsmanager.html#SecretsManager.Client.get_secret_value). It would be nice if these secrets would be securely injected as environment variables, which AWS Batch supports:

https://docs.aws.amazon.com/batch/latest/userguide/specifying-sensitive-data-secrets.html#secrets-envvar

# Solution
> We can enable this for AWS Batch. We have a similar functionality for Kubernetes. 

[Outerbounds Slack thread](https://outerbounds-community.slack.com/archives/C02116BBNTU/p1635958797165600?thread_ts=1635958686.164300&cid=C02116BBNTU)

erichhhhho:
Not able to access the outerbounds thread. May I know if there is any on this issue?

savingoyal:
https://docs.metaflow.org/scaling/secrets

PR #976, opened 2022-03-11
Inject secrets from AWS Secrets Manager into AWS Batch job

stevenhoelscher:
Resolves https://github.com/Netflix/metaflow/issues/814

## Feature Overview

[AWS Batch allows for the injection of secrets into jobs](https://docs.aws.amazon.com/batch/latest/userguide/specifying-sensitive-data-secrets.html) by pulling them from AWS Secrets Manager and exposing them as environment variables.

Users would need to provide an environment variable name where the secret should be exposed and an Amazon Resource Name (ARN) of the Secrets Manager secret. They should be able to provide multiple `environment_variable:secret_arn` key pairs.

This feature should help users eliminate any code calling out to Secrets Manager directly, and instead pull a secret from an environment variable directly.
 
### AWS Batch

```
python main.py --no-pylint run --with batch:secrets='{"SNOWFLAKE_USER": "arn:aws:secretsmanager:region:123:secret:MyFlow-abc:SNOWFLAKE_USER::", "SNOWFLAKE_PASSWORD": "arn:aws:secretsmanager:region:123:secret:MyFlow-abc:SNOWFLAKE_PASSWORD::"}'
```

✅ 

### AWS Step Functions

```
python src/main.py --no-pylint --with batch:secrets='{"SNOWFLAKE_USER": "arn:aws:secretsmanager:region:123:secret:MyFlow-abc:SNOWFLAKE_USER::", "SNOWFLAKE_PASSWORD": "arn:aws:secretsmanager:region:123:secret:MyFlow-abc:SNOWFLAKE_PASSWORD::"}' step-functions create
```

❌

### DIFF
--- a/metaflow/plugins/aws/batch/batch_cli.py
+++ b/metaflow/plugins/aws/batch/batch_cli.py
@@ -149,6 +149,12 @@ def kill(ctx, run_id, user, my_runs):
     type=int,
     help="Number of parallel nodes to run as a multi-node job.",
 )
+@click.option(
+    "--secrets",
+    multiple=True,
+    default=None,
+    help="Secrets for AWS Batch job"
+)
 @click.pass_context
 def step(
     ctx,
@@ -169,6 +175,7 @@ def step(
     swappiness=None,
     host_volumes=None,
     num_parallel=None,
+    secrets=None,
     **kwargs
 ):
     def echo(msg, stream="stderr", batch_id=None):
@@ -294,6 +301,7 @@ def step(
                 attrs=attrs,
                 host_volumes=host_volumes,
                 num_parallel=num_parallel,
+                secrets=secrets,
             )
     except Exception as e:
         traceback.print_exc()
--- a/metaflow/plugins/aws/batch/batch_cli.py
+++ b/metaflow/plugins/aws/batch/batch_cli.py
@@ -6,7 +6,7 @@ import traceback
 
 from distutils.dir_util import copy_tree
 
-from metaflow import util
+from metaflow import JSONType, util
 from metaflow import R
 from metaflow.exception import CommandException, METAFLOW_EXIT_DISALLOW_RETRY
 from metaflow.metadata.util import sync_local_metadata_from_datastore
@@ -151,9 +151,9 @@ def kill(ctx, run_id, user, my_runs):
 )
 @click.option(
     "--secrets",
-    multiple=True,
+    type=JSONType,
     default=None,
-    help="Secrets for AWS Batch job"
+    help="Secrets to inject into AWS Batch job",
 )
 @click.pass_context
 def step(
--- a/metaflow/plugins/aws/batch/batch_client.py
+++ b/metaflow/plugins/aws/batch/batch_client.py
@@ -263,20 +263,11 @@ class BatchJob(object):
                 )
 
         if secrets:
-            job_definition["containerProperties"]["secrets"] = []
-            if isinstance(secrets, str):
-                secrets = [secrets]
-            for secret in secrets:
-                delimiter = ","
-                if delimiter not in secret:
-                    raise ValueError(
-                        "Secrets should be of the format `environment_key_name,value_from`\n"
-                        "got %s" % secret
-                    )
-                name, value_from = secret.split(delimiter)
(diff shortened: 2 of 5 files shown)


metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -261,6 +262,23 @@ def _register_job_definition(
                     {"sourceVolume": name, "containerPath": host_path}
                 )
 
+        if secrets:
+            job_definition["containerProperties"]["secrets"] = []
+            if isinstance(secrets, str):
+                secrets = [secrets]
+            for secret in secrets:
+                delimiter = ","
```
stevenhoelscher, line 270: I'm not sure what best practice for user-supplied key-value pairs is.

I primarily worked with this example when testing:

```
python flow.py run --with batch:secrets=secret_a,arn:aws:secretsmanager:secret_a
```

I originally tried `=` and `->` but both proved problematic with how Metaflow seems to be handling incoming arguments.

metaflow/plugins/aws/batch/batch_cli.py
```diff
@@ -149,6 +149,12 @@ def kill(ctx, run_id, user, my_runs):
     type=int,
     help="Number of parallel nodes to run as a multi-node job.",
 )
+@click.option(
+    "--secrets",
+    multiple=True,
```
stevenhoelscher, line 154: Does this apply to the `--with batch` syntax?

I haven't been able to get any of these to work:

```
python flow.py run --with batch:secrets=secret_a,arn:aws:secretsmanager:secret_a,secrets=secret_b,arn:aws:secretsmanager:secret_b
```

```
python flow.py run --with batch:secrets=secret_a,arn:aws:secretsmanager:secret_a;secret_b,arn:aws:secretsmanager:secret_b
```

metaflow/plugins/aws/batch/batch_client.py
savingoyal: @stevenhoelscher Would it make sense to do `python flow.py run --with batch:secrets=arn:aws:secretsmanager:secret_a` which will expose `secret_a`? That way we don't have to maintain a mapping. Wdyt?

savingoyal: Happy to chat live in the `#dev-metaflow` channel in chat.metaflow.org.

stevenhoelscher: Based on [Slack conversation](https://outerbounds-community.slack.com/archives/C020U025QJK/p1647038067402739), updated secrets type to `JSONType` / dictionary

metaflow/plugins/aws/batch/batch_cli.py
stevenhoelscher: This isn't applicable anymore with the move to a dictionary-like type

stevenhoelscher:
Closing this out because I don't have enough understanding of the Metaflow codebase yet.

For anyone that wants to revive this work: the outstanding issue is that `run` and `step-functions create` hit two different code paths. Despite passing the `--with batch:secrets=...` flag for both, the step decoration behaves differently.

My understanding is that [`run`](https://github.com/Netflix/metaflow/blob/master/metaflow/cli.py#L753) will create a Worker to [launch a task via the subprocess](https://github.com/Netflix/metaflow/blob/540f21133b08108f7129ce42b1c6a24fd9175b2f/metaflow/runtime.py#L994-L1028), and by doing so will hit the [`batch_cli.step`](https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/aws/batch/batch_cli.py#L153), which is responsible for type coercion.

But [`step-functions create`](https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/aws/step_functions/step_functions_cli.py#L120) will [only attach decorators in-memory](https://github.com/Netflix/metaflow/blob/540f21133b08108f7129ce42b1c6a24fd9175b2f/metaflow/plugins/aws/step_functions/step_functions_cli.py#L271-L274), so the `batch_cli.step` is never invoked.

Downstream, this means the `BatchClient` will receive a JSON-encoded string from `step-functions create` and a dictionary from `run`.

PR #983, opened 2022-03-15
Add optional setting for max cpu and max memory to config

oavdeev:
There is a recurring problem with remote execution environments, where Metaflow task may request more resources that can be provisioned, and that causes task be stuck. These issues tend to be hard to debug -- for example, AWS Batch will not return an error in this case. 

Generally speaking, it is hard for Metaflow to determine _exactly_ what is maximum amount of memory or cpu that can be used at any moment, since it depends on autoscaling configuration of AWS Batch or K8S, that can be pretty complex. However, usually the person who configured the cluster can come up a decent upper bound. For example, there is no chance a task requesting 100GB of RAM will be scheduled in a cluster that only has instances with 32GB RAM.

This PR allows Metaflow config to specify that upper bound for CPU and Memory limits so that Metaflow will immediately reject tasks that request more with a nice error message. It is intended to be set by Metaflow admins, and distributed to end users as part of the config profile.

### DIFF
--- a/metaflow/plugins/aws/aws_utils.py
+++ b/metaflow/plugins/aws/aws_utils.py
@@ -1,6 +1,7 @@
 import re
 
 from metaflow.exception import MetaflowException
+from metaflow.metaflow_config import MAX_MEMORY_PER_TASK, MAX_CPU_PER_TASK
 
 
 def get_docker_registry(image_uri):
@@ -54,7 +55,7 @@ def get_docker_registry(image_uri):
     return registry
 
 
-def compute_resource_attributes(decos, compute_deco, resource_defaults):
+def compute_resource_attributes(decos, compute_deco, step_name, resource_defaults):
     """
     Compute resource values taking into account defaults, the values specified
     in the compute decorator (like @batch or @kubernetes) directly, and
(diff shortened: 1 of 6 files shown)


metaflow/plugins/aws/aws_utils.py
```diff
@@ -101,4 +102,17 @@ def compute_resource_attributes(decos, compute_deco, resource_defaults):
         if compute_deco.attributes.get(k) is not None:
             result[k] = str(compute_deco.attributes[k] or "0")
 
+    if "cpu" in result and MAX_CPU_PER_TASK:
```
romain-intel, line 105: nit: I would reverse the check order to benefit from lazy evaluation (checking a bool is much faster).

metaflow/plugins/aws/aws_utils.py
```diff
@@ -101,4 +102,17 @@ def compute_resource_attributes(decos, compute_deco, resource_defaults):
...
+                "Step %s requires %s CPU units, but you cannot use more than %s per step in this environment"
+                % (step_name, result["cpu"], MAX_CPU_PER_TASK)
+            )
+    if "memory" in result and MAX_MEMORY_PER_TASK:
+        if float(result["memory"]) > float(MAX_MEMORY_PER_TASK):
+            raise MetaflowException(
+                "Step %s requires %s CPU units, but you cannot use more than %s per step in this environment"
+                % (step_name, result["memory"], MAX_MEMORY_PER_TASK)
+            )
+
```
romain-intel, line 117: I wonder if a more flexible approach here would be to have this block for all attributes in the `resource_defaults` or something like that. This seems very specific to just two elements for some reason.

oavdeev: Yes, its mostly because I thought I may generate more user-friendly error messages if I don't start with a generic approach, but looking at it they're not that different. Then again, can always make it more generic later

savingoyal:
closing this PR for now since we can rely on server side checks today

PR #984, opened 2022-03-16
Add timezone to returned datetime for core objects

romain-intel:


Shriprasad-P (commented):
This looks like the right fix for #13399 — returning early for missing Form fields lets Pydantic own defaults so `model_fields_set` stays accurate, while Query/Header/Cookie keep the previous default-injection behavior.

A few review notes:

1. **Behavioral change in error payloads** — `tests/test_forms_single_model.py` now expects `"input": {}` instead of the prior dict with defaults. That is consistent with the Form path, but it is a user-visible validation error shape change. Calling that out in the PR description (and maybe a brief docs note if Form vs JSON error examples exist) would help maintainers and users.
2. **Explicit default sent as the default value** — the new tests cover empty/partial/all provided well. One more case worth adding: posting `field_2` with the literal default string and asserting it *is* in `fields_set` (provided vs omitted).
3. **Nested / embedded Form models** — if `_get_multidict_value` is on the hot path for nested models, a quick note on whether this early-return still applies recursively would be useful for future readers.

Overall the change is small and the regression suite is a solid addition.

Shriprasad-P:
Follow-up to my review: looking again at the diff in `metaflow/client/core.py`, `_created_at` is assigned with `tz=timezone.utc` and then immediately overwritten by a naive `fromtimestamp(...)` on the next line. The timezone-aware assignment looks unintentional / dead. Suggest removing the naive overwrite (or the first line) so the intended aware UTC value actually sticks, plus a small assertion that `created_at.tzinfo` is not `None`.

Shriprasad-P (commented):
Re-reviewing with the actual diff in mind (apologies if an earlier comment looked off-topic).

In `metaflow/client/core.py` the change currently does:

```python
self._created_at = datetime.fromtimestamp(
    self._object["ts_epoch"] / 1000.0, tz=timezone.utc
)
self._created_at = datetime.fromtimestamp(self._object["ts_epoch"] / 1000.0)
```

The second assignment overwrites the timezone-aware value with a naive one, so the `timezone` import and the new aware line have no effect. Suggest deleting the naive overwrite so `_created_at` stays UTC-aware.

Also worth checking:

1. **Call sites** — any code comparing `created_at` to naive datetimes will start raising (Py3) or behaving oddly once this sticks; a changelog note about aware returns would help.
2. **Tests** — assert `obj.created_at.tzinfo is not None` (and ideally `== timezone.utc`) on the core objects this path covers.
3. **Other constructors** — if sibling types still use naive `fromtimestamp`, this PR may want to update them in the same pass for consistency.

Happy to look again after the double-assignment is cleaned up.

PR #986, opened 2022-03-16
Fix in determining root of some extensions.

romain-intel:
Also tiny fix in test of extensions import


romain-intel:
Superseded by #989

PR #987, opened 2022-03-18
Add an option to pass a `role` argument to S3() to provide a role

romain-intel:


romain-intel:
I have not tested this yet. Just putting out here for early feedback. It turns out it was a fairly simple change. @oavdeev, since you are much more knowledgeable than me regarding AWS, please make sure that the assume-role function in the get_aws_client function is up to snuff.

romain-intel:
Need to fix something to do with pickling of the op_* functions.

metaflow/datatools/s3.py
```diff
@@ -277,7 +280,7 @@ def error(self):
         return self._s3_error
 
     def reset_client(self):
-        self._s3_client, self._s3_error = get_s3_client()
+        self._s3_client, self._s3_error = get_s3_client(role=self._s3_role)
```
oavdeev, line 283: as a mandatory naming nitpick for this PR, consider calling this `s3_role_arn` to make it clear that you have to pass the ARN :)

metaflow/plugins/aws/aws_client.py
```diff
@@ -45,12 +47,26 @@ def get_client(module, with_error=False, params={}):
...
+                source_credentials=session._session.get_credentials(),
+                role_arn=role,
+                extra_args={},
+            )
+            creds = botocore.credentials.DeferredRefreshableCredentials(
+                method="assume-role", refresh_using=fetcher.fetch_credentials
+            )
+            botocore_session = botocore.session.Session()
+            botocore_session._credentials = creds
+            session = boto3.session.Session(botocore_session=botocore_session)
```
oavdeev, line 63: Would this still work if original creds are temporary creds from IMDS? (I _think_ it would but not 100% sure)

metaflow/datatools/s3.py
```diff
@@ -1019,6 +1025,8 @@ def _s3op_with_retries(self, mode, **options):
                     cmdline.append("--no-%s" % key)
             else:
                 cmdline.extend(("--%s" % key, value))
+        if self._s3_role is not None:
+            cmdline.extend(("--s3role", _quote(self._s3_role)))
```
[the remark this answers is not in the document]
romain-intel: I can probably get rid of it. It should work both ways I think. I may have been a little more on the careful side.

metaflow/plugins/aws/aws_client.py
```diff
@@ -45,12 +47,26 @@ def get_client(module, with_error=False, params={}):
             return boto3.session.Session(**cached_aws_sandbox_creds).client(
                 module, **params
             )
+        session = boto3.session.Session()
+        if s3_role_arn:
```
savingoyal, line 51: Minor nit - we can rename this from `s3_role_arn` to just `role_arn` - there isn't anything s3 specific here.

PR #988, opened 2022-03-20
Use importlib_metadata 2.8.3 for Python 3.4, 3.5 but 4.8.3 for Python 3.6+

romain-intel:
This is required because in importlib_metadata 3.4.0, a field called `_normalized_name`
was introduced and is relied on by later versions of importlib_metadata. When
importlib_metadata looks for distributions, it queries all registered/installed
importlib_metadata and if an older version (like 2.8.3) returns something that
does not have this field, it causes a crash.


romain-intel:
Literally tried this in 5 minutes. Pushing out so you can tell me if this approach is OK with you. Want to try to get this out early next week so please let me know what you think; I will adjust tomorrow.

Issue #990, opened 2022-03-22
Specify base directory for artifact pulls


kgullikson88:
I am using metaflow locally but with the AWS service (e.g. the actual compute is happening locally rather than in AWS batch but the metadata is using AWS). When I access the run data through `run.data` I get new local directories with names like `metaflow.s3.w3efey1k`, which I presume is because metaflow pulls from S3 into that directory, and then un-pickles the result from there. Is there a way to specify the root directory for those "temporary" directories? They are "polluting" my working directory and I would prefer they go in `/tmp` or something.

It seems like adding one of [these things](https://github.com/Netflix/metaflow/blob/c903953db34b18da07499112188f0a81b538d326/metaflow/metaflow_config.py#L56) to the `.metaflowconfig` file would do it, but am not sure which one.

kgullikson88:
@romain-intel I can take a stab. So basically would need to

1. Add a field in the metaflow_config.py relating to this, with default=os.getcwd()
2. Replace all the `tmproot=os.getcwd()` with `tmproot=THAT_CONFIG` in the file you linked?

PR #991, opened 2022-03-22
Configurable temp root when pulling artifacts from s3

kgullikson88:
This addresses https://github.com/Netflix/metaflow/issues/990 by

1. Adds `METAFLOW_ARTIFACT_LOCALROOT` to the configuration. It defaults to `os.getcwd()`
2. When instantiating the `S3` object in `metaflow/datastore/s3_storage.py`, use that configurable value instead of `os.getcwd()`

I tested this locally by adding `"METAFLOW_ARTIFACT_LOCALROOT": "/tmp/"` to my `~/.metaflowconfig/config.json` file, and then pulling some artifacts via `task.data`. It did indeed create a `/tmp/metaflow.s3.br9v3iit` instead of going in the current directory.

### DIFF
--- a/metaflow/datastore/s3_storage.py
+++ b/metaflow/datastore/s3_storage.py
@@ -3,7 +3,7 @@ import os
 from itertools import starmap
 
 from ..datatools.s3 import S3, S3Client, S3PutObject
-from ..metaflow_config import DATASTORE_SYSROOT_S3
+from ..metaflow_config import DATASTORE_SYSROOT_S3, ARTIFACT_LOCALROOT
 from .datastore_storage import CloseAfterUse, DataStoreStorage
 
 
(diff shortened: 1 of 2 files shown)


romain-intel (approved):
This LGTM. Have you confirmed that it works the way you want it to for your use case?

kgullikson88:
> Have you confirmed that it works the way you want it to for your use case?

Yes it does

savingoyal (approved):

talebzeghmi:
Is there reason `METAFLOW_ARTIFACT_LOCALROOT` wasn't applied to `S3(tmproot)`?  
Is it because of breaking change? What are your thoughts for the default to be `METAFLOW_ARTIFACT_LOCALROOT` if defined else `"."`?

see 
https://github.com/Netflix/metaflow/blob/master/metaflow/datatools/s3.py#L290

romain-intel:
Yes, we could do that. I like the option of using the env value and defaulting to getcwd o/w.

PR #998, opened 2022-04-05
Upgrade black to 22.3.0

cloudw:
### Why
The black version `21.9b0` doesn’t work with new [click](https://pypi.org/project/click/#history) version 8.1.0 released on Mar 28, and it’s fixed by [black](https://pypi.org/project/black/#history) version 22.3.0  released on the same day. Upgrading black in this PR.

The new black version no longer support python 2 formats however. Specifically string is assumed to be unicode. Therefore this PR assumes that **python 2 compatibility is dropped**.

### What
- Upgrade black to 22.3.0 and run `pre-commit run --all-files`
- Remove python 2 compatibility code as they no longer make sense.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ repos:
       - id: check-yaml
       - id: check-json
   - repo: https://github.com/ambv/black
-    rev: 21.9b0
+    rev: 22.3.0
     hooks:
       - id: black
         language_version: python3
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -1,5 +1,5 @@
 import inspect
-import os
+import pickle
 import sys
 import traceback
 from datetime import datetime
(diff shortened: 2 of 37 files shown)


savingoyal:
@cloudw Thanks for the PR. We are trying to make sure that we don't actively break support for Python 2. Is there a specific version of black we can pin against - say a `pre-22.1` version? Ideally, it would be nice if somehow we could ensure that all PRs have an auto-generated new commit applied on top with the correct version of black - without requiring folks to rely on pre-commit hooks (like what we attempted in #771 - but it doesn't fork on PRs generated from forks) - would you be interested in pursuing that direction?

cloudw:
> @cloudw Thanks for the PR. We are trying to make sure that we don't actively break support for Python 2. Is there a specific version of black we can pin against - say a `pre-22.1` version? Ideally, it would be nice if somehow we could ensure that all PRs have an auto-generated new commit applied on top with the correct version of black - without requiring folks to rely on pre-commit hooks (like what we attempted in #771 - but it doesn't fork on PRs generated from forks) - would you be interested in pursuing that direction?

Can do. Let me find out the last version of black that was compatible with py2, and a compatible click version.

PR #1000, opened 2022-04-06
Standardize on $METAFLOW_DATATOOLS_SYSROOT_S3

### LINKED ISSUE #999: Where "METAFLOW_DATATOOLS_SYSROOT_S3" config is used?
When configuring aws variables with `metaflow configure aws`, we set `METAFLOW_DATATOOLS_SYSROOT_S3`.
But I couldn't find where this config is used. It's only used in [cli](https://github.com/Netflix/metaflow/blob/62f5e52ebce755d9130287c1576011cb056e0e3d/metaflow/main_cli.py#L490).
I guess [this line](https://github.com/Netflix/metaflow/blob/master/metaflow/metaflow_config.py#L66) should be changed to `METAFLOW_DATATOOLS_SYSROOT_S3` from `METAFLOW_DATATOOLS_S3ROOT`.

kevsmith:
* Made main_cli.py and metaflow_config.py use same env var for data tools' S3 root
* Updated config var to agree with env var name along w/all users

Fixes #999


oavdeev:
I think we need to make it backwards compatible so existing configs with `METAFLOW_DATATOOLS_S3ROOT` continue to work, and then merge

PR #1002, opened 2022-04-06
Add metadata to cloned tasks

romain-intel:


metaflow/task.py
```diff
@@ -275,24 +275,63 @@ def clone_only(self, step_name, run_id, task_id, clone_origin_task, retry_count)
         )
         metadata_tags = ["attempt_id:{0}".format(retry_count)]
         output.clone(origin)
+        # Get original metadata and extract fields we care about
+        metadata_fields = [
+            "conda_env_id",
```
savingoyal, line 280: Why do we need `conda_env_id`? We are also leaking information about a plugin here.

metaflow/task.py
```diff
@@ -275,24 +275,63 @@ def clone_only(self, step_name, run_id, task_id, clone_origin_task, retry_count)
...
+            ),
+            MetaDatum(
+                field="origin-run-id",
+                value=str(origin_run_id),
+                type="origin-run-id",
+                tags=metadata_tags,
+            ),
+            MetaDatum(
+                field="attempt",
+                value=str(retry_count),
```
savingoyal, line 324: Is it correct to assume that the `retry_count` is set to 0 for all cloned tasks?

metaflow/task.py
```diff
@@ -275,24 +275,63 @@ def clone_only(self, step_name, run_id, task_id, clone_origin_task, retry_count)
...
         output.clone(origin)
+        # Get original metadata and extract fields we care about
+        metadata_fields = [
+            "conda_env_id",
+            "code_package",
+            "ds-type",
+            "ds-root",
+            "attempt_ok",
+        ]
+        origin_metadata = self.metadata.get_object(
```
savingoyal, line 286: This will couple user code execution with metadata reads from the service which doesn't happen to be the case today and we would ideally like to avoid adding. Rather than adding metadata fields here, what do you think about instead following the trail (through references to `origin_run_id`, `origin_task_id`) in the hosting client?

metaflow/task.py
romain-intel: I don't think so. It would be rare but there could be cases where the scheduler has to try twice and if the retry count comes from there, it may be non zero.

metaflow/task.py
romain-intel: That should work yes. I am not sure I see the issue with reading from the metadata (we write to it just there which implies the service is definitely up and we also somewhat read from it when we get it to mint a new ID. In that scenario, would we, for now, be adding just the metadata that we list in `metadata_fields` (ie: if there is an origin-task-id, go fetch the latest metadata from that task, extract the ones for metadata_fields and return them)? We would have to properly duplicate them?

romain-intel:
Replaced by #1003.

PR #1003, opened 2022-04-07
Alternate way of adding metadata to cloned tasks

romain-intel:
This adds it on the client side

PR #1006, opened 2022-04-08
Pin older click compatible to old black

cloudw:
Alternative solution to https://github.com/Netflix/metaflow/pull/998

### Why
The black version 21.9b0 doesn’t work with new [click](https://pypi.org/project/click/#history) version 8.1.0 released on Mar 28, and it’s fixed by [black](https://pypi.org/project/black/#history) version 22.3.0 released on the same day. 

However black version 22.3.0 doesn't support python 2. Instead pinning an older version of click in pre-commit hook allow us to keep python 2 compatibility.

### What
Pin latest version of click that stays compatible with black 21.9b0

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,8 +6,9 @@ repos:
       - id: check-yaml
       - id: check-json
   - repo: https://github.com/ambv/black
-    rev: 21.9b0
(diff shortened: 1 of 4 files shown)


savingoyal:
@cloudw you can cherry-pick the changes from #1003 to get this PR going.

PR #1012, opened 2022-04-15
Fix var collision for ARGO_WORKFLOW_NAME

cloudw:
ARGO_WORKFLOW_NAME is added in two places: all-pods-default and new argo code. This issue is reported in https://gitlab.zgtools.net/analytics/artificial-intelligence/ai-platform/aip-infrastructure/kubeflow-k8s-profiles/-/merge_requests/531

Meanwhile `ARGO_WORKFLOW_NAME` and `MF_ARGO_WORKFLOW_NAME` are duplicate of each other since they both have reference to `workflows.argoproj.io/workflow`. Switch to use MF_ARGO_WORKFLOW_NAME solves the problem.

Note on roll out:
- By removing `ARGO_WORKFLOW_NAME`` from all-pods-default we are forcing an upgrade on zillow-metaflow code to customer teams. This PR is thus a blocking one for argo upgrade.


cloudw:
Created by mistake. This is only intended for zillow/metaflow fork.

Issue #1017, opened 2022-04-20
Get the value of reserved_params inside flow step


TonyIvchenko:
Is that possible to get the value or `reserved_params` inside `FlowSpec` step? As for instance, how I can get the value of `max-workers` param if it is set or default value if not?

savingoyal:
Not in a straightforward manner. There is an open ticket #571 to add them to the `current` object. One immediate workaround would be to launch your workflow as `METAFLOW_MAX_WORKERS=100 python flow.py run` and sniff the value of `METAFLOW_MAX_WORKERS` in `os.environ` but this approach wouldn't work when running on AWS Batch where the local environment variables are not passed all the way to AWS Batch.

TonyIvchenko:
Thank you, unfortunately we need it in AWS Batch, do you have an ETA for #571

savingoyal:
I can get a PR out next week? Would that work?

TonyIvchenko:
Absolutely, thank you!

PR #1019, opened 2022-04-21
Add tags to `current` singleton

sam-watts:


sam-watts:
Tests run for all graphs on updated `CurrentSingletonTest` at least, didn't have time to run all tests locally

savingoyal:
@sam-watts Thanks for the PR. It looks good to go and we can merge it as soon as the tests pass. One point that I would like to call out is that we will soon allow tag mutations and theoretically, the tags can be modified for an inflight run but the new values won't show up in the `current` object.

savingoyal (approved):

savingoyal:
@sam-watts In case you are interested - #1017 is a very similar issue that needs to be resolved :) Would you like to make a PR for that as well?

sam-watts:
@savingoyal I want to get familiar with the codebase so more than happy to have a crack at https://github.com/Netflix/metaflow/issues/1017 too 😄

PR #1020, opened 2022-04-22
add a way for aws client provider to configure creds for code package downloader

oavdeev:


savingoyal:
Not needed anymore either.

PR #1021, opened 2022-04-22
move k8s credentials provider to a separate module

oavdeev:


savingoyal:
This isn't needed any more.

PR #1026, opened 2022-04-28
Add execution context to `current`

### LINKED ISSUE #571: Add execution context in `current` singleton
Currently, the only way to check whether a task is executing `locally`, on `batch` or with `step-functions` is to sniff some environment variables - we can make it a bit easier by including the execution context in the current singleton.

https://outerboundsco.slack.com/archives/C02116BBNTU/p1623947638238600

### LINKED ISSUE #1017: Get the value of reserved_params inside flow step

sam-watts:
This PR will fix #1017, fix #571 

Currently implemented:
- [x] runtime_environment - this can be one of [local, aws-batch, kubernetes]
- [x] runtime_name - this can be one of [None, step-functions]
- [x] sfn state machine name - the name of the state machine if the flow is running on sfn 

To add:
- [ ] max_num_splits
- [ ] max_workers

These are trickier, as they cannot be accessed inside `task.Task.run_step`. I've looked at importing `current` inside `runtime.py`, however as tasks are spun up in subprocesses, the instance of `current` will be a new one, so the values will not actually be reference-able from inside the step

### DIFF
--- a/metaflow/current.py
+++ b/metaflow/current.py
@@ -15,6 +15,7 @@ class Current(object):
         self._namespace = None
         self._username = None
         self._is_running = False
+        self._tags = None
 
         def _raise(ex):
             raise ex
@@ -49,6 +50,10 @@ class Current(object):
         self._username = username
         self._is_running = is_running
         self._tags = tags
+        self._runtime_environment = os.environ.get(
+            "METAFLOW_RUNTIME_ENVIRONMENT", "local"
+        )
+        self._runtime_name = os.environ.get("METAFLOW_RUNTIME_NAME", None)
 
     def _update_env(self, env):
         for k, v in env.items():
@@ -111,6 +116,14 @@ class Current(object):
     @property
     def tags(self):
         return self._tags
+
+    @property
+    def runtime_environment(self):
+        return self._runtime_environment
+
+    @property
+    def runtime_name(self):
+        return self._runtime_name
 
 
 # instantiate the Current singleton. This will be populated
--- a/metaflow/current.py
+++ b/metaflow/current.py
@@ -18,7 +18,6 @@ class Current(object):
         self._tags = None
         self._runtime_environment = None
         self._runtime_name = None
-        self._sfn_state_machine_name = None
         self._max_workers = None
         self._max_num_splits = None
 
@@ -61,7 +60,6 @@ class Current(object):
             "METAFLOW_RUNTIME_ENVIRONMENT", "local"
         )
         self._runtime_name = os.environ.get("METAFLOW_RUNTIME_NAME", None)
-        self._sfn_state_machine_name = os.environ.get("SFN_STATE_MACHINE", None)
         self._max_workers = os.environ.get("CURRENT_MAX_WORKERS", None)
         self._max_num_splits = os.environ.get("CURRENT_MAX_NUM_SPLITS", None)
 
@@ -134,10 +132,6 @@ class Current(object):
     @property
     def runtime_name(self):
         return self._runtime_name
-
-    @property
-    def sfn_state_machine_name(self):
-        return self._sfn_state_machine_name
 
     @property
     def max_workers(self):
--- a/metaflow/current.py
+++ b/metaflow/current.py
@@ -62,8 +62,8 @@ class Current(object):
         )
         self._runtime_name = os.environ.get("METAFLOW_RUNTIME_NAME", None)
         self._sfn_state_machine_name = os.environ.get("SFN_STATE_MACHINE", None)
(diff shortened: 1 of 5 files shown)


sam-watts:
This PR will address #1017 and #571

savingoyal:
Thanks for the PR @sam-watts! Let me know when you would like me to start reviewing.

metaflow/current.py
```diff
@@ -15,6 +15,12 @@ def __init__(self):
         self._namespace = None
         self._username = None
         self._is_running = False
+        self._tags = None
+        self._runtime_environment = None
+        self._runtime_name = None
+        self._sfn_state_machine_name = None
```
romain-intel, line 21: This one seems very plugin specific.

metaflow/runtime.py
```diff
@@ -134,6 +134,10 @@ def __init__(
         # finished.
         self._control_num_splits = {}  # control_task -> num_splits mapping
 
+        # to allow the current singleton access within Tasks (which are in subprocesses)
+        os.environ["MAX_WORKERS"] = str(max_workers)
```
romain-intel, line 138: I would make these names a bit more specific to avoid the possibility of conflicts.

romain-intel (commented):
Minor comments.

metaflow/current.py
sam-watts: I agree, at some point it might make sense for the plugins to have their own implementations of `current` with inheritance of the common attributes? Don't know if that is a huge priority for now

metaflow/runtime.py
sam-watts: Good point, will change

sam-watts:
@savin good to go for a review :)

test/core/run_tests.py
```diff
@@ -258,7 +258,7 @@ def run_test_cases(args):
     "--tests",
     default="",
     type=str,
-    help="A comma-separate list of graphs to include (default: all).",
+    help="A comma-separate list of tests to include (default: all).",
```
savingoyal, line 261: Spurious commit?

metaflow/runtime.py
```diff
@@ -134,6 +134,10 @@ def __init__(
         # finished.
         self._control_num_splits = {}  # control_task -> num_splits mapping
 
+        # to allow the current singleton access within Tasks (which are in subprocesses)
+        os.environ["CURRENT_MAX_WORKERS"] = str(max_workers)
+        os.environ["CURRENT_MAX_NUM_SPLITS"] = str(max_num_splits)
```
savingoyal, line 139: Would this work with AWS Batch or Kubernetes?

test/core/run_tests.py
sam-watts: The help string was wrong 🤷‍♂️

metaflow/runtime.py
savingoyal: @sam-watts any updates? we would like to ship this PR soon.

metaflow/current.py
sam-watts: good idea, thanks for the tip!

metaflow/runtime.py
sam-watts: @savingoyal hoping to get back to this within the next week

metaflow/current.py
sam-watts: This is now resolved :) A related thing I was thinking about is if the environment for `current` is altered by different plugins, what is the best way to surface to users what is accessible from `current`. Right now that takes a bit of digging to work out!

sam-watts:
@savingoyal both `max_workers` and `max_num_splits` are proving trickier than I thought at first. Will need some more time to work those out. If you want to ship the rest first, I could create a separate PR for that?

sam-watts:
^ failing test looks like an unrelated issue with an R dependency

```
The downloaded binary packages are in
	/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T//RtmpCHGeu8/downloaded_packages
Error in loadNamespace(j <- i[[1L]], c(lib.loc, .libPaths()), versionCheck = vI[[j]]) : 
  there is no package called ‘memoise’
Calls: loadNamespace ... loadNamespace -> withRestarts -> withOneRestart -> doWithOneRestart
Execution halted
```

sam-watts:
@savingoyal let me know if you have thoughts! Would prefer this to get merged in some form rather than going to the PR graveyard 😄

Issue #1011, opened 2022-04-13
bug in `current` singleton for property `pathspec` when used outside of a flow


bergdavidj:
When accessing  the `pathspec` property of the `current` singleton outside of a flow I get the following error:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/apps/python3/lib/python3.7/site-packages/metaflow/current.py", line 91, in pathspec
    return "/".join((self._flow_name, self._run_id, self._step_name, self._task_id))
TypeError: sequence item 0: expected str instance, NoneType found

I expect the behavior to be the same as for other properties that return `None` if not in a flow.

PR #1033, opened 2022-05-03
current.pathspec to return None when used outside Flow

jackie-ob:
Address issue #1011 

While `current` usage outside of a running Flow is not part of [normal usage](https://docs.metaflow.org/metaflow/tagging#accessing-current-ids-in-a-flow), it is trivial to change to achieve more consistent behavior.

### DIFF
--- a/metaflow/current.py
+++ b/metaflow/current.py
@@ -90,7 +90,12 @@ class Current(object):
 
     @property
     def pathspec(self):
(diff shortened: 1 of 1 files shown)


savingoyal:
LGTM! Can you please format the PR with `black`?

savingoyal (approved):

PR #1034, opened 2022-05-05
[WIP] Address issues with S3 get and ranges

romain-intel:
This addresses two issues:
  - get and get_many would not properly populate the range_info in S3Object
  - get_many on different ranges of the same file would return incorrect results


romain-intel:
Need to add more tests.

romain-intel:
The testing failures are an artifact of the fact that I still have my own versions of the s3_data and test_s3 files (I am going to fix that). Could someone review please?

metaflow/datatools/s3.py
```diff
@@ -922,6 +972,8 @@ def _one_boto_op(self, op, url, create_tmp_file=True):
                     raise MetaflowS3NotFound(url)
                 elif error_code == 403:
                     raise MetaflowS3AccessDenied(url)
+                elif error_code == 601:
+                    raise MetaflowS3InvalidRange(err)
```
savingoyal, line 976: Isn't the error code - 416? https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html#RESTErrorResponses

savingoyal: not a big deal since it's an internal detail but to keep it consistent with S3's interface we may want to use 416

test/data/s3/test_s3.py
```diff
@@ -351,6 +389,13 @@ def test_init_options(s3root, pathspecs, expected):
         assert_results(s3objs, expected)
         assert_results(s3.get_all(), expected, info_should_be_empty=True)
 
+    # option 5) run object
+    if DO_TEST_RUN:
```
savingoyal, line 393: What's the intention behind this conditional?

metaflow/datatools/s3.py
romain-intel: I'll change. Ya, it was internal but I can use the same one. I always get lost in AWS documentation.

test/data/s3/test_s3.py
romain-intel: Test does not work in OSS (works internally). I merged all test files for S3 now so this was one of the differences and this seemed like an easy enough way to add it without being too problematic. I can add a comment if you want.

romain-intel:
I'll make the two small changes. Apart from that, are we good to merge?

savingoyal:
> I'll make the two small changes. Apart from that, are we good to merge?

Yep!

romain-intel:
THe 1 failure is a known thing so can ignore. This is good to go as far as I am concerned. Please stamp and we can merge :).

PR #1037, opened 2022-05-09
add list runs command

### LINKED ISSUE #602: Feature Request : Printing information about recent runs via CLI
Context : 

The `resume` functionality in Metaflow has a few nuances that get noted once used often. When resuming a flow, Metaflow clones the artifacts of the "resuming" flow. If the flow is really large and I cancel the resume before it cloned everything then the immediate next `python myflow.py resume` will resume a partially "resumed" flow.

User Story : 

Sometimes when I am resuming a flow, I would suddenly have a new idea or realize some mistake. At such a moment I would cancel that resume without paying any attention to the `run-id` (And many times even clear my shell before I remember that I need the right run-id to resume again). At such times when I immediately resume the flow, it wouldn't be the one I "intended" to resume. This has sometimes caused a cascade of wrong resumes. At that point, I have to go to my notebook/python and run some custom code to check which run-id should I resume from. 

This is a pain point I faced many times because I don't pay attention to `run-id`.  A CLI command which shows all recent runs makes this problem go away :)

hamelsmu:
~TODO: I need to add tests~

@valayDave @savingoyal this is ready for review

This will close #602

### DIFF
--- a/metaflow/plugins/list_cli.py
+++ b/metaflow/plugins/list_cli.py
@@ -1,5 +1,5 @@
 from metaflow._vendor import click
-from metaflow import Flow, Run
+from metaflow import Flow, namespace
 from metaflow import util
 from metaflow.exception import MetaflowNotFound, CommandException
 import json
@@ -15,7 +15,7 @@ def list():
     pass
 
 
-def _fetch_runs(flow_name, num_runs, namespace):
+def _fetch_runs(flow_name, num_runs):
     counter = 1
     try:
         flow = Flow(flow_name)
@@ -26,19 +26,17 @@ def _fetch_runs(flow_name, num_runs, namespace):
         for run in Flow(flow_name).runs():
             if counter > num_runs:
                 break
-            tags = [t for t in run.tags]
-            if namespace is None or namespace in tags:
-                counter += 1
-                run_list.append(
-                    dict(
-                        created=str(run.created_at),
-                        name=flow_name,
-                        id=run.id,
-                        status=run.successful,
-                        finished=run.finished,
-                        tags=tags,
-                    )
+            counter += 1
+            run_list.append(
+                dict(
+                    created=str(run.created_at),
+                    name=flow_name,
+                    id=run.id,
+                    status=run.successful,
+                    finished=run.finished,
+                    tags=[t for t in run.tags],
                 )
+            )
     return run_list
 
 
@@ -60,29 +58,35 @@ def _fetch_runs(flow_name, num_runs, namespace):
     help="Save the run list to file as json.",
 )
 @click.option("--user", default=None, help="List runs for the given user.")
-@click.option("--namespace", default=None, help="List runs for the given namespace.")
 @click.option(
-    "--my-runs",
+    "--namespace", "ns", default=None, help="List runs only for the given namespace."
+)
+@click.option(
+    "--show-all",
     default=False,
     is_flag=True,
-    help="List all my runs.",
+    help="List runs from the global namespace instead of the current user.",
 )
 @list.command(help="List recent runs for your flow.")
 @click.pass_context
-def runs(ctx, num_runs, as_json, file, user, my_runs, namespace):
-    if user and my_runs:
-        raise CommandException("--user and --my-runs are mutually exclusive.")
-    if user and namespace:
+def runs(ctx, num_runs, as_json, file, user, show_all, ns):
+    if user and show_all:
+        raise CommandException("--user and --show-all are mutually exclusive.")
+    if user and ns:
         raise CommandException("--user and --namespace are mutually exclusive.")
-    if my_runs and namespace:
-        raise CommandException("--my-runs and --namespace are mutually exclusive.")
+    if show_all and ns:
+        raise CommandException("--show-all and --namespace are mutually exclusive.")
 
-    if my_runs:
-        namespace = "user:{}".format(util.get_username())
-    if user:
-        namespace = "user:{}".format(user)
+    if show_all:
+        namespace(None)
+    elif user:
+        namespace("user:{}".format(user))
+    elif ns:
+        namespace(ns)
+    else:
+        namespace("user:{}".format(util.get_username()))
 
-    run_list = _fetch_runs(ctx.obj.flow.name, num_runs, namespace)
+    run_list = _fetch_runs(ctx.obj.flow.name, num_runs)
     if not run_list:
         ctx.obj.echo("No runs found for flow: {name}".format(name=ctx.obj.flow.name))
         return
--- a/metaflow/plugins/list_cli.py
+++ b/metaflow/plugins/list_cli.py
@@ -67,3 +67,10 @@ def runs(ctx, num_runs, as_json, file):
             json.dump(run_list, f)
     if as_json:
         ctx.obj.echo_always(json.dumps(run_list, indent=4), err=False)
+    else:
+        for run in run_list:
+            ctx.obj.echo(
+                "{created} {name} [{id}] (Successful:{status} Finished:{finished})".format(
+                    **run
+                )
+            )
--- a/metaflow/plugins/list_cli.py
+++ b/metaflow/plugins/list_cli.py
@@ -1,0 +1,52 @@
+from metaflow._vendor import click
+from metaflow import Flow
+from metaflow.exception import MetaflowNotFound
+
+
+@click.group()
+def cli():
+    pass
+
+
+@cli.group(help="List objects pertaining to your flow.")
+def list():
+    pass
+
+
+def _execute_cmd(echo, flow_name, num_runs):
+    found = False
+    counter = 1
+    try:
+        flow = Flow(flow_name)
+    except MetaflowNotFound:
+        flow = None
+
+    if flow:
+        for run in Flow(flow_name).runs():
+            found = True
+            if counter > num_runs:
+                break
+            counter += 1
+            echo(
+                "{created} {name} [{id}] (Successful:{status} Finished:{finished})".format(
+                    created=run.created_at,
+                    name=flow_name,
+                    id=run.id,
+                    status=run.successful,
+                    finished=run.finished,
+                )
+            )
+
+    if not found:
+        echo("No runs found for flow: {name}".format(name=flow_name))
+
+
+@click.option(
+    "--num-runs",
+    default=10,
+    help="Number of runs to show.",
+)
+@list.command(help="List recent runs for your flow.")
+@click.pass_context
+def runs(ctx, num_runs):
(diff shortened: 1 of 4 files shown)


metaflow/plugins/list_cli.py
```diff
@@ -0,0 +1,69 @@
...
+@click.option(
+    "--as-json",
+    default=False,
+    is_flag=True,
+    help="Print run list as a JSON object",
+)
+@click.option(
+    "--file",
+    default=None,
+    help="Save the run list to file as json.",
```
romain-intel, line 55: The code makes it look like if you don't specify anything it doesn't print as json nor does it output to a file? There doesn't seem to also respect the quiet flag.

metaflow/plugins/list_cli.py
```diff
@@ -0,0 +1,69 @@
...
+)
+@list.command(help="List recent runs for your flow.")
+@click.pass_context
+def runs(ctx, num_runs, as_json, file):
+    run_list = _fetch_runs(ctx.obj.flow.name, num_runs)
+    if not run_list:
+        ctx.obj.echo("No runs found for flow: {name}".format(name=ctx.obj.flow.name))
```
romain-intel, line 62: Nit: this message is incorrect (potentially) if num_runs <= 0

romain-intel (commented):
I don't have a huge issue with this PR but it does seem somewhat of a point solution and maybe we should think about the "list" API a bit more generally. It may be that we want to add more types of "lists" after (not saying to do it now but at least to think about it a bit).

metaflow/plugins/list_cli.py
hamelsmu: I thought [empty lists are falsy](https://docs.python.org/3/library/stdtypes.html#truth-value-testing)?  I have a feeling I'm missing something else, just trying to learn, thanks.

metaflow/plugins/list_cli.py
hamelsmu: Great catch, I accidentally forgot to add that back in while refactoring!  I will fix this!

hamelsmu: Ok I fixed this

metaflow/plugins/list_cli.py
hamelsmu: Ohh you mean if the cli argument itself is <= 0 !!!! I didn't think about this.  Okay

hamelsmu: ok I fixed this via https://github.com/Netflix/metaflow/pull/1037/commits/fbba6be401b9f4a50f3accaabcf20b85dad4d443

hamelsmu:
Oh I need to add the bit about the `--quiet` flag

metaflow/plugins/list_cli.py
```diff
@@ -0,0 +1,77 @@
...
+    help="Print run list as a JSON object",
+)
+@click.option(
+    "--file",
+    default=None,
+    help="Save the run list to file as json.",
+)
+@list.command(help="List recent runs for your flow.")
+@click.pass_context
+def runs(ctx, num_runs, as_json, file):
```
savingoyal, line 60: Can you also add `--user` and `--my-runs` flags as well? - take a look at `batch_cli.py`

metaflow/plugins/list_cli.py
```diff
@@ -0,0 +1,77 @@
...
+@cli.group(help="List objects pertaining to your flow.")
+def list():
+    pass
+
+
+def _fetch_runs(flow_name, num_runs):
+    counter = 1
+    try:
+        flow = Flow(flow_name)
+    except MetaflowNotFound:
```
savingoyal, line 21: We need to set the correct namespace here to ensure that the `Flow` or `Run` can be found.

metaflow/plugins/list_cli.py
hamelsmu: :thumbs_up:

hamelsmu: Thanks I implemented this in https://github.com/Netflix/metaflow/pull/1037/commits/e0682b15a576989cfd408db7708e26870bc282ed

hamelsmu:
@savingoyal I was discussin this with @valayDave and some questions:

1. Do we want to add a `--namespace` flag as well?
2. If we want to add 1, should we not have the `--user` flag and let people pass the namespace `user:<username>` instead?
3. If we want to keep both a `--namepsace` and `--user` flag, should they be mutually exclusive, or how do we want to handle collisions in this case, should we throw an error if there if there is a `--namespace` argument with `user:<username>` etc?  

Thanks

savingoyal:
How about we add both? It's usually a lot easier as a user for me to do `python flow.py list --my runs` or `python flow.py list --user=hamel runs`. We can also support `namespace` and make the set - `my`, `user` and `namespace` mutually exclusive. wdyt?

valayDave:
Ok. Here are some key questions that would need decisions to quickly converge on code changes.
1. What are the levels of filters were wish to support. Some I can think of are: `tags`, `namespace`, `user`, `date`, `branch`/`project`, `success`/`failure`.  
2. When options for these filters are provided from CLI, then will the resultant runs be filtered based on the union of filters or will there be some compatibility Constraints. An example constraint: You can provide `date` or `tags` but not both. Not saying we need to place this constraint. Just giving an example.  
3. Should the namespace command be `list` or something else ? Since we right now have something like `card list` or `batch list-jobs` etc. Similarly why not `runs list`. I agree that it is quite close to `run` making it a little confusing for someone at some point. 
4. Given the convergence on the above decisions, what is the final command spec. (Example command with all arguments)

hamelsmu:
@savingoyal I have added the `--namespace` flag so that we have something to look at and consider, but I think we should definitely discuss @valayDave 's [comments above](https://github.com/Netflix/metaflow/pull/1037#issuecomment-1122977965).   Please LMK your thoughts

romain-intel:
The namespace is implicitly the user (which I thought was the initial intent) but yes, this is exactly the type of conversation I was mentioning we should have around this more general "listing" API. These types of higher-level options (independent in effect of whether you are listing runs, steps, artifacts, etc) is useful.

metaflow/plugins/list_cli.py
```diff
@@ -0,0 +1,101 @@
...
+        raise CommandException("--user and --my-runs are mutually exclusive.")
+    if user and namespace:
+        raise CommandException("--user and --namespace are mutually exclusive.")
+    if my_runs and namespace:
+        raise CommandException("--my-runs and --namespace are mutually exclusive.")
+
+    if my_runs:
+        namespace = "user:{}".format(util.get_username())
+    if user:
+        namespace = "user:{}".format(user)
```
romain-intel, line 83: default should probably be my-runs (which is what it was before).

metaflow/plugins/list_cli.py
```diff
@@ -0,0 +1,101 @@
...
+    try:
+        flow = Flow(flow_name)
+    except MetaflowNotFound:
+        flow = None
+    run_list = []
+    if flow:
+        for run in Flow(flow_name).runs():
+            if counter > num_runs:
+                break
+            tags = [t for t in run.tags]
```
romain-intel, line 29: you don't need this. the call to `namespace(namespace)` should take care of filtering for you (it was already filtered by username by default).

metaflow/plugins/list_cli.py
hamelsmu: @romain-intel Just to clarify, I think what you are saying is that because of the call to `namespace(namespace)` in `cli.py` makes the `--my-runs` flags necessary?  Is my understanding correct? 

cc: @savingoyal incase I misunderstood your instructions earlier to add this flag as well

metaflow/plugins/list_cli.py
hamelsmu: @romain-intel perhaps you can shed some light here as I think I am confused b/w what Savin is saying and what you are saying elsewhere?  

Happy to have a sync conversation as well around this.  Thank you both for your help

metaflow/plugins/list_cli.py
romain-intel: no, I meant that right now, if you don't specify `my-runs` or `user`, it will do `namespace(None)` (or at least you should do that -- see my comment below) which means "no namespace filter". I am saying that if there is nothing specified, we should default to the behavior of `my-runs` because that is probably the most natural thing. We can still have an option (to be explicit). Of course that does bring the "how do you list all runs across namespaces" but I still feel that having the `my-runs` be the "default" behavior is probably more natural. At least it is the one that best answers the user story that @valayDave had in the original issue.

hamelsmu:
ok @romain-intel I made the changes (If I understood correctly) [in this commit](https://github.com/Netflix/metaflow/pull/1037/commits/742b080a7b45f6be699c1539ba215ca881d51604)

Please let me know if I am understanding correctly, hopefully, I'm getting closer!

hamelsmu:
ok @romain-intel I made the changes (If I understood correctly) [in this commit](https://github.com/Netflix/metaflow/pull/1037/commits/742b080a7b45f6be699c1539ba215ca881d51604)

I made the following changes:

1. removed the `--my-runs` flag and made that the default behavior
2. added a `--show-all` flag that calls `namespace(None)` to allow you to access the global namespace
3. Filtered runs by calling `namespace(<namespace>)`, instead of iterating through tags of each run

Please let me know if I am understanding correctly, hopefully, I'm getting closer!

We still have to have a discussion of this interface in general.    Note: this PR was also meant as a "Good First Issue" for me tackle with regards to learning the code base of Metaflow (just to let you know the dual purpose of this PR)

metaflow/plugins/list_cli.py
hamelsmu: ok I resolved this [with these changes](https://github.com/Netflix/metaflow/pull/1037#issuecomment-1123825173)

metaflow/plugins/list_cli.py
hamelsmu: Gotcha, ok I changed it to setting the namespace before retrieving runs, instead of manually filtering tags [as described here](https://github.com/Netflix/metaflow/pull/1037#issuecomment-1123825173)

hamelsmu:
Feedback from @savingoyal 

> My take is that we can expand list in the future to include cards etc. and supporting just --namespace (and maybe --my and --user as well) for now is a good start. We can always introduce more flags as we learn usage patterns.  --namespace already takes care of tags , user , date , project etc.

One minor TODO for @hamelsmu requested by @savingoyal 

> can we do list --all runs instead of list --show-all runs

savingoyal:
Converted the PR to draft since it is not ready for review yet.

PR #1039, opened 2022-05-10
[tag-mutation-project] Stash system info in task metadata

jackie-ob:
This is part of project to add tag mutation support. RFC [here](https://www.notion.so/outerbounds/RFC-Implementing-tag-mutation-98ef6ba39aee4aeeb9c6150b4cffea2a) (see "Consolidate tagging to Runs")

A side effect of consolidating tagging to Runs on the read paths is that Task-level tags and system_tags are no longer accessible over the Metaflow client API - we will get only Run-level tags and system tags.  In most cases this is fine because usually tags are the same between a Run and its descendant Tasks.  The exception is Task-level system tags which may diverge from the Run's system tags in some rare situations.

In this PR:
* We explicitly name the information being stored as system tags (including at Task-level) "system information"
* For tasks, we _continue_ to store this "system information" as system tags.  This is for backwards compatibility.  Eventually we we can look to stop doing so (e.g. Deprecation policy TBD....).
* For tasks, we also stash the same "system information" in a Task's metadata.  This allows Task-divergent system information previously accessed in a task's tags to become accessible within a Task's metadata instead.

We added the concept of "system information" to maintain a single info source driving both "system tags" today as well as what to save as Task metadata.  I.e. we don't want to be go from dictionary-like information to a list of `"%s:%s"` strings and then back again to construct Task metadata.

For simplicity, we also just stash away all existing system information items as Task metadata, rather than picking / excluding certain items. This removes the dependence for detailed knowledge about exactly which items are important to stash or which can possibly diverge between Run and Task, both today and in future.

Drive-by: we also remove `date` as a piece of system information.  I.e. it will no longer show up as a system tag for ALL objects,  It is too-generically named and similar information is available in client API as `created_at`.

This PR took some cues from this [older PR](https://github.com/Netflix/metaflow/pull/917).

### DIFF
--- a/metaflow/metadata/metadata.py
+++ b/metaflow/metadata/metadata.py
@@ -6,8 +6,7 @@ from collections import namedtuple
 from datetime import datetime
 
 from metaflow.exception import MetaflowInternalError
-from metaflow.util import get_username, resolve_identity
-
+from metaflow.util import get_username, resolve_identity_as_tuple
 
 DataArtifact = namedtuple("DataArtifact", "name ds_type ds_root url type sha")
 
@@ -507,8 +506,8 @@ class MetadataProvider(object):
         sys_info["runtime"] = env["runtime"]
         sys_info["python_version"] = env["python_version_code"]
         sys_info["date"] = datetime.utcnow().strftime("%Y-%m-%d")
-        identity_parts = resolve_identity().split(":", maxsplit=1)
-        sys_info[identity_parts[0]] = identity_parts[1]
+        identity_type, identity_value = resolve_identity_as_tuple()
+        sys_info[identity_type] = identity_value
         if env["metaflow_version"]:
             sys_info["metaflow_version"] = env["metaflow_version"]
         if "metaflow_r_version" in env:
@@ -517,10 +516,10 @@ class MetadataProvider(object):
             sys_info["r_version"] = env["r_version_code"]
         return sys_info
 
-    def _get_system_info_as_tag_set(self):
-        return set(
+    def _get_system_info_as_tags(self):
+        return [
             "{}:{}".format(k, v) for k, v in self._get_system_info_as_dict().items()
-        )
+        ]
 
     def _get_system_info_as_metadatum_list(self):
         metadata = []
@@ -640,4 +639,4 @@ class MetadataProvider(object):
         self._monitor = monitor
         self._environment = environment
         self._runtime = os.environ.get("METAFLOW_RUNTIME_NAME", "dev")
-        self.add_sticky_tags(sys_tags=self._get_system_info_as_tag_set())
+        self.add_sticky_tags(sys_tags=self._get_system_info_as_tags())
--- a/metaflow/metadata/metadata.py
+++ b/metaflow/metadata/metadata.py
@@ -515,12 +515,14 @@ class MetadataProvider(object):
             sys_info["r_version"] = env["r_version_code"]
         return sys_info
 
-    def _get_system_info_as_tags(self):
+    def _get_system_tags(self):
+        """Convert system info dictionary into a list of system tags"""
         return [
             "{}:{}".format(k, v) for k, v in self._get_system_info_as_dict().items()
         ]
 
     def _register_system_metadata(self, run_id, step_name, task_id, attempt):
+        """Gather up system and code packaging info and register them as task metadata"""
         metadata = []
         # Take everything from system info and store them as metadata
         sys_info = self._get_system_info_as_dict()
@@ -626,4 +628,4 @@ class MetadataProvider(object):
         self._monitor = monitor
         self._environment = environment
         self._runtime = os.environ.get("METAFLOW_RUNTIME_NAME", "dev")
-        self.add_sticky_tags(sys_tags=self._get_system_info_as_tags())
+        self.add_sticky_tags(sys_tags=self._get_system_tags())
--- a/metaflow/metadata/metadata.py
+++ b/metaflow/metadata/metadata.py
@@ -497,28 +497,52 @@ class MetadataProvider(object):
             for datum in metadata
         ]
 
-    def _tags(self):
+    def _get_system_info(self):
+        sys_info = dict()
         env = self._environment.get_environment_info()
-        tags = [
-            resolve_identity(),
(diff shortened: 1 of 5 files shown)


metaflow/metadata/metadata.py
```diff
@@ -497,41 +496,77 @@ def _metadata_to_json(self, run_id, step_name, task_id, metadata):
...
         env = self._environment.get_environment_info()
-        tags = [
-            resolve_identity(),
-            "runtime:" + env["runtime"],
-            "python_version:" + env["python_version_code"],
-            "date:" + datetime.utcnow().strftime("%Y-%m-%d"),
-        ]
+        sys_info["runtime"] = env["runtime"]
+        sys_info["python_version"] = env["python_version_code"]
+        sys_info["date"] = datetime.utcnow().strftime("%Y-%m-%d")
```
oavdeev, line 508: see https://github.com/Netflix/metaflow/pull/917#discussion_r791079881

jackie-ob: Removed, updated PR summary with:
```
Drive-by: we also remove date as a piece of system information. I.e. it will no longer show up as a system tag for ALL objects, It is too-generically named and similar information is available in client API as created_at.
```

oavdeev:
lgtm 

@savingoyal can you take a quick 👀 too?

oavdeev (approved):

metaflow/metadata/metadata.py
```diff
@@ -604,4 +638,4 @@ def __init__(self, environment, flow, event_logger, monitor):
         self._monitor = monitor
         self._environment = environment
         self._runtime = os.environ.get("METAFLOW_RUNTIME_NAME", "dev")
-        self.add_sticky_tags(sys_tags=self._tags())
+        self.add_sticky_tags(sys_tags=self._get_system_info_as_tags())
```
romain-intel, line 641: Won't this still add tags to the task and lower levels? Is that the intent? Is the goal to still support the older metadata services with newer versions? We could maybe check the service version and only push sticky tags below runs if older one. It would not cover case of newer service going down to older service but we said that wouldn't be a supported path.

jackie-ob: Re: [service downgrade support.](https://www.notion.so/outerbounds/RFC-Implementing-tag-mutation-98ef6ba39aee4aeeb9c6150b4cffea2a#29aa525b0a664420a6d855db91c09ab2).

The option I was considering was to just keep writing task tags here until a longer time in the future, when vast majority of services in the wild have upgraded (e.g. 1 year later). This will offer a soft landing for users in the service downgrade scenario in the meantime and it doesn't seem like a big lift to do it this way (in fact, less work).  There is also UI backend service which is deployed separately and has a separate direct DB read path.  Same tradeoff there.

Open to doing version check too if folks feel strongly - I think if we want to take a harder line about one-way service upgrade right now.  If we go that way, we would probably do it a bit closer to where we construct the payload to send to the metadata provider.

metaflow/metadata/metadata.py
```diff
@@ -497,41 +496,76 @@ def _metadata_to_json(self, run_id, step_name, task_id, metadata):
...
+            code_ds = os.environ.get("METAFLOW_CODE_DS")
             metadata.append(
                 MetaDatum(
                     field="code-package",
                     value=json.dumps(
                         {"ds_type": code_ds, "sha": code_sha, "location": code_url}
                     ),
                     type="code-package",
-                    tags=["attempt_id:{0}".format(attempt)],
+                    tags=[],
```
savingoyal, line 561: Why are we removing this tag here?

metaflow/plugins/metadata/local.py
```diff
@@ -80,7 +80,9 @@ def register_task_id(
         except ValueError:
             self._new_task(run_id, step_name, task_id, attempt, tags, sys_tags)
         else:
-            self._register_code_package_metadata(run_id, step_name, task_id, attempt)
+            self._register_system_info_and_code_packaging_metadata(
```
savingoyal, line 83: Minor nit - can we simply rename this to `_register_system_metadata`?

metaflow/metadata/metadata.py
```diff
@@ -497,41 +496,76 @@ def _metadata_to_json(self, run_id, step_name, task_id, metadata):
...
+            sys_info["r_version"] = env["r_version_code"]
+        return sys_info
+
+    def _get_system_info_as_tags(self):
+        return [
+            "{}:{}".format(k, v) for k, v in self._get_system_info_as_dict().items()
+        ]
 
-    def _register_code_package_metadata(self, run_id, step_name, task_id, attempt):
+    def _get_system_info_as_metadatum_list(self):
```
savingoyal, line 523: Minor nit - can we move this logic inside `_register_system_metadata`? It seems that's the only place this method is being used.

metaflow/metadata/metadata.py
```diff
@@ -497,41 +496,76 @@ def _metadata_to_json(self, run_id, step_name, task_id, metadata):
...
         if "metaflow_r_version" in env:
-            tags.append("metaflow_r_version:" + env["metaflow_r_version"])
+            sys_info["metaflow_r_version"] = env["metaflow_r_version"]
         if "r_version_code" in env:
-            tags.append("r_version:" + env["r_version_code"])
-        return tags
+            sys_info["r_version"] = env["r_version_code"]
+        return sys_info
+
+    def _get_system_info_as_tags(self):
```
savingoyal, line 518: Just - `_get_system_tags`?

savingoyal (approved):
LGTM, very minor comments. Let's 🚢 it!

metaflow/plugins/metadata/local.py
jackie-ob: Don't feel strongly... it is mainly for clarity.  We can use the shorter name - I can write a comment instead.

metaflow/metadata/metadata.py
jackie-ob: I'll update it and add a comment.

metaflow/metadata/metadata.py
jackie-ob: OK

PR #1043, opened 2022-05-12
[tag-mutation-project] Local metadata provider - object read paths to return ancestral Run tags

jackie-ob:
This is part of project to add tag mutation support. RFC [here](https://www.notion.so/outerbounds/RFC-Implementing-tag-mutation-98ef6ba39aee4aeeb9c6150b4cffea2a) (see "Consolidate tagging to Runs")

This is analogous to [this change](https://github.com/Netflix/metaflow-service/pull/315) on metadata service.

The main changes are on `LocalMetadataProvide._get_object_internal`.  We retrieve the ancestral Run of every object being returned, and replace `tags` and `system_tags` fields in the objects with the Run's tags before returning.

`LocalMetadataProvider._get_object_internal` returns a collection of objects of a certain type (controlled by `sub_type`), under a certain scope (controlled by `obj_type`).  There are multiple rules governing how `obj_type` and `sub_type` relate to each other.  

In this PR, we add some assertion statements to clarify the logic to make the code easier to reason about.  Most assertions are about control flow logic - except for asserting that ancestral runs can be read successfully from the user's local filesystem when we need it.

There was a complication with the unbounded foreach (UBF) feature.  In UBF there is a concept of a "control task".  We used to indicate that a task is a "control task" by adding a system tag on the task.  Here we use the [approach of stashing task-specific info as task metadata](https://github.com/Netflix/metaflow/pull/1039). New clients will look in tags (good for old tasks) AND in task metadata (for new tasks).

### DIFF
--- a/metaflow/client/core.py
+++ b/metaflow/client/core.py
@@ -1488,10 +1488,7 @@ class Step(MetaflowObject):
         Task
             A control task in the step
         """
-        children = super(Step, self).__iter__()
-        for t in children:
-            if CONTROL_TASK_TAG in t.tags:
-                return t
+        return next(self.control_tasks(), None)
 
     def control_tasks(self, *tags):
         """
@@ -1509,11 +1506,23 @@ class Step(MetaflowObject):
             Iterator over Control Task objects in this step
         """
         children = super(Step, self).__iter__()
-        filter_tags = [CONTROL_TASK_TAG]
-        filter_tags.extend(tags)
         for child in children:
-            if all(tag in child.tags for tag in filter_tags):
+            # first filter by standard tag filters
+            if not all(tag in child.tags for tag in tags):
+                continue
+            # Then look for control task indicator in one of two ways
+            # Look in tags - this path will activate for metadata service
+            # backends that pre-date tag mutation release
+            if CONTROL_TASK_TAG in child.tags:
                 yield child
+            else:
+                # Look in task metadata
+                for task_metadata in child.metadata:
+                    if (
+                        task_metadata.name == "internal_task_type"
+                        and task_metadata.value == CONTROL_TASK_TAG
+                    ):
+                        yield child
 
     def __iter__(self):
         children = super(Step, self).__iter__()
--- a/metaflow/metadata/metadata.py
+++ b/metaflow/metadata/metadata.py
@@ -46,6 +46,33 @@ def with_metaclass(mcls):
     return decorator
 
 
+class ObjectOrder:
+    # Consider this list a constant that should never change.
+    # Lots of code depend on the membership of this list as
+    # well as exact ordering
+    _order_as_list = [
+        "root",
+        "flow",
+        "run",
+        "step",
+        "task",
+        "artifact",
+        "metadata",
+        "self",
+    ]
+    _order_as_dict = {v: i for i, v in enumerate(_order_as_list)}
+
+    @staticmethod
+    def order_to_type(order):
+        if order < len(ObjectOrder._order_as_list):
+            return ObjectOrder._order_as_list[order]
+        return None
+
+    @staticmethod
+    def type_to_order(obj_type):
+        return ObjectOrder._order_as_dict.get(obj_type)
+
+
 @with_metaclass(MetadataProviderMeta)
 class MetadataProvider(object):
     @classmethod
@@ -354,22 +381,12 @@ class MetadataProvider(object):
             object or list :
                 Depending on the call, the type of object return varies
         """
-        obj_order = {
-            "root": 0,
-            "flow": 1,
-            "run": 2,
-            "step": 3,
-            "task": 4,
-            "artifact": 5,
-            "metadata": 6,
-            "self": 7,
-        }
-        type_order = obj_order.get(obj_type)
-        sub_order = obj_order.get(sub_type)
+        type_order = ObjectOrder.type_to_order(obj_type)
+        sub_order = ObjectOrder.type_to_order(sub_type)
 
         if type_order is None:
             raise MetaflowInternalError(msg="Cannot find type %s" % obj_type)
-        if type_order > 5:
+        if type_order >= ObjectOrder.type_to_order("metadata"):
             raise MetaflowInternalError(msg="Type %s is not allowed" % obj_type)
 
         if sub_order is None:
@@ -399,7 +416,7 @@ class MetadataProvider(object):
         pre_filter = cls._get_object_internal(
             obj_type, type_order, sub_type, sub_order, filters, attempt_int, *args
         )
-        if attempt_int is None or sub_order != 6:
+        if attempt_int is None or sub_type != "metadata":
             # If no attempt or not for metadata, just return as is
             return pre_filter
         return MetadataProvider._reconstruct_metadata_for_attempt(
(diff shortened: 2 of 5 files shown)


metaflow/client/core.py
```diff
@@ -1509,10 +1506,24 @@ def control_tasks(self, *tags):
...
+            # first filter by standard tag filters
+            if not all(tag in child.tags for tag in tags):
+                continue
+            # then look for control task indicator in one of two ways
+            is_control_task = False
+            # Look in tags - this path will activate for metadata service backends that pre-date tag mutation release
+            if CONTROL_TASK_TAG in child.tags:
+                is_control_task = True
+            # Look in task metadata
+            for task_metadata in child.metadata:
```
romain-intel, line 1519: nit: fast path can skip this if `is_control_task` is already true.

metaflow/client/core.py
```diff
@@ -1509,10 +1506,24 @@ def control_tasks(self, *tags):
...
-        filter_tags = [CONTROL_TASK_TAG]
-        filter_tags.extend(tags)
         for child in children:
-            if all(tag in child.tags for tag in filter_tags):
+            # first filter by standard tag filters
+            if not all(tag in child.tags for tag in tags):
+                continue
+            # then look for control task indicator in one of two ways
+            is_control_task = False
+            # Look in tags - this path will activate for metadata service backends that pre-date tag mutation release
```
romain-intel, line 1515: nit: these comment lines look quite long. Not sure why black isn't shortening them.

metaflow/metadata/metadata.py
```diff
@@ -47,6 +47,32 @@ def decorator(cls):
     return decorator
 
 
+class ObjectOrder:
+    # Consider this list a constant that should never change. Lots of code depend on the membership
```
romain-intel, line 51: Ditto on long lines.

metaflow/metadata/metadata.py
```diff
@@ -47,6 +47,32 @@ def decorator(cls):
     return decorator
 
 
+class ObjectOrder:
```
romain-intel, line 50: Ooh, much cleaner. I like this :)

metaflow/plugins/metadata/local.py
```diff
@@ -109,28 +112,62 @@ def register_metadata(self, run_id, step_name, task_id, metadata):
     def _get_object_internal(
         cls, obj_type, obj_order, sub_type, sub_order, filters, attempt, *args
     ):
+        # This is guaranteed by MetaflowProvider.get_object(), sole intended caller
+        if obj_type in ("metadata", "self"):
+            raise MetaflowInternalError(msg="Type %s is not allowed" % obj_type)
+
+        assert obj_type in ("root", "flow", "run", "step", "task", "artifact"), (
```
romain-intel, line 119: Why not keep it a `MetaflowInternalError`? I think we have tried to avoid asserts in general as they don't really give a huge amount of information. Am fine either ways but for consistency, it seems preferable to return an exception than have an assert (yes yes, asserts raise exceptions).

romain-intel (commented):
A first detailed pass on this: no major issues. I do want to run tests (but don't foresee any problems). I am still merging changes from 1039 (should be done soon) and then this will be good to go. A few minor comments but nothing major.

metaflow/plugins/metadata/local.py
jackie-ob: Don't feel strongly about it, updated!

jackie-ob:
@savingoyal @oavdeev do we need to do something on our side to run @romain-intel 's tests?  "testable" label?

savingoyal (dismissed):
LGTM! Some very minor stylistic comments

romain-intel:
Tests won't pass because I already moved to the `_register_system_metadata` here: https://github.com/Netflix/metaflow/blob/master/metaflow/metadata/metadata.py#L524. Please make sure it is up to date with that before running tests.

metaflow/plugins/metadata/local.py
```diff
@@ -116,28 +119,75 @@ def register_metadata(self, run_id, step_name, task_id, metadata):
     def _get_object_internal(
         cls, obj_type, obj_order, sub_type, sub_order, filters, attempt, *args
     ):
+        # This is guaranteed by MetaflowProvider.get_object(), sole intended caller
+        if obj_type in ("metadata", "self"):
+            raise MetaflowInternalError(msg="Type %s is not allowed" % obj_type)
+
+        if obj_type not in ("root", "flow", "run", "step", "task", "artifact"):
+            raise MetaflowInternalError(msg="unexpected object type %s" % obj_type)
```
romain-intel, line 127: nit: inconsistent capitalization (with message above)

metaflow/plugins/metadata/local.py
```diff
@@ -116,28 +119,75 @@ def register_metadata(self, run_id, step_name, task_id, metadata):
...
+                    )
+
+                obj["tags"] = run.get("tags", [])
+                obj["system_tags"] = run.get("system_tags", [])
+                return obj
             return None
 
         if sub_type == "artifact":
+            if obj_type not in ("root", "flow", "run", "step", "task"):
+                raise MetaflowInternalError(msg="unexpected object type %s" % obj_type)
```
romain-intel, line 189: ditto

metaflow/plugins/metadata/local.py
```diff
@@ -116,28 +119,75 @@ def register_metadata(self, run_id, step_name, task_id, metadata):
...
+                if obj_type not in ("step", "task"):
+                    raise MetaflowInternalError(
+                        msg="unexpected object type %s" % obj_type
+                    )
+                run = LocalMetadataProvider.get_object(
+                    "run", "self", {}, None, *args[:RUN_ORDER]  # *[flow_id, run_id]
+                )
+                if not run:
+                    raise MetaflowInternalError(
+                        msg="could not find run %s" % str(args[:RUN_ORDER])
```
romain-intel, line 179: ditto

metaflow/plugins/metadata/local.py
```diff
@@ -116,28 +119,75 @@ def register_metadata(self, run_id, step_name, task_id, metadata):
...
                     [LocalMetadataProvider._read_json_file(self_file)], filters
                 )[0]
+                # For non-descendants of a run, we are done
+
+                if obj_order <= RUN_ORDER:
+                    return obj
+
+                if obj_type not in ("step", "task"):
+                    raise MetaflowInternalError(
+                        msg="unexpected object type %s" % obj_type
```
romain-intel, line 172: ditto

metaflow/plugins/metadata/local.py
```diff
@@ -163,11 +213,27 @@ def _get_object_internal(
...
                 for obj in glob.iglob(artifact_files):
                     result.append(LocalMetadataProvider._read_json_file(obj))
+
+            # We are getting artifacts. We should overlay with ancestral run's tags
+            run = LocalMetadataProvider.get_object(
+                "run", "self", {}, None, *args[:RUN_ORDER]  # *[flow_id, run_id]
+            )
+            if not run:
+                raise MetaflowInternalError(
+                    msg="could not find run %s" % str(args[:RUN_ORDER])
```
romain-intel, line 223: ditto

romain-intel (dismissed):
This looks fine. I will re-trigger the tests but previous failures seem to have been due to some flakiness (I was mucking around with the hosting test environment around that time). I will confirm when the tests have re-run.

romain-intel (approved):
LGTM.

PR #1041, opened 2022-05-12
Fix minio access for remote runs

sappier:
Local runs use METAFLOW_S3_ENDPOINT_URL config to set the
"--endpoint-url" parameter required to access a minio server.
Remote containers miss this part, hence can't download a package
archive to start a step execution.

This fix adds "--endpoint-url" to the awscli command if
METAFLOW_S3_ENDPOINT_URL envar is defined in the remote container.

PR #1045, opened 2022-05-12
Enable MinIO/EMC-ECS as blob stores for Metaflow

savingoyal:


savingoyal:
Builds on #1041 AND #964

PR #1049, opened 2022-05-18
[tag-mutation-project] Add tag mutation support to LocalMetadataProvider and a tag CLI

jackie-ob:
This is part of project to add tag mutation support. RFC [here](https://www.notion.so/outerbounds/RFC-Implementing-tag-mutation-98ef6ba39aee4aeeb9c6150b4cffea2a) (see "Tagging CLI")

We add a new public method `MetadataProvider.mutate_user_tags_for_run()`.  This superclass implementation performs some generic input validation before delegating to a protected method `_mutate_user_tags_for_run`, to be implemented by each subclass.   We provide an implementation for `LocalMetadataProvider` in this PR.

Our intention is to match the client API as defined in this [PR](https://github.com/Netflix/metaflow/pull/547). This means exposing tag mutation support on the client API `Run` object as follows:
* `Run.add_tag`
* `Run.remove_tag`
* `Run.replace_tag`
Today, these functions will support both single tags as well as tag iterables (for backward compat at Netflix).  Tag iterable use cases are expected to be migrated to the functions below at some point, whence we will remove this unpublished functionality.

In addition we add these plural functions:
* `Run.add_tags`
* `Run.remove_tags`
* `Run.replace_tags` (all other methods above delegate to this!)
These will only support tag iterables.

A new core test (`TagMutationTest`) has been added to cover new tag mutation functionality. It runs only for local contexts at the moment.  We can enable the test globally once `ServiceMetadataProvider` and the underlying Metadata service has also gained support for this (coming shortly).

A new CLI command `tag` is included here.  Similar to client API, we aim to match as much as we can in the CLI UX in this [PR](https://github.com/Netflix/metaflow/pull/547).  To that end, the code for `tag_cli.py` was largely lifted from that same PR (`tagging_cli.py` there).

CLI differences:
* `tag replace` no longer accepts position args `TAG_TO_REMOVE TAG_TO_ADD`.  Instead we go with one or more `--add`, `--remove` options.  This seamlessly supports replacing more than one tag.
* We no longer explicitly check for (and ban) runs that belong to namespaces.  Instead we simple require the user to be in the right namespace (through `--namespace` option).
* `tag list --by-run / --by-tag` becomes `tag list --group-by-run / --group-by-tag`.

Otherwise outward UX is expected to be the same.

Note that the client API `MetaflowObject` base class gains two new properties: `user_tags` and `system_tags`.  With introduction of tag mutation functionality, it is now quite important to be able to distinguish between them (e.g. for `tag` CLI reporting, and necessary user understanding).  Nice side effect is it also makes testing easier (e.g. we need to know that a user tag is added correctly, and only to `user_tags`, etc.).

### DIFF
--- a/test/core/metaflow_test/cli_check.py
+++ b/test/core/metaflow_test/cli_check.py
@@ -31,7 +31,9 @@ class CliCheck(MetaflowCheck):
         cmd.extend([opt for opt in self.cli_options if opt != "--quiet"])
         cmd.extend(args)
 
-        return subprocess.run(cmd, capture_output=True, check=True)
+        return subprocess.run(
+            cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True
+        )
 
     def assert_artifact(self, step, name, value, fields=None):
         for task, artifacts in self.artifact_dict(step, name).items():


metaflow/plugins/__init__.py
```diff
@@ -84,6 +84,7 @@ def get_plugin_cli():
     from .aws.step_functions import step_functions_cli
     from .argo import argo_workflows_cli
     from .cards import card_cli
+    from . import tag_cli
```
jackie-ob, line 87: I updated the module name to `tag_cli` to follow the `[command]_cli` pattern

test/core/tests/tag_mutation.py
```diff
@@ -0,0 +1,88 @@
...
+        tags_to_remove = self.added_tags_at_start[:10]
+        tags_to_remain = self.added_tags_at_start[10:]
+        run.remove_tags(tags_to_remove)
+
+        for tag in tags_to_remain:
+            assert tag in run.tags
+        for tag in tags_to_remove:
+            assert tag not in run.tags
+
+        # TODO let's verify Task tags match up with run tags here
```
jackie-ob, line 47: Once [this PR](https://github.com/Netflix/metaflow/pull/1043) merges, we will rebase this PR and add extra checks here.

metaflow/plugins/tag_cli.py
```diff
@@ -0,0 +1,531 @@
...
+    user_namespace = resolve_identity() if user_namespace is None else user_namespace
+    run = _get_client_run_obj(obj, run_id, user_namespace)
+
+    run.replace_tag(tags[0], tags[1])
+
+    obj.echo("Operation successful. New tags:")
+    _print_tags_for_one_run(obj, run)
+
+
+# TODO(jackie) confirm with team on whether this can supercede "replace" command above
```
jackie-ob, line 348: @romain-intel  @savingoyal  Most of this CLI was lifted from this [old PR](https://github.com/Netflix/metaflow/pull/547) (as suggested by @romain-intel  to be the latest local implementation in use).  Mostly minor changes (vendor click, some variable renames, style stuff).

In this draft PR, there are two subcommands related to tag replacement:

* `tag replace tag_to_add tag_to_remove` - comes from this [PR](https://github.com/Netflix/metaflow/pull/547)
* `tag replace-many [--add tag_to_add]+ [--remove tag_to_remove]` - a suggested substitute for the `tag replace` above.

For OSS we were thinking it would be good to release a single subcommand called `replace`. but taking on the logic of `replace-many`.  @romain-intel would this CLI interface change vs your PR be palatable?

We could, of course have a `replace` that can do positional arguments and/or the `--add` / `--remove` at the same time.  However that seems a bit ugly and confusing IMO.  My hope is that it's not too late in the game to switch to a `replace-many` type command.

metaflow/plugins/tag_cli.py
```diff
@@ -0,0 +1,531 @@
...
+
+    # handle error messaging for three cases
+    # 1. our user makes a typo in --run-id
+    # 2. our user tries to mutate tags on a meson/production run
+    # 3. our user's --run-id does not exist in the default/specified namespace
+    try:
+        namespace(user_namespace)
+        run = Run(pathspec=run_name)
+    except MetaflowNotFound:
+        raise CommandException("No run *%s* found" % run_name)
```
savingoyal, line 224: To make it consistent with other error messages - `No run *<run>* found for *<flow>* `

metaflow/plugins/tag_cli.py
```diff
@@ -0,0 +1,531 @@
...
+    except MetaflowNamespaceMismatch:
+        namespace(None)
+        run = Run(pathspec=run_name)
+
+        if any([_is_prod_token(tag) for tag in run.system_tags]):
```
savingoyal, line 229: We can remove this conditional

metaflow/plugins/tag_cli.py
```diff
@@ -0,0 +1,531 @@
...
+            raise CommandException(
+                "Modifying a scheduled run's tags is currently not allowed."
+            )
+
+        msg = "Run *%s* does not belong to namespace *%s*\n" % (
```
savingoyal, line 234: `Run *<run>* for *<flow>* ....`

metaflow/plugins/tag_cli.py
```diff
@@ -0,0 +1,531 @@
...
+            run_name,
+            user_namespace,
+        )
+
+        user_identifier_tags = [
+            tag for tag in run.system_tags if tag.startswith("user:")
```
savingoyal, line 240: Not every run will have a "user:.." tag - all runs executed on top of SFN/Argo/Airflow for example. How about we drop this functionality altogether?

metaflow/plugins/tag_cli.py
```diff
@@ -0,0 +1,531 @@
...
+    type=str,
+    help="Run ID of the specific run to tag. [required]",
+)
+@click.option(
+    "--namespace",
+    "user_namespace",
+    required=False,
+    default=None,
+    type=str,
+    help="Change namespace from the default (your username) to the one specified.",
```
savingoyal, line 278: @romain-intel @tuulos should we support the global namespace?

metaflow/plugins/tag_cli.py
```diff
@@ -0,0 +1,531 @@
...
+        Flow(pathspec=flow_name)
+    except MetaflowNotFound:
+        raise CommandException(
+            "No execution found for *%s*. Please run the flow before tagging."
+            % flow_name
+        )
+
+    except MetaflowNamespaceMismatch:
+        raise CommandException(
+            "No execution found for *%s* in namespace *%s*."
```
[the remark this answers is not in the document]
savingoyal: Also, we can add - `You can switch the namespace using --namespace`.

metaflow/plugins/tag_cli.py
```diff
@@ -0,0 +1,531 @@
...
+            % (flow_name, user_namespace)
+        )
+
+    # throw an error with message to include latest run-id when run_id is None
+    if run_id is None:
+        latest_run_id = Flow(pathspec=flow_name).latest_run.id
+        msg = (
+            "Please specify a run-id using --run-id.\n"
+            "*%s*'s latest execution in namespace *%s* has id *%s*."
```
savingoyal, line 206: `..latest run...`

metaflow/plugins/tag_cli.py
```diff
@@ -0,0 +1,531 @@
...
+)
+@click.option(
+    "--hide-system-tags",
+    required=False,
+    is_flag=True,
+    default=False,
+    help="Hide system tags.",
+)
+@click.option(
+    "--by-tag",
```
savingoyal, line 441: `--group-by-tag` and similar for below?

metaflow/plugins/tag_cli.py
```diff
@@ -0,0 +1,531 @@
...
+    obj, run_id, hide_system_tags, list_all, my_runs, by_tag, by_run, arg_run_id
+):
+    if run_id is None and arg_run_id is None and not list_all and not my_runs:
+        # Assume list_all by default
+        list_all = True
+
+    if list_all and my_runs:
+        raise CommandException(
+            "Option --all cannot be used together with --my-runs.\n"
+            "You can run the flow with commands 'tag list --help' "
```
savingoyal, line 466: We can safely remove L:466 and 467.

metaflow/plugins/tag_cli.py
```diff
@@ -0,0 +1,531 @@
...
+            "to check the meaning of the two options."
+        )
+
+    if run_id is not None and arg_run_id is not None:
```
savingoyal, line 470: Is `arg_run_id` for nflx backwards compatibility? If so, can we notate that fact in the code?

savingoyal (approved):
LGTM, some minor changes.

metaflow/metadata/metadata.py
```diff
@@ -423,11 +484,17 @@ def _flow_to_json(self):
         return {"flow_id": self._flow_name, "ts_epoch": int(round(time.time() * 1000))}
 
     def _run_to_json(self, run_id=None, tags=None, sys_tags=None):
+        return MetadataProvider._run_to_json_static(
```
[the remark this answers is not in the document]
jackie-ob: It is a `@staticmethod`, distinct from the existing instance method.  We needed a static version of `_run_to_json` in order to keep `mutate_user_run_tags` a class method (i.e. we don't have an instance of any MetadataProvider, only the class).

Suggestions for better names welcome!

metaflow/metadata/metadata.py
```diff
@@ -406,10 +406,71 @@ def get_object(cls, obj_type, sub_type, filters, attempt, *args):
...
+        run_id: str
+            Run id, together with flow_id, that identifies the specific Run whose tags to mutate
+        """
+        # perform common validation, across all provider implementations
+        if tags_to_remove is None:
+            tags_to_remove = []
+        if tags_to_add is None:
+            tags_to_add = []
+        if not tags_to_add and not tags_to_remove:
+            return
```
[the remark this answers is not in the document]
jackie-ob: Good catch.  I will throw an error instead.  This helps us keep our return promise and will match the existing API in any case.

metaflow/plugins/tag_cli.py
jackie-ob: cc @romain-intel re: `arg_run_id` - should we keep it and add a comment, or can we remove it?

metaflow/plugins/tag_cli.py
jackie-ob: It seems reasonable to drop - unless it is well-beloved by existing users (cc @romain-intel )

metaflow/plugins/tag_cli.py
jackie-ob: OK!  cc @romain-intel minor API change here.

romain-intel:
Let me look at this but the tests won't pass. There is some top level error: `Tags must be strings: got ['mftest_driver']`.

metaflow/client/core.py
```diff
@@ -1717,6 +1747,83 @@ def end_task(self):
 
         return end_step.task
 
+    def add_tag(self, tag):
```
romain-intel, line 1750: This is the issue compared to the current version we have. This version makes more sense but may break a bunch of our flows. For consistency, we may be able to break our users but maybe it would be better to add an error message to indicate what changed. I am not a fan of this solution but I understand the desire to give the function a better name. And it was what I was going for in 547.

metaflow/metadata/metadata.py
```diff
@@ -406,10 +406,75 @@ def get_object(cls, obj_type, sub_type, filters, attempt, *args):
...
+        Run tags after mutation operations
+        """
+        # perform common validation, across all provider implementations
+        if tags_to_remove is None:
+            tags_to_remove = []
+        if tags_to_add is None:
+            tags_to_add = []
+        if not tags_to_add and not tags_to_remove:
+            raise MetaflowTaggingError("Must add or remove at least one tag")
+        if isinstance(tags_to_add, str):
```
romain-intel, line 435: We do check things here so maybe we can do something to keep both modes?

metaflow/plugins/metadata/local.py
```diff
@@ -112,6 +115,35 @@ def register_metadata(self, run_id, step_name, task_id, metadata):
         }
         self._save_meta(meta_dir, metadict)
 
+    @classmethod
+    def _mutate_user_tags_for_run(
```
romain-intel, line 119: This operation is non atomic. We should guarantee sequential ordering of tagging operations IMHO

metaflow/plugins/metadata/local.py
```diff
@@ -112,6 +115,35 @@ def register_metadata(self, run_id, step_name, task_id, metadata):
...
+        cls, flow_id, run_id, tags_to_add=None, tags_to_remove=None
+    ):
+        # get existing tags
+        run = LocalMetadataProvider.get_object("run", "self", {}, None, flow_id, run_id)
+        if not run:
+            raise MetaflowTaggingError(msg="Run not found (%s, %s)" % (flow_id, run_id))
+        existing_user_tags = run["tags"]
+        existing_system_tags = run["system_tags"]
+
+        for tag in tags_to_remove:
```
romain-intel, line 129: nit: can also use a set operation here (since you form the sets anyways below)

metaflow/plugins/metadata/local.py
```diff
@@ -112,6 +115,35 @@ def register_metadata(self, run_id, step_name, task_id, metadata):
...
+            if tag in existing_system_tags:
+                raise MetaflowTaggingError(
+                    msg="Cannot remove a tag that is an existing system tag (%s)" % tag
+                )
+
+        # make sure no existing system tags get added as a user tag
+        tags_to_add_set = set(tags_to_add) - set(existing_system_tags)
+        # remove tags first, then add
+        next_user_tags_set = (
```
romain-intel, line 138: So here, removing and adding the same tag would result in a no-op which I guess is what we want but do we warn the user?

metaflow/plugins/tag_cli.py
```diff
@@ -0,0 +1,531 @@
...
+
+from metaflow._vendor import click
+
+
+def _print_tags_for_runs_by_groups(
+    obj, system_tags_by_group, all_tags_by_group, by_tag
+):
+    header = [
+        "",
+        "System tags and *tags*; only *tags* can be modified",
```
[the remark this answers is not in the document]
romain-intel: The latest version I have in 547 is system tags and tags so I *think* that's what we agreed on (I vaguely remember a previous comment of yours saying there was no real notion of a user tag)

metaflow/plugins/tag_cli.py
romain-intel: It's a breaking API for us internally. I am fine with the replace-many syntax but would definitely prefer an optional thing ie that the replace <to-replace> <replacement> still works. If it is too complicated, we can see about breaking things internally. I do not think this command is very used so it may be ok. I'll check.

metaflow/plugins/tag_cli.py
romain-intel: We do have this currently so yes, it would be for backward compatibility. It seems easy enough to keep so would prefer if possible.

metaflow/plugins/tag_cli.py
```diff
@@ -0,0 +1,508 @@
...
+)
+@click.option(
+    "--hide-system-tags",
+    required=False,
+    is_flag=True,
+    default=False,
+    help="Hide system tags.",
+)
+@click.option(
+    "--group-by-tag",
```
romain-intel, line 413: We currently have this called `--by-tag` and `--by-run` instead of `--group-by-tag` and `--group-by-run`. Not sure how big a deal.

metaflow/plugins/tag_cli.py
romain-intel: You mean like `namespace = None` == `all namespaces` ? I don't mind but it seems a little possibly dangerous.

romain-intel (commented):
A few comments. I need to figure out the tests as well and limit breaking changes :)

metaflow/plugins/tag_cli.py
jackie-ob: It's not complicated to code up the args + opts way.  IMO the consideration here is: is this a good UX to expose in OSS (which gets us even deeper into the backward compatibility game, after this point):

* `tag replace --remove R1 --remove R2 --add A1 --add A2  R3 A3`

If we are able to accept the short term annoyance of dropping `R3 A3` above... I think there are long term benefits globally:
* I think it's confusing for CLI users to have to remember the ordering of R2, A2 above.
* Two ways to do the same thing (A1 and A2).

A middle way could be:
* Disallow simultaneous usage of options and positional args (throw error)
* Hide the positional way in usage / help print, so it becomes a grandfathered hidden feature.  New users of this will naturally only use the options.

What is folks' preferences?  cc @romain-intel @savingoyal @tuulos

metaflow/plugins/metadata/local.py
jackie-ob: Could you elaborate?  It looks atomic to me (We write to temporary file and then atomic rename it into `_self.json`).  Last write wins, all or nothing.

We will potentially drop concurrent updates.  E.g. two processes: one adds A, one adds B.... we may only end up seeing one of them in final result.  This is unavoidable without a serialization mechanism (e.g. locks).   IMO it is risky to do this, doesn't seem worth it here.  The Metadata service mutation operation will be atomic AND serializable (coming up soon).

metaflow/metadata/metadata.py
jackie-ob: We will do that per https://outerboundsco.slack.com/archives/C023Y9JH36W/p1653349445888399

metaflow/plugins/tag_cli.py
savingoyal: I am okay with the middle way if there are users of this functionality at Netflix.Otherwise we can drop `R3 A3`. @romain-intel can you confirm if there are any users using this feature?

metaflow/plugins/tag_cli.py
savingoyal: more like `--global-namespace`. We don't have to do it now - if a use case presents itself, we can always add it later.

savingoyal:
> Let me look at this but the tests won't pass. There is some top level error: `Tags must be strings: got ['mftest_driver']`.

@romain-intel this may be coming from the `is_stringish` check that this PR introduces.

metaflow/plugins/metadata/local.py
jackie-ob: Sounds good.

metaflow/plugins/metadata/local.py
jackie-ob: Discussed in this [slack thread](https://outerboundsco.slack.com/archives/C023Y9JH36W/p1653416648537399?thread_ts=1653349445.888399&cid=C023Y9JH36W).

As a way to provide a more predictable UX (no dropped changes) whilst avoiding locks, I have added a simple retry with backoff until the callers sees what it expects post operation.

metaflow/plugins/tag_cli.py
jackie-ob: OK resolving this convo.  We know where to look later.

metaflow/plugins/tag_cli.py
```diff
@@ -0,0 +1,484 @@
...
+    multiple=True,
+    default=None,
+    help="Add this tag to a run. Must specify one or more tags to add.",
+)
+@click.option(
+    "--remove",
+    "tags_to_remove",
+    multiple=True,
+    default=None,
+    help="Remove this tag to a run. Must specify one or more tags to remove.",
```
savingoyal, line 324: `from a run`

metaflow/plugins/tag_cli.py
```diff
@@ -0,0 +1,484 @@
...
+        raise CommandException(msg)
+    return run
+
+
+@click.group()
+def cli():
+    pass
+
+
+@cli.group(help="Commands related to Metaflow tagging.")
```
tuulos, line 231: let's have "Commands related to tagging" - Metaflow is quite implied here :)

metaflow/metadata/metadata.py
```diff
@@ -423,10 +424,86 @@ def get_object(cls, obj_type, sub_type, filters, attempt, *args):
...
+                return False
+
+        if not _is_iterable(tags_to_add):
+            raise MetaflowTaggingError("tags_to_add must be iterable")
+        if not _is_iterable(tags_to_remove):
+            raise MetaflowTaggingError("tags_to_remove must be iterable")
+
+        # check each tag is valid
+        for tag in chain(tags_to_add, tags_to_remove):
+            validate_tag(tag)
```
jackie-ob, line 478: This is generic tag value validation, cuts across all `MetadataProviders`

metaflow/plugins/metadata/local.py
```diff
@@ -115,6 +119,88 @@ def register_metadata(self, run_id, step_name, task_id, metadata):
...
+                existing_user_tag_set - tags_to_remove_set
+            ) | tags_to_add_set
+            # we think it will be a no-op, so let's return right away
+            if next_user_tags_set == existing_user_tag_set:
+                return MutationResult(
+                    tags=next_user_tags_set,
+                    tags_are_consistent=True,
+                )
+
+            validate_tags(next_user_tags_set, existing_tags=existing_user_tag_set)
```
jackie-ob, line 164: `LocalMetadataProvider` validates tag set sizing.  Yes we will implicitly check all the tag values again.  Not a big deal, keeps code simple.

jackie-ob:
I tested argo and step functions 'create' step manually - it works fine.  Note we fail as early as possible - the moment we receive the tags tuple from click.

PR #1051, opened 2022-05-24
Run id file should be written prior to execution when resuming

jackie-ob:
This matches behavior for a non-resume (regular `flow.py run`).  This way is simply more useful, following user feedback.

Core test added.


savingoyal (approved):
LGTM!

jackie-ob:
+ @romain-intel NFLX tests show one failure.  Could you please confirm if things are OK with this change?

romain-intel (approved):
LGTM. Ignore test failure; it's a known flaky thing which should go away at some point.

PR #1053, opened 2022-05-26
Fix instance metadata calls for IMDSV2

savingoyal:


russellbrooks (approved):
Worked like a charm after building a test docker image with these changes, syncing to ECR, and redeploying one of our SFNs with the `@batch` decorators specifying to use the test image :rocket:

PR #1056, opened 2022-05-27
try data tests on 3.10 ubuntu

jackie-ob:


.github/workflows/s3_test.yml
```diff
@@ -17,17 +17,17 @@ jobs:
...
-        os: [macos-latest]
+        os: [macos-latest, ubuntu-latest]
 
     steps:
     - uses: actions/checkout@v2
       with:
         ref: refs/pull/${{ github.event.pull_request.number }}/merge
         submodules: recursive
-    - name: Set up Python 3.7
+    - name: Set up Python 3.10
```
savingoyal, line 27: We can consider running the tests on multiple Python versions.

savingoyal:
there is #699 that can be closed once this PR makes its way in.

.github/workflows/s3_test.yml
jackie-ob: Yes good idea.

I was debugging why the tests were erroring out for me, hence trying this 3.10 thing out.  Turns out it was something else.

I will close this draft.

PR #1057, opened 2022-06-02
Improve sidecar message handling

romain-intel:
Several changes:
  - allow side-passing of a static "context" to the sidecar process to keep
    messages as small as possible
  - clean up the interfaces and make it clearer what is what (ie: there is
    an emitter inside the process and then a sidecar that runs outside)
  - better error checking
  - better handling of shutdown (sidecars will now properly get a shutdown
    message giving them the opportunity to clean up -- this may add a bit
    of latency at the end of the program execution.

Also improved message when plugins don't load in case of an issue with
Metaflow extensions

### DIFF
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -943,11 +943,11 @@ def start(
     ctx.obj.environment.validate_environment(echo)
 
     ctx.obj.event_logger = EventLogger(
-        event_logger, ctx.obj.environment, ctx.obj.flow.name
+        event_logger, ctx.obj.environment, flow_name=ctx.obj.flow.name
     )
     ctx.obj.event_logger.start()
 
-    ctx.obj.monitor = Monitor(monitor, ctx.obj.environment, ctx.obj.flow.name)
+    ctx.obj.monitor = Monitor(monitor, ctx.obj.environment, flow_name=ctx.obj.flow.name)
     ctx.obj.monitor.start()
 
     ctx.obj.metadata = [m for m in METADATA_PROVIDERS if m.TYPE == metadata][0](
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -937,12 +937,15 @@ def start(
     ctx.obj.package_suffixes = package_suffixes.split(",")
     ctx.obj.reconstruct_cli = _reconstruct_cli
 
-    ctx.obj.event_logger = EventLogger(event_logger)
-
     ctx.obj.environment = [
         e for e in ENVIRONMENTS + [MetaflowEnvironment] if e.TYPE == environment
     ][0](ctx.obj.flow)
     ctx.obj.environment.validate_environment(echo)
+
+    ctx.obj.event_logger = EventLogger(
+        event_logger, ctx.obj.environment, ctx.obj.flow.name
+    )
+    ctx.obj.event_logger.start()
 
     ctx.obj.monitor = Monitor(monitor, ctx.obj.environment, ctx.obj.flow.name)
     ctx.obj.monitor.start()
@@ -1116,3 +1119,8 @@ def main(flow, args=None, handle_exceptions=True, entrypoint=None):
             sys.exit(1)
         else:
             raise
+    finally:
+        if hasattr(state, "monitor") and state.monitor is not None:
+            state.monitor.terminate()
+        if hasattr(state, "event_logger") and state.event_logger is not None:
+            state.event_logger.terminate()
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -44,8 +44,6 @@ from .metaflow_config import (
 )
 from .metaflow_environment import MetaflowEnvironment
 from .pylint_wrapper import PyLint
-from .event_logger import EventLogger
-from .monitor import Monitor
 from .R import use_r, metaflow_r_version
 from .mflog import mflog, LOG_SOURCES
 from .unbounded_foreach import UBF_CONTROL, UBF_TASK
@@ -942,12 +940,12 @@ def start(
     ][0](ctx.obj.flow)
     ctx.obj.environment.validate_environment(echo)
 
-    ctx.obj.event_logger = EventLogger(
-        event_logger, ctx.obj.environment, flow_name=ctx.obj.flow.name
+    ctx.obj.event_logger = LOGGING_SIDECARS[event_logger](
+        ctx.obj.flow, ctx.obj.environment
     )
     ctx.obj.event_logger.start()
 
-    ctx.obj.monitor = Monitor(monitor, ctx.obj.environment, flow_name=ctx.obj.flow.name)
+    ctx.obj.monitor = MONITOR_SIDECARS[monitor](ctx.obj.flow, ctx.obj.environment)
     ctx.obj.monitor.start()
 
     ctx.obj.metadata = [m for m in METADATA_PROVIDERS if m.TYPE == metadata][0](
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -941,11 +941,13 @@ def start(
     ctx.obj.environment.validate_environment(echo)
 
     ctx.obj.event_logger = LOGGING_SIDECARS[event_logger](
-        ctx.obj.flow, ctx.obj.environment
+        flow=ctx.obj.flow, env=ctx.obj.environment
     )
     ctx.obj.event_logger.start()
 
-    ctx.obj.monitor = MONITOR_SIDECARS[monitor](ctx.obj.flow, ctx.obj.environment)
+    ctx.obj.monitor = MONITOR_SIDECARS[monitor](
+        flow=ctx.obj.flow, env=ctx.obj.environment
+    )
     ctx.obj.monitor.start()
 
     ctx.obj.metadata = [m for m in METADATA_PROVIDERS if m.TYPE == metadata][0](
--- a/metaflow/event_logger.py
+++ b/metaflow/event_logger.py
@@ -17,17 +17,24 @@ class NullEventLogger(object):
 
 
 class EventLogger(NullEventLogger):
-    def __init__(self, logger_type):
-        # type: (str) -> None
+    def __init__(self, logger_type, env, flow_name):
+        # type: (str, str, str) -> None
         self.sidecar_process = None
         self.logger_type = logger_type
+        self.env_info = env.get_environment_info()
+        self.env_info["flow_name"] = flow_name
 
     def start(self):
-        self.sidecar_process = SidecarSubProcess(self.logger_type)
+        if self.sidecar_process is None:
+            self.sidecar_process = SidecarSubProcess(
+                self.logger_type, {"env": self.env_info}
+            )
 
     def log(self, payload):
-        msg = Message(MessageTypes.LOG_EVENT, payload)
-        self.sidecar_process.msg_handler(msg)
+        if self.sidecar_process is not None:
+            msg = Message(MessageTypes.LOG_EVENT, payload)
+            self.sidecar_process.send(msg)
 
     def terminate(self):
-        self.sidecar_process.kill()
+        if self.sidecar_process is not None:
+            self.sidecar_process.kill()
--- a/metaflow/event_logger.py
+++ b/metaflow/event_logger.py
@@ -9,6 +9,9 @@ class NullEventLogger(object):
     def start(self):
         pass
 
+    def add_context(self, **kwargs):
+        pass
+
     def log(self, payload):
         pass
 
@@ -17,18 +20,26 @@ class NullEventLogger(object):
 
 
 class EventLogger(NullEventLogger):
-    def __init__(self, logger_type, env, flow_name):
+    def __init__(self, logger_type, env, **kwargs):
         # type: (str, str, str) -> None
         self.sidecar_process = None
         self.logger_type = logger_type
-        self.env_info = env.get_environment_info()
-        self.env_info["flow_name"] = flow_name
+        self._context = {"env": env.get_environment_info()}
+        # The additional keywords are added to the context for the monitor.
+        # One use case is adding tags like flow_name etc.
+        if kwargs:
+            # Make sure we copy it as we don't want it to change
+            self._context["user_context"] = dict(kwargs)
 
     def start(self):
         if self.sidecar_process is None:
-            self.sidecar_process = SidecarSubProcess(
-                self.logger_type, {"env": self.env_info}
-            )
+            self.sidecar_process = SidecarSubProcess(self.logger_type, self._context)
+
+    def add_context(self, **kwargs):
+        self._context["user_context"].update(kwargs)
+        if self.sidecar_process is not None:
+            msg = Message(MessageTypes.UPDATE_CONTEXT, self._context)
+            self.sidecar_process.send(msg)
 
     def log(self, payload):
         if self.sidecar_process is not None:
--- a/metaflow/event_logger.py
+++ b/metaflow/event_logger.py
@@ -1,21 +1,31 @@
 from metaflow.sidecar import Message, MessageTypes, Sidecar
 
 
-class BaseEventLogger(Sidecar):
+class BaseEventLogger(object):
+    TYPE = "nullSidecarLogger"
+
+    def __init__(self, flow, env):
+        self._sidecar = Sidecar(self.TYPE)
+
+    def start(self):
+        return self._sidecar.start()
+
+    def terminate(self):
+        return self._sidecar.terminate()
+
+    def add_context(self, context):
+        pass
+
     def log(self, payload):
-        if self._is_active:
-            msg = Message(MessageTypes.EVENT, payload)
-            self._send(msg)
+        if self._sidecar.is_active:
+            msg = Message(MessageTypes.BEST_EFFORT, payload)
+            self._sidecar.send(msg)
+
+    @classmethod
+    def get_worker(cls):
+        return None
 
 
 # Backward compatible name
 class EventLogger(BaseEventLogger):
     pass
-
-
-class NullLogger(BaseEventLogger):
-    TYPE = "nullSidecarLogger"
-
-    @classmethod
-    def get_sidecar_worker_class(cls):
-        return None
--- a/metaflow/event_logger.py
+++ b/metaflow/event_logger.py
@@ -4,7 +4,7 @@ from metaflow.sidecar import Message, MessageTypes, Sidecar
 class BaseEventLogger(object):
     TYPE = "nullSidecarLogger"
 
-    def __init__(self, flow, env):
+    def __init__(self, flow=None, env=None):
         self._sidecar = Sidecar(self.TYPE)
 
     def start(self):
@@ -13,7 +13,7 @@ class BaseEventLogger(object):
     def terminate(self):
         return self._sidecar.terminate()
 
-    def add_context(self, context):
+    def add_to_context(self, **kwargs):
         pass
 
     def log(self, payload):
--- a/metaflow/event_logger.py
+++ b/metaflow/event_logger.py
@@ -1,10 +1,11 @@
 from metaflow.sidecar import Message, MessageTypes, Sidecar
 
 
-class BaseEventLogger(object):
+class NullEventLogger(object):
     TYPE = "nullSidecarLogger"
 
-    def __init__(self, flow=None, env=None):
+    def __init__(self, *args, **kwargs):
+        # Currently passed flow and env in kwargs
         self._sidecar = Sidecar(self.TYPE)
 
     def start(self):
@@ -13,7 +14,9 @@ class BaseEventLogger(object):
     def terminate(self):
         return self._sidecar.terminate()
 
-    def add_to_context(self, **kwargs):
+    def send(self, msg):
+        # Arbitrary message sending. Useful if you want to override some different
+        # types of messages.
         pass
 
     def log(self, payload):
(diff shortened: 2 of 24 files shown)


metaflow/plugins/debug_logger.py
```diff
@@ -1,17 +1,19 @@
...
+        pass
 
     def process_message(self, msg):
         # type: (Message) -> None
-        self.log(msg.payload)
+        if msg.msg_type == MessageTypes.SHUTDOWN:
+            print("Debug event: got shutdown!")
+            self._shutdown()
+        elif msg.msg_type == MessageTypes.LOG_EVENT:
+            print("Debug event logging %s" % str(msg.payload), file=sys.stderr)
```
savingoyal, line 16: minor nit - `Debug event: %s`

metaflow/plugins/debug_logger.py
```diff
@@ -1,17 +1,19 @@
...
-        sys.stderr.write("event_logger: " + str(msg) + "\n")
+class DebugEventLoggerSidecar(object):
+    def __init__(self):
+        pass
 
     def process_message(self, msg):
         # type: (Message) -> None
-        self.log(msg.payload)
+        if msg.msg_type == MessageTypes.SHUTDOWN:
+            print("Debug event: got shutdown!")
```
savingoyal, line 13: Minor nit - `Debug[shutdown]: got shutdown!`

metaflow/event_logger.py
```diff
@@ -17,17 +20,32 @@ def terminate(self):
 
 
 class EventLogger(NullEventLogger):
-    def __init__(self, logger_type):
-        # type: (str) -> None
+    def __init__(self, logger_type, env, **kwargs):
+        # type: (str, str, str) -> None
         self.sidecar_process = None
         self.logger_type = logger_type
+        self._context = {"env": env.get_environment_info()}
```
savingoyal, line 27: Can't this be done in the constructor for `KeyStoneLogger`?

metaflow/sidecar_worker.py
```diff
@@ -11,33 +11,32 @@
...
             else:
-                raise WorkershutdownError()
-        except WorkershutdownError:
-            break
-        except Exception as e:  # todo handle other possible exceptions gracefully
-            print(traceback.format_exc())
+                break
+
+        except:  # todo handle other possible exceptions gracefully
+            print(traceback.format_exc(), file=sys.stderr)
```
savingoyal, line 39: minor nit - `[sidecar_process:worker-type] exception`

metaflow/sidecar_worker.py
```diff
@@ -11,33 +11,32 @@
...
         try:
             msg = sys.stdin.readline().strip()
             if msg:
-                parsed_msg = deserialize(msg)
-                if parsed_msg.msg_type == MessageTypes.SHUTDOWN:
-                    raise WorkershutdownError()
+                parsed_msg = Message.deserialize(msg)
+                if parsed_msg.msg_type == MessageTypes.INVALID:
+                    print(
+                        "[sidecarProcess:%s] Invalid message -- skipping: %s"
```
savingoyal, line 27: minor nit - `sidecar_process...`

metaflow/cli.py
```diff
@@ -937,14 +937,17 @@ def start(
...
     ctx.obj.environment = [
         e for e in ENVIRONMENTS + [MetaflowEnvironment] if e.TYPE == environment
     ][0](ctx.obj.flow)
     ctx.obj.environment.validate_environment(echo)
 
-    ctx.obj.monitor = Monitor(monitor, ctx.obj.environment, ctx.obj.flow.name)
+    ctx.obj.event_logger = EventLogger(
+        event_logger, ctx.obj.environment, flow_name=ctx.obj.flow.name
+    )
+    ctx.obj.event_logger.start()
```
savingoyal, line 948: For my own curiosity - what additional events do we now capture by starting the logging here compared to before?

metaflow/sidecar.py
```diff
@@ -48,109 +49,184 @@ def poll(self, timeout):
...
-        self.__process = None
-        self.__poller = None
+    def __init__(self, worker_type, context=None):
+        # type: (str, dict) -> None
+        self._worker_type = worker_type
+        self._process = None
+        self._poller = None
+        self._context = context
+        self._send_context_remaining_tries = (
+            CONTEXT_RETRY_TIMES if self._context is not None else 0
```
savingoyal, line 59: This seems highly specific to the logger use case here.

metaflow/sidecar_messages.py
```diff
@@ -4,21 +4,26 @@
...
+    INVALID, START, UPDATE_CONTEXT, SHUTDOWN, LOG_EVENT = range(1, 6)
 
 
 class Message(object):
     def __init__(self, msg_type, payload):
         self.msg_type = msg_type
         self.payload = payload
 
-    def serialize(self):
+    def serialize(self, clear_previous=False):
```
savingoyal, line 15: `clear_previous`? Wherever you are using the output of (erstwhile)`serialize`, can't you just prepend a new line right there rather than bleeding in the logic here?

metaflow/event_logger.py
```diff
@@ -17,17 +20,32 @@ def terminate(self):
...
     def start(self):
-        self.sidecar_process = SidecarSubProcess(self.logger_type)
+        if self.sidecar_process is None:
+            self.sidecar_process = SidecarSubProcess(self.logger_type, self._context)
+
+    def add_context(self, **kwargs):
+        self._context["user_context"].update(kwargs)
+        if self.sidecar_process is not None:
+            msg = Message(MessageTypes.UPDATE_CONTEXT, self._context)
+            self.sidecar_process.send(msg)
```
savingoyal, line 42: Can't this be something that the `KeyStoneLogger` can handle by itself at startup?

metaflow/monitor.py
```diff
@@ -34,49 +36,59 @@ def terminate(self):
...
 class Monitor(NullMonitor):
-    def __init__(self, monitor_type, env, flow_name):
-        # type: (str) -> None
+    def __init__(self, monitor_type, env, **kwargs):
+        # type: (str, str, str) -> None
         self.sidecar_process = None
         self.monitor_type = monitor_type
-        self.env_info = env.get_environment_info()
-        self.env_info["flow_name"] = flow_name
+        self._context = {"env": env.get_environment_info()}
```
savingoyal, line 43: Similar to the logger feedback - can the `AtlasMonitor` assume the responsibility for setting the right context from env and what not?

metaflow/sidecar.py
```diff
@@ -48,109 +49,184 @@ def poll(self, timeout):
...
                 # unable to start subprocess, fallback to Null sidecar
-                self.logger(
-                    "unable to start subprocess for sidecar %s" % self.__worker_type
-                )
-                self.__poller = NullPoller()
+                self._logger("Unable to start subprocess")
+                self._poller = NullPoller()
+            self._starting = False
+
+    def update_context(self, new_context):
```
savingoyal, line 114: Are the `context` related segments really needed in this module?

metaflow/sidecar.py
```diff
@@ -48,109 +49,184 @@ def poll(self, timeout):
...
+                self._send_context_remaining_tries = 0
+            else:
+                self._send_context_remaining_tries -= 1
+                if self._send_context_remaining_tries == 0:
+                    # Mark as "failed after try"
+                    self._send_context_remaining_tries = -1
+
+    def _emit_msg(self, msg):
+        # If the previous message had an error, we want to prepend a "\n" to this message
+        # to maximize the chance of this message being valid (for example, if the
```
savingoyal, line 211: not sure I follow the reasoning (as well as why this is needed)? This seems like an issue with constructing the payload. Can't the logger/debugger implementations take care of this?

metaflow/event_logger.py
romain-intel: of course. I was just trying to be consistent with the monitor. I am fine either doing both in subclasses or both in the "base classes"

metaflow/cli.py
romain-intel: Right now, nothing since we log all of two events. We are working on logging a lot more to get more insight on what is happening in the runtime. It's also more consistent with the monitor part.

metaflow/sidecar.py
romain-intel: How so? It basically ensures that the context is sent prior to sending a message. Or at least gives it a real go at sending it.

metaflow/sidecar_messages.py
romain-intel: Yep, can do that. I thought this was a bit cleaner but happy to move it out.

metaflow/event_logger.py
romain-intel: The use case for this is to update the context after creation so it needs to be somewhat in the base class. Sidecars are free to ignore though. I am fine if the specific context (like environment) is in the subclass but the notion of a context should be base class stuff IMHO.

metaflow/monitor.py
romain-intel: yes, am fine moving both out. We had it this way already so my modus operandi was "don't remove stuff" but am happy to remove.

metaflow/sidecar_messages.py
```diff
@@ -4,21 +4,26 @@
 # Unfortunately we can't use enums because they are not supported
 # officially in Python2
 class MessageTypes(object):
-    SHUTDOWN, LOG_EVENT = range(1, 3)
+    INVALID, START, UPDATE_CONTEXT, SHUTDOWN, LOG_EVENT = range(1, 6)
```
savingoyal, line 7: Remove `START`

savingoyal: And change `UPDATE_CONTEXT` to `CONTEXT` and `LOG_EVENT` to `EVENT`

metaflow/event_logger.py
romain-intel: I have refactored all this.

metaflow/event_logger.py
```diff
@@ -1,33 +1,21 @@
...
 
 
-class NullEventLogger(object):
-    def __init__(self, *args, **kwargs):
-        pass
-
-    def start(self):
-        pass
-
+class BaseEventLogger(Sidecar):
```
savingoyal, line 4: Can this be just `EventLogger`?

savingoyal: Also, what necessitated changes in this class?

metaflow/metadata/heartbeat.py
```diff
@@ -20,27 +20,30 @@ def __init__(self, msg):
...
         self.hb_url = None
 
     def process_message(self, msg):
         # type: (Message) -> None
         if msg.msg_type == MessageTypes.SHUTDOWN:
-            # todo shutdown doesnt do anything yet? should it still be called
-            self.shutdown()
-        if (not self.req_thread.is_alive()) and msg.msg_type == MessageTypes.LOG_EVENT:
+            self._shutdown()
+        if (not self.req_thread.is_alive()) and msg.msg_type == MessageTypes.CONTEXT:
```
savingoyal, line 32: What happens if there is no `Context` message ever?

metaflow/plugins/metadata/service.py
```diff
@@ -137,8 +127,16 @@ def _start_heartbeat(
...
+        # start sidecar
+        if self.version() is None or LooseVersion(self.version()) < LooseVersion(
+            "2.0.4"
+        ):
+            # if old version of the service is running
+            # then avoid running real heartbeat sidecar process
+            self.sidecar_process = SidecarSubProcess("nullSidecarHeartbeat")
+        else:
+            self.sidecar_process = SidecarSubProcess("heartbeat")
+        self.sidecar_process.send(Message(MessageTypes.CONTEXT, payload))
```
savingoyal, line 139: Isn't the payload an `Event` here?

metaflow/event_logger.py
```diff
@@ -1,33 +1,21 @@
...
     def log(self, payload):
-        pass
+        if self._is_active:
```
savingoyal, line 6: All methods in `*_logger` and `*_monitor` have this conditional. Is this needed? In what scenario would a logger be assigned but disabled?

metaflow/mflog/save_logs_periodically.py
```diff
@@ -16,10 +15,12 @@ def __init__(self):
         self._thread.start()
 
     def process_message(self, msg):
-        pass
+        if msg.msg_type == MessageTypes.SHUTDOWN:
```
savingoyal, line 18: What's the intention behind this change? In other words - why get rid of the explicit `shutdown` method?

metaflow/sidecar/sidecar.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class Sidecar(object):
+    TYPE = ""
+
+    def __init__(self, flow, env):
+        self.sidecar_process = None
+
+    def start(self):
+        if not self._is_active and self.get_sidecar_worker_class() is not None:
```
savingoyal, line 13: When is `self.get_sidecar_worker_class()` `None`?

metaflow/event_logger.py
romain-intel: It's more consistent with BaseMonitor but it's not really an issue. I alias it because for some reason it is exposed at the TL (I don't think *anyone* uses it since it is totally undocumented).

metaflow/mflog/save_logs_periodically.py
romain-intel: it was never called.

metaflow/plugins/metadata/service.py
romain-intel: It doesn't really matter in this case. Happy to make it an EVENT. I considered it to be the start context basically since there are no real events but it doesn't really matter.

metaflow/sidecar/sidecar.py
romain-intel: for the Null* ones. This is an optimization of sorts -- we don't start the subprocess. It was like that before so kept that behavior.

metaflow/event_logger.py
romain-intel: The Null loggers.

metaflow/metadata/heartbeat.py
```diff
@@ -20,27 +20,30 @@ def __init__(self, msg):
...
         self.hb_url = None
 
     def process_message(self, msg):
         # type: (Message) -> None
         if msg.msg_type == MessageTypes.SHUTDOWN:
-            # todo shutdown doesnt do anything yet? should it still be called
-            self.shutdown()
-        if (not self.req_thread.is_alive()) and msg.msg_type == MessageTypes.LOG_EVENT:
+            self._shutdown()
+        if (not self.req_thread.is_alive()) and msg.msg_type == MessageTypes.MUST_SEND:
```
savingoyal, line 32: The heartbeats are best effort. `BEST_EFFORT_SEND`?

metaflow/monitor.py
```diff
@@ -2,170 +2,158 @@
...
 
-class NullMonitor(object):
-    def __init__(self, *args, **kwargs):
-        pass
+class BaseMonitor(object):
+    TYPE = "nullSidecarMonitor"
 
-    def start(self):
-        pass
+    def __init__(self, flow=None, env=None):
```
savingoyal, line 15: can this be `(self, *args, **kwargs)`?

metaflow/metadata/heartbeat.py
romain-intel: this is the heartbeat configuration. It doesn't really make a difference though so I can change it. Semantically I thought that the configuration was more of a MUST_SEND but since there are no other messages sent anyways, it has no impact.

metaflow/monitor.py
```diff
@@ -2,170 +2,158 @@
...
 COUNTER_TYPE = "COUNTER"
 GAUGE_TYPE = "GAUGE"
-MEASURE_TYPE = "MEASURE"
 TIMER_TYPE = "TIMER"
 
 
-class NullMonitor(object):
-    def __init__(self, *args, **kwargs):
-        pass
+class BaseMonitor(object):
```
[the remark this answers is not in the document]
romain-intel: I didn't like inheriting from NullMonitor since it was kind of a bit weird but I don't have a strong preference.

metaflow/monitor.py
romain-intel: sure. I can make it more generic (and for the logger).

savingoyal: `flow` and `env` are not at all used anywhere in this constructor so it's a bit weird seeing them here.

metaflow/monitor.py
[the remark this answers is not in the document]
savingoyal: Yeah, I have a slight preference for `NullMonitor` since this is how it is invoked on the CLI currently.

metaflow/monitor.py
```diff
@@ -2,170 +2,158 @@
...
-        yield
-
-    @contextmanager
-    def measure(self, name):
-        yield
-
-    def gauge(self, gauge):
-        pass
+    def start(self):
+        return self._sidecar.start()
```
savingoyal, line 19: Wouldn't you want to define the count, measure, gauge methods in this super class?

metaflow/event_logger.py
```diff
@@ -1,33 +1,31 @@
...
 
+    def add_to_context(self, **kwargs):
+        pass
 
-class EventLogger(NullEventLogger):
-    def __init__(self, logger_type):
-        # type: (str) -> None
-        self.sidecar_process = None
-        self.logger_type = logger_type
+    def log(self, payload):
```
savingoyal, line 19: we can combine `log` and `add_to_context` into `log(self, message)`

metaflow/monitor.py
```diff
@@ -2,170 +2,158 @@
...
-        # type: (str) -> None
-        self.sidecar_process = None
-        self.monitor_type = monitor_type
-        self.env_info = env.get_environment_info()
-        self.env_info["flow_name"] = flow_name
-
-    def start(self):
-        if self.sidecar_process is None:
-            self.sidecar_process = SidecarSubProcess(self.monitor_type)
+    def add_to_context(self, **kwargs):
```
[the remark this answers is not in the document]
savingoyal: see https://github.com/Netflix/metaflow/pull/1057/files/25fd4246ffb5311a2d5af2771875086c6e775a45#r898620435

metaflow/metadata/heartbeat.py
savingoyal: in that case, do you need to check for the message type here?

metaflow/plugins/debug_logger.py
```diff
@@ -1,17 +1,28 @@
...
+        pass
 
     def process_message(self, msg):
         # type: (Message) -> None
-        self.log(msg.payload)
+        if msg.msg_type == MessageTypes.SHUTDOWN:
+            print("Debug[shutdown]: got shutdown!", file=sys.stderr)
+            self._shutdown()
+        elif msg.msg_type == MessageTypes.BEST_EFFORT:
+            print("Debug[event]: %s" % str(msg.payload), file=sys.stderr)
```
savingoyal, line 25: also print the `MUST_SEND_EVENT` message?

metaflow/plugins/debug_monitor.py
```diff
@@ -1,43 +1,35 @@
...
-        else:
-            pass
-
-    def shutdown(self):
-        sys.stderr.flush()
-
-    def logger(self, msg):
-        print("local_monitor: %s" % msg, file=sys.stderr)
+        if msg.msg_type == MessageTypes.MUST_SEND:
+            self._context = msg.payload
```
savingoyal, line 22: We can just print both the `MUST_SEND` and `BEST_EFFORT` messages here in this implementation

metaflow/plugins/metadata/service.py
```diff
@@ -137,8 +127,17 @@ def _start_heartbeat(
...
+        if self.version() is None or LooseVersion(self.version()) < LooseVersion(
+            "2.0.4"
+        ):
+            # if old version of the service is running
+            # then avoid running real heartbeat sidecar process
+            self.sidecar = Sidecar("none")
+        else:
+            self.sidecar = Sidecar("heartbeat")
+        self.sidecar.start()
+        self.sidecar.send(Message(MessageTypes.MUST_SEND, payload))
```
savingoyal, line 140: `BEST_EFFORT` perhaps?

metaflow/plugins/debug_monitor.py
romain-intel: fixed.

metaflow/plugins/debug_logger.py
romain-intel: fixed.

metaflow/event_logger.py
romain-intel: As discussed, I kept log as is and added send(self, msg). Also added send(self, msg) to the monitor.

metaflow/monitor.py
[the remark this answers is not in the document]
romain-intel: resolved.

metaflow/monitor.py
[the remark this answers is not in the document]
romain-intel: I changed.

metaflow/metadata/heartbeat.py
romain-intel: not really. I'll remove.

metaflow/plugins/metadata/service.py
romain-intel: changed. makes no real difference.

metaflow/monitor.py
romain-intel: They are, they start at line 28 in this new version of the file.

PR #1059, opened 2022-06-03
Add two options to resume: the ability to specify a run-id and the ab…

romain-intel:
…ility to only clone tasks

The scenario for this use is mainly for external schedulers that are
capable of "resuming" failed runs by only re-executing the ones that failed
or didn't execute the first time. We need a way for Metaflow to "clone" the
tasks that are not re-executed.

Also took the opportunity to fix a few TODOs related to status updates for local runs
as well as some tiny cleanups.

### DIFF
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -674,6 +674,18 @@ def common_run_options(func):
     help="ID of the run that should be resumed. By default, the "
     "last run executed locally.",
 )
+@click.option(
+    "--run-id",
+    default=None,
+    help="Run ID for the new run. By default, a new run-id will be generated",
+    hidden=True,
+)
+@click.option(
+    "--clone-only/--no-clone-only",
+    default=False,
+    show_default=True,
+    help="Only clone tasks without continuing execution",
+)
 @click.argument("step-to-rerun", required=False)
 @cli.command(help="Resume execution of a previous run of this flow.")
 @common_run_options
@@ -683,6 +695,8 @@ def resume(
     tags=None,
     step_to_rerun=None,
     origin_run_id=None,
+    run_id=None,
+    clone_only=False,
     max_workers=None,
     max_num_splits=None,
     max_log_size=None,
@@ -712,6 +726,17 @@ def resume(
             )
         clone_steps = {step_to_rerun}
 
+    if run_id:
+        # Run-ids that are provided by the metadata service are always integers.
+        # External providers or run-ids (like external schedulers) always need to
+        # be non-integers to avoid any clashes. This condition ensures this.
+        try:
+            int(run_id)
+        except:
+            pass
+        else:
+            raise CommandException("run-id %s cannot be an integer" % run_id)
+
     runtime = NativeRuntime(
         obj.flow,
         obj.graph,
@@ -723,7 +748,9 @@ def resume(
         obj.entrypoint,
         obj.event_logger,
         obj.monitor,
+        run_id=run_id,
         clone_run_id=origin_run_id,
+        clone_only=clone_only,
         clone_steps=clone_steps,
         max_workers=max_workers,
         max_num_splits=max_num_splits,
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -700,6 +700,7 @@ def common_run_options(func):
     default=False,
     show_default=True,
     help="Only clone tasks without continuing execution",
+    hidden=True,
 )
 @click.option(
     "--reentrant/--no-reentrant",
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -252,7 +252,7 @@ def dump(obj, input_path, private=None, max_value_size=None, include=None, file=
         run_id, step_name, task_id = parts
     else:
         raise CommandException(
-            "input_path should either be run_id/step_name" "or run_id/step_name/task_id"
+            "input_path should either be run_id/step_name or run_id/step_name/task_id"
         )
 
     datastore_set = TaskDataStoreSet(
@@ -398,7 +398,7 @@ def logs(obj, input_path, stdout=None, stderr=None, both=None, timestamps=False)
                     log = ds.load_log_legacy(stream)
                     if log and timestamps:
                         raise CommandException(
-                            "We can't show --timestamps for " "old runs. Sorry!"
+                            "We can't show --timestamps for old runs. Sorry!"
                         )
                     echo_unicode(log, nl=False)
     else:
@@ -428,7 +428,7 @@ def logs(obj, input_path, stdout=None, stderr=None, both=None, timestamps=False)
 )
 @click.option(
     "--input-paths",
-    help="A comma-separated list of pathspecs " "specifying inputs for this step.",
+    help="A comma-separated list of pathspecs specifying inputs for this step.",
 )
 @click.option(
     "--split-index",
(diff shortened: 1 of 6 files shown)


metaflow/runtime.py
```diff
@@ -241,6 +288,11 @@ def execute(self):
         # assert that end was executed and it was successful
         if ("end", ()) in self._finished:
             self._logger("Done!", system_msg=True)
+        elif self._clone_only:
+            self._logger(
+                "Clone-only resume complete -- some steps did not execute!",
```
savingoyal, line 293: No steps should have executed?

romain-intel: I will rephrase but basically it didn't reach end. So the steps that "executed" are only clones.

romain-intel:
Rebased and addressed both comments.

romain-intel:
To add a bit of context of how this is used. When the scheduler resumes a flow, it has a "resume from incomplete" mode which effectively "skips" everything that completed successfully and creates a graph that effectively starts with node(s) that were not successful the last time around. The way we do this in Metaflow is to create a new run and effectively "resume" the old one but only the successful parts so that the scheduler can then schedule the previously unsuccessful parts. The options present here therefore enable us to:
- specify a new run id (given by the scheduler) for the new run (that will partially clone the old one)
- allow the resume command to basically just do the cloning part and return so that the scheduler may take over and do the rest.

We add this to the command line prior to actually executing the step. So the command line looks semantically equivalent to:
```
if need_to_resume:
  myflow.py resume --clone-only --run-id <new id> --origin-run-id <old id>
end
myflow.py step my_step ....
```

The re-entrant bit comes in because the resumed graph may have multiple heads now (think of a foreach where more than one of the branches in the foreach need to be restarted, or, more simply, of just a simple branch where both sides failed). With the command line above, the logic would have the resume code run twice. We actually don't want to run twice BUT we want to make sure that we block until the clone is complete. The re-entrant flag basically does this (it is more computationally expensive and should really not be used by users) and forces the various resumes to synchronize so that only one of them does the actual cloning; all the others will verify (and wait if needed) that the clone happened prior to continuing.

Hopefully this makes a bit of sense as to how we are using it. I have incorporated this in the scheduler and it does what it needs to (at least for some small testing).

I made another minor tweak which consists of adding a line in the stderr (probably need to move to stdout) of cloned tasks saying that they were just cloned. It gives a better experience in the UI and better explains why you have a completed task with no logs :).

metaflow/runtime.py
```diff
@@ -756,6 +823,67 @@ def log(self, msg, system_msg=False, pid=None, timestamp=True):
...
+                    sys_tags=tags,
+                )
+                # As part of tag mutation project, we will be overlaying Run's tags
+                # (user and system) on top of Task tags.  I.e. a Task will no longer have
+                # its own tags distinct from their ancestral Run.  It means we should not
+                # use tags to indicate if a task is a control task in future.
+                #
+                # This is the main PR that implements the same "stashing to task metadata"
+                # concept:
+                #     https://github.com/Netflix/metaflow/pull/1039
```
savingoyal, line 861: Given that this is merged and released, we can update the comment.

metaflow/runtime.py
```diff
@@ -698,10 +675,100 @@ def __init__(
...
         self._results_ds = None
 
+        origin = None
         if clone_run_id and may_clone:
-            self._is_cloned = self._attempt_clone(clone_run_id, join_type)
+            origin = self._find_origin_task(clone_run_id, join_type)
+        if origin and origin["_task_ok"]:
+            # At this point, we know we are going to clone
+            self._is_cloned = True
+            if reentrant:
```
savingoyal, line 684: Can we add a comment here to note the expected reentrant behavior?

metaflow/task.py
```diff
@@ -300,6 +331,20 @@ def clone_only(self, step_name, run_id, task_id, clone_origin_task, retry_count)
...
             ],
         )
+        # We write a single line to the log to give a better experience to the user
+        # (particularly in the UI)
+        output.save_logs(
+            TASK_LOG_SOURCE,
+            {
+                "stdout": BytesIO(
+                    bytes(
+                        "Task was cloned from %s/%s"
```
savingoyal, line 341: Do we strictly need this?

metaflow/cli.py
```diff
@@ -674,6 +689,25 @@ def wrapper(*args, **kwargs):
...
+@click.option(
+    "--run-id",
+    default=None,
+    help="Run ID for the new run. By default, a new run-id will be generated",
+    hidden=True,
+)
+@click.option(
+    "--clone-only/--no-clone-only",
+    default=False,
+    show_default=True,
```
savingoyal, line 701: hidden=True

metaflow/task.py
savingoyal: This modifies the resume-from-local CLI logs as well and now we will have two log lines that say the same thing in the user logs

metaflow/runtime.py
romain-intel: Sure. I had just copy pasted. I'll remove it.

metaflow/runtime.py
romain-intel: Sure, would this work:
```
If `reentrant`, multiple processes may be trying to "clone" at the same time. In this case, we want only one to perform the actual clone and the others to wait for the clone to be done prior to proceeding.
```

metaflow/task.py
romain-intel: one is not persisted though. We can remove the other one if you want to remove one. It's not just the UI, it's also the client for the logs.

metaflow/cli.py
romain-intel: will do.

metaflow/task.py
tuulos: @romain-intel in general we should avoid polluting the user's `stdout` and `stderr` with our logging noise unless it is something critical that they must see.

In this case the contents of `Cloning results of a previously run task` is already available in `task.metadata_dict['origin-task-id']` if you need the info, so I don't think we should output it in the persisted logs unnecessarily.

The ephemeral output on the console is different, since it serves the role of informing the user that something is happening since resuming might take a while.

romain-intel: I am not married to it. If you like that better I'll remove it. From the UI's perspective, it's just non-obvious both during the running of the clone and after:
- during the running of the clone, nothing will be printed (to your point of "it may take a while" that may be confusing although in practice it does not take long).
- after the cloning, you have to know that something that says "No logs found" isn't actually an error or a sign that something went wrong in the case of a clone and you have to know where to look to determine if it is a clone.

metaflow/runtime.py
```diff
@@ -698,10 +675,119 @@ def __init__(
...
+                if not self._wait_for_clone:
+                    # Clone in place without relying on run_queue.
+                    self.new_attempt()
+                    self._ds.clone(origin)
+                    self._ds.done()
+                else:
+                    # TODO: There is a bit of a duplication with the task.py
+                    # clone_only function here
+                    self.log(
+                        "Waiting for clone of _parameters step to occur...",
```
savingoyal, line 747: Where would these logs show up?

romain-intel: Same as "Cloning results of a previously run task". We didn't use to show anything on the parameters task but here it may take some time. I can strip out the message on sleep and just add "(this may take some time)" here.

PR #1061, opened 2022-06-08
Fix periodic messages printed at runtime

romain-intel:


savingoyal (dismissed):

metaflow/runtime.py
```diff
@@ -219,14 +225,52 @@ def execute(self):
...
+                        ]
+                    )
+                    if self._active_tasks[0] == 0:
+                        msg = "No tasks are running."
+                    else:
+                        if self._active_tasks[0] == 1:
+                            msg = "1 task is running: "
+                        else:
+                            msg = "%d tasks are running: " % self._active_tasks[0]
+                        msg += "%s." % tasks_print
```
savingoyal, line 242: Do we need to list _all_ running/pending tasks in the queue?

metaflow/runtime.py
```diff
@@ -219,14 +224,45 @@ def execute(self):
...
+
                     self._logger(msg, system_msg=True)
-                    msg = "%d steps are pending: %s." % (0, "e.g. ...")  # TODO
+                    if len(self._unprocessed_steps) == 0:
+                        msg = "No steps are pending."
+                    else:
+                        if len(self._unprocessed_steps) == 1:
+                            msg = "1 step is pending: "
+                        else:
+                            msg = "%d steps are pending: " % len(
```
savingoyal, line 262: What is the definition of a step being pending? Is a step pending if no task for that step has started? Or is it the case that not all tasks have been finished thus far? I am not sure of the utility of printing out pending steps in either of the cases though.

metaflow/runtime.py
```diff
@@ -606,7 +650,13 @@ def _launch_worker(self, task):
         for fd in worker.fds():
             self._workers[fd] = worker
             self._poll.add(fd)
-        self._num_active_workers += 1
+        active_step_counts = self._active_tasks.setdefault(task.step, [0, 0])
+
+        # We have an additional task for this step running
+        active_step_counts[0] += 1
```
savingoyal, line 656: Where is this used? We can set the defaults as `[1,0]` directly?

savingoyal (approved):
minor nit - but good to go.

metaflow/runtime.py
romain-intel: Pending is steps that haven't started running. It gives an idea of where you are in the "execution frontier". I can rephrase to say %d steps not started

metaflow/runtime.py
romain-intel: We could initialize to [1, 0] but then would have to be a different path for doesn't exist and exists. Here, the active_step_counts always points to what needs to be updated. It's used to print the number of tasks running/done.

savingoyal: How is replacing with `self._active_tasks.setdefault(task.step, [1, 0])` necessitating a different path?

romain-intel: self._active_tasks[task.step][0] by adding one whether or not task.step is already in self._active_tasks. The above code is equivalent to:
```
if task.step in self._active_tasks:
  self._active_tasks[task.step][0] += 1
else:
  self._active_tasks[task.step] = [1, 0]
```
with the two paths. The code simplifies that a bit IMHO.

romain-intel:
For some reasno, it didn't trigger the test -- re-running.

PR #1063, opened 2022-06-11
This is the Jackie's tag changes with an additional test and a fix.

romain-intel:


metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -139,6 +140,8 @@ def create(
     workflow_timeout=None,
     workflow_priority=None,
 ):
+    validate_tags(tags)
```
savingoyal, line 143: @valayDave You would need to make a similar addition to the Airflow PR once this PR is merged.

savingoyal (approved):

PR #1080, opened 2022-07-08
Added runtime context in current.

valayDave:


metaflow/plugins/aws/batch/batch_decorator.py
```diff
@@ -206,6 +206,7 @@ def task_pre_step(
     ):
         self.metadata = metadata
         self.task_datastore = task_datastore
+        current._update_env({"runtime_context": "batch"})
```
savingoyal, line 209: `aws-batch`

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -267,6 +268,7 @@ def task_pre_step(
     ):
         self.metadata = metadata
         self.task_datastore = task_datastore
+        current._update_env({"runtime_context":"kubernetes"})
```
savingoyal, line 271: This hook is not guaranteed to always execute on Kubernetes.

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -3,6 +3,7 @@
 import sys
 
 import requests
+from metaflow import current
```
savingoyal, line 6: Can you also handle the scenario when the context is local?

valayDave:
Can we merge this ?

PR #1089, opened 2022-07-20
Fix fractional resoure handling for batch

oavdeev:


metaflow/plugins/aws/aws_utils.py
```diff
@@ -79,7 +80,15 @@ def compute_resource_attributes(decos, compute_deco, resource_defaults):
                     continue
                 if my_val is not None and v is not None:
                     try:
-                        result[k] = str(max(int(my_val or 0), int(v or 0)))
+                        # Use Decimals to compare and convert to string here so
+                        # that numbers that can't be exactly represented as
+                        # floats (e.g. 0.8) still look "nice". We don't care
+                        # about precision more that .001 for resources anyway.
+                        result[k] = str(
```
savingoyal, line 87: wouldn't `str(max(float(my_val or 0), float(v or 0)))` have worked?

oavdeev: Yeah probably this stuff i did with decimals isn't necessary; i just get uneasy when there are back-and-forth conversions from floats to str, since it depends on the default precision that `str` uses for formatting, so I wanted to be more explicit.

savingoyal: we do this type of conversion (`str(max(float(my_val or 0), float(v or 0)))`) at a bunch of places - it's better to keep it consistent and simple. The precision stays the same for string formatting in this case (which needs to be handled explicitly otherwise in the decimal case).

oavdeev: Tried this and remembered the reason for Decimal stuff; if I do `str(max(float(my_val or 0), float(v or 0)))` it forces the conversion to float, so even if you specify `memory=1024` AND you hit that `if` branch, it gets converted to `1024.0` which is kinda ugly and makes unit testing finnicky. 

So depending on where you specify it (resources decorator or batch decorator or both) you may get either "1024.0" or "1024" in the end. It doesn't matter a ton for the users (i think) just harder to test. 

I'll change it anyway since kubernetes plugin does the same thing with float() but we should prob unify this logic and actually start running those unit tests..

PR #1091, opened 2022-07-20
Adding support for Azure Blob Storage as a datastore

jackie-ob:
The primary change is implementing `AzureStorage` (analogous to existing `S3Storage`, `LocalStorage`).  We are consciously deferring the decision of having first class "data tools" support for Azure.

There are some necessary changes to ensure full Azure support on all Metaflow surfaces:
* includefile
* conda
* cards
* mflog
* kubernetes
* argo

We take care to ensure there is no cross disruption to users not using Azure.  More specifically:
* Users need to setup AWS dependencies (boto3, config params), iff they are using AWS.
* Users need to setup Azure dependencies (azure SDK libs, config params), iff they are using Azure.

We aggressively use local imports to achieve this.

Much effort was also spent to ensure good performance of Metaflow's usage of Azure Blob Storage.   See context docs for more details.

Some docs for context:
* Overall [Metaflow-on-Azure](https://docs.google.com/document/d/1_W7zDiImLtxY0Xq47fohGvtEPrd3KWV4ULfOxYNjlhs/edit#heading=h.6xrup8gk65q4) effort.
* AzureStorage [implementation](https://docs.google.com/document/d/18_WiFdagdaV0ATfHcK7CfG6wMTpap691hAoMxlfCMEg/edit#heading=h.ikim125j24da) notes (may not match up 100% to code in PR in minor ways, but the bones should be consistent).
  * [Section on performant implementation.](https://docs.google.com/document/d/18_WiFdagdaV0ATfHcK7CfG6wMTpap691hAoMxlfCMEg/edit#heading=h.x77ca5o2wmep)
* [Performance discovery](https://docs.google.com/document/d/1PkR_0Fz9B3OJukyIKwD-QiUC4jMeAGx9Snh_cUQyfNg/edit#heading=h.z4v8dso6utye) (pre-implementation).


Mikkolehtimaki:
Hi there, I'm eager to try this out. I have some basic questions.

1. Is there a simple "getting started" or "howto" example for setting up Azure support?
2. Now that there will be different dependencies for different backends, have you considered using install extras offered by setuptools? It would offer a neat way to specify versions of the dependencies. I mean these: https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies but maybe this is handled elsewhere in the program (only reference I found was in pinned conda libs). 
3. `metaflow configure` does not hint at any Azure options, is this intended for now?

Thanks!

jackie-ob:
Thank you for your interest!

1. How-to guide.  We do have some WIP instructions and templates.  It’s in private preview state.  We will share it the Google docs with you.  Please DM us directly in the community Slack!

2. “Pinned packages”.   To share some context, even prior to the Azure project, we already have optional dependencies, though we handle it in a somewhat low-tech way (IMO).  We use soft imports - i.e. only when users choose to use a feature requiring the optional dependencies do we actually import it.  We do this for Kubernetes integration, and we do it for S3 (boto3).  Note that for boto3 we actually do list boto3 as non-optional - but if user ends up not using S3 we won’t actually import boto3 at runtime.  

We plan to take the same approach for Azure.  I.e. we will end up importing Azure deps IFF user really uses Azure.  If those Azure deps are missing at runtime, we take care to having friendly messaging to ask the user to install them at that point.  

Setuptools’s optional dependencies feature does seem like a promising option to handle dependencies more gracefully.  If we do go down this route, we would likely want to handle Kubernetes, S3, azure (and any other optional features) in a uniform way.  We would want to think through what the UX will be like for folks who need 0 extras, 1 extra, and 2>=extras.  Plus whether the user base at large understands what the “square brackets” mean exactly, etc, etc.

3. “metaflow configure azure” was added late last week.  Please check it out!

metaflow/datastore/azure_storage.py
```diff
@@ -0,0 +1,550 @@
+import math
```
romain-intel, line 1: I did not review this file.

metaflow/includefile.py
```diff
@@ -80,6 +83,93 @@ def path(self):
         return self._path
 
 
+class Azure(object):
```
romain-intel, line 86: nit: maybe this whole this can move over to the azure folder to keep this a little cleaner (kind of like S3 doesn't show up here except in the dictionary?

metaflow/metaflow_config.py
```diff
@@ -36,8 +36,50 @@ def init_config():
 METAFLOW_CONFIG = init_config()
 
 
-def from_conf(name, default=None):
-    return os.environ.get(name, METAFLOW_CONFIG.get(name, default))
+def from_conf(name, default=None, from_env_only=False, validate_fn=None):
```
romain-intel, line 39: from_env_only does not seem to be used here (or I missed it), what is the use case?

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -6,12 +6,11 @@
```
romain-intel, line 6: did not review this file.

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -328,8 +328,10 @@ def make_flow(
 ):
```
romain-intel, line 328: did not review this file.

metaflow/plugins/aws/batch/batch.py
```diff
@@ -8,19 +8,17 @@
```
romain-intel, line 8: did not review this file.

metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -417,6 +417,8 @@ def gpu(self, gpu):
         return self
```
romain-intel, line 417: did not review this file.

metaflow/plugins/aws/step_functions/step_functions.py
```diff
@@ -629,6 +629,10 @@ def _batch(self, node):
                 )
```
romain-intel, line 629: did not review this file.

metaflow/plugins/azure/azure_exceptions.py
```diff
@@ -0,0 +1,13 @@
+from metaflow.exception import MetaflowException
```
romain-intel, line 1: did not review anything in the azure directory.

metaflow/plugins/kubernetes/kubernetes.py
```diff
@@ -1,34 +1,32 @@
-import hashlib
 import json
```
romain-intel, line 1: did not review anything in kubernetes.

romain-intel (commented):
Some minor comments. Nothing seems to be problematic. Let me run my full tests on it (as soon as I get all that straightened out) but this looks good overall.

romain-intel (changes requested):
It's on me but there is a small change that seems that is needed for this to work with our internal version. I will make that change and test again.

Apart from that, minor comments. Performance-wise, no issues -- load times are comparable.

metaflow/includefile.py
jackie-ob: The intent is that this should stay logically private for includefile usage ONLY.  I think we can do something like `metaflow/plugins/azure/includefile_support.py` which would be sufficiently explicit about it.

metaflow/metaflow_config.py
jackie-ob: I'll remove this. It was used for a security token value  ("shared access signature") that should not live in the config JSON - but we decided not to support signature at all.

romain-intel:
Testing is fine now on my end. The one failure is transient. Small comments remain and we can then go ahead and merge.

romain-intel (dismissed):
Tests are fine. All comments addressed. Can merge

savingoyal (dismissed):

romain-intel:
NOTE: Not compatible with include file changes. This PR can be made compatible fairly simply by adding the relevant methods in the includefile_support.py file.

romain-intel:
I'll re-run the tests as well.

romain-intel:
Hun, for some reason there is an issue now (can't find `JSONType`). I'll take a look in a bit.

romain-intel (approved):
I have not reviewed the azure implementation in detail but I have verified that this does not break anything else and does not have any performance implications. This is good to merge for me.

PR #1094, opened 2022-07-29
Airflow Support

valayDave:
# Key changes
The following are the key user-related constructs exposed by metaflow to support the Airflow universe:

- A CLI command to create airflow workflow DAG file from metaflow `FlowSpec`. 

## How it works

Metaflow users can now easily compile Metaflow workflows into Airflow compatible workflows. Every Metaflow `FlowSpec` file exposes a `airflow create` command to compile the workflow into an Airflow DAG file. Under the hood metaflow packages the code packages, stores them in S3 and then compiles the code into a airflow compatible python file. The generated file can be moved to the Airflow scheduler's DAGs folder. From there the scheduler will run the workflow as a native airflow workflow. 

Metaflow requires Kubernetes as Airflow's compute-execution medium. Metaflow leverages the Airflow `KuberneterPodOperator` to run different `@step`s on Kubernetes via Airflow. 

## Decorator Support 
`@conda`,`@catch`, `@retry`,`@card`,`@project`,`@environment`,`@timeout`,`@environment`,`@kubernetes`,`@resources`
## Version Support 
A minimum Airflow version of `2.2.0` is required to support basic Metaflow decorators and flows with nested branches. _*Foreach's are not supported with Airflow version 2.2.0*_. Decorators like `@batch` are also not supported with Airflow. 

Airflow introduced [dynamic task mapping](https://airflow.apache.org/docs/apache-airflow/stable/concepts/dynamic-task-mapping.html) in version 2.3.0. Metaflow will very soon support foreach's which can work with Airflow version `2.3.0`. _*Foreach support will also require that the airflow installation has [apache-airflow-providers-cncf-kubernetes](https://airflow.apache.org/docs/apache-airflow-providers-cncf-kubernetes/stable/index.html) provider installed at minimum of version 4.2.0*_.  

## CLI Usage
Metaflow exposes an `airflow create` command from the FlowSpec file's cli. This command can help convert a FlowSpec file into a Airflow DAG.  
```
Usage: python myflow.py airflow create [OPTIONS] FILE

  Compile a new version of this flow to Airflow DAG.

Options:
  --authorize TEXT            Authorize using this production token. You need
                              this when you are re-deploying an existing flow
                              for the first time. The token is cached in
                              METAFLOW_HOME, so you only need to specify this
                              once.

  --generate-new-token        Generate a new production token for this flow.
                              This will move the production flow to a new
                              namespace.

  --new-token TEXT            Use the given production token for this flow.
                              This will move the production flow to the given
                              namespace.

  --tag TEXT                  Annotate all objects produced by Airflow DAG
                              executions with the given tag. You can specify
                              this option multiple times to attach multiple
                              tags.

  --is-paused-upon-creation   Generated Airflow DAG is paused/unpaused upon
                              creation.

  --namespace TEXT            Change the namespace from the default to the
                              given tag. See run --help for more information.

  --max-workers INTEGER       Maximum number of parallel processes.  [default:
                              100]

  --workflow-timeout INTEGER  Workflow timeout in seconds. Enforced only for
                              scheduled DAGs.

  --worker-pool TEXT          Worker pool for Airflow DAG execution.
  --help                      Show this message and exit
```

## Scheduling Airflow Dags
Metaflow compile airflow DAGs can be scheduled via [@schedule](http://docs.metaflow.org/going-to-production-with-metaflow/scheduling-metaflow-flows/scheduling-with-argo-workflows#scheduling-a-flow) decorator. 

# Resources 
1. [Airflow Task Instance](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/models/taskinstance/index.html#airflow.models.taskinstance.TaskInstance)
2. [Airflow Task Object](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/models/taskmixin/index.html)
3. [Airflow Dag Run](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/models/dagrun/index.html#airflow.models.dagrun.DagRun)
4. [Running Airflow with Metaflow on Minikube](https://github.com/outerbounds/airflow-on-minikube)

### DIFF
--- a/metaflow/decorators.py
+++ b/metaflow/decorators.py
@@ -100,6 +100,8 @@ class Decorator(object):
 
     name = "NONAME"
     defaults = {}
+    # `allow_multiple` allows setting many decorators of the same type to a step/flow.
+    allow_multiple = False
 
     def __init__(self, attributes=None, statically_defined=False):
         self.attributes = self.defaults.copy()
@@ -225,9 +227,6 @@ class StepDecorator(Decorator):
                   step.__name__ etc., so that we don't have to
                   pass them around with every lifecycle call.
     """
-
-    # `allow_multiple` allows setting many decorators of the same type to a step.
-    allow_multiple = False
 
     def step_init(
         self, flow, graph, step_name, decorators, environment, flow_datastore, logger
@@ -374,12 +373,17 @@ def _base_flow_decorator(decofunc, *args, **kwargs):
         if isinstance(cls, type) and issubclass(cls, FlowSpec):
             # flow decorators add attributes in the class dictionary,
             # _flow_decorators.
-            if decofunc.name in cls._flow_decorators:
+            if decofunc.name in cls._flow_decorators and not decofunc.allow_multiple:
                 raise DuplicateFlowDecoratorException(decofunc.name)
             else:
-                cls._flow_decorators[decofunc.name] = decofunc(
-                    attributes=kwargs, statically_defined=True
-                )
+                deco_instance = decofunc(attributes=kwargs, statically_defined=True)
+                if decofunc.allow_multiple:
+                    if decofunc.name not in cls._flow_decorators:
+                        cls._flow_decorators[decofunc.name] = [deco_instance]
+                    else:
+                        cls._flow_decorators[decofunc.name].append(deco_instance)
+                else:
+                    cls._flow_decorators[decofunc.name] = deco_instance
         else:
             raise BadFlowDecoratorException(decofunc.name)
         return cls
@@ -469,11 +473,26 @@ def _attach_decorators_to_step(step, decospecs):
 def _init_flow_decorators(
     flow, graph, environment, flow_datastore, metadata, logger, echo, deco_options
 ):
+    # Certain decorators can be specified multiple times and exist as lists in the _flow_decorators dictionary
     for deco in flow._flow_decorators.values():
-        opts = {option: deco_options[option] for option in deco.options}
-        deco.flow_init(
-            flow, graph, environment, flow_datastore, metadata, logger, echo, opts
-        )
+        if type(deco) == list:
+            for rd in deco:
+                opts = {option: deco_options[option] for option in rd.options}
+                rd.flow_init(
+                    flow,
+                    graph,
+                    environment,
+                    flow_datastore,
+                    metadata,
+                    logger,
+                    echo,
+                    opts,
+                )
+        else:
+            opts = {option: deco_options[option] for option in deco.options}
+            deco.flow_init(
+                flow, graph, environment, flow_datastore, metadata, logger, echo, opts
+            )
 
 
 def _init_step_decorators(flow, graph, environment, flow_datastore, logger):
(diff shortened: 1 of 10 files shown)


romain-intel (commented):
This should be fine. I'll also run the tests on it.

Generally though, I am starting to wonder if all this adds significant bloat when booting (in terms of startup time). I will investigate this and see if I can provide an option to remove some things (which may be useful in general). I am not sure but metaflow bootup time was one of the overhead costs for S3 in particular and I wonder if this adds a lot or not. If you already have numbers, I'd be curious but o/w will try to look into it too.

metaflow/decorators.py
```diff
@@ -469,11 +473,26 @@ def _attach_decorators_to_step(step, decospecs):
 def _init_flow_decorators(
     flow, graph, environment, flow_datastore, metadata, logger, echo, deco_options
 ):
+    # Certain decorators can be specified multiple times and exist as lists in the _flow_decorators dictionary
     for deco in flow._flow_decorators.values():
-        opts = {option: deco_options[option] for option in deco.options}
-        deco.flow_init(
-            flow, graph, environment, flow_datastore, metadata, logger, echo, opts
-        )
+        if type(deco) == list:
```
romain-intel, line 478: nit: use isinstance.

metaflow/decorators.py
```diff
@@ -469,11 +473,26 @@ def _attach_decorators_to_step(step, decospecs):
...
+                    flow,
+                    graph,
+                    environment,
+                    flow_datastore,
+                    metadata,
+                    logger,
+                    echo,
+                    opts,
+                )
+        else:
```
romain-intel, line 491: as a small nit, maybe we just always make it a list. It would simplify this and the other change. A few more changes would be required when `_flow_decorators` is accessed but IMHO, it is nicer if there is a single type and we don't have to assume the number of decorators of a given type.

metaflow/decorators.py
```diff
@@ -374,12 +373,17 @@ def _base_flow_decorator(decofunc, *args, **kwargs):
         if isinstance(cls, type) and issubclass(cls, FlowSpec):
             # flow decorators add attributes in the class dictionary,
             # _flow_decorators.
-            if decofunc.name in cls._flow_decorators:
+            if decofunc.name in cls._flow_decorators and not decofunc.allow_multiple:
```
romain-intel, line 376: I am a little confused as to why step and flow decorators are treated differently wrt to the allow_multiple.

romain-intel (approved):
A few minor comments mostly around consistency in handling decorators. I am also not sure of the use of multiple decorators at the flow level but if we have multiple decorators at both flow and step, we should be consistent with them.

No issues with tests on my end -- this merges cleanly.

Performance wise, no issues either -- the additional loads are good and trivial.

valayDave:
@romain-intel : The changes to `decorators.py` were not immediately used with the Airflow integration (in this PR). We were planning to use them a little later when we introduce Foreach's and other airflow constructs as features. 

We can incorporate this change later in a separate PR which is very tightly scoped. 

Now nothing in the PR touches the core :)

romain-intel:
Am fine if it is a separate pr or this one. My main point was that it looks like the step/flow decorator's behavior for allow_multiple is different which was a bit surprising to me.

PR #1098, opened 2022-08-03
Move sys.path insert earlier in s3op.py

romain-intel:


romain-intel:
ok, I am running the tests. Please run yours as well so we can make sure this also works.

PR #1100, opened 2022-08-03
Fix for env_escape bug when importing local packages

hunsdiecker:
This PR aims to solve a bug related to importing multiple packages using `env_escape` when one of these packages are available locally. The steps to reproduce are given below:

```
import package A # available locally
class Flow():
       import package B # not available locally
```

`env_escape` fails for package B since we fail to update the `sys.path` when using the package A that is available locally.


romain-intel (commented):
LGTM. Running tests.

romain-intel (approved):

PR #1103, opened 2022-08-06
Add a flag to overwrite config for CLI configure sandbox command

oavdeev:


savingoyal (approved):

metaflow/main_cli.py
```diff
@@ -437,8 +437,16 @@ def import_from(profile, input_filename):
...
-    overwrite_config(profile)
+@click.option(
+    "--overwrite/--no-overwrite",
+    "-o/",
+    default=False,
+    show_default=True,
+    help="Overwrite profile configuration without asking",
+)
+def sandbox(profile, overwrite):
+    if not overwrite:
```
savingoyal, line 448: Shouldn't this be the other way round?

savingoyal (changes requested):

metaflow/main_cli.py
oavdeev: No, its just `overwrite_confg()` is a little confusingly named. It should really be `ask_to_overwrite_config()`. The idea is that if `--overwrite` is set, it doesn't get called, and the config is overwritten without asking.

Maybe the flag should be `--quiet-overwrite/--no-quiet-overwrite` 🤔

savingoyal: can we just change `overwrite_confg` to `ask_to_overwrite_config` and ship this PR?

PR #1126, opened 2022-09-01
Card bug fix when task-ids are non-unique

valayDave:


metaflow/plugins/cards/card_datastore.py
```diff
@@ -74,6 +79,20 @@ def __init__(self, flow_datastore, pathspec=None):
         self._run_id = run_id
         self._step_name = step_name
         self._pathspec = pathspec
+
+        # Save the task's metaflow version so that we can use it in the `_make_path` function.
+        # Since metaflow version >= 2.7.8 had a change in the directory structure for cards, CardDatastore object needs this information to determine the correct directory structure for retrieving/storing cards.
+        from metaflow.client import Run
```
savingoyal, line 85: Can we avoid making the call to the metadata service from within this module? Is this call strictly needed?

valayDave: I spoke to Romain and I will change this. I will be doing a small refactor to remove the `Run` object altogether.

metaflow/plugins/cards/card_datastore.py
```diff
@@ -89,10 +108,21 @@ def _make_path(self, base_pth, pathspec=None):
         sysroot = base_pth
 
         if pathspec is not None:
+            # since most cards are at a task level there will always be 4 non-none values returned
             flow_name, run_id, step_name, task_id = path_spec_resolver(pathspec)
-
-        # For task level cards the flow_name and run_id and task_id are required
-        if flow_name is not None and run_id is not None and task_id is not None:
+        if abs_version(self._metaflow_version) >= 2780:
```
savingoyal, line 113: Rather than depending explicitly on metaflow-version, we can also first check for the card's existence in the new location and fall back on the old location if the card doesn't exist. There is no guarantee that this PR will ship with Metaflow 2.7.8.

metaflow/plugins/cards/card_datastore.py
```diff
@@ -85,14 +85,28 @@ def get_card_location(cls, base_path, card_name, card_html, card_id=None):
...
         sysroot = base_pth
-
         if pathspec is not None:
+            # since most cards are at a task level there will always be 4 non-none values returned
             flow_name, run_id, step_name, task_id = path_spec_resolver(pathspec)
 
-        # For task level cards the flow_name and run_id and task_id are required
-        if flow_name is not None and run_id is not None and task_id is not None:
+        # We have a condition that checks for `with_steps` because when cards were introduced there was an assumption made about task-ids being unique.
+        # This assumption is incorrect since pathspec needs to be unique but there is no such gaurentees on task-ids
```
romain-intel, line 95: nit: guarantees. Also shorter lines.

romain-intel (approved):
The changes look good on paper. I haven't tested in my specific use case yet but don't foresee any issues. I'll run the tests on them and if all looks good, I would be in favor of merging.

metaflow/plugins/cards/card_datastore.py
```diff
@@ -102,16 +116,15 @@ def _make_path(self, base_pth, pathspec=None):
...
         if sysroot == "" or sysroot == None:
             pth_arr.pop(0)
         return os.path.join(*pth_arr)
 
-    def _get_card_path(self, base_pth=""):
-        return self._make_path(
-            base_pth,
-            pathspec=self._pathspec,
-        )
+    def _get_write_path(self, base_pth=""):
```
savingoyal, line 123: To ensure that this is not a breaking change right away, we can write the cards to both the locations (with and without step name) and deprecate the second write path (without step name) 3 months from now. The reads still look for the path with the step name first.

savingoyal: If we don't do this, the UI will start breaking almost immediately for all deployments.

romain-intel: That would work. The UI would indeed not show cards until updated. Writing to both locations would solve this and doesn't ahve a huge cost. Maybe protect it with a flag though (hidden). That way I can disable it because I don't really need it (and would update the UI right away).

metaflow/plugins/cards/card_datastore.py
```diff
@@ -153,25 +173,68 @@ def card_info_from_path(path):
...
+        # It would especially happen for users who are accessing cards created by a newer
+        # MF client from an older version of MF client.
+        # It will also easily end up breaking the metaflow-ui (which maybe using a client from an older version).
+        # Hence we are writing cards to both paths so that we can introduce breaking changes later in the future.
+        card_path_with_steps = self.get_card_location(
+            self._get_write_path(), card_file_name, card_html, card_id=card_id
         )
-        self._backend.save_bytes(
-            [(card_path, BytesIO(bytes(card_html, "utf-8")))], overwrite=overwrite
+        card_path_without_steps = self.get_card_location(
```
romain-intel, line 196: As a small nit, this will compute the card hash twice. Not a major deal since it's a hack.

metaflow/plugins/cards/card_datastore.py
```diff
@@ -153,25 +173,68 @@ def card_info_from_path(path):
...
+            self._get_read_path(with_steps=False),
+            card_file_name,
+            card_html,
+            card_id=card_id,
         )
-        return self.card_info_from_path(card_path)
+        if hasattr(metaflow_config, "_HACK_SKIP_CARD_DUALWRITE"):
```
romain-intel, line 202: Perfect! Thank you, that will allow me to turn this off.

metaflow/plugins/cards/card_datastore.py
```diff
@@ -153,25 +173,68 @@ def card_info_from_path(path):
...
-                self._pathspec,
-                card_hash=card_hash,
-                card_type=card_type,
+        # Check for new cards first
+        card_paths = []
+        card_paths_with_steps = self._backend.list_content(
+            [self._get_read_path(with_steps=True)]
+        )
+
+        if len(card_paths_with_steps) == 0:
```
romain-intel, line 222: This will "penalize" tasks that don't have cards at all (since you look in two locations). I think this is OK given that it is most likely not something you should be doing. The only possible issue is the UI which may take a tad longer to determine that there are no cards. I don't believe this is an issue either. I don't have a good solution to have a penalty free approach in all cases.

romain-intel (dismissed):

romain-intel:
Running tests. If all good. We can merge and release.

metaflow/plugins/cards/card_datastore.py
valayDave: This exception is how the user is notified that there are no cards when accessing things from the cli. Card tests also rely on these exceptions to make decisions if tests passed or failed. 

Removing this will create a lot bigger refactor , I think we can leave it here for now :)

metaflow/plugins/cards/card_datastore.py
valayDave: Great catch! Since cards can dump a lot of information in a JSON, you can be hashing huge files. Will fix.

metaflow/plugins/cards/card_datastore.py
romain-intel: Sorry, wasn't suggesting to not return the exception, was just thinking if we could do so more cheaply but I think this is fine.

metaflow/metaflow_config.py
```diff
@@ -201,6 +201,8 @@ def _validate_choice(name, value):
 )
 CARD_NO_WARNING = from_conf("METAFLOW_CARD_NO_WARNING", False)
 
+SKIP_CARD_DUAL_WRITE = from_conf("METAFLOW_SKIP_CARD_DUALWRITE", False)
```
romain-intel, line 204: This works for me but as a quick note, this setting would drop off at some point (so maybe we should put a note that it will become True by default and then drop off).

romain-intel (approved):

romain-intel:
The failures seem to be transient failures with the card tests. I have seen this a few times. I can never seem to figure out what exactly is causing the error. I will merge and release.

romain-intel (approved):

PR #1132, opened 2022-09-06
Merging internal conda plugin to OSS version

romain-intel:
This cleans up a few things and also introduces the
possibility of installing a pre-configured Conda distribution
to use on both machines that launch Metaflow flows and machines
that execute them (like batch).


romain-intel:
Replaced by https://github.com/Netflix/metaflow-nflx-extensions

PR #1138, opened 2022-09-12
Argo Events support

kevsmith:
This commit introduces event-triggering support for Argo Worklows orchestrated runs based on the status of another run by using Argo Events.

Given two flows, A and B, users can express:

* When A succeeds B should run
* When A fails B should run

When event support is enabled all flows will emit `metaflow_flow_run_succeeded` and `metaflow_flow_run_failed` events. Downstream flows, such as B in the above example, can declare a dependency on their upstream flow's success or failure by using the new `@trigger_on decorator`. Eventing support can be temporarily disabled with the `--ignore-events` CLI flag.

Argo Events support requirements:

* Argo Events installed and configured correctly
* Argo Event event source in the target K8s namespace (NATS and webhook event sources are supported)
* Ability to pull 3rd party images from Docker Hub

The last requirement is necessary due to Argo Events' design. It provides an API for creating event sources and sensors which listen for events from sources but assumes events will be sent to the source by an external system. Source for the event trigger image used here can be found at https://github.com/kevsmith/event-trigger.

### DIFF
--- a/metaflow/metaflow_config.py
+++ b/metaflow/metaflow_config.py
@@ -363,6 +363,31 @@ if AWS_SANDBOX_ENABLED:
 
 KUBERNETES_SANDBOX_INIT_SCRIPT = from_conf("METAFLOW_KUBERNETES_SANDBOX_INIT_SCRIPT")
 
+# Argo Events
+EVENT_SOURCE_NAME = from_conf(
+    "METAFLOW_EVENT_SOURCE_NAME"
+)  # Required name of Argo Event source
+EVENT_SOURCE_URL = from_conf(
+    "METAFLOW_EVENT_SOURCE_URL"
+)  # Required: HTTP(S) or NATS URL
+EVENT_SERVICE_ACCOUNT = from_conf(
+    "METAFLOW_EVENT_SERVICE_ACCOUNT"
+)  # Required K8s service account
+EVENT_SOURCE_AUTH_SECRET = from_conf(
+    "METAFLOW_EVENT_AUTH_SECRET"
+)  # Optional name of K8s secret
+EVENT_SOURCE_AUTH_KEY = from_conf(
+    "METAFLOW_EVENT_AUTH_KEY"
+)  # Optional path to auth value within K8s secret
+EVENT_SOURCE_AUTH_TOKEN = from_conf(
+    "METAFLOW_EVENT_SOURCE_AUTH_TOKEN"
+)  # Optional string literal of auth token
+# Note: If auth is enabled on the underlying event source then either both AUTH_SECRET
+# and AUTH_KEY must be populated or AUTH_TOKEN.
+EVENT_DISPATCH_IMAGE = from_conf(
+    "METAFLOW_EVENT_DISPATCH_IMAGE", "kevinob/event-trigger"
+)  # Image used by event dispatch pods
+
 # MAX_ATTEMPTS is the maximum number of attempts, including the first
 # task, retries, and the final fallback task and its retries.
 #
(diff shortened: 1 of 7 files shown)


metaflow/metaflow_config.py
```diff
@@ -363,6 +363,31 @@ def _validate_choice(name, value):
 
 KUBERNETES_SANDBOX_INIT_SCRIPT = from_conf("METAFLOW_KUBERNETES_SANDBOX_INIT_SCRIPT")
 
+# Argo Events
```
romain-intel, line 366: I would put "ARGO" in the name somewhere. This seems a bit too generic given that it is specifically for argo.

romain-intel (commented):
Minor comment. Should not impact anything on our end.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -954,6 +934,92 @@ def _container_templates(self):
...
+                    yield self._make_lifecycle_hook_container_template(
+                        self._make_env(node), "failed"
+                    )
+
+    def _make_lifecycle_hook_container_template(self, env, status):
+        t = LifecycleHookContainerTemplate(current_flow_name(), status)
+        t.env(env)
+        return t
+
+    def _make_env(self, node, retry_count=None):
```
savingoyal, line 953: It doesn't seem like any of these env variables are needed by the docker container responsible for emitting the event (besides flow name and run id). 

Currently, that docker container is responsible for 
1. emitting the metadata (which we can move to the triggered workflow as part of a decorator lifecycle - take a look at argo_workflows_decorator)
2. emitting the actual event (we should explore embedding the logic within the entry point in itself to save an extra image's hurdle)

savingoyal (commented):
First pass code feedback

metaflow/plugins/argo/argo_client.py
```diff
@@ -203,3 +203,146 @@ def schedule_workflow_template(self, name, schedule=None):
             raise ArgoClientException(
                 json.loads(e.body)["message"] if e.body is not None else e.reason
             )
+
+    def register_trigger_on_template(self, name, sensor_template):
```
savingoyal, line 207: minor nit - `register_event_sensor`

metaflow/plugins/argo/argo_client.py
```diff
@@ -203,3 +203,146 @@ def schedule_workflow_template(self, name, schedule=None):
...
+            return (False,)
+        except self._client.rest.ApiException as e:
+            if e.status == 404:
+                return (False,)
+
+            raise ArgoClientException(
+                json.loads(e.body)["message"] if e.body is not None else e.reason
+            )
+
+    def remove_existing_workflow_template(self, flow_name, dry_run=False):
```
savingoyal, line 308: If we drop `explicit deletes` for now, we should be able to drop these methods as well

metaflow/plugins/argo/argo_client.py
```diff
@@ -203,3 +203,146 @@ def schedule_workflow_template(self, name, schedule=None):
...
+                plural="sensors",
+                name=name,
+                body=sensor_template,
+            )
+        except self._client.rest.ApiException as e:
+            raise ArgoClientException(
+                json.loads(e.body)["message"] if e.body is not None else e.reason
+            )
+
+    def remove_existing_trigger_on_template(self, name, dry_run=False):
```
savingoyal, line 258: If we drop `explicit deletes` for now, we should be able to drop these methods as well

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -114,19 +131,41 @@ def __init__(
...
                 self.name, self._workflow_template.to_json()
             )
+            if are_events_configured():
+                if self._trigger_on_template is not None:
+                    client.register_trigger_on_template(
+                        self._trigger_on_template.name(),
+                        self._trigger_on_template.to_json(),
+                    )
+                else:
+                    client.remove_existing_trigger_on_template(
```
savingoyal, line 166: Is there a way to disable the trigger rather than deleting it?

PR #1139, opened 2022-09-12
User-specified events, event data, and parameter support

kevsmith:
This PR adds two new features: ad-hoc user event support via the new `@emit_event` decorator and arbitrary JSON-encoded user event data.

`@emit_event` allows users to emit ad-hoc events from within flows based on the success/failure status of a step. `@trigger_on` has been modified to take either an event name or a flow name so an appropriate trigger can be set up for user events. For now, it's an error to emit an event within a foreach loop. This limitation exists to simplify the initial effort and help users avoid accidentally spamming the underlying message bus or webhook with messages.

Users can specify arbitrary data to be included with an event via `@emit_event`'s `data` argument. Names and values of user data are automatically mapped to parameters of any downstream workflows triggered by the event.


metaflow/plugins/argo/event_decorators.py
```diff
@@ -0,0 +1,247 @@
...
+                        "message": ERR_MSG,
+                        "headline": ("@%s requires eventing support" % self.name),
+                    }
+                }
+            )
+
+
+class EmitEventDecorator(StepDecorator):
+
+    name = "emit_event"
```
savingoyal, line 151: rather than a decorator, we can make this a function. That way the user retains more control of the payload going into the event (it can be a function of user code within a step).

```
@step
def foo(self):
    x = get_some_output()
    event.emit(..., x, ...)
```

metaflow/plugins/argo/event_decorators.py
```diff
@@ -0,0 +1,247 @@
...
+    @property
+    def event(self):
+        return self.attributes["event"]
+
+    @property
+    def data(self):
+        return self.attributes.get("data")
+
+
+class AnnotateLifecycleDecorator(FlowDecorator):
```
savingoyal, line 190: What is the intended use case for this?

metaflow/plugins/argo/event_decorators.py
```diff
@@ -0,0 +1,247 @@
...
+    for key in data.keys():
+        if type(key) != str:
+            raise MetaflowException(
+                msg="@%s requires data to be a dict with string keys: %s"
+                % (deco_name, str(key))
+            )
+    return data
+
+
+class TriggerOnDecorator(FlowDecorator):
```
savingoyal, line 68: Can we decouple this to explicit `TriggerOnFinish` and `Trigger` decorators? That way, we can have a proper delineation that `TriggerOnFinish` is a first-class metaflow concept that underlying workflow orchestrators may not support out of the box. For example - Airflow doesn't have any notion of events, but someday we would like to introduce `TriggerOnFinish` for Airflow (native support for Airflow events is unlikely since they don't exist at the moment).

savingoyal (changes requested):
First pass UX review

metaflow/plugins/argo/event_decorators.py
```diff
@@ -0,0 +1,247 @@
...
+
+    name = "trigger_on"
+    defaults = {
+        "flow": None,
+        "status": "succeeded",
+        "event": None,
+        "data": None,
```
savingoyal, line 75: uninvited guest to the party?

metaflow/plugins/argo/event_decorators.py
savingoyal: it is quite likely that we may need some customization for Argo Workflows - let's chat with the [Deliveroo folks](https://outerbounds-community.slack.com/archives/C02116BBNTU/p1663344880751489?thread_ts=1663253873.389359&cid=C02116BBNTU) independent of this effort.

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -171,10 +193,14 @@ def create(
         max_workers,
         workflow_timeout,
         workflow_priority,
+        ignore_events,
     )
 
     if only_json:
         obj.echo_always(str(flow), err=False, no_bold=True)
+        if flow.trigger_on_template() is not None:
```
savingoyal, line 201: We can skip printing the sensor template or introduce a CLI arg that prints just that.

metaflow/plugins/argo/argo_client.py
```diff
@@ -203,3 +203,100 @@ def schedule_workflow_template(self, name, schedule=None):
             raise ArgoClientException(
                 json.loads(e.body)["message"] if e.body is not None else e.reason
             )
+
+    def register_trigger_on_template(self, name, sensor_template):
```
savingoyal, line 207: maybe just `register_sensor`

metaflow/plugins/argo/argo_client.py
```diff
@@ -203,3 +203,100 @@ def schedule_workflow_template(self, name, schedule=None):
...
+                plural="sensors",
+                name=name,
+                body=sensor_template,
+            )
+        except self._client.rest.ApiException as e:
+            raise ArgoClientException(
+                json.loads(e.body)["message"] if e.body is not None else e.reason
+            )
+
+    def disable_existing_trigger_on_template(self, name):
```
savingoyal, line 258: maybe just `disable_sensor`?

savingoyal: Also, wouldn't we need the same concurrency control measure for this method as with all the other ones in this file?

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -360,6 +449,8 @@ def _compile(self):
                 )
                 # Set the entrypoint to flow name
                 .entrypoint(self.flow.name)
+                # Set succeeded lifecycle hook
+                .lifecycle_hooks(self.flow.name, ignore_events=self._ignore_events)
```
savingoyal, line 453: nit - we can skip setting the lifecycle hook if `ignore_events` is set to True rather than embedding the skip logic inside the method.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -772,17 +857,19 @@ def _container_templates(self):
...
                         "ARGO_WORKFLOW_TEMPLATE": self.name,
                         "ARGO_WORKFLOW_NAME": "{{workflow.name}}",
                         "ARGO_WORKFLOW_NAMESPACE": KUBERNETES_NAMESPACE,
+                        "METAFLOW_EVENT_SOURCE": EVENT_SOURCE_URL,
+                        "METAFLOW_EVENT_SOURCE_NAME": EVENT_SOURCE_NAME,
                     },
                     **self.metadata.get_runtime_environment("argo-workflows"),
                 }
             )
+            if retry_count is not None:
```
savingoyal, line 871: What's the reason for this change?

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -234,14 +316,21 @@ def _process_parameters(self):
...
                 )
-            value = deploy_time_eval(param.kwargs.get("default"))
-            # If the value is not required and the value is None, we set the value to
-            # the JSON equivalent of None to please argo-workflows.
-            if not is_required or value is not None:
-                value = json.dumps(value)
-            parameters.append(
-                dict(name=param.name, value=value, description=param.kwargs.get("help"))
-            )
+            if include_values:
```
savingoyal, line 319: nit - `self.parameters` is already available to `SensorTemplate` - we can ignore the values set for parameters and save a duplicate evaluation of parameter names.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -954,6 +1041,28 @@ def _container_templates(self):
...
+                            "/", ""
+                        )
+                    else:
+                        template_env["METAFLOW_EVENT_SOURCE_URL"] = EVENT_SOURCE_URL
+                    yield self._make_lifecycle_hook_container_template(
+                        template_env, "succeeded"
+                    )
+
+    def _make_lifecycle_hook_container_template(self, env, status):
+        t = WorkflowLifecycleHookContainerTemplate(current_flow_name(), status)
```
savingoyal, line 1062: All invocations of `current_flow_name` can be replaced with `self.name`

metaflow/plugins/argo/util.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def current_flow_name():
+    flow_name = current.get("project_flow_name")
+    if flow_name is None:
+        flow_name = current.flow_name
+    return flow_name.lower()
+
+
+def project_and_branch():
```
savingoyal, line 40: how about `return (current.get("project_name"), current.get("branch_name"))`

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -114,19 +134,43 @@ def __init__(
...
             )
+            if are_events_configured():
+                if self._trigger_on_template is not None:
+                    client.register_trigger_on_template(
+                        self._trigger_on_template.name,
+                        self._trigger_on_template.to_json(),
+                    )
+                else:
+                    client.disable_existing_trigger_on_template(
+                        format_sensor_name(current_flow_name())
```
savingoyal, line 172: we can replace `format_sensor_name(current_flow_name())` with `self.name`

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -140,10 +145,20 @@ def create(
...
     workflow_priority=None,
+    ignore_events=False,
 ):
     validate_tags(tags)
 
     obj.echo("Deploying *%s* to Argo Workflows..." % obj.workflow_name, bold=True)
+    if ignore_events:
+        obj.echo("Workflow trigger setup and event dispatch will be skipped.")
+    else:
+        data = current.get("event_decorator_error")
```
savingoyal, line 156: Does this still apply?

metaflow/plugins/argo/eventing.py
```diff
@@ -0,0 +1,169 @@
...
+            "event_type": event_type,
+            "data": event_data,
+            "pathspec": current.pathspec,
+            "timestamp": int(datetime.utcnow().timestamp()),
+        }
+    }
+
+
+def send_event(event_name, event_data={}):
+    flow_name = current_flow_name()
```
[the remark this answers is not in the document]
savingoyal: Also, we account for name length limits for the workflow template name - so just using them here will ensure that we don't have to worry about them much here.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -1,10 +1,13 @@
...
 import json
 import os
 import shlex
 import sys
+import time
 from collections import defaultdict
+from urllib import parse
 
-from metaflow import current
+from metaflow.current import current
```
savingoyal, line 10: this shouldn't be needed?

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -72,6 +89,7 @@ def __init__(
         max_workers=None,
         workflow_timeout=None,
         workflow_priority=None,
+        ignore_events=False,
```
savingoyal, line 92: the CLI should be able to determine whether the event backend is configured or not and pass `ignore_events=True` if the event backend is missing? that will remove the need for `are_events_configured` method's usage in this file.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -114,19 +132,42 @@ def __init__(
...
                 self.name, self._workflow_template.to_json()
             )
+            if are_events_configured():
+                if self._sensor_template is not None:
+                    client.register_sensor_template(
+                        self._sensor_template.name,
+                        self._sensor_template.to_json(),
+                    )
+                else:
+                    client.disable_sensor(format_sensor_name(self.name))
```
savingoyal, line 170: nit - sensor name could be simply `self.name` and (minor nit)`client.register_sensor_template(name, payload=None)` can perhaps take care of unsetting the sensor if the payload is missing.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -189,6 +230,51 @@ def trigger_explanation(self):
                 "This workflow triggers automatically via the CronWorkflow *%s*."
                 % self.name
             )
+        elif len(self._triggering_flows) > 0 or len(self._triggering_events) > 0:
```
savingoyal, line 233: it seems from this block that `@schedule` and `@trigger...` both are not supported concurrently, but I don't see that enforced anywhere in the code.

metaflow/plugins/argo/argo_workflows.py
kevsmith: Replaced `current_flow_name()` refs with `self.name` but left the separate function in place. I think it makes it easier to grok the overall flow.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -1392,3 +1545,278 @@ def to_json(self):
...
     def __str__(self):
         return json.dumps(self.payload, indent=4)
+
+
+class WorkflowLifecycleHookContainerTemplate(object):
+    def __init__(self, flow_name, lifecycle_event):
+        tree = lambda: defaultdict(tree)
+        self.payload = tree()
+        self._requests = {"requests": {"cpu": "1", "memory": "512M"}}
+        self._image = "ubuntu:22.04"
```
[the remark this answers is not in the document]
kevsmith: Changed to DEFAULT_CONTAINER_IMAGE

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -171,10 +186,14 @@ def create(
...
         workflow_timeout,
         workflow_priority,
+        ignore_events,
     )
 
     if only_json:
         obj.echo_always(str(flow), err=False, no_bold=True)
+        if flow.sensor_template() is not None:
+            obj.echo_always("", err=False, no_bold=True)
+            obj.echo_always(str(flow.sensor_template()), err=False, no_bold=True)
```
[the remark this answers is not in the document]
kevsmith: Added a new flag `--include-events` which is only evaluated in conjunction with `--only-json` which defaults to False. This preserves the existing behavior and still provides a way to dump event JSON for troubleshooting.

metaflow/plugins/argo/event_decorators.py
```diff
@@ -0,0 +1,320 @@
...
+
+class TriggerOnFinishDecorator(TriggerDecorator):
+
+    name = "trigger_on_finish"
+    defaults = {"flow": None, "flows": [], "reset_at": ""}
+    options = {
+        "flow": dict(
+            is_flag=False,
+            show_default=True,
+            help="Trigger the current flow when the named flow completes.",
```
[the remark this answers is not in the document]
kevsmith: Done

metaflow/plugins/argo/event_decorators.py
```diff
@@ -0,0 +1,320 @@
...
+            else:
+                self.attributes["trigger_set"] = TriggerSet(None, None)
+            for flow in flows:
+                info = TriggerInfo(TriggerInfo.LIFECYCLE_EVENT)
+                info.name = flow
+                info.status = "succeeded"
+                self.attributes["trigger_set"].append(info)
+        else:
+            # Defer raising an error in case user has specified --ignore-triggers
+            self.error = {
```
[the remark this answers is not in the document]
kevsmith: Yeah, probably. Will fix.

metaflow/plugins/argo/argo_workflows.py
kevsmith: Good catch. Rolled back.

savingoyal (changes requested):

romain-intel (commented):
A few other top-level comments:
- what is the rationale for distinguishing "flows" and "events" in the TriggerOn decorator. I can see TriggerOnFinish as a syntactic sugar that is useful if it takes something different than the TriggerOn decorator. Internally, for example, TriggerOnFinish takes a flow name but that gets effectively converted to a `TriggerOn` depending on an event which is derived from the flow name. At that point, the `TriggerOn` can only depend on events and the "a flow finished" is an event that has a well defined name.

I am also curious as to how you envision the signal data to mesh with the "required" and "default" parameter. Specifically:
- for required=False; default=<something>: if the event does not have the field mapping to the parameter, I suppose the default is used, is this the case?
- for required=False; no default: if the event does not have the field, I suppose the flow still proceeds with a value of "None", is this the case?
- for required=True, default is ignored: if the event does not have the field, the flow is not triggered. Is this correct? (or maybe it is an error reported to the user).

metaflow/plugins/argo/event_decorators.py
```diff
@@ -0,0 +1,320 @@
...
+            is_flag=False,
+            show_default=True,
+            help="Trigger the current flow when the named flow completes.",
+        ),
+        "flows": dict(
+            is_flag=False,
+            show_default=False,
+            help="Trigger the current flow when all named flows complete.",
+        ),
+        "reset_at": dict(
```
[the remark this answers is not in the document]
savingoyal: If you are merging two events, then one needs to specify a window for the merge. The `resest_at` field starts a timer when the first event shows up for the other events to introduce themselves.

metaflow/plugins/argo/event_decorators.py
```diff
@@ -0,0 +1,320 @@
...
+                self.attributes["parsed_reset_at"] = parsed_time
+            except ValueError:
+                try:
+                    parsed_time = strptime(reset_at, "%I:%M%p")
+                    self.attributes["parsed_reset_at"] = parsed_time
+                except ValueError:
+                    raise BadResetAfterException(msg=BAD_RESET_AFTER_MSG)
+
+
+class TriggerOnFinishDecorator(TriggerDecorator):
```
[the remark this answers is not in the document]
savingoyal: It's simply a syntactic sugar and is easy for Metaflow users to wrap their head around.

metaflow/plugins/argo/event_decorators.py
```diff
@@ -0,0 +1,320 @@
...
+
+
+class TriggerOnDecorator(TriggerDecorator):
+
+    name = "trigger_on"
+    defaults = {
+        "flow": None,
+        "flows": [],
+        "event": None,
+        "events": [],
```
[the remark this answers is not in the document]
savingoyal: @romain-intel As discussed, can you share what your internal APIs/UX look like?

leeyh20:
Will this allow us to trigger AWS step functions / argo workflow when there is a change in an AWS bucket?

metaflow/plugins/argo/argo_workflows.py
kevsmith: Cleaned this up a bit. We can't use `self.name` for the sensor name when NATS or Jetstream is enabled. Argo Events uses the sensor name to establish a connection which also seems to allocate some per-client storage in the messaging cluster (buffer perhaps?). In any case, the messaging cluster complains when the sensor name contains `.`. 

As a workaround I changed the sensor naming code to just replace `.` with `-` and dropped the `mf` and `trigger` prefix and suffix.

metaflow/plugins/argo/event_decorators.py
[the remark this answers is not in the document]
kevsmith: This is actually a wall clock time at which to reset the trigger. I agree some kind of interval -- "reset if all events aren't received within 4 hours" -- is vastly better. Unfortunately Argo Events only exposes a [cron-like reset mechanism](https://argoproj.github.io/argo-events/sensors/trigger-conditions/#conditions-reset).

Sensor are also evaluated in a stateless environment which complicates storing and retrieving the state necessary to implement reset intervals.

metaflow/plugins/argo/argo_workflows.py
kevsmith: Will fix

metaflow/plugins/argo/eventing.py
```diff
@@ -0,0 +1,169 @@
...
+            )
+        ).replace(".", "-")
+    else:
+        event_name = event_name.replace(".", "-")
+    if re.fullmatch("[a-z0-9\-_\.]+", event_name) is None:
+        raise BadEventNameException(
+            ("Attempted to send '%s'. " % event_name)
+            + "Event names can only contain lowercase characters, digits, '.', '_', and '-'."
+        )
+    body = make_event_body(event_name, "metaflow_user", event_data=event_data)
```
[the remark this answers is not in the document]
kevsmith: `metaflow_user` and `metaflow_system` are intended to partition the event namespace such that users can't accidentally (or maliciously) impersonate system events. Renaming `metaflow_system` to `metaflow_system_finished` could make sense.

metaflow/plugins/argo/event_decorators.py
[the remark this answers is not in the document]
kevsmith: Last commit fully extracts all trigger on functionality into `TriggerOnDecorator` leaving `TriggerOnFinishDecorator` as only syntactic sugar.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -1397,3 +1593,422 @@ def to_json(self):
...
     def __str__(self):
         return json.dumps(self.payload, indent=4)
+
+
+class WorkflowMetadataHookContainerTemplate:
+    def __init__(self, parameters):
+        self._name = "on-run-started"
+        self._env = dict()
+        event_parameters = [
+            p["name"] for p in parameters if p["name"].startswith("mf-event")
```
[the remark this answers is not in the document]
kevsmith: I'm not sure I follow. I'm able to trigger flows with and without parameters. Could you provide an example?

metaflow/plugins/argo/util.py
```diff
@@ -0,0 +1,122 @@
...
+    return (current.get("project_name"), current.get("branch_name"))
+
+
+# Creates the Argo Event sensor name for a given workflow
+def format_sensor_name(flow_name):
+    # Argo's internal NATS bus throws an error when attempting to
+    # create a persistent queue for the sensor when the sensor name
+    # contains '.' which is NATS' topic delimiter.
+    return flow_name.replace("_", "").replace(".", "-").lower()
```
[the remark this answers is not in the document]
kevsmith: Good point. Will add.

metaflow/plugins/argo/event_decorators.py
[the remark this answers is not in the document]
romain-intel: although I think it is useful to think how ti would map to argo (or others), we should think as well more generally so the UX is stable and different implemenations may or may not implement everything. I do understand the concept now. We don't have anything like this internally. I am, however, not 100% sure if a wallclock time is very useful (most flows are periodic so a wall-clock time is a bit weird).

metaflow/plugins/argo/event_decorators.py
```diff
@@ -0,0 +1,320 @@
...
+
+    name = "trigger_on_finish"
+    defaults = {"flow": None, "flows": [], "reset_at": ""}
+    options = {
+        "flow": dict(
+            is_flag=False,
+            show_default=True,
+            help="Trigger the current flow when the named flow completes.",
+        ),
+        "flows": dict(
```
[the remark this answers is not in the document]
romain-intel: it's fine if it is not the focus, but if we are going to do it (I think we should), we can have it in the UX. This also goes back to the somewhat hazy separation between trigger_on_finish and trigger_on_event. IMHO, the trigger_on_finish is effectively a sugary wrapper around the trigger_on_event and I think it is useful but should provide some real functionality.

metaflow/plugins/argo/event_decorators.py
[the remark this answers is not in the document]
romain-intel: done

metaflow/plugins/argo/event_decorators.py
```diff
@@ -0,0 +1,242 @@
...
+            if type(v) != str:
+                raise BAD_MAPPING_ERROR
+    return mappings
+
+
+class TriggerOnDecorator(FlowDecorator):
+
+    name = "trigger_on"
+    defaults = {
+        "flow": None,
```
romain-intel, line 102: @savingoyal : I don't think we need to have flow and flows and event and events. We can stick to just one and take a list (or just consider a single element to be a list of a single element). IMHO, the API should be fairly straightfoward.

I would also not stick to just "flows" because we can trigger at the task level (task is not user visible but we should be able to allow users to trigger on steps at the very least).

romain-intel: @savingoyal : I also suppose this is an "and" type of situation (ie: the trigger only happens if everything specified here matches). Is this the assumption?

metaflow/plugins/argo/event_decorators.py
```diff
@@ -0,0 +1,242 @@
...
+    def _report_empty_inputs(self):
+        raise MetaflowException(
+            msg=(MISSING_INPUTS_MSG.format(self.name, "one flow or event name"))
+        )
+
+
+class TriggerOnFinishDecorator(TriggerOnDecorator):
+
+    name = "trigger_on_finish"
+    defaults = {"flow": None, "flows": [], "event": None, "options": dict()}
```
romain-intel, line 221: @savingoyal : why do we have an event here? It should be just flows (or steps as mentioned earlier).

romain-intel: @savingoyal : we also need cross branch/project support. This is fairly key to be able to "branch" off of a production flow and is used for debugging and experimentation.

metaflow/plugins/argo/event_decorators.py
romain-intel: @savingoyal : these fields also need to be deploy-time evaluated. I don't think this is a huge problem but mentioning it as this is a fairly key feature in some of our workflows.

romain-intel (commented):
A few more high-level comments:
  - we need to be clear on OR vs AND semantics. My understanding is that within a single decorator, it's an AND semantic and across decorators that would be an OR
  - we need to be clear on failure conditions. For example, what happens if a signal contains a field that is not mapped (that should probably be ok and ignored). What happens if a signal is missing a field that is mapped (this should probably revert to what the parameter has -- if a default, use that, if not, error out). I don't think we have done a good job of this in the past so we should clearly spell out the various conditions.

savingoyal:
superceded by https://github.com/Netflix/metaflow/pull/1271

PR #1143, opened 2022-09-20
Add `cmd` extension point to allow MF extensions to extend the `metaf…

romain-intel:
…low` command

Simply add a `cmd` directory in your extension and the usual __init__.py or mfextinit_*.py file containing a function called `get_cmd_clis` which returns a list of CLIs to add.

Example:

@click.group()
@click.pass_context
def cli(ctx):
  pass

@cli.group(help="My commands")
@click.pass_context
def foobar(ctx):
  pass

@cli.command(help="Overrides provided `status` command") @click.pass_context
def status(ctx):
  print("I am overriding the usual status command")

@foobar.command(help="Some other command")
@click.pass_context
def baz(ctx):
  print("Hi!")

def get_cmd_clis():
  return [cli]

### DIFF
--- a/metaflow/main_cli.py
+++ b/metaflow/main_cli.py
@@ -1009,7 +1009,7 @@ try:
 except Exception as e:
     _ext_debug("\tWARNING: ignoring all plugins due to error during import: %s" % e)
     print(
-        "WARNING: Plugins did not load -- ignoring all of them which may not "
+        "WARNING: Command extensions did not load -- ignoring all of them which may not "
         "be what you want: %s" % e
     )
     _clis = []
--- a/metaflow/main_cli.py
+++ b/metaflow/main_cli.py
@@ -1,3 +1,5 @@
+import builtins
+import traceback
 from metaflow._vendor import click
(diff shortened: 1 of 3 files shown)


metaflow/main_cli.py
```diff
@@ -31,50 +33,10 @@ def echo_always(line, **kwargs):
...
-            "status": "Display the current working tree.",
-            "help": "Show all available commands to run.",
-        }
-
-        echo("Commands:", bold=False)
-
-        for cmd, desc in short_help.items():
-            echo("  metaflow {0:<10} ".format(cmd), fg="cyan", bold=False, nl=False)
-
-            echo("%s" % desc)
```
savingoyal, line 77: Any reason to remove the commands section?

romain-intel: Yes, it doesn't cover the commands added as part of extensions making it confusing. Further, click does this just fine as well.

romain-intel:
Tests are OK -- failure is flaky.

PR #1149, opened 2022-09-25
bump patch version for release

savingoyal:


romain-intel:
Re-running tests -- the failure is not one I was familiar with. I expect it to be transient though so let's see.

PR #1151, opened 2022-09-27
Remove message introduced in #1061

romain-intel:


metaflow/runtime.py
```diff
@@ -253,17 +253,15 @@ def execute(self):
                             )
 
                     self._logger(msg, system_msg=True)
-                    if len(self._unprocessed_steps) == 0:
-                        msg = "All steps have started."
-                    else:
+                    if len(self._unprocessed_steps) > 0:
                         if len(self._unprocessed_steps) == 1:
                             msg = "1 step has not started: "
```
savingoyal, line 258: We can further improve the messaging to "foo has not started yet." but it's okay if you want to ship this PR as is.

savingoyal (approved):

PR #1152, opened 2022-09-28
Pass datastore_type to validate_environment

romain-intel:


romain-intel:
For additional context, going by the lifecycle.png document: https://github.com/Netflix/metaflow/blob/master/docs/lifecycle.png, you can see that `validate_environment` is always called but *not* `init_environment`. In the new conda implementation, I need to know the datastore type (so I can properly figure out where to get things from) in all cases so need the information in `validate_environment`.

savingoyal (approved):

PR #1155, opened 2022-09-29
Simplifying Airflow parameter creation

valayDave:
This PR Simplifies how parameters are parsed. It also moves around the code that raises errors during dag compilation to raising errors before DAG compilation (if there are any).

### DIFF
--- a/metaflow/plugins/airflow/airflow.py
+++ b/metaflow/plugins/airflow/airflow.py
@@ -172,42 +172,18 @@ class Airflow(object):
         return None
 
     def _process_parameters(self):
-        seen = set()
         airflow_params = []
         type_transform_dict = {
             int.__name__: "integer",
             str.__name__: "string",
             bool.__name__: "string",
             float.__name__: "number",
-            JSONTypeClass.name: "string",
-            FilePathClass.name: "string",
-        }
-
-        type_parser = {
-            bool.__name__: lambda v: str(v),
         }
 
         for var, param in self.flow._get_parameters():
-            # Throw an exception if the parameter is specified twice.
-            norm = param.name.lower()
-            if norm in seen:
-                raise MetaflowException(
-                    "Parameter *%s* is specified twice. "
-                    "Note that parameter names are "
-                    "case-insensitive." % param.name
-                )
-            seen.add(norm)
-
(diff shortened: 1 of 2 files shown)


metaflow/plugins/airflow/airflow.py
```diff
@@ -172,63 +172,40 @@ def _get_retry_delay(self, node):
...
-                if param_type_name in type_transform_dict:
-                    airflow_param["type"] = type_transform_dict[param_type_name]
-                    if param_type_name in type_parser and value is not None:
-                        airflow_param["default"] = type_parser[param_type_name](value)
+
+            # Since we will always have a default value and `deploy_time_eval` resolved that to an actual value
+            # we can just use the `default` to infer the object's type.
+            # This avoids parsing/indentifying types like `JSONType` or `FilePathClass`
+            # which are returned by calling `param.kwargs.get("type")`
+            param_type = type(airflow_param["default"])
```
romain-intel, line 199: Does JSONType return a string or can it return a dict? I suspect it returns the dict which may require a dumps at least so maybe something that does a dumps if it is not in param_type?

valayDave: Isn't the default value of JSONType if supposed to be a JSON string?

valayDave: My assumption is that JSONType's default value should be a string.

valayDave: I don't want to run json.dumps because _I don't know_ if it may mess up parsing when the user has already run json.dumps when feeding the default.

valayDave: Our [Docs](http://docs.metaflow.org/metaflow/basics#how-to-define-parameters-for-flows) expect JSON strings

savingoyal (approved):

metaflow/plugins/airflow/airflow.py
romain-intel: Yes, and I verified the evaluation of the default will return a string and not actually do the loads; only the conversion function does the loads so this is all good. Sorry for the confusion.

valayDave:
This PR can be merged.

PR #1156, opened 2022-09-30
Metaflow Proposal format

romain-intel:


docs/mfp/mfp-1.md
```diff
@@ -0,0 +1,84 @@
+# MFP-1 - Metaflow Proposal Format
```
savingoyal, line 1: Can we use a google doc instead? Easier to comment and collaborate on compared to PRs.

romain-intel: Sure, I don't mind a google doc. The only issue is permissions. Here it's a PR and everyone can submit PRs to it but google doc has a bit less version control. I'll think about it a bit.

PR #1158, opened 2022-10-05
Added typing information for all public APIs

romain-intel:
The docstrings were also slightly more standardized.

Only public facing APIs had the type information added to them.

There are a few minor bug fixes in the default values for some functions (namely using [] as a default value (replaced with None)) as well as removing default values where it does not make sense (S3.info() for example). There should be no other functional change.


romain-intel:
I have a typo -- will fix.

romain-intel:
I am still working on this to make it fully compatible with the dev command being developed. It's mostly formatting the strings and what not but want to do it only once :).

metaflow/includefile.py
```diff
@@ -300,7 +301,13 @@ class IncludeFile(Parameter):
     """
 
     def __init__(
-        self, name, required=False, is_text=True, encoding=None, help=None, **kwargs
+        self,
+        name: str,
+        required: bool = False,
+        is_text: bool = True,
+        encoding: str = None,
+        help: str = None,
```
phitoduck, line 309: I saw you used `Optional` in other instances of parameters which could be `None`. Should `Optional` be used here as well?

metaflow/datatools/s3.py
```diff
@@ -449,7 +471,13 @@ def get_root_from_config(cls, echo, create_on_absent=True):
         return DATATOOLS_S3ROOT
 
     def __init__(
-        self, tmproot=".", bucket=None, prefix=None, run=None, s3root=None, **kwargs
+        self,
+        tmproot: str = ".",
+        bucket: Optional[str] = None,
+        prefix: Optional[str] = None,
+        run: Optional[Union[FlowSpec, TRun]] = None,
```
phitoduck, line 478: Been staring at this for a moment. Why the use of a TypeVar? I know these can be used for generics to enhance autocompletion when using subclasses.

Example:

```python
from typing import Generic, TypeVar
from dataclasses import dataclass
from .some_file import Weapon

TWeapon = TypeVar("TWeapon", bound=Weapon)

@dataclass
class Enemy(Generic[TWeapon]):
    holding: TWeapon

enemy = Enemy(holding=...some weapon subclass instance)
enemy.holding # gives autocompletion for the specific subclass of Weapon in holding
```

Is that somehow what's going on with `TRun`?  Does `Union[type, type]` simply require TypeVar objects rather than the string representation, i.e. `"Run"`?

metaflow/plugins/cards/card_modules/components.py
```diff
@@ -103,15 +112,15 @@ def __init__(self, data=[[]], headers=[]):
             self._data = data
 
     @classmethod
-    def from_dataframe(cls, dataframe=None, truncate=True):
+    def from_dataframe(cls, dataframe=None, truncate: bool = True):
```
phitoduck, line 115: Was `dataframe: pd.DataFrame` not used here because it would require a new import statement `import pandas as pd`?

metaflow/flowspec.py
```diff
@@ -354,7 +358,12 @@ def _find_input(self, stack_index=None):
                     )
             return self._cached_input[stack_index]
 
-    def merge_artifacts(self, inputs, exclude=[], include=[]):
+    def merge_artifacts(
+        self,
+        inputs: Sequence[TFlowSpec],
```
phitoduck, line 363: Similar question to elsewhere: why `TFlowSpec` instead of `FlowSpec`? Feel free to ignore this question if answered elsewhere.

phitoduck:
@romain-intel thanks for taking this on! Fingers crossed this gets accepted :)

metaflow/includefile.py
romain-intel: Good point. Yes, it probably should. I'll fix it.

metaflow/datatools/s3.py
romain-intel: Let me try with "Run". I have to check it works on 3.5 too. IIRC that may be why I went this route but may have been delusional :) .

metaflow/plugins/cards/card_modules/components.py
romain-intel: Yes, I would like to avoid the import. Maybe I can use a string here too.

romain-intel:
Thanks for the review. I need to rebase this and clean it up some more but yes, I am also hoping we can get this in. @tuulos , what do you think?

romain-intel:
sorry for the delay @phitoduck  -- I am back to working on this to address your comments, update and hopefully merge soon :).

metaflow/flowspec.py
```diff
@@ -264,12 +265,12 @@ def input(self):
...
         Returns
         -------
-        object
+        object, optional
             Input passed to the foreach task.
         """
         return self._find_input()
 
-    def foreach_stack(self):
+    def foreach_stack(self) -> List[Tuple[int, int, Any]]:
```
tuulos, line 273: shouldn't this be `Optional[List[Tuple[int, int, Any]]]`?

metaflow/flowspec.py
```diff
@@ -386,7 +392,7 @@ def merge_artifacts(self, inputs, exclude=[], include=[]):
 
         Parameters
         ----------
-        inputs : List[Steps]
+        inputs : List[FlowSpec]
```
tuulos, line 395: technically inputs is not a list of `FlowSpec` but a list of `Input` objects that get passed to the join step

metaflow/parameters.py
```diff
@@ -246,21 +246,51 @@ class MyFlow(FlowSpec):
...
+                Dict[str, Any],
+                Callable[[], Union[str, float, int, bool, Dict[str, Any]]],
+            ]
+        ] = None,
+        type: Optional[
+            Union[Type[str], Type[float], Type[int], Type[bool], JSONTypeClass]
+        ] = None,
+        help: Optional[str] = None,
+        required: bool = False,
+        show_default: bool = True,
```
tuulos, line 279: this changes the semantics a bit since e.g. `required` is `False` if not defined, not `None` as before. I hope it doesn't change the behavior in `click` but needs extra attention

metaflow/plugins/cards/card_modules/components.py
```diff
@@ -11,6 +12,12 @@
 from .convert_to_native_type import TaskToDict, _full_classname
 from .renderer_tools import render_safely
 
+if TYPE_CHECKING:
+    from pandas import DataFrame
```
tuulos, line 16: doesn't this introduce a hard dependency to `PIL`, `pandas`, `matplotlib` which is not desirable?

metaflow/parameters.py
```diff
@@ -246,21 +246,51 @@ class MyFlow(FlowSpec):
         indicate that the value must be a valid JSON object. A function
         implies that the parameter corresponds to a *deploy-time parameter*.
         The type of the default value is used as the parameter `type`.
-    type : type
+    type : Type, default: None
```
tuulos, line 249: `default: str`, right?

metaflow/plugins/cards/card_modules/components.py
```diff
@@ -103,15 +118,17 @@ def __init__(self, data=[[]], headers=[]):
...
+    def from_dataframe(
+        cls, dataframe: Optional["DataFrame"] = None, truncate: bool = True
+    ):
         """
         Create a `Table` based on a Pandas dataframe.
 
         Parameters
         ----------
-        dataframe : pandas.DataFrame
+        dataframe : pandas.DataFrame, optional
```
tuulos, line 129: `dataframe=None` is a bit incorrect in the function signature - `dataframe` is not `optional` here

metaflow/plugins/cards/card_modules/components.py
```diff
@@ -303,16 +320,18 @@ def from_pil_image(cls, pilimage, label=None):
             )
 
     @classmethod
-    def from_matplotlib(cls, plot, label=None):
+    def from_matplotlib(
+        cls, plot: Union["Figure", "Axes"], label: Optional[str] = None
```
tuulos, line 324: docs say that `AxesSubplot` is ok too

metaflow/plugins/datatools/s3/s3.py
```diff
@@ -64,17 +68,32 @@ def ensure_unicode(x):
...
+    [
+        ("key", str),
+        ("value", Optional[PutValue]),
+        ("path", Optional[str]),
+        ("content_type", Optional[str]),
+        ("metadata", Optional[Dict[str, str]]),
+    ],
     defaults=(None, None, None, None),
 )
+S3PutObject.__module__ = __name__
```
tuulos, line 89: what's this for?

metaflow/plugins/datatools/s3/s3.py
```diff
@@ -682,7 +707,7 @@ def _list(keys):
 
         return list(starmap(S3Object, _list(keys)))
 
-    def info(self, key=None, return_missing=False):
```
tuulos, line 685: hmm, `.info()` with `key=None` should work, right?

metaflow/flowspec.py
romain-intel: I think it's just `Inputs` actually. Corrected this. Thanks for catching this.

metaflow/parameters.py
romain-intel: Looking at the page here: https://click.palletsprojects.com/en/7.x/api/#click.Parameter, required defaults to False when not specified. For type, we do default to str (so made it that explicitly), for show_default, we had a line that set it to True if None. This seems to make it clearer to the user (hopefully).

metaflow/plugins/cards/card_modules/components.py
romain-intel: I think it adds a dependency only for type resolution. 
```
>>> import PIL
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'PIL'
>>> from metaflow.plugins.cards.card_modules.components import Table
>>>
```

metaflow/plugins/datatools/s3/s3.py
romain-intel: It was when I was generating the .pyi. I think basically since `namedtuple_with_defaults` is in another module, it was placing those objects there too.

metaflow/plugins/datatools/s3/s3.py
romain-intel: It has very little value but I'll switch it back.

metaflow/client/core.py
```diff
@@ -2004,17 +1926,106 @@ def runs(self, *tags):
...
         Returns
         -------
-        Iterator[Run]
+        Iterable[Run]
             Iterator over `Run` objects in this flow.
         """
         return self._filtered_children(*tags)
 
 
+class Metaflow(object):
```
savingoyal, line 1940: Any reasons to move this object to the end? Are there any diffs here that I should be looking into?

metaflow/plugins/aws/batch/batch_decorator.py
```diff
@@ -33,39 +33,35 @@ class BatchDecorator(StepDecorator):
...
+        Number of GPUs required for this step. If `@resources` is
         also present, the maximum value from all decorators is used.
-    memory : int
-        Memory size (in MB) required for this step. Defaults to 4096 (4GB). If
+    memory : int, default: 4096
+        Memory size (in MB) required for this step. If
         `@resources` is also present, the maximum value from all decorators is
         used.
-    image : string
+    image : str, optional, default: None
```
savingoyal, line 46: the default isn't technically None here

metaflow/plugins/aws/batch/batch_decorator.py
```diff
@@ -33,39 +33,35 @@ class BatchDecorator(StepDecorator):
...
         AWS IAM role that AWS Batch container uses to access AWS cloud resources.
-        Defaults to the one specified by the configuration variable `METAFLOW_ECS_S3_ACCESS_IAM_ROLE`.
-    execution_role : string
+    execution_role : str, default: METAFLOW_ECS_FARGATE_EXECUTION_ROLE
         AWS IAM role that AWS Batch can use [to trigger AWS Fargate tasks]
         (https://docs.aws.amazon.com/batch/latest/userguide/execution-IAM-role.html).
-        Defaults to the one determined by the configuration variable
-        `METAFLOW_ECS_FARGATE_EXECUTION_ROLE`.
-    shared_memory : int
+    shared_memory : int, optional, default: None
```
savingoyal, line 56: do we need to specify default here and below?

metaflow/plugins/aws/step_functions/schedule_decorator.py
```diff
@@ -9,13 +9,13 @@ class ScheduleDecorator(FlowDecorator):
...
+    hourly : bool, default: False
+        Run the workflow hourly.
+    daily : bool, default: True
+        Run the workflow daily.
+    weekly : bool, default: False
+        Run the workflow weekly.
+    cron : str, optional
         Run the workflow at [a custom Cron schedule](https://docs.aws.amazon.com/eventbridge/latest/userguide/scheduled-events.html#cron-expressions)
         specified by this expression.
     timezone : str
```
savingoyal, line 21: timezone is also optional

metaflow/util.py
```diff
@@ -50,19 +50,14 @@ def unquote_bytes(x):
 
     from shlex import quote as _quote
 
-if sys.version_info >= (3, 7):
```
savingoyal, line 53: what's the context for this change?

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -39,29 +39,25 @@ class KubernetesDecorator(StepDecorator):
...
         `@resources` is also present, the maximum value from all decorators is
         used.
-    disk : int
-        Disk size (in MB) required for this step. Defaults to 10GB. If
+    disk : int, default: 10240
+        Disk size (in MB) required for this step. If
         `@resources` is also present, the maximum value from all decorators is
         used.
-    image : str
+    image : str, optional, default: None
```
savingoyal, line 53: this default isn't wholly accurate

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -39,29 +39,25 @@ class KubernetesDecorator(StepDecorator):
...
-    namespace : str
-        Kubernetes namespace to use when launching pod in Kubernetes. If
-        not specified, the value of `METAFLOW_KUBERNETES_NAMESPACE` is used
-        from Metaflow configuration.
-    secrets : List[str]
+    service_account : str, default: METAFLOW_KUBERNETES_SERVICE_ACCOUNT
+        Kubernetes service account to use when launching pod in Kubernetes.
+    namespace : str, default: METAFLOW_KUBERNETES_NAMESPACE
+        Kubernetes namespace to use when launching pod in Kubernetes.
+    secrets : List[str], optional, default: None
```
savingoyal, line 60: the default would be METAFLOW_KUBERNETES_SECRETS

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -39,29 +39,25 @@ class KubernetesDecorator(StepDecorator):
...
         Kubernetes secrets to use when launching pod in Kubernetes. These
         secrets are in addition to the ones defined in `METAFLOW_KUBERNETES_SECRETS`
         in Metaflow configuration.
```
savingoyal, line 63: default for tolerations would be METAFLOW_KUBERNETES_TOLERATIONS

metaflow/client/core.py
romain-intel: it needed `Flow` iirc.

romain-intel: There should be no change.

metaflow/plugins/aws/batch/batch_decorator.py
romain-intel: good catch. I'll change to default: BATCH_CONTAINER_IMAGE (I went by the default below but this is in the code for some reason.

metaflow/plugins/aws/batch/batch_decorator.py
romain-intel: I'll remove default: None in both cases.

metaflow/plugins/aws/step_functions/schedule_decorator.py
romain-intel: ah right, it was added after this PRR was done. I'll add it.

metaflow/plugins/kubernetes/kubernetes_decorator.py
romain-intel: I'll change.

metaflow/plugins/kubernetes/kubernetes_decorator.py
romain-intel: I'll change.

metaflow/util.py
romain-intel: so I could use `NamedTuple` and have the proper types.

metaflow/plugins/kubernetes/kubernetes_decorator.py
romain-intel: I left to optional beceause the doc string and code actually add it to KUBERNETES_SECRETS

metaflow/plugins/kubernetes/kubernetes_decorator.py
romain-intel: I put it as optional and described tthe behavior in the doc string.

metaflow/plugins/aws/batch/batch_decorator.py
romain-intel: I actually left it as optional and added an explanation.

metaflow/plugins/kubernetes/kubernetes_decorator.py
savingoyal: the behavior is symmetric to batch - so the comment should be quite similar here

savingoyal (commented):
just a couple of nits waiting to be resolved. once those are addressed, we should be good to merge.

metaflow/plugins/kubernetes/kubernetes_decorator.py
romain-intel: stupid me -- I had modified it but forgot to save the file.

romain-intel: stupid me -- I had modified it but forgot to save the file.

PR #1162, opened 2022-10-11
Add back Azure support

jackie-ob:
As title.

### DIFF
--- a/metaflow/plugins/azure/includefile_support.py
+++ b/metaflow/plugins/azure/includefile_support.py
@@ -57,13 +57,17 @@ class Azure(object):
         with storage.load_bytes([short_key]) as load_result:
             for _, tmpfile, _ in load_result:
                 if tmpfile is None:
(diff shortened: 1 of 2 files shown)


metaflow/plugins/azure/includefile_support.py
```diff
@@ -57,13 +57,17 @@ def get(self, key=None, return_missing=False):
         with storage.load_bytes([short_key]) as load_result:
```
romain-intel, line 57: As a quick note from the previous lines: why do you only support return_missing=True?

romain-intel (approved):
I haven't tested it but it LGTM (one minor comment). If this is tested on your end, you can merge.

metaflow/plugins/azure/includefile_support.py
jackie-ob: That is a good question - and it took me a while to remember.  I added a comment inline for posterity.

PR #1171, opened 2022-10-21
Move plugins, datastore and datatools and allow user choice of plugins.

romain-intel:
When reviewing, please review each commit separately as some are purely moving things around and do not need a review (but they account for a lot of the changed lines).


metaflow/metaflow_config.py
```diff
@@ -74,296 +96,285 @@ def _validate_choice(name, value):
...
 # Cards related config variables
-DATASTORE_CARD_SUFFIX = "mf.cards"
-DATASTORE_CARD_LOCALROOT = from_conf("METAFLOW_CARD_LOCALROOT")
-DATASTORE_CARD_S3ROOT = from_conf(
-    "METAFLOW_CARD_S3ROOT",
-    os.path.join(from_conf("METAFLOW_DATASTORE_SYSROOT_S3"), DATASTORE_CARD_SUFFIX)
-    if from_conf("METAFLOW_DATASTORE_SYSROOT_S3")
-    else None,
+CARD_SUFFIX = "mf.cards"
+CARD_LOCALROOT = from_conf("METAFLOW_CARD_LOCALROOT", propagate=False)
```
valayDave, line 206: Should this be `CARD_LOCALROOT` ?

metaflow/plugins/conda/batch_bootstrap.py
```diff
@@ -60,16 +60,17 @@ def download_conda_packages(flow_name, env_id, datastore_type):
                     )
         else:
             # Import DATASTORES dynamically... otherwise, circular import
-            from metaflow.datastore import DATASTORES
+            from metaflow.plugins import DATASTORES
 
-            if datastore_type not in DATASTORES:
+            datastore_potentials = [d for d in DATASTORES if d == datastore_type]
```
savingoyal, line 65: why datastore_potentials and not just datastore?

savingoyal: or storage_impl?

metaflow/metaflow_config.py
```diff
@@ -255,6 +255,54 @@
 for typ in DEBUG_OPTIONS:
     vars()["DEBUG_%s" % typ.upper()] = from_conf("DEBUG_%s" % typ.upper())
 
+###
+# Plugin configuration
+# List of plugins that are enabled.
```
savingoyal, line 260: what is the rationale behind listing everything as enabled here? Why can't we have everything enabled by default and only disable certain things within the extensions?

metaflow/__init__.py
```diff
@@ -108,7 +108,11 @@ class and related decorators.
 JSONType = JSONTypeClass()
 
 # data layer
-from .datatools import S3
+# For historical reasons, we make metaflow.plugins.datatools accessible as
+# metaflow.datatools. S3 is also a tool that has historically been available at the
+# TL so keep as is.
+lazy_load_aliases({"metaflow.datatools": "metaflow.plugins.datatools"})
```
savingoyal, line 114: should we move the method lazy_load_aliases out of `extension_support` now?

savingoyal: and can we start expanding TL as top level here and every where else? trips me up every time.

metaflow/plugins/conda/batch_bootstrap.py
romain-intel: sure. whatever name you want :).

metaflow/__init__.py
romain-intel: I'll expand TL. I don't know if we should move it out though; I've tried to keep everything related to extensions (aliasing basically) in one place.

metaflow/metaflow_config.py
romain-intel: The rationale was @tuulos 's comment about having a more traditional approach where you list the plugins you want (so basically have an ENABLED_ list. However, that was a bit hard to square with the notion of "no change from what we currently have" so this is what i came up with: you do have an ENABLED list where users can explicitly select what they want but the default is "everything" (as opposed to nothing). I am fine switching to something else but that was why I did it this way. Initially, iirc, I just had a notion of toggle and changed it to this based on feedback.

romain-intel:
Replaced by #1212

PR #1174, opened 2022-10-24
Fix `._orig` access for submodules for MF extensions

romain-intel:
This addresses an issue where submodules like `mymodule._orig.submodule` would not load properly.

This will properly load things like `from .foofa import xyz` from a module or sub-module in the original location but currently does not work if the import path is absolute.

### DIFF
--- a/metaflow/extension_support.py
+++ b/metaflow/extension_support.py
@@ -924,15 +924,22 @@ class _AliasLoader(Loader):
 
 
 class _OrigLoader(Loader):
-    def __init__(self, fullname, orig_loader, previously_loaded_module=None):
+    def __init__(
+        self,
+        fullname,
+        orig_loader,
+        previously_loaded_module=None,
+        previously_loaded_parent_module=None,
+    ):
         self._fullname = fullname
         self._orig_loader = orig_loader
         self._previously_loaded_module = previously_loaded_module
+        self._previously_loaded_parent_module = previously_loaded_parent_module
 
     def create_module(self, spec):
         _ext_debug(
-            "Loading original module '%s' (will be loaded at '%s')"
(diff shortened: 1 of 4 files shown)


roofurmston (commented):
Not an expert with the inner working of the Python import system, so learnt a lot from the PR. :)

the PR looks good to me. 

The one thing I wonder is if the changes merit some additional tests. You think it is worth adding some? Would be happy to have a stab at adding some, if you think it is worthwhile.

romain-intel:
Yep. I think tests would be nice. THere are some unit tests that test the extension mechanism. We can just add an extension that overrides something simple and see if we can access _orig. Would you be able to do that? Happy to provide pointers. And don't feel obligated, if you don't have time, I can do it too :).

roofurmston:
> Yep. I think tests would be nice. THere are some unit tests that test the extension mechanism. We can just add an extension that overrides something simple and see if we can access _orig. Would you be able to do that? Happy to provide pointers. And don't feel obligated, if you don't have time, I can do it too :).

Yeah, would be very happy to have a go at it :)

I'll have a crack at it next week.

tfurmston:
OK, so I've had a go at adding a test case. I picked the most innocuous looking module to extend.

The attempt is here, if you want to have a look - [https://github.com/Netflix/metaflow/pull/1192](https://github.com/Netflix/metaflow/pull/1192). Still waiting for the tests to start in CI, but I checked the `ExtensionsTest` locally and it went through fine. 

One thing I wasn't sure about, will the `frameworks` module be extended just for the `ExtensionsTest` test? Or will it be for all test casses?

savingoyal (approved):

PR #1183, opened 2022-11-03
Clean up configuration variables

romain-intel:


metaflow/metadata/heartbeat.py
```diff
@@ -4,7 +4,7 @@
 
 from threading import Thread
 from metaflow.sidecar import MessageTypes, Message
-from metaflow.metaflow_config import METADATA_SERVICE_HEADERS
```
valayDave, line 7: [NIT]: what was the problem with `METADATA_SERVICE_HEADERS`? Not a blocker but just curious.

metaflow/metaflow_config_funcs.py
```diff
@@ -0,0 +1,128 @@
...
+    # By default, we just return non-null values that are marked as "propagate" and that
+    # are not default. This is the common use case because in all other cases, the code
+    # is sufficient to recreate the value (ie: there is no external source for the value)
+    for name, config_value in _all_configs.items():
+        if (
+            (config_value.value is not None or include & NULL_VALUES)
+            and (config_value.is_propagate or include & NON_PROPAGATED_VALUES)
+            and (not config_value.is_default or include & NON_CHANGED_VALUES)
+        ):
+            yield name, config_value.serializer(config_value.value)
```
valayDave, line 52: Will this function be used by downstream integrations like `batch`, `sfn`, `argo`,`airflow` etc to just get all the config values for metaflow? That seems like a very good step forward! It makes any new external remote integration so much easier!.

metaflow/metaflow_config_funcs.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def from_conf(name, default=None, validate_fn=None, propagate=True):
```
valayDave, line 55: I really like the idea of `propagate`. Makes things much simpler when integrating with new schedulers/compute/datastore etc. Awesome stuff @romain-intel

metaflow/plugins/airflow/airflow.py
```diff
@@ -349,8 +349,8 @@ def _to_job(self, node):
             "METAFLOW_CODE_URL": self.code_package_url,
             "METAFLOW_CODE_DS": self.flow_datastore.TYPE,
             "METAFLOW_USER": user,
-            "METAFLOW_SERVICE_URL": BATCH_METADATA_SERVICE_URL,
-            "METAFLOW_SERVICE_HEADERS": json.dumps(BATCH_METADATA_SERVICE_HEADERS),
+            "METAFLOW_SERVICE_URL": SERVICE_URL,
+            "METAFLOW_SERVICE_HEADERS": json.dumps(SERVICE_HEADERS),
```
valayDave, line 353: I was thinking that why not call that `config_values` function here so that most of these can be removed directly and any new integration won't require changing multiple places. If that makes sense, I can throw a PR to add that here and a few other places.

metaflow/plugins/airflow/airflow.py
```diff
@@ -373,7 +373,7 @@ def _to_job(self, node):
             "METAFLOW_AZURE_STORAGE_BLOB_SERVICE_ENDPOINT"
         ] = AZURE_STORAGE_BLOB_SERVICE_ENDPOINT
         env["METAFLOW_DATASTORE_SYSROOT_AZURE"] = DATASTORE_SYSROOT_AZURE
-        env["METAFLOW_DATASTORE_CARD_AZUREROOT"] = DATASTORE_CARD_AZUREROOT
+        env["METAFLOW_CARD_AZUREROOT"] = CARD_AZUREROOT
```
valayDave, line 376: The same applies here too. If we already `propagate` then why not use the `config_values` function to get all the appropriate config values.

metaflow/metadata/heartbeat.py
romain-intel: The external user-facing env name was METAFLOW_SERVICE_HEADERS so I standardized to that (kind of needed to make it work :( ).

metaflow/metaflow_config_funcs.py
romain-intel: If you want it to be. I have not included it in any of the OSS integrations (`batch`, `sfn`, `argo`, `airflow`) because @savingoyal mentioned it would take a *lot* longer to test. I am using this internally for exactly the reason you specify (I am guaranteed that all configuration values are passed remotely and if a new one shows up, it is automatically passed along). But again, it's up to you if you want to use it or not. This PR does not make that change.

metaflow/plugins/airflow/airflow.py
romain-intel: That was what my first version did. I removed that to reduce the testing area. I am perfectly happy if you want to do that but to get this PR through quickly, I would rather that be a separate PR that you guys own and test out.

metaflow/plugins/airflow/airflow.py
romain-intel: Same response as above :).

romain-intel (commented):
Thanks for the review. It looks like we are mostly on the same page and there are no actionable changes that I see. If that's the case, let me know when you are done with any test you want to do and then we can merge and I can rebase my other ones on top of this :).

metaflow/metadata/heartbeat.py
savingoyal: Can you add some color on why `METAFLOW_` needs to be stripped out for what you intend to achieve?

metaflow/metaflow_config.py
```diff
@@ -7,363 +6,291 @@
...
-DEFAULT_ENVIRONMENT = from_conf("METAFLOW_DEFAULT_ENVIRONMENT", "local")
-DEFAULT_EVENT_LOGGER = from_conf("METAFLOW_DEFAULT_EVENT_LOGGER", "nullSidecarLogger")
-DEFAULT_METADATA = from_conf("METAFLOW_DEFAULT_METADATA", "local")
-DEFAULT_MONITOR = from_conf("METAFLOW_DEFAULT_MONITOR", "nullSidecarMonitor")
-DEFAULT_PACKAGE_SUFFIXES = from_conf("METAFLOW_DEFAULT_PACKAGE_SUFFIXES", ".py,.R,.RDS")
-DEFAULT_AWS_CLIENT_PROVIDER = from_conf("METAFLOW_DEFAULT_AWS_CLIENT_PROVIDER", "boto3")
+
+# Propagate defaults to be able to read what the user sets in configuration files/envs;
+# this will *not* propagate things set on the command line
+DEFAULT_DATASTORE = from_conf("DEFAULT_DATASTORE", "local")
```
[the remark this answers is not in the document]
romain-intel: like `namedtuple` and I didn't want to have METAFLOW_ all over the code. We can add a note if you want to say that the environment variables are all METAFLOW_. Similar to how click does it so shouldn't be too foreign.

metaflow/plugins/airflow/airflow.py
```diff
@@ -11,17 +11,17 @@
 from metaflow.decorators import flow_decorators
 from metaflow.exception import MetaflowException
 from metaflow.metaflow_config import (
-    BATCH_METADATA_SERVICE_HEADERS,
-    BATCH_METADATA_SERVICE_URL,
-    DATASTORE_CARD_S3ROOT,
+    SERVICE_HEADERS,
+    SERVICE_URL,
```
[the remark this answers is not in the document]
romain-intel: ah, forgot the sfn, argo and kube. updated in batch. i'll change.

metaflow/metaflow_config_funcs.py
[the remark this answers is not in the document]
romain-intel: this is removed. they all propagage by default but you don't have to use the function.

metaflow/metaflow_config_funcs.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+# Initialize defaults required to setup environment variables.
+METAFLOW_CONFIG = init_config()
+
+_all_configs = {}
+
+
+def config_values(include=0):
+    # By default, we just return non-null values that are marked as "propagate" and that
```
savingoyal, line 42: NIT - the comment needs to be updated

metaflow/metaflow_config.py
```diff
@@ -7,363 +6,291 @@
...
 ###
-DEFAULT_DATASTORE = from_conf("METAFLOW_DEFAULT_DATASTORE", "local")
-DEFAULT_ENVIRONMENT = from_conf("METAFLOW_DEFAULT_ENVIRONMENT", "local")
-DEFAULT_EVENT_LOGGER = from_conf("METAFLOW_DEFAULT_EVENT_LOGGER", "nullSidecarLogger")
-DEFAULT_METADATA = from_conf("METAFLOW_DEFAULT_METADATA", "local")
-DEFAULT_MONITOR = from_conf("METAFLOW_DEFAULT_MONITOR", "nullSidecarMonitor")
-DEFAULT_PACKAGE_SUFFIXES = from_conf("METAFLOW_DEFAULT_PACKAGE_SUFFIXES", ".py,.R,.RDS")
-DEFAULT_AWS_CLIENT_PROVIDER = from_conf("METAFLOW_DEFAULT_AWS_CLIENT_PROVIDER", "boto3")
+
+# Propagate defaults to be able to read what the user sets in configuration files/envs;
```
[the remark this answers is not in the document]
savingoyal: I believe this comment in the code no longer holds and can be cleaned out

metaflow/metaflow_config.py
```diff
@@ -7,363 +6,278 @@
...
+# this will *not* propagate things set on the command line
+DEFAULT_DATASTORE = from_conf("DEFAULT_DATASTORE", "local")
```
savingoyal, line 22: Can we add a top level comment to this file that calls out that the actual env vars are `METAFLOW_*`? Alternatively, we can have from_conf("METAFLOW_DEFAULT_ENVIRONMENT") etc. so that it is easier for metaflow users to find out what exact env vars are used with Metaflow.

metaflow/metaflow_config_funcs.py
```diff
@@ -0,0 +1,120 @@
...
+                raise ValueError(
+                    "Expected a %s for %s, got: %s" % (type(default), env_name, value)
+                )
+        else:
+            raise RuntimeError(
+                "Default of type %s for %s is not supported" % (type(default), env_name)
+            )
+    _all_configs[env_name] = ConfigValue(
+        value=value,
+        serializer=str,
```
savingoyal, line 103: are we sure that the serializer has to be `str` always? Currently, in the tip of master - there are a few config values that are cast to int/bool.

savingoyal (changes requested):

metaflow/metaflow_config.py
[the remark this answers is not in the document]
savingoyal: there is also another facet to this - today, it's easy for me to grep where all an env var is being used (batch, sfn, k8s, config etc.) - now using grep will become slightly tenuous (only for config though).

metaflow/metaflow_config.py
```diff
@@ -7,363 +6,278 @@
...
 # Datastore local cache
 ###
+# Cache is only local on the host machine so do not propagate values
+
 # Path to the client cache
-CLIENT_CACHE_PATH = from_conf("METAFLOW_CLIENT_CACHE_PATH", "/tmp/metaflow_client")
+CLIENT_CACHE_PATH = from_conf("CLIENT_CACHE_PATH", "/tmp/metaflow_client")
 # Maximum size (in bytes) of the cache
-CLIENT_CACHE_MAX_SIZE = int(from_conf("METAFLOW_CLIENT_CACHE_MAX_SIZE", 10000))
+CLIENT_CACHE_MAX_SIZE = from_conf("CLIENT_CACHE_MAX_SIZE", 10000)
```
savingoyal, line 51: This value is no longer an int. rather than changing the names and types everywhere, we could have instead just emitted the env vars into a dict as a side effect from the `from_conf` method. That would have truly minimized the overall effort this PR has needed.

romain-intel: Why is it no longer an int? It most definitely is still an int. The type of the value is now determined based on the type of the default. So in this case, 10000 is an int so the overall result will be an int. The env_vars are emitted in a dict of sort that is accessible using the `config_values` function. I can change it but at least for now `config_values` emits things as strings but the from_conf returns the proper type (serializing and deserializing happen automatically).

metaflow/metaflow_config_funcs.py
romain-intel: see below comment but yes, we deserialize from str to the type (int/bool) and serialize back to str when emitting back out as env var. It's either str or json.loads/dumps

metaflow/metaflow_config.py
romain-intel: I will add a comment.

metaflow/metaflow_config.py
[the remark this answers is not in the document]
romain-intel: yep. will update.

metaflow/metaflow_config.py
savingoyal: perfect!

metaflow/metaflow_config.py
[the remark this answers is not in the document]
romain-intel: I added a comment. THe grep is now just hte name of the var so should be not too too hard.

savingoyal (approved):

Issue #1172, opened 2022-10-24
Verify code SHA hash before running it


ruial:
Hi all,

Metaflow stores the ```METAFLOW_CODE_SHA``` environment variable on Step Functions but does not use it to check the downloaded code artifact from S3. This can lead to undesired remote code execution where multiple teams share a single Metaflow instance/bucket and have different compute clusters (which may have different IAM permissions per team). Although access to the bucket is secured via IAM, it would be safer for the init script to verify the file hash after downloading it.

To reproduce this issue, change the hash to any value and the flow runs without any errors. The Python docker image contains the ```sha1sum``` tool. To mitigate this, an additional check script could be executed in https://github.com/Netflix/metaflow/blob/master/metaflow/metaflow_environment.py

Best regards.

ruial:
Closing this as sharing S3 buckets is not really required in a multi-tenant setup anymore with the latest versions, as each tenant can use his own S3 bucket and security is managed by the S3 policies.

PR #1184, opened 2022-11-03
Verify code hash

ruial:
This PR addresses issue https://github.com/Netflix/metaflow/issues/1172

By verifying the code SHA after it is downloaded, we have additional guarantees that the pipeline code was not tampered with. I've tested this simple implementation with success, both sha1sum and cut are available in the default Python docker image.

### DIFF
--- a/metaflow/metaflow_environment.py
+++ b/metaflow/metaflow_environment.py
@@ -144,6 +144,9 @@ class MetaflowEnvironment(object):
             "mflog 'Failed to download code package from %s "
             "after 6 tries. Exiting...' && exit 1; "
             "fi" % code_package_url,
+            "if [ `sha1sum job.tar | cut -d ' ' -f1` != $METAFLOW_CODE_SHA ]; then "
+            "mflog 'Hash does not match for code package from %s. Exiting...' && exit 1; "
+            "fi" % code_package_url,
             "TAR_OPTIONS='--warning=no-timestamp' tar xf job.tar",
             "mflog 'Task is starting.'",
         ]


savingoyal:
@ruial Can you provide a bit more context on the tamper-proof guarantees that you are seeking from Metaflow?

ruial:
Hey @savingoyal I expect Metaflow to not execute the code from S3 if it doesn't match the hash defined in the step functions, as mentioned in the issue. This situation can happen when Metaflow is deployed in a shared AWS environment, where multiple teams use the same UI and S3 bucket but have the Batch clusters and Step functions in different accounts. If I wanted, I could replace other teams code and it will be executed in their compute without this fix.

Another great feature for Metaflow would be to allow people to optionally specify a different S3 bucket per flow/experiment, like MLflow does, providing additional security and control over object lifecycle for teams. I would be happy to talk more about this topic and other possible improvements via Slack if you'd like.

romain-intel:
As a quick note, you can specify the bucket you want to use by using the `METAFLOW_DATASTORE_SYSROOT_S3` environment variable.

ruial:
Thanks for the note @romain-intel. I assumed the Metaflow UI was only able to read from a single S3 bucket because a single bucket was set for the UI on the Terraform module https://github.com/outerbounds/terraform-aws-metaflow/blob/master/modules/ui/locals.tf#L26. Does the UI support multiple buckets? I don't remember if I tested it but I will soon.

I still think the code tarball hash should still be checked, as it is already there anyway.

romain-intel:
The UI should work. I remember we ran into this issue a while back and fixed it but can't find the relevant PR. Let me know if it doesn't work.

ruial:
I can confirm that the latest UI works with different S3 buckets, the ```METAFLOW_``` variables are not needed on the backend service. 

This is great, as I no longer have a reason to share an S3 bucket with other accounts/teams. We have a central Metaflow UI, however each team has their own compute/s3/scheduler. By sharing read access to their S3 buckets, other teams can allow us to see their logs/cards from a single UI.

Thanks you guys. Since S3 write access is not shared anymore, there is no risk of someone changing the code tarballs. It may still be considered a good practice to check the hash, but I no longer see this as a security issue for my setup, feel free to close or merge this.

savingoyal:
@ruial That's good to hear. One issue that needs to be fixed for this PR is checking whether sha1sum is indeed available in the image -it is available for the default python image but might not be available in the image the user might be using as an override. This can be a potentially breaking change. Let's think through the universe of use cases we would want to support for code verification and we can resuscitate this PR after we have a good body of user stories.

PR #1186, opened 2022-11-09
Deal with transient errors (like SlowDowns) more effectively for S3

romain-intel:
In the previous incantantion, SlowDown errors were treated as a regular error and everything was retried (making things worse). With this change, we continue retrying the operations that were unsuccessful (and only those).

Also use a better internal boto retry policy (unless specified by the user).

Also fix a bug in the current code when puts with retries could return the incorrect files uploaded.

Modified the tests to inject failures to be able to test the functionality. Tests now have a failure injection rate of 0, 10, 50 or 90 percent.


metaflow/datatools/s3op.py
```diff
@@ -97,14 +109,41 @@ def format_triplet(prefix, url="", local=""):
 def normalize_client_error(err):
     error_code = err.response["Error"]["Code"]
     try:
-        return int(error_code)
+        err_code = int(error_code)
+        # We "normalize" retriable server errors to 503. These are also considered
+        # transient by boto3 (see:
+        # https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html)
+        if err_code in (500, 502, 504):
+            err_code = 503
```
savingoyal, line 117: rather than swallowing the original error code, can we just retry for 500, 502, 5043 and 504 (in L:184)

metaflow/datatools/s3.py
```diff
@@ -1348,32 +1368,228 @@ def _s3op_with_retries(self, mode, **options):
...
+                # This is the first time around; we make the list big enough. Typically,
+                # there is nothing in out_lines but in some cases (a retry after a
+                # partial result), there may be stuff in it
+                out_lines.extend([None] * (len(ok_lines) - len(out_lines)))
+            for l in ok_lines:
+                idx, rest = l.split(b" ", maxsplit=1)
+                if rest.decode(encoding="utf-8") != TRANSIENT_RETRY_LINE_CONTENT:
+                    out_lines[int(idx.decode(encoding="utf-8"))] = rest
+
+        def _internal(last_ok_count, pending_retries, out_lines, inject_failures):
```
savingoyal, line 1412: Can you make this function name a bit more descriptive and some context to aid in readability?

metaflow/datatools/s3op.py
romain-intel: Are you suggesting I have multiple "values" for transient errors? As of now, in line 184, we return just a ERROR_TRANSIENT_RETRY so the exact code is anyways swallowed there. I could have TRANSIENT_RETRY_500, TRANSIENT_RETRY_502, TRANSIENT_RETRY_504 and TRANSIENT_RETRY_503 if you want but it seems like since the behavior is the same going back out, it won't matter much.

metaflow/datatools/s3.py
romain-intel: sure, I can do that. The TL;DR was just that I wanted to factor out this from the outer while loop below since it got to be a bit large. I can put it all inline too if that helps.

metaflow/datatools/s3.py
```diff
@@ -1348,32 +1368,245 @@ def _s3op_with_retries(self, mode, **options):
...
+        def _get_inject_failures():
+            # list mode does not do retries on transient failures (there is no
+            # SlowDown handling) so we never inject a failure rate
+            if mode == "list":
+                return 0
+            # Otherwise, we cap the failure rate at 90%
+            return min(90, self._s3_inject_failures)
+
+        retry_count = 0  # Number of retries (excluding transient failures)
+        transient_retry_count = 0  # Number of transient retries (per TL retry)
```
savingoyal, line 1380: Maybe I am missing something, but why do we have a distinction between a normal retry and a transient retry?

metaflow/datatools/s3op.py
```diff
@@ -85,10 +94,19 @@ def __str__(self):
 ERROR_VERIFY_FAILED = 9
 ERROR_LOCAL_FILE_NOT_FOUND = 10
 ERROR_INVALID_RANGE = 11
-
-
-def format_triplet(prefix, url="", local=""):
-    return " ".join(url_quote(x).decode("utf-8") for x in (prefix, url, local))
+ERROR_TRANSIENT_RETRY = 12
```
savingoyal, line 97: A better description is simply - `ERROR_SLOW_DOWN` or `ERROR_TRANSIENT`?

metaflow/datatools/s3op.py
savingoyal: No - in L:190 you could simply do 
```
elif error_code in (500, 502, 503, 504):
                to_return = {"error": ERROR_TRANSIENT, "raise_error": err}

```

metaflow/datatools/s3op.py
```diff
@@ -85,10 +94,19 @@ def __str__(self):
...
+
+
+def format_triplet(idx, prefix, url="", local=""):
```
savingoyal, line 100: it's no longer a triplet, eh?

savingoyal (commented):
still reviewing - but added some larger questions

metaflow/datatools/s3op.py
romain-intel: Sure, I can make that change and return 503 for all the text and then catch all these other ones explicitly.

metaflow/datatools/s3.py
```diff
@@ -1277,7 +1288,7 @@ def _read_many_files(self, op, prefixes_and_ranges, **options):
                 )
             )
             inputfile.flush()
-            stdout, stderr = self._s3op_with_retries(
+            stdout_lines, stderr = self._s3op_with_retries(
```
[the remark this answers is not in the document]
romain-intel: It just adds unnecessary time complexity. Given that the lines of stdout can now be generated out of order (due to retries), I have to collect it as an array there. Later in this code, we basically do a splitlines to get each line individually so we would go from array to string to array which was a bit not really required. Same thing for stderr (but opposite: it's collected as string and used as string so making it transition to an array seems unnecessary).

metaflow/datatools/s3.py
```diff
@@ -1348,32 +1368,245 @@ def _s3op_with_retries(self, mode, **options):
         if self._s3_client_params is not None:
             cmdline.extend(("--s3clientparams", json.dumps(self._s3_client_params)))
 
-        for i in range(S3_RETRY_COUNT + 1):
+        def _get_inject_failures():
```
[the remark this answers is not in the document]
romain-intel: I can make it `_inject_failure_rate`. I want the fact that it is injected to be clear.

metaflow/datatools/s3.py
```diff
@@ -1348,32 +1368,245 @@ def _s3op_with_retries(self, mode, **options):
...
+            # list mode does not do retries on transient failures (there is no
```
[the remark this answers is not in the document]
romain-intel: list does not deal with transient failures. It was too complicated to add iirc and I don't think it is worth it. The rationale for capping at 90 (and not 100) is that you could otherwise fail to make progress and have more flaky tests (there is a cap on the number of full retries and no progress counts as making a full retry).

savingoyal (approved):

metaflow/metaflow_config.py
```diff
@@ -127,6 +127,8 @@ def _validate_choice(name, value):
 # so setting it to 0 means each operation will be tried once.
 S3_RETRY_COUNT = int(from_conf("METAFLOW_S3_RETRY_COUNT", 7))
 
+S3_TRANSIENT_RETRY_COUNT = int(from_conf("METAFLOW_S3_TRANSIENT_RETRY_COUNT", 20))
```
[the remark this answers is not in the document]
romain-intel: you typically want it higher since you are still making progress. The typical retry count of 3 would be a little low for this.

metaflow/datatools/s3.py
romain-intel: added a very detailed comment.

PR #1192, opened 2022-11-12
Add test that orig module is accessible

tfurmston:
This PR adds a test that checks that the `_orig` module is accessible when a module is overwritten through extensions framework.


romain-intel:
LGTM

PR #1199, opened 2022-11-21
Update black to latest version

savingoyal:


metaflow/plugins/aws/batch/batch_cli.py
```diff
@@ -201,7 +201,7 @@ def echo(msg, stream="stderr", batch_id=None):
     if num_parallel and num_parallel > 1:
         # For multinode, we need to add a placeholder that can be mutated by the caller
         step_args += " [multinode-args]"
-    step_cli = u"{entrypoint} {top_args} step {step} {step_args}".format(
```
romain-intel, line 204: Of note, the removal of the "u" prefix. It should not matter since we are now 3.5+ but just pointing it out as this is technically a semantic change for some versions of Python.

romain-intel (approved):
LGTM

PR #1201, opened 2022-11-28
Pin GH tests to Ubuntu 20.04

savingoyal:
Ubuntu latest which maps to 22.04 doesn't support Py 3.5/3.6 on GH actions


romain-intel (approved):

savingoyal:
should be rolled back once https://github.com/actions/setup-python/issues/544 is addressed

PR #1202, opened 2022-11-28
Set gpu resources correctly "--with kubernetes"

shrinandj:
GPU resources were not getting propagated when used "--with kubernetes". Now, it does.

Testing Done:

- Reproduced the problem creating a step with a decorator @resources(gpu=1) and running it "--with kubernetes". The pods were scheduled without gpus.
- Verified that after the fix, the pods were scheduled with gpus (request and limit).

### DIFF
--- a/metaflow/plugins/kubernetes/kubernetes_decorator.py
+++ b/metaflow/plugins/kubernetes/kubernetes_decorator.py
@@ -145,6 +145,12 @@ class KubernetesDecorator(StepDecorator):
                     "@kubernetes does not support parallel execution currently."
                 )
 
+            # If GPU count is specified, explicitly set it in self.attributes.
+            for k, v in deco.attributes.items():
+                if k == "gpu" and v != None:
+                    self.attributes['gpu'] = v
+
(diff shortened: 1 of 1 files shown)


metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -145,6 +145,12 @@ def step_init(self, flow, graph, step, decos, environment, flow_datastore, logge
                     "@kubernetes does not support parallel execution currently."
                 )
 
+            # If GPU count is specified, explicitly set it in self.attributes.
+            for k, v in deco.attributes.items():
+                if k == "gpu" and v != None:
```
savingoyal, line 150: we can move this and the next line underneath L:166

savingoyal: we can also do away with the code comment in L:165

shrinandj: Done

PR #1206, opened 2022-12-04
Fix/move data files

romain-intel:


romain-intel:
Tests are ok internally. Merging.

PR #1207, opened 2022-12-05
Add support for Kubernetes tolerations

odracci:
This PR adds support for `tolerations` in the Kubernetes and Argo plugins.

Testing Done:

Using the script `flow.py`
```python
DEFAULT_RESOURCES = {"cpu": "1", "gpu": "0", "memory": "1"}

from metaflow import (
    FlowSpec,
    step,
    timeout,
    kubernetes,
    resources,
)


class TestTolerationsFlow(FlowSpec):
    @resources(**DEFAULT_RESOURCES)
    @step
    def start(self):
        """
        This is the 'start' step. All flows must have a step named 'start' that
        is the first step in the flow.

        """
        self.next(self.end)

    @resources(**DEFAULT_RESOURCES)
    @step
    def end(self):
        """
        This is the 'end' step. All flows must have an 'end' step, which is the
        last step in the flow.

        """
        print("TestTolerationsFlow is all done.")


if __name__ == "__main__":
    TestTolerationsFlow()
```
Tested the following commands:
- `python3 flow.py run --with kubernetes`
- `python3 flow.py run --with kubernetes:node_selector=app=cpu`
- `python3 flow.py run --with kubernetes:node_selector=app=cpu,tolerations='[{"key":"app","value":"cpu","effect":"NoSchedule"}]'`
- `METAFLOW_KUBERNETES_TOLERATIONS='[{"key":"app","value":"cpu","effect":"NoSchedule"}]' python3 flow.py run --with kubernetes`
- `METAFLOW_KUBERNETES_TOLERATIONS='[{"key":"app","value":"cpu","effect":"NoSchedule"}]' METAFLOW_KUBERNETES_NODE_SELECTOR='app=cpu' python3 flow.py run --with kubernetes`
- `METAFLOW_KUBERNETES_TOLERATIONS='[{"key":"app","value":"cpu","effect":"NoSchedule"}]' METAFLOW_KUBERNETES_NODE_SELECTOR='app=cpu' python3 flow.py argo-workflows create`
- `python3 flow.py argo-workflows trigger`

With the script `flow_decorator.py`
```python
DEFAULT_RESOURCES = {"cpu": "1", "gpu": "0", "memory": "1"}
NODE_GROUP = "cpu"
KUBERNETES_NODE_SELECTOR = f"app={NODE_GROUP}"
KUBERNETES_TOLERATIONS = [{"key": "app", "effect": "NoSchedule", "value": NODE_GROUP}]

from metaflow import (
    FlowSpec,
    step,
    timeout,
    kubernetes,
    resources,
)


class TestTolerationsFlow(FlowSpec):
    @resources(**DEFAULT_RESOURCES)
    @kubernetes(node_selector=KUBERNETES_NODE_SELECTOR, tolerations=KUBERNETES_TOLERATIONS)
    @step
    def start(self):
        """
        This is the 'start' step. All flows must have a step named 'start' that
        is the first step in the flow.

        """
        self.next(self.end)

    @resources(**DEFAULT_RESOURCES)
    @kubernetes(node_selector=KUBERNETES_NODE_SELECTOR, tolerations=KUBERNETES_TOLERATIONS)
    @step
    def end(self):
        """
        This is the 'end' step. All flows must have an 'end' step, which is the
        last step in the flow.

        """
        print("TestTolerationsFlow is all done.")


if __name__ == "__main__":
    TestTolerationsFlow()
```

- `python3 flow_decorator.py run`
- `python3 flow_decorator.py argo-workflows create`
- `python3 flow_decorator.py argo-workflows trigger`

In all the executions, verified that the pods have the expected `Tolerations` and `nodeSelectors`

### DIFF
--- a/metaflow/metaflow_config.py
+++ b/metaflow/metaflow_config.py
@@ -242,7 +242,7 @@ KUBERNETES_NAMESPACE = from_conf("KUBERNETES_NAMESPACE", "default")
 KUBERNETES_SERVICE_ACCOUNT = from_conf("KUBERNETES_SERVICE_ACCOUNT")
 # Default node selectors to use by K8S jobs created by Metaflow - foo=bar,baz=bab
 KUBERNETES_NODE_SELECTOR = from_conf("KUBERNETES_NODE_SELECTOR", "")
-KUBERNETES_TOLERATIONS = from_conf("METAFLOW_KUBERNETES_TOLERATIONS")
+KUBERNETES_TOLERATIONS = from_conf("METAFLOW_KUBERNETES_TOLERATIONS", "")
 KUBERNETES_SECRETS = from_conf("KUBERNETES_SECRETS", "")
 # Default GPU vendor to use by K8S jobs created by Metaflow (supports nvidia, amd)
 KUBERNETES_GPU_VENDOR = from_conf("KUBERNETES_GPU_VENDOR", "nvidia")
--- a/metaflow/metaflow_config.py
+++ b/metaflow/metaflow_config.py
@@ -242,7 +242,7 @@ KUBERNETES_NAMESPACE = from_conf("KUBERNETES_NAMESPACE", "default")
 KUBERNETES_SERVICE_ACCOUNT = from_conf("KUBERNETES_SERVICE_ACCOUNT")
 # Default node selectors to use by K8S jobs created by Metaflow - foo=bar,baz=bab
 KUBERNETES_NODE_SELECTOR = from_conf("KUBERNETES_NODE_SELECTOR", "")
-KUBERNETES_TOLERATIONS = from_conf("METAFLOW_KUBERNETES_TOLERATIONS", "")
+KUBERNETES_TOLERATIONS = from_conf("KUBERNETES_TOLERATIONS", "")
 KUBERNETES_SECRETS = from_conf("KUBERNETES_SECRETS", "")
 # Default GPU vendor to use by K8S jobs created by Metaflow (supports nvidia, amd)
 KUBERNETES_GPU_VENDOR = from_conf("KUBERNETES_GPU_VENDOR", "nvidia")
--- a/metaflow/metaflow_config.py
+++ b/metaflow/metaflow_config.py
@@ -242,6 +242,7 @@ KUBERNETES_NAMESPACE = from_conf("KUBERNETES_NAMESPACE", "default")
 KUBERNETES_SERVICE_ACCOUNT = from_conf("KUBERNETES_SERVICE_ACCOUNT")
 # Default node selectors to use by K8S jobs created by Metaflow - foo=bar,baz=bab
 KUBERNETES_NODE_SELECTOR = from_conf("KUBERNETES_NODE_SELECTOR", "")
+KUBERNETES_TOLERATIONS = from_conf("METAFLOW_KUBERNETES_TOLERATIONS")
 KUBERNETES_SECRETS = from_conf("KUBERNETES_SECRETS", "")
 # Default GPU vendor to use by K8S jobs created by Metaflow (supports nvidia, amd)
 KUBERNETES_GPU_VENDOR = from_conf("KUBERNETES_GPU_VENDOR", "nvidia")
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -140,7 +140,9 @@ class ArgoWorkflows(object):
         return name.replace("_", "-")
 
     @classmethod
-    def trigger(cls, name, parameters={}):
+    def trigger(cls, name, parameters=None):
+        if parameters is None:
+            parameters = {}
         try:
             workflow_template = ArgoClient(
                 namespace=KUBERNETES_NAMESPACE
@@ -378,7 +380,7 @@ class ArgoWorkflows(object):
 
     # Visit every node and yield the uber DAGTemplate(s).
     def _dag_templates(self):
-        def _visit(node, exit_node=None, templates=[], dag_tasks=[]):
+        def _visit(node, exit_node=None, templates=None, dag_tasks=None):
             # Every for-each node results in a separate subDAG and an equivalent
             # DAGTemplate rooted at the child of the for-each node. Each DAGTemplate
             # has a unique name - the top-level DAGTemplate is named as the name of
@@ -386,6 +388,10 @@ class ArgoWorkflows(object):
             # of the for-each node.
 
             # Emit if we have reached the end of the sub workflow
+            if dag_tasks is None:
+                dag_tasks = []
+            if templates is None:
+                templates = []
             if exit_node is not None and exit_node is node.name:
                 return templates, dag_tasks
 
@@ -839,6 +845,22 @@ class ArgoWorkflows(object):
             env_without_none_values = {k: v for k, v in env.items() if v is not None}
             del env
 
+            node_selector = []
+            if resources.get("node_selector"):
+                try:
+                    node_selector = {
+                        str(k.split("=", 1)[0]): str(k.split("=", 1)[1])
+                        for k in resources.get("node_selector").split(",")
+                    }
+                except IndexError:
+                    raise ArgoWorkflowsException(f"Unable to parse node_selector {node_selector}")
+            elif KUBERNETES_NODE_SELECTOR:
+                node_selector = KUBERNETES_NODE_SELECTOR.split(",")
+
+            tolerations = None
+            if resources.get("tolerations"):
+                tolerations = json.loads(resources.get("tolerations"))
+
             # Create a ContainerTemplate for this node. Ideally, we would have
             # liked to inline this ContainerTemplate and avoid scanning the workflow
             # twice, but due to issues with variable substitution, we will have to
@@ -871,19 +893,8 @@ class ArgoWorkflows(object):
                 # Set emptyDir volume for state management
                 .empty_dir_volume("out")
                 # Set node selectors
-                .node_selectors(
-                    {
-                        str(k.split("=", 1)[0]): str(k.split("=", 1)[1])
-                        for k in (
-                            resources.get("node_selector")
-                            or (
-                                KUBERNETES_NODE_SELECTOR.split(",")
-                                if KUBERNETES_NODE_SELECTOR
-                                else []
-                            )
-                        )
-                    }
-                )
+                .node_selectors(node_selector)
+                .tolerations(tolerations)
                 # Set container
                 .container(
                     # TODO: Unify the logic with kubernetes.py
@@ -1234,8 +1245,12 @@ class Template(object):
 
     def node_selectors(self, node_selectors):
         if "nodeSelector" not in self.payload:
-            self.payload["labels"] = {}
+            self.payload["nodeSelector"] = {}
         self.payload["nodeSelector"].update(node_selectors)
+        return self
+
+    def tolerations(self, tolerations):
+        self.payload["tolerations"] = tolerations
         return self
 
     def to_json(self):
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -845,24 +845,6 @@ class ArgoWorkflows(object):
             env_without_none_values = {k: v for k, v in env.items() if v is not None}
             del env
 
-            node_selector = []
-            if resources.get("node_selector"):
-                try:
-                    node_selector = {
-                        str(k.split("=", 1)[0]): str(k.split("=", 1)[1])
-                        for k in resources.get("node_selector").split(",")
-                    }
-                except IndexError:
-                    raise ArgoWorkflowsException(
-                        f"Unable to parse node_selector {node_selector}"
-                    )
-            elif KUBERNETES_NODE_SELECTOR:
-                node_selector = KUBERNETES_NODE_SELECTOR.split(",")
-
-            tolerations = None
-            if resources.get("tolerations"):
-                tolerations = json.loads(resources.get("tolerations"))
-
             # Create a ContainerTemplate for this node. Ideally, we would have
             # liked to inline this ContainerTemplate and avoid scanning the workflow
             # twice, but due to issues with variable substitution, we will have to
@@ -896,8 +878,8 @@ class ArgoWorkflows(object):
                 # Set emptyDir volume for state management
                 .empty_dir_volume("out")
                 # Set node selectors
-                .node_selectors(node_selector)
-                .tolerations(tolerations)
+                .node_selectors(resources.get("node_selector"))
+                .tolerations(resources.get("tolerations"))
                 # Set container
                 .container(
                     # TODO: Unify the logic with kubernetes.py
@@ -1249,7 +1231,8 @@ class Template(object):
     def node_selectors(self, node_selectors):
         if "nodeSelector" not in self.payload:
             self.payload["nodeSelector"] = {}
-        self.payload["nodeSelector"].update(node_selectors)
+        if node_selectors:
+            self.payload["nodeSelector"].update(node_selectors)
         return self
 
     def tolerations(self, tolerations):
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -853,7 +853,9 @@ class ArgoWorkflows(object):
                         for k in resources.get("node_selector").split(",")
                     }
                 except IndexError:
-                    raise ArgoWorkflowsException(f"Unable to parse node_selector {node_selector}")
+                    raise ArgoWorkflowsException(
+                        f"Unable to parse node_selector {node_selector}"
+                    )
             elif KUBERNETES_NODE_SELECTOR:
                 node_selector = KUBERNETES_NODE_SELECTOR.split(",")
 
@@ -881,7 +883,8 @@ class ArgoWorkflows(object):
                 .retry_strategy(
                     times=total_retries,
                     minutes_between_retries=minutes_between_retries,
-                ).metadata(
+                )
+                .metadata(
                     ObjectMeta().annotation("metaflow/step_name", node.name)
                     # Unfortunately, we can't set the task_id since it is generated
                     # inside the pod. However, it can be inferred from the annotation
--- a/metaflow/plugins/kubernetes/kubernetes_cli.py
+++ b/metaflow/plugins/kubernetes/kubernetes_cli.py
@@ -11,6 +11,7 @@ from metaflow.metaflow_config import DATASTORE_LOCAL_DIR
 from metaflow.mflog import TASK_LOG_SOURCE
 
 from .kubernetes import Kubernetes, KubernetesKilledException
+from .kubernetes_decorator import KubernetesDecorator
 
 
 @click.group()
@@ -172,6 +173,9 @@ def step(
     )
     stdout_location = ds.get_log_location(TASK_LOG_SOURCE, "stdout")
     stderr_location = ds.get_log_location(TASK_LOG_SOURCE, "stderr")
+
+    # Here node_selector is a tuple of strings, convert it to a dictionary
+    node_selector = KubernetesDecorator.parse_node_selector(node_selector)
 
     def _sync_metadata():
         if ctx.obj.metadata.TYPE == "local":
--- a/metaflow/plugins/kubernetes/kubernetes_cli.py
+++ b/metaflow/plugins/kubernetes/kubernetes_cli.py
@@ -3,7 +3,7 @@ import sys
 import time
 import traceback
 
-from metaflow import util
+from metaflow import util, JSONTypeClass
 from metaflow._vendor import click
 from metaflow.exception import METAFLOW_EXIT_DISALLOW_RETRY, CommandException
 from metaflow.metadata.util import sync_local_metadata_from_datastore
@@ -84,6 +84,12 @@ def kubernetes():
     default=5 * 24 * 60 * 60,  # Default is set to 5 days
     help="Run time limit in seconds for Kubernetes pod.",
 )
+@click.option(
+    "--tolerations",
+    default=None,
+    type=JSONTypeClass(),
+    multiple=False,
+)
 @click.pass_context
 def step(
     ctx,
@@ -102,6 +108,7 @@ def step(
     gpu=None,
     gpu_vendor=None,
     run_time_limit=None,
+    tolerations=None,
     **kwargs
 ):
     def echo(msg, stream="stderr", job_id=None):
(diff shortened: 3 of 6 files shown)


metaflow/plugins/argo/argo_workflows.py
```diff
@@ -1234,10 +1245,14 @@ def empty_dir_volume(self, name):
 
     def node_selectors(self, node_selectors):
         if "nodeSelector" not in self.payload:
-            self.payload["labels"] = {}
+            self.payload["nodeSelector"] = {}
```
odracci, line 1248: I think `labels` is a typo

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -140,7 +140,9 @@ def _sanitize(name):
         return name.replace("_", "-")
 
     @classmethod
-    def trigger(cls, name, parameters={}):
+    def trigger(cls, name, parameters=None):
+        if parameters is None:
```
savingoyal, line 144: Any specific reason for this change here and in L:383?

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -839,6 +845,22 @@ def _container_templates(self):
             env_without_none_values = {k: v for k, v in env.items() if v is not None}
             del env
 
+            node_selector = []
+            if resources.get("node_selector"):
```
savingoyal, line 849: Any reason for moving this code segment out from L:896?

shrinandj:
Looks great overall! Can you add some details about the testing you did with this change? It would make reviewing a little easier (see the `Testing Done` section in [this](https://github.com/Netflix/metaflow/pull/1202) PR as an example)

metaflow/plugins/argo/argo_workflows.py
savingoyal: good catch!

metaflow/plugins/kubernetes/kubernetes_job.py
```diff
@@ -71,6 +71,13 @@ def create(self):
         # Note: This implementation ensures that there is only one unique Pod
         # (unique UID) per Metaflow task attempt.
         client = self._client.get()
+        tolerations = []
+        for toleration in self._kwargs.get("tolerations") or []:
```
savingoyal, line 75: to maintain the same coding style in this file, would you mind in-lining this in L:198 instead?

savingoyal (changes requested):
some comments re: code organization. @shrinandj is doing a full review.

metaflow/plugins/kubernetes/kubernetes_cli.py
```diff
@@ -84,6 +84,12 @@ def kubernetes():
     default=5 * 24 * 60 * 60,  # Default is set to 5 days
     help="Run time limit in seconds for Kubernetes pod.",
 )
+@click.option(
+    "--tolerations",
+    default=None,
+    type=JSONTypeClass(),
+    multiple=False,
```
shrinandj, line 91: Shouldn't this be `true`? 

`secrets` and `node_selector` also have `multiple=true`.

metaflow/metaflow_config.py
```diff
@@ -242,6 +242,7 @@
 KUBERNETES_SERVICE_ACCOUNT = from_conf("KUBERNETES_SERVICE_ACCOUNT")
 # Default node selectors to use by K8S jobs created by Metaflow - foo=bar,baz=bab
 KUBERNETES_NODE_SELECTOR = from_conf("KUBERNETES_NODE_SELECTOR", "")
+KUBERNETES_TOLERATIONS = from_conf("METAFLOW_KUBERNETES_TOLERATIONS")
```
shrinandj, line 245: This needs a default value of `""` similar to `KUBERNETES_NODE_SELECTOR` and `KUBERNETES_SECRETS`.

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -79,6 +78,7 @@ class KubernetesDecorator(StepDecorator):
         "namespace": None,
         "gpu": None,  # value of 0 implies that the scheduled node should not have GPUs
         "gpu_vendor": None,
+        "tolerations": None,  # e.g., [{"key": "arch", "operator": "Equal", "value": "amd"}]
```
shrinandj, line 81: Two questions:

- The map could also have a key called `effect`. So should the comment be changed to something like `# e.g., [{"key": "arch", "operator": "Equal", "value": "amd", "effect":"NoSchedule"}]`. e.g. [this](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/)
- There's an additional comment above for allowing multiple tolerations. In that case, would users express the tolerations as something like:
`[{"key": "arch", "operator": "Equal", "value": "amd"}],[{"key": "arch", "operator": "Equal", "value": "arm"}]`

metaflow/plugins/argo/argo_workflows.py
odracci: Just removing mutable default arguments. [Doc](https://docs.python-guide.org/writing/gotchas/)

metaflow/plugins/argo/argo_workflows.py
odracci: The problem is how Metaflow parses `node_selector`.

When set with the decorator (`@kubernetes(node_selector=["key=value", "foo=bar"])`), `resources.get("node_selector")` is an array of strings.
When set using the environment variable `METAFLOW_KUBERNETES_NODE_SELECTOR`, both `resources.get("node_selector")` and `KUBERNETES_NODE_SELECTOR` are strings like `key=value,foo=bar`
The previous code raises an exception:
```
IndexError: list index out of range
```

In addition, the code now raises an exception in case of malformed input. It took me a while to figure out what was the error

metaflow/plugins/kubernetes/kubernetes_job.py
odracci: I extracted the code to handle errors if the input is invalid. I can't do it in-line.
I can extract it in a function, WDYT?

metaflow/plugins/kubernetes/kubernetes_cli.py
odracci: I decided to use `JSONType` because it is easy to pass an object.
Or do you mean to use something like `--tolerations='{"key": "arch", "operator": "Equal", "value": "amd"}' --tolerations='{"key": "foo", "operator": "Equal", "value": "bar"}'`?

shrinandj: So, I think if there are multiple tolerations, they'd look something like:
`[{"key": "arch", "operator": "Equal", "value": "amd"}, {"key": "foo", "operator": "Equal", "value": "bar"}]`

If that's correct, then your approach is totally valid.

odracci: yes, correct. The input is an array of objects

metaflow/plugins/kubernetes/kubernetes_job.py
```diff
@@ -71,6 +69,21 @@ def create(self):
...
+            except TypeError:
+                raise KubernetesJobException("Toleration definition contains invalid keys: %s" % toleration.keys())
+
+        try:
+            node_selector = {
+                str(k.split("=", 1)[0]): str(k.split("=", 1)[1])
+                for k in self._kwargs.get("node_selector")
+            }
+        except IndexError:
+            raise KubernetesJobException(f"Unable to parse node_selector {self._kwargs.get('node_selector')}")
```
odracci, line 85: @savingoyal I moved `node_selector` as well because now `KUBERNETES_NODE_SELECTOR` is set in `kubernetes_job.py:97`

metaflow/plugins/argo/argo_workflows.py
savingoyal: @odracci we should ensure that the node_selector is properly formatted in `kubernetes_decorator.py` instead of this file. You would notice that this file doesn't do any validation of user-provided inputs for the decorators and assumes that that has already happened upstreams.

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -79,6 +78,8 @@ class KubernetesDecorator(StepDecorator):
         "namespace": None,
         "gpu": None,  # value of 0 implies that the scheduled node should not have GPUs
         "gpu_vendor": None,
+        "tolerations": None,  # e.g., [{"key": "arch", "operator": "Equal", "value": "amd"}],
```
savingoyal, line 81: @odracci can you verify how tolerations would be passed through CLI - `python flow.py run --with kubernetes:tolerations=blah..`

metaflow/plugins/kubernetes/kubernetes_job.py
savingoyal: we can move the input verification inside the `@kubernetes` decorator module - currently all validations (we don't have many) reside there. Also, the `@kubernetes` decorator hooks are guaranteed to execute for both `@kubernetes` and `argo-workflows create` while `kubernetes_job` only gets invoked for `python flow.py run` - so this toleration check would not happen for argo workflows.

metaflow/plugins/argo/argo_workflows.py
odracci: @savingoyal it is not possible because `resources` is extracted on L:730 in this file. It requires a major refactor

savingoyal: @odracci we should be able to set the `node_selector` attribute for the kubernetes decorator properly in L:97 of `kubernetes_decorator.py` and any necessary validation there or in `step_init`

metaflow/plugins/kubernetes/kubernetes_job.py
odracci: in `KubernetesDecorator`, `KUBERNETES_TOLERATIONS` is a string and not a parsed JSON.
I can try to move the logic in `Kubernetes.create_job()`, WDYT?

odracci: The problem is this piece of code in `KubernetesDecorator`
```python
if not self.attributes["node_selector"] and KUBERNETES_NODE_SELECTOR:
    self.attributes["node_selector"] = KUBERNETES_NODE_SELECTOR
```
`self.attributes["node_selector"]` is a python object set as a parameter to the `@kubernetes` decorator, but `KUBERNETES_NODE_SELECTOR` is a string set using an environment variable.
`KUBERNETES_NODE_SELECTOR` is parsed as a python object in `kubernetes_cli`, which is executed afterward.

One solution would be to parse `KUBERNETES_NODE_SELECTOR` in `KubernetesDecorator` but it would duplicate the logic done by `@click.option`

metaflow/plugins/argo/argo_workflows.py
odracci: same issue described in this [comment](https://github.com/Netflix/metaflow/pull/1207/files#r1040209561)

metaflow/plugins/kubernetes/kubernetes_decorator.py
odracci: it doesn't work. I think there is a bug in `BaseCommand.invoke`.
I ran `python flow.py run --with 'kubernetes:tolerations=[{"key":"app","effect":"NoSchedule","value":"foo"}]'` and it fails. Logs:
```
2022-12-06 15:54:56.596 Workflow starting (run-id 1030):
2022-12-06 15:54:58.335 [1030/start/73637 (pid 25832)] Task is starting.
2022-12-06 15:55:00.021 [1030/start/73637 (pid 25832)] Usage: workflow_simple.py kubernetes step [OPTIONS] STEP_NAME CODE_PACKAGE_SHA
2022-12-06 15:55:00.300 [1030/start/73637 (pid 25832)] CODE_PACKAGE_URL
2022-12-06 15:55:00.300 [1030/start/73637 (pid 25832)] Try 'workflow_simple.py kubernetes step --help' for help.
2022-12-06 15:55:00.301 [1030/start/73637 (pid 25832)]
2022-12-06 15:55:00.301 [1030/start/73637 (pid 25832)] Error: Invalid value for '--tolerations': {'key': 'app', 'effect': 'NoSchedule', 'value': 'foo'} is not a valid JSON object
2022-12-06 15:55:00.436 [1030/start/73637 (pid 25832)] Task failed.
```

I debugged the script. The main thread has the correct JSON string, then when Metaflow invokes a new process (pid 25832), the parameter becomes a string like `{'key': 'app', 'effect': 'NoSchedule', 'value': 'foo'}`, it replaces double quotes with quotes

shrinandj:
The PR itself looks good to me. Can you confirm that at least the following scenarios have been tested:
- Basic test for tolerations. Create a flow `--with kubernetes:tolerations=[{"key": "arch", "operator": "Equal", "value": "amd"}]`. It should run successfully. 
- Basic test for node_selectors since that code was also touched. Create a flow `--with kubernetes:tolerations=[{"key": "arch", "operator": "Equal", "value": "amd"}],node_selector="mynode=mylabel"`. It should run successfully. 
- Ensuring other options haven't regressed. Create a flow `--with kubernetes:cpu=2,tolerations=[{"key": "arch", "operator": "Equal", "value": "amd"}]`. It should run successfully.
- Ensuring that default K8s backend works as expected with just... `--with kubernetes`

The above tests with argo-workflows.
And some basic smoke tests WITHOUT kubernetes or Argo.

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -248,6 +258,10 @@ def runtime_step_cli(
             for k, v in self.attributes.items():
                 if k == "namespace":
                     cli_args.command_options["k8s_namespace"] = v
+                elif k == "node_selector":
+                    cli_args.command_options[k] = ",".join(["=".join([key, str(val)]) for key, val in v.items()])
+                elif k == "tolerations":
+                    cli_args.command_options[k] = json.dumps(v)
```
odracci, line 264: Convert the options back to strings. This is needed by the `kubernetes_cli`

metaflow/plugins/argo/argo_workflows.py
odracci: fixed

metaflow/plugins/kubernetes/kubernetes_job.py
odracci: Fixed

shrinandj:
> Could you please review my latest commits?

I will look into these latest commits by later tonight.

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -108,9 +108,15 @@ def __init__(self, attributes=None, statically_defined=False):
...
+            attribute_map = [
+                "effect",
+                "key",
+                "operator",
+                "toleration_seconds",
+                "value",
+            ]
             for toleration in self.attributes["tolerations"]:
-                invalid_keys = [k for k in toleration.keys() if k not in V1Toleration.attribute_map.keys()]
+                invalid_keys = [k for k in toleration.keys() if k not in attribute_map]
```
shrinandj, line 121: Why did this have to change? As K8s changes, these attributes could change, right?

odracci: When it runs on Argo, the module `kubernetes` is unavailable.
I can make the validation optional, given that it is required only when the flow runs locally and `KubernetesClient` raises an exception if the module is not installed. WDYT?
```python
if self.attributes["tolerations"]:
    try:
        from kubernetes.client import V1Toleration
        for toleration in self.attributes["tolerations"]:
            invalid_keys = [k for k in toleration.keys() if k not in V1Toleration.attribute_map.keys()]
            if len(invalid_keys) > 0:
                raise KubernetesException(
                    "Tolerations parameter contains invalid keys: %s" % invalid_keys
                )
    except (NameError, ImportError):
        pass
```

metaflow/plugins/kubernetes/kubernetes_decorator.py
odracci: solved

metaflow/plugins/kubernetes/kubernetes_decorator.py
shrinandj: The above looks like a happy compromise to me in which we try to do our best to validate AND keep up with upstream changes in K8s.

shrinandj (approved):
LGTM! Great work!!

metaflow/plugins/kubernetes/kubernetes_cli.py
```diff
@@ -166,6 +174,9 @@ def echo(msg, stream="stderr", job_id=None):
     stdout_location = ds.get_log_location(TASK_LOG_SOURCE, "stdout")
     stderr_location = ds.get_log_location(TASK_LOG_SOURCE, "stderr")
 
+    # `node_selector` is a tuple of strings, convert it to a dictionary
+    node_selector = KubernetesDecorator.parse_node_selector(node_selector)
```
savingoyal, line 178: is this needed anymore?

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -93,9 +99,27 @@ def __init__(self, attributes=None, statically_defined=False):
...
+        if isinstance(self.attributes["node_selector"], str):
+            self.attributes["node_selector"] = self.parse_node_selector(
+                self.attributes["node_selector"].split(",")
+            )
 
-        # TODO: Handle node_selector in a better manner. Currently it is special
-        #       cased in kubernetes_client.py
+        if self.attributes["tolerations"]:
+            try:
+                from kubernetes.client import V1Toleration
```
savingoyal, line 114: I understand that the rationale for including this check in `_init_` is to ensure that this check is invoked for `argo-workflows` too. However, this check will fail if the user hasn't installed the python package `kubernetes` yet - which is checked in `package_init` - that check should technically happen before the check for tolerations.

savingoyal (commented):
LGTM! Just one minor issue needs to be addressed. We should be good to merge and release right after.

metaflow/plugins/kubernetes/kubernetes_cli.py
odracci: Yes, it is. At this stage, `node_selector` is a tuple of strings. `kubernetes.launch_job` expects a dictionary

metaflow/plugins/kubernetes/kubernetes_decorator.py
odracci: It is checked in `kubernetes_cli`.
The idea is that the check is invoked only if required, which means the python package `kubernetes` must be installed. I think the order of the execution doesn't matter.
If `kubernetes` is not available, `self.attributes["tolerations"]` is not being used, then the check is not required.
Does it make sense to you?

savingoyal: When the user pip installs metaflow, we don't install Kubernetes python package. It's only when the user starts executing a flow that involves `@kubernetes` or argo - we throw a nice warning asking them to install the python package. Now, if that first flow has `tolerations` defined, then the user will instead get an error saying `no module named Kubernetes`

metaflow/plugins/kubernetes/kubernetes_cli.py
savingoyal: Any reason not to handle this parsing within `kubernetes_job` - the actual format is dictated by the kubernetes SDK and that's why currently all the Kubernetes-related formatting is happening within the KubernetesJob object. As the SDK evolves, any changes would be isolated to that object.

metaflow/plugins/kubernetes/kubernetes_decorator.py
odracci: That import is inside a `try` block with
```
except (NameError, ImportError):
  pass
```
It should not raise any errors related to the missing module, is it correct?

metaflow/plugins/kubernetes/kubernetes_cli.py
odracci: `node_selector` contains the value generated by
```python
@click.option(
    "--node-selector",
    multiple=True,
    default=None,
    help="NodeSelector for Kubernetes pod.",
)
```
which is a tuple of strings
```python
('key'='val','foo=bar')
```

`kubernetes_job` expect a dictionary like
```python
{
  "key": "val",
  "foo": "bar",
}
```
`parse_node_selector` converts the tuple of strings to a dictionary compatible with the Kubernetes SDK.
Does it make sense?

odracci:
@shrinandj @savingoyal I improved the error handling in https://github.com/Netflix/metaflow/pull/1207/commits/a1711b6971387bdd39faa1a8a3a52a9c797cbc32

odracci:
> The PR itself looks good to me. Can you confirm that at least the following scenarios have been tested:
> 
> * Basic test for tolerations. Create a flow `--with kubernetes:tolerations=[{"key": "arch", "operator": "Equal", "value": "amd"}]`. It should run successfully.
> * Basic test for node_selectors since that code was also touched. Create a flow `--with kubernetes:tolerations=[{"key": "arch", "operator": "Equal", "value": "amd"}],node_selector="mynode=mylabel"`. It should run successfully.
> * Ensuring other options haven't regressed. Create a flow `--with kubernetes:cpu=2,tolerations=[{"key": "arch", "operator": "Equal", "value": "amd"}]`. It should run successfully.
> * Ensuring that default K8s backend works as expected with just... `--with kubernetes`
> 
> The above tests with argo-workflows. And some basic smoke tests WITHOUT kubernetes or Argo.

@shrinandj Tests done, I've updated the description of this PR.

shrinandj:
I just realized that this PR would be a great reference for implementing some of the other features for K8s support (e.g. volume support).

shrinandj (approved):

odracci:
@shrinandj @savingoyal can you please approve the GitHub workflow?

metaflow/plugins/kubernetes/kubernetes_decorator.py
savingoyal: Yes - but it's the round about way this check is implemented which is my concern. We can ship this and come back to clean it up.

savingoyal (approved):

bbrandt:
_This seems to be the only documentation for Metaflow's Kubernetes tolerations support, so I'll add my note here._

To allow a Metaflow to run on an [Azure Spot node pool in AKS](https://learn.microsoft.com/en-us/azure/aks/spot-node-pool), add this to your Metaflow config.json:
```json
"METAFLOW_KUBERNETES_TOLERATIONS":"[{\"key\":\"kubernetes.azure.com/scalesetpriority\",\"value\":\"spot\",\"effect\":\"PreferNoSchedule\"},{\"key\":\"kubernetes.azure.com/scalesetpriority\",\"value\":\"spot\",\"effect\":\"NoSchedule\"}]"
```

This will allow flows to prefer running on a spot instance, but fallback to a more expensive node pool when your spot instance is not available.

PR #1213, opened 2022-12-09
Fix CVE-2007-4559 (tar.extractall)

romain-intel:
See #1177 for more details

### DIFF
--- a/metaflow/client/core.py
+++ b/metaflow/client/core.py
@@ -759,7 +759,8 @@ class MetaflowCode(object):
 
     You can extract code in the directory `snapshot` like so:
     ```
-    Run(...).code.tarball.extractall(path='snapshot')
+    from metaflow.util import tar_safe_extract
+    tar_safe_extract(tar=Run(...).code.tarball, path='snapshot')
     ```
 
     Attributes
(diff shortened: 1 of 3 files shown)


shrinandj (dismissed):

metaflow/client/core.py
```diff
@@ -759,7 +759,8 @@ class MetaflowCode(object):
 
     You can extract code in the directory `snapshot` like so:
     ```
-    Run(...).code.tarball.extractall(path='snapshot')
+    from metaflow.util import tar_safe_extract
```
savingoyal, line 762: is this needed? we don't expose any util methods from metaflow.util to the end user. instead we can recommend any method that the user has at their disposal to do tar extraction.

savingoyal (changes requested):

metaflow/client/core.py
romain-intel: Sure. I can remove the entire comment and just say that the tarball is given by Run(...).code.tarball and just leave it at that. I didn't want to recommend something that is potentially vulnerable (I know the risk is very very small).

savingoyal: That works!

savingoyal: Please merge the PR after that change is in :)

savingoyal (approved):

PR #1224, opened 2023-01-09
Support for multi-flow decorators.

valayDave:
PR Stack Merge Order (#1252 --> #1234 --> #1226 --> #1225 --> #1224)
- This change is required to support Airflow Sensors. 
- Airflow Sensors will support multiple flow decorators per flow 
- This will also be the base PR for the stacked Airflow Foreach + Sensors + Airflow GCP Support PRs

### DIFF
--- a/metaflow/decorators.py
+++ b/metaflow/decorators.py
@@ -517,11 +517,11 @@ def _init_flow_decorators(
                     "`allow_mutiple=True` are not allowed to have options"
                     % (deco.name, deco.__class__.__name__)
                 )
-            else:
-                # Each "non-multiple" flow decorator is only allowed to have one set of options
-                deco_flow_init_options = {
-                    option: deco_options[option] for option in deco.options
-                }
+        else:
+            # Each "non-multiple" flow decorator is only allowed to have one set of options
+            deco_flow_init_options = {
+                option: deco_options[option] for option in deco.options
+            }
         for deco in decorators:
             deco.flow_init(
                 flow,


savingoyal (dismissed):

metaflow/decorators.py
```diff
@@ -503,11 +504,20 @@ def _attach_decorators_to_step(step, decospecs):
...
     flow, graph, environment, flow_datastore, metadata, logger, echo, deco_options
 ):
+    # Since all flow decorators are stored as `{key:[deco]}` we iterate through each of them.
     for deco in flow._flow_decorators.values():
-        opts = {option: deco_options[option] for option in deco.options}
-        deco.flow_init(
-            flow, graph, environment, flow_datastore, metadata, logger, echo, opts
-        )
+        for rd in deco:
+            opts = {option: deco_options[option] for option in rd.options}
```
romain-intel, line 510: This should be hoisted out of the inner `for` loop because it makes it look like there are different options for each `rd` which is *not* the case (they are shared for a given type of decorator).

romain-intel (commented):
A few minor nits.

metaflow/decorators.py
valayDave: Won't the decorator options be different per decorator instance. For example : 

```python
@airflow_sensor_a(myopt1="DEF")
@airflow_sensor_a(myopt1="ABC")
class MyFlow(FlowSpec):
```

romain-intel: Yes, but those are *attributes* and not *options* (I got confused too). The attributes are things extracted from the decorator itself and the optiosn are extracted from the command line (so for project, for example, you have name as an attribute and production and branch as options).

To remove this confusion, when we were talking with @savingoyal today, we thought that maybe we could assert that if allow_multiple is true, the options dict is empty. An even better way of doing it is ensuring this at runtime too but it's a bit more complicated (ie: ensuring that it can't be filled in later too). This will prevent some weird things and ppl getting surprised.

valayDave: Refactored it. Ready for review.

romain-intel (approved):
LGTM. Gimme until monday to make my changes and test it out but should be good.

PR #1227, opened 2023-01-11
add parameter_names property to current with docstring

sam-watts:
See https://outerbounds-community.slack.com/archives/C020U025QJK/p1670331712790799

This will require changes to the docs site too - https://github.com/Netflix/metaflow-docs/blob/master/docs/api/current.ipynb. Happy to open a PR for that after this is merged :)

### DIFF
--- a/metaflow/current.py
+++ b/metaflow/current.py
@@ -16,6 +16,7 @@ class Current(object):
         self._username = None
         self._metadata_str = None
         self._is_running = False
+        self._parameter_names = None
 
         def _raise(ex):
             raise ex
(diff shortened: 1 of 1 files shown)


metaflow/current.py
```diff
@@ -220,6 +221,18 @@ def parallel(self):
...
+        """
+        The names of the Parameters defined in the current flow.
+
+        Returns
+        -------
+        List[str]
+            Parameter names.
+        """
+        return self._parameter_names
+
```
sam-watts, line 235: On reflection, this doesn't make sense. As `parameter_names` is updated directly by the `_update_env` method, which overwrites this property. This also means that the instance var I added `_parameter_names` is never used

Issue #1010, opened 2022-04-13
Add delete calls to S3 client


romain-intel:
We currently have read and write capabilities but do not support deleting. We could add a few calls like delete delete_all and some recursive way of deleting.

vijal-patel:
@romain-intel could I work on this?

romain-intel:
of course!

deepthi-kk:
Is this still open?

rajveer43:
can I work on this?

Shrujanapatil:
I'd like to work on this!

rupneel:
I am working on it

PR #1228, opened 2023-01-12
feat: delete, delete_many, delete_recursive to s3 wrapper

mSounak:
Reference to Issue #1010

1. Added support to delete one object using s3 client
2. Added support to delete multiple objects.
3. Added support to delete objects recursively.

### DIFF
--- a/metaflow/plugins/airflow/airflow.py
+++ b/metaflow/plugins/airflow/airflow.py
@@ -109,11 +109,10 @@ class Airflow(object):
     @classmethod
     def get_existing_deployment(cls, name, flow_datastore):
         _backend = flow_datastore._storage_impl
-        token_paths = _backend.list_content([cls.get_token_path(name)])
-        if len(token_paths) == 0:
+        token_exists = _backend.is_file([cls.get_token_path(name)])
+        if not token_exists[0]:
             return None
-
-        with _backend.load_bytes([token_paths[0]]) as get_results:
+        with _backend.load_bytes([cls.get_token_path(name)]) as get_results:
             for _, path, _ in get_results:
                 if path is not None:
                     with open(path, "r") as f:
@@ -125,12 +124,12 @@ class Airflow(object):
         return os.path.join(cls.TOKEN_STORAGE_ROOT, name)
 
     @classmethod
-    def save_deployment_token(cls, owner, token, flow_datastore):
+    def save_deployment_token(cls, owner, name, token, flow_datastore):
         _backend = flow_datastore._storage_impl
         _backend.save_bytes(
             [
                 (
-                    cls.get_token_path(token),
+                    cls.get_token_path(name),
                     BytesIO(
                         bytes(
                             json.dumps({"production_token": token, "owner": owner}),
--- a/metaflow/plugins/airflow/airflow_cli.py
+++ b/metaflow/plugins/airflow/airflow_cli.py
@@ -101,7 +101,7 @@ def resolve_token(
                 )
         obj.echo("")
         obj.echo("A new production token generated.")
-        Airflow.save_deployment_token(get_username(), token, obj.flow_datastore)
+        Airflow.save_deployment_token(get_username(), name, token, obj.flow_datastore)
     else:
         token = prev_token
 


mSounak:
@romain-intel can you please review this?

metaflow/plugins/datatools/s3/s3.py
```diff
@@ -792,6 +792,31 @@ def _head():
 
         return list(starmap(S3Object, _head()))
 
+    def delete(self, key=None, return_info=True):
```
romain-intel, line 795: This API seems a little bit confusing:
  - I don't think `key=None` makes sense (or maybe is too dangerous)
  - similarly, return_info also does not make sense (there is nothing to be deleted).

One parameter that may make sense is `error_on_absent` or something like that, similar to the overwrite parameter for writes in a way.

It would also be nice to use the `_one_boto_op` function like all other functions to benefit from the same retry logic as those.

Finally, and I know this is harder but we should consider a `delete_many` and, possibly, a `delete_recursive`.

romain-intel:
Thanks for taking this up. I've posted a few comments. Happy to provide more details if you ahve questions.

metaflow/plugins/datatools/s3/s3.py
mSounak: I have added `delete_many`. Could you explain what should happen in `delete_recursive`?

metaflow/plugins/datatools/s3/s3.py
```diff
@@ -792,6 +792,59 @@ def _head():
...
+
+        url = self._url(key)
+        src = urlparse(url)
+
+        def _delete(s3, tmp):
+            resp = s3.delete_object(Bucket=src.netloc, Key=src.path.lstrip('/'))
+
+            return resp
+
+        self._one_boto_op(_delete, url)
```
romain-intel, line 813: we can probably set `create_tmp_file=False` here since we don't need any temporary file.

metaflow/plugins/datatools/s3/s3.py
```diff
@@ -792,6 +792,59 @@ def _head():
...
+        ----------
+        key : str
+            Object to delete. It can be an S3 url or a path suffix.
+        """
+
+        url = self._url(key)
+        src = urlparse(url)
+
+        def _delete(s3, tmp):
+            resp = s3.delete_object(Bucket=src.netloc, Key=src.path.lstrip('/'))
```
romain-intel, line 809: can simplify to be `return s3.delete_object(...)`

romain-intel: also need to protect against key = None; it should be legit (and to follow the other calls, we can make key optional) and corresponds to deleting the root.

metaflow/plugins/datatools/s3/s3.py
```diff
@@ -792,6 +792,59 @@ def _head():
...
+
+            return resp
+
+        self._one_boto_op(_delete, url)
+
+    def delete_many(self, keys):
```
romain-intel, line 815: So for this one, I am not sure what the best way forward is:
- either use _one_boto_op and call the `delete_objects` as you do here
- go with the subprocess approach we take for get/put.

I can see pros/cons for both. Let me talk about this internally a little bit.

I haven't reviewed the function in detail yet.

romain-intel (commented):
For the `delete_recursive`, I was thinking of something like the `get_recursive` and somewhat equivalent to a `rm -r` in Linux (basically go down the "prefix" and delete everything in it).

mSounak:
> For the `delete_recursive`, I was thinking of something like the `get_recursive` and somewhat equivalent to a `rm -r` in Linux (basically go down the "prefix" and delete everything in it).

ok got it.

metaflow/plugins/datatools/s3/s3.py
mSounak: Any update on this?

mSounak:
@romain-intel I have also added `delete_recursive`.

romain-intel:
Hello. Apologies, I have been running late on a few other tasks. I will take a look hopefully soon. I am on vacation next week but will try to make some time in the evenings and get back to you.

Issue #1232, opened 2023-01-13
Card display in vscode notebook takes too much space


obgibson:
Cards shown in vscode notebooks have a huge amount of whitespace at the bottom of them.

Research showed that the size of the card's `output_container` was 1.67E7px.

![](https://files.slack.com/files-pri/T01NV728P9P-F04KCQPD2P2/screen_shot_2023-01-12_at_10.52.21_am.png)

To replicate

* Run a flow with a card
* Get the cards for the flow inside a vscode notebook e.g.
```
from metaflow.cards import get_cards
flow_name = 'MovieStatsFlow'
run_id = 28
step_name = 'start'
task_id = 219
get_cards(task=f'{flow_name}/{run_id}/{step_name}/{task_id}')
```
* Scroll down the card and see the huge amount of whitespace at the bottom.

PR #1233, opened 2023-01-13
Bug/long card

obgibson:
Fixing: https://github.com/Netflix/metaflow/issues/1232

Hypothesis was that a process in VSCode's notebook was setting the height of the `output_container` to 1.67E7. And that was because it thought the height of the card was that much because it did not understand `100vh` in the card's CSS.

This fix replaces `100vh` by `12rem` for cases where the card is embedded (using `embed` class as a selector).


seethroughdev (approved):
LGTM.

valayDave (approved):

Issue #1235, opened 2023-01-16
Support for labels with @kubernetes


dhpollack:
It would be nice if additional labels were supported.  This can be done pretty easily since the code already supports launching pods with labels except the labels are always empty.

shrinandj:
The `labels` feature in Kubernetes is definitely very useful and powerful for filtering, querying, attribution, etc. Metaflow also has the notion of `

dhpollack:
> The `labels` feature in Kubernetes is definitely very useful and powerful for filtering, querying, attribution, etc.
> 
>     * Metaflow also has the notion of `tags` which allow users to attach arbitrary metadata to runs so as to get similar functionality.
> 
>     * Metaflow adds annotations for common metadata: run-id, task-id, etc. to pods that are run in the Kubernetes cluster.
> 
> 
> Can you elaborate a bit about the use-cases where tags may not be sufficient?
> 
> Also, could having both `labels` and `tags` be confusing?

I'll answer your second question first.  I believe having both metaflow tags and kubernetes labels would not be confusing.  They basically have two different audiences.  Metaflow tags are for data scientists and users of metaflow.  You guys do a great job in your docs explaining how those can be used.  Kubernetes labels are used by infrastructure teams to do things that don't really concern data scientists.

Importantly, many tools that utilize kubernetes labels are not metaflow-aware and often times are not even python-based.  I would say the most well known would be using prometheus metrics and grafana dashboards to monitor all pods with a certain kubernetes labels.  These kubernetes labels are automatically scraped from kubernetes labels.  Importantly, these metrics are part of everything that runs in kubernetes, regardless of if it's a flow in metaflow, a microservice written in Go, or a build run as part of a CI/CD pipeline.  For these things it would be difficult or impossible to get any metadata from metaflow.

One other thing that I want to mention is that you already support adding kubernetes labels and add a few labels to jobs launched by metaflow, which is great.  But it's currently impossible to add custom labels, although not very difficult to implement because most of the work has already been done since kubernetes labels are already being used.

shrinandj:
Thanks a lot for clarifying this. I agree that use-cases such as adding prometheus specific labels or labels for similar use-cases would require having K8s specific labels.

> Metaflow tags are for data scientists and users of metaflow...Kubernetes labels are used by infrastructure teams to do things that don't really concern data scientists.

[...]

PR #1236, opened 2023-01-16
Add kubernetes labels

dhpollack:
This adds support for adding kubernetes labels to pods.  Either via the kubernetes decorator with `labels={"label": "value", "label2": "value2"}`, with an env var `METAFLOW_KUBERNETES_LABELS="label=value,label2=value2"`, or add to the with command `python hello_world.py run --with kubernetes:labels="cli_label=cli_value"`.  I also created an issue for this:

Netflix/metaflow#1235


metaflow/plugins/kubernetes/kubernetes_cli.py
```diff
@@ -175,7 +182,13 @@ def echo(msg, stream="stderr", job_id=None):
     stderr_location = ds.get_log_location(TASK_LOG_SOURCE, "stderr")
 
     # `node_selector` is a tuple of strings, convert it to a dictionary
-    node_selector = KubernetesDecorator.parse_node_selector(node_selector)
+    node_selector = KubernetesDecorator.parse_kube_list(node_selector)
+
+    # `labels` is a tuple of strings or a tuple with a single comma separated string
+    # convert it to a dict
+    labels = KubernetesDecorator.parse_kube_list(
```
dhpollack, line 189: The string generated in the `runtime_step_cli` in the `kubernetes_decorator.py` ends up looking something like:

```
labels = ("label1=val1,label2=val2",)
```

This seems somewhat by design for the `node_selector` field.  Hence there is a bit of additional logic here.

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -103,9 +109,16 @@ def __init__(self, attributes=None, statically_defined=False):
             self.attributes["node_selector"] = KUBERNETES_NODE_SELECTOR
         if not self.attributes["tolerations"] and KUBERNETES_TOLERATIONS:
             self.attributes["tolerations"] = json.loads(KUBERNETES_TOLERATIONS)
+        if not self.attributes["labels"] and KUBERNETES_LABELS:
+            self.attributes["labels"] = KUBERNETES_LABELS
+
+        if isinstance(self.attributes["labels"], str):
+            self.attributes["labels"] = self.parse_kube_list(
+                self.attributes["labels"].split(","), False
+            )
```
odracci, line 118: I would validate the labels, using something like https://github.com/apache/airflow/blob/207f65b542a8aa212f04a9d252762643cfd67a74/airflow/kubernetes/pod_generator.py#L56-L74
[Kubernetes documentation](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set)

dhpollack:
> Hey @dhpollack, thanks for looking into this. Metaflow also has the concept of `tags` (https://docs.metaflow.org/scaling/tagging).
> 
> Is there any overlap between `tags` and labels? And could there be some option to simply use one to avoid confusion? WDYT?

We use kubernetes labels everywhere in our organization for cost tracking and tagging. Some of these labels are not very useful as metaflow tags.

dhpollack:
> Hey @dhpollack, thanks for looking into this. Metaflow also has the concept of `tags` (https://docs.metaflow.org/scaling/tagging).
> 
> Is there any overlap between `tags` and labels? And could there be some option to simply use one to avoid confusion? WDYT?

But specifically about overlap... There is a little but they are not the same nor do I think one should try to use them in the same. 

For one, your tags are strings, while the kubernetes labels are key-valur pairs. But even from a user perspective, metaflow tags seem designed for data scientists to track experiments in metaflow. While kubernetes labels are what a lot of the kubernetes universe uses for filtering and organizing things that happen on kubernetes...

If you changed metaflow tags to mirror kubernetes labels (key-valur, restrictions on length and valid characters, etc etc) then i think you could simply use one of them. But that's a lot more work for you and I think people that want to add kubernetes labels to their pods generally know what kubernetes labels are and wouldn't be too bothered by the additional complexity.

dhpollack:
@shrinandj can you re-run the failed CI test?  It looks like it failed but shouldn't have.  Probably just a hiccup on the machine.

metaflow/plugins/kubernetes/kubernetes_decorator.py
dhpollack: done

metaflow/plugins/kubernetes/kubernetes_cli.py
dhpollack: https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/kubernetes/kubernetes_decorator.py#L288

I suspect this should not have `",".join(...)`.  I will test if removing that fixes the issue.

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -384,13 +401,34 @@ def _save_package_once(cls, flow_datastore, package):
             )[0]
 
     @staticmethod
-    def parse_node_selector(node_selector: list):
+    def parse_kube_list(items: Union[str, List[str]], requires_both: bool = True):
```
shrinandj, line 404: Is the name `parse_kube_list` seem confusing? Would something like `parse_kube_keyvalue_list` be better? It also makes the function argument `requires_both` more intuitive.

WDYT?

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -384,13 +401,34 @@ def _save_package_once(cls, flow_datastore, package):
...
-            raise KubernetesException(
-                "Unable to parse node_selector: %s" % node_selector
-            )
+            raise KubernetesException("Unable to parse kubernetes list: %s" % items)
+
+    @staticmethod
+    def clean_kube_labels(
+        labels: Optional[Dict[str, Optional[str]]],
+        max_len: int = 63,
+        regex_sub: str = r"^[^a-z0-9A-Z]*|[^a-zA-Z0-9_\-\.]|[^a-z0-9A-Z]*$",
```
shrinandj, line 420: Can you confirm that this was tested thoroughly?

shrinandj:
Generally looks good. Thanks for doing this @dhpollack! Can you add some details of the tests that were run for testing this PR? Especially, did you create argo workflows with this code and ensure that the pods created by Argo have the correct labels?

dhpollack:
So to test this I used something like the following flow:

```python
from metaflow import FlowSpec, step, kubernetes

class TestFlow(FlowSpec):
    @step
    def start(self):
        self.my_var = "hello world"
        self.next(self.a)

    @kubernetes(labels={"my_label": "my_value", "label2": None, "label3": "val"})
    @step
    def a(self):
        print("the data artifact is: %s" % self.my_var)
        self.next(self.b)

    @kubernetes(
        labels={
            "my_label": "superlonglabelthatshouldgethashedbytheautomatedhashingfunctionorsomethinglikethat",  # noqa: E501
            "my_label2": "ein bißchen falsch",
        }
    )
    @step
    def b(self):
        pass
        self.next(self.end)

    @step
    def end(self):
        print("the data artifact is still: %s" % self.my_var)

if __name__ == "__main__":
    TestFlow()
```

Then I ran the following (note image each space is a new terminal):

```bash
export METAFLOW_KUBERNETES_LABELS='env_label=env_var`
python test_workflow.py run  --with kubernetes

python test_workflow.py run

python test_workflow.py run --with kubernetes:labels="label1=val1,label2=val2,label3,label4=val4"

# remove kubernetes decorators from the flow (this is actually how we generally would use this)
export METAFLOW_KUBERNETES_LABELS='env_label=env_var`
python test_workflow.py run --with kubernetes

export METAFLOW_KUBERNETES_LABELS='env_label=env_var`
python test_workflow.py argo-workflows create && python test_workflows argo-workflows trigger
```

There are a few quirks.

1. If you set an env var and use `--with kubernetes` the steps decorated with the kubernetes labels overwrite the env var labels completely.
2. The argo labels seem to work, but it feels a bit like magic because metaflow automatically imports the kubernetes decorator and that adds the labels.  If you look at the argo template that gets generated, the labels are not in the argo metadata.  But I feel like that was ok, because if you wanted to set those then perhaps you'd have a separate parameter for the argo decorator that is a flow decorator rather than a step decorator.

metaflow/plugins/kubernetes/kubernetes_decorator.py
dhpollack: Ok, I added some tests for this function.  I also added an assertion error if the label is entirely composed of non-valid characters.  This seems like an edge case that would fail while launching the job.

metaflow/plugins/kubernetes/kubernetes_decorator.py
dhpollack: cool, I made this change.

test/unit/test_kubernetes_decorator.py
```diff
@@ -0,0 +1,33 @@
+import pytest
+
+from metaflow.plugins.kubernetes.kubernetes_decorator import KubernetesDecorator
+
+
+@pytest.mark.parametrize(
```
shrinandj, line 6: This is awesome! Thank you so much for doing this!

shrinandj:
Thanks for the details about testing. It definitely increases the reviewers confidence in the commits.

Just a couple of other tests:
- Can you test with K8s labels AND nodeSelectors since this code touches that functionality as well.
- Can you test original `@kubernetes` and `--with kubernetes` functionality without labels. Basically, to ensure that nothing has regressed because of this change.

dhpollack:
> Thanks for the details about testing. It definitely increases the reviewers confidence in the commits.
> 
> Just a couple of other tests:
> 
>     * Can you test with K8s labels AND nodeSelectors since this code touches that functionality as well.
> 
>     * Can you test original `@kubernetes` and `--with kubernetes` functionality without labels. Basically, to ensure that nothing has regressed because of this change.

I forgot to mention that in my comment about the tests.  I also tested without labels and with node selectors in different combinations as well (alone, via the decorator, env vars, and using `--with kubernetes:...`.  The main change was the change in the [`runtime_step_cli`](https://github.com/Netflix/metaflow/pull/1236/files#diff-61935adb175a915b5065b6d681245a71e224de77f911687c1ffa3f48d5542aeaR304).  Previously, if you had more than one node selector, then it would create a tuple with a single item like `('label1=val1,label2=val2',)` and now it will do `('label1=val1', 'label2=val2')` which I think was just a bug in the previous code.

shrinandj (dismissed):

dhpollack:
@savingoyal Anything you would like done here?

shrinandj:
I download the source code locally and tried to test this change.

* Ran a flow with Kubernetes and a single label. This worked just fine.
```
$ python3 /tmp/a.py run --with kubernetes:labels="abc=pqr"
...
```

* Ran a flow with Kubernetes decorator and added a label there. This worked just fine.
```
    @kubernetes(labels={"abc":"pqr"})
```

* Ran a flow with Kubernetes decorator and add two labels. This worked just fine too.
```
    @kubernetes(labels={"abc":"pqr", "lmn":"xyz"})
```

* Ran a flow with Kubernetes on cli with two labels. **THIS FAILED** with the following error
```
$ python3 a.py run --with kubernetes:labels="abc=pqr,lmn=xyz"
Metaflow 2.7.19 executing StarterFlow for user:shri
    Unknown decorator attribute:
    Decorator 'kubernetes' does not support the attribute 'lmn'. These attributes are supported: cpu, memory, disk, image, service_account, secrets, node_selector, labels, namespace, gpu, gpu_vendor, tolerations.
```

* exported the env variable METAFLOW_KUBERNETES_LABELS with one label and ran a flow. This worked just fine.

* exported the env variable METAFLOW_KUBERNETES_LABELS with two labels and ran a flow. This worked just fine too.

* Ran a few other flows with Kubernetes specifying cpu, memory, etc. on the cli. They worked just fine.

dhpollack:
@shrinandj looking at how metaflow creates the `--with kubernetes` string for argo.  It actually does something like `--with kubernetes:labels='{"abc":"pqr", "lmn":"xyz"}'`.  I am not sure if this is undesired behavior or bad documentation.  The way that I had it before was based on the node selector, but that also did not work correctly when having multiple node selectors.

https://github.com/Netflix/metaflow/pull/1236#discussion_r1071193731

It's related to this issue which I brought up earlier.

dhpollack:
The following code will split `"labels=l1=v1,l2=v2"` as `[("labels", "l1=v1"), ("l2", "v2")]`.  It looks like the only way to have multiple objects in this format is to use a json object.  The other way would be to write a much more complicated regex that searches for valid keys and then rewrites that string as a json.

https://github.com/Netflix/metaflow/blob/master/metaflow/decorators.py#L132

```
    @classmethod
    def _parse_decorator_spec(cls, deco_spec):
        if len(deco_spec) == 0:
            return cls()

        attrs = {}
        # TODO: Do we really want to allow spaces in the names of attributes?!?
        for a in re.split(""",(?=[\s\w]+=)""", deco_spec):
            name, val = a.split("=", 1)
            ...
```

shrinandj:
There's some quirkiness involved related to the shell and the OS involved when there are multiple labels to be specified on the CLI. Using the json style input works just fine in that case.

Verified this with:
```
python a.py run --with kubernetes:labels='{"l1":"v1","l2":"v2"}'
```

This change should be good to go!

shrinandj (approved):

metaflow/plugins/kubernetes/kubernetes.py
```diff
@@ -9,6 +9,8 @@
 from metaflow.metaflow_config import (
     SERVICE_HEADERS,
     SERVICE_INTERNAL_URL,
+    CARD_AZUREROOT,
```
dhpollack, line 12: should this be removed?

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -1,8 +1,10 @@
+import hashlib
 import json
 import os
 import platform
-import requests
+import re
 import sys
```
dhpollack, line 6:
```suggestion
import requests
```

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -399,14 +414,61 @@ def _save_package_once(cls, flow_datastore, package):
...
+                ret[str(item[0])] = str(item[1]) if len(item) > 1 else None
+            return ret
         except (AttributeError, IndexError):
-            raise KubernetesException(
-                "Unable to parse node_selector: %s" % node_selector
-            )
+            raise KubernetesException("Unable to parse kubernetes list: %s" % items)
+
+    @staticmethod
+    def clean_kube_labels(
```
savingoyal, line 453: @dhpollack what's the intention behind this? Ideally if we are supporting labels we would want to pass them as is to kubernetes.

savingoyal: Mutating labels may catch the user off guard when the desired label isn't attached as is - we could rely on runtime checks disallowing certain labels instead - I agree with you that that will lead to an error further down during runtime but will be a better no-surprises approach. Also, we do expect that OPA may intercept the request and not allow certain kinds of labels be attached to the pod - so there is that use case. Would it be useful to jump on a quick 15-30 min call to quickly go over all open questions so that we can push this PR out asap? Let me know if any time [here](https://calendar.app.google/7wjKcswSBVdMgMbNA) works.

dhpollack: I have changed this to throw an exception now.

shrinandj (approved):

shrinandj (approved):

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -387,14 +403,59 @@ def _save_package_once(cls, flow_datastore, package):
...
+        labels: Optional[Dict[str, Optional[str]]],
+        regex_match: str = r"^(([A-Za-z0-9][-A-Za-z0-9_.]{0,61})?[A-Za-z0-9])?$",
+    ):
+        def validate_label(s: Optional[str]):
+            if not s:
+                # allow empty label
+                return s
+            if not re.search(regex_match, s):
+                # this is the same message kubernetes itself returns
+                raise Exception(
```
saikonen, line 452: use the MetaflowException wrapper to get a cleaner error output for the CLI

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -98,9 +105,17 @@ def __init__(self, attributes=None, statically_defined=False):
...
         if not self.attributes["tolerations"] and KUBERNETES_TOLERATIONS:
             self.attributes["tolerations"] = json.loads(KUBERNETES_TOLERATIONS)
+        if not self.attributes["labels"] and KUBERNETES_LABELS:
+            self.attributes["labels"] = KUBERNETES_LABELS
+
+        if isinstance(self.attributes["labels"], str):
+            self.attributes["labels"] = self.parse_kube_keyvalue_list(
+                self.attributes["labels"].split(","), False
+            )
+        self.attributes["labels"] = self.validate_kube_labels(self.attributes["labels"])
```
saikonen, line 115: Nit: relies on `validate_kube_labels` to return something, which given its name is unexpected. As the validator is more concerned about raising exceptions, it doesn't necessarily need to return anything. Could we instead simply call it to validate the labels in the attributes?

Based on the [lifecycle graph](https://github.com/Netflix/metaflow/blob/master/docs/lifecycle.png) , one place to call this could be in the `step_init()` phase

not a dealbreaker though so feel free to ignore this, as the current implementation also works :)

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -387,14 +403,59 @@ def _save_package_once(cls, flow_datastore, package):
...
+                raise Exception(
+                    f'Invalid value: "{s}": a valid label must be an empty string or '
+                    "consist of alphanumeric characters, '-', '_' or '.', and must "
+                    "start and end with an alphanumeric character (e.g. 'MyValue', "
+                    "or 'my_value', or '12345', regex used for validation is "
+                    "'^(([A-Za-z0-9][-A-Za-z0-9_.]{0,61})?[A-Za-z0-9])?$'"
+                )
+            return s
+
+        return {k: validate_label(v) for k, v in labels.items()} if labels else labels
```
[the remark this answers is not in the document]
dhpollack: The keys would need to pass a more complicated regex.  Basically, you can have a prefix for keys that is a valid DSN and can be up to 256 characters.  But also, one reason I am not validating the keys is that in theory the key is more static than the value.  I would just let bad keys fail and let the user get the kubernetes message.  If you want me to look for those reserved prefixes (`kubernetes.io` and `k8s.io`), I could easily do that.  I don't think it is worth it tho.

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -387,14 +403,59 @@ def _save_package_once(cls, flow_datastore, package):
...
+                # allow empty label
+                return s
+            if not re.search(regex_match, s):
+                # this is the same message kubernetes itself returns
+                raise Exception(
+                    f'Invalid value: "{s}": a valid label must be an empty string or '
+                    "consist of alphanumeric characters, '-', '_' or '.', and must "
+                    "start and end with an alphanumeric character (e.g. 'MyValue', "
+                    "or 'my_value', or '12345', regex used for validation is "
+                    "'^(([A-Za-z0-9][-A-Za-z0-9_.]{0,61})?[A-Za-z0-9])?$'"
```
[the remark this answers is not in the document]
dhpollack: This is like the exact message that the kubernetes library gives you.  Would you prefer something like this:

```
                raise MetaflowException(
                    f'Invalid value: "{s}": a valid label must be an empty string or '
                    "consist of alphanumeric characters, '-', '_' or '.', must "
                    "start and end with an alphanumeric character (e.g. 'MyValue', "
                    "or 'my_value', or '12345'), and be no longer than 63 characters"
                )
```

metaflow/plugins/kubernetes/kubernetes_decorator.py
dhpollack: Gonna use the `KubernetesException` which is a child of `MetaflowException`

metaflow/plugins/kubernetes/kubernetes_decorator.py
dhpollack: So I changed the function to return a bool or throw an exception.

metaflow/plugins/kubernetes/kubernetes_decorator.py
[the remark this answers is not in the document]
saikonen: It's a bit easier on the eyes without the regex in my opinion. I would maybe consider splitting the conditions into a list like
```
raise KubernetesException(
                    f'Invalid value: "{s}"\n'
                    "A valid label is an empty string or one that\n"
                    "   - Consists of alphanumeric, '-', '_', or '.' characters\n"
                    "   - Begins and ends with an alphanumeric character\n"
                    "   - Is at most 63 characters long"
                )
```
but the current change is good as well.

Note that there's still the comment about the exception being identical to what kubernetes throws which should be cleaned up.

metaflow/plugins/kubernetes/kubernetes_decorator.py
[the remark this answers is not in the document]
saikonen: Agree that it might become unnecessarily complex with the key validation, we can skip this for the time being then. Maybe good to write a comment in the code for future developers as to why we only validate the values.

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -391,14 +407,80 @@ def _save_package_once(cls, flow_datastore, package):
...
-                str(k.split("=", 1)[0]): str(k.split("=", 1)[1])
-                for k in node_selector or []
-            }
+            ret = {}
+            for item_str in items:
+                item = item_str.split("=", 1)
+                if requires_both:
+                    item[1]  # raise IndexError
+                if str(item[0]) in ret:
+                    raise KubernetesException("Duplicate key found: %s" % str(item[0]))
```
saikonen, line 450: In its current form the exception does not  convey where the duplicate key was found, as it makes no mention of 'labels' when I set duplicate labels.

As this is used for checks other than the labels as well, A more suitable place to perform the duplicate label check might be in the `validate_kube_labels` instead. That way the exception can be more specific as well (Duplicate *label* found)

dhpollack: This function is used on both labels and node selectors and putting the check here ensures that we catch duplicate keys in both sets.

savingoyal (approved):

savingoyal:
@dhpollack looks like this PR is backward incompatible - 

Screen Shot 2023-04-17 at 3 32 17 PM

I have created #1359 which reverts this PR so that releases are unblocked. Can you please address this issue and resubmit the PR?

dhpollack:
> @dhpollack looks like this PR is backward incompatible -
> 
> Screen Shot 2023-04-17 at 3 32 17 PM
> 
> I have created #1359 which reverts this PR so that releases are unblocked. Can you please address this issue and resubmit the PR?

Yea, looks like it should be a quick fix.  I'll work on it tomorrow.

PR #1242, opened 2023-01-22
fix #1241

### LINKED ISSUE #1241: Error in Tutorial 4 (04-playlist-plus) caused by incompatibility with latest pandas version.
Error message:
`AttributeError: Can't get attribute '_unpickle_block' on <module 'pandas._libs.internals' from ...`
 

I was following the tutorial for the metaflow project and encountered an error in Tutorial 4. The issue is caused by a compatibility issue between the pinned version of pandas used in Tutorial 4 and the latest pandas.

Specifically, the code in Tutorial 4 attempts to pickle the output from Tutorial 3 but raises an error due to incompatibility. This is preventing me from progressing in the tutorial and I cannot continue.

A solution is to update the pinned version of pandas to the latest version:
```

@conda_base(libraries={'pandas': "1.5.3"})
class PlayListFlow(FlowSpec):

```

eracle:
fix #1241

### DIFF
--- a/metaflow/tutorials/04-playlist-plus/playlist.py
+++ b/metaflow/tutorials/04-playlist-plus/playlist.py
@@ -48,7 +48,7 @@ class PlayListFlow(FlowSpec):
         default=5,
     )
 
-    @conda(libraries={"pandas": "1.3.3"})
+    @conda(libraries={"pandas": "1.5.3"})
     @step
     def start(self):
         """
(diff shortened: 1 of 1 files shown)


shrinandj (approved):

savingoyal:
This PR is no longer needed since the tutorials have been changed to remove dependency on pandas

PR #1234, opened 2023-01-14
Kubernetes Variables for AWS MWAA support with Airflow

valayDave:
Stacked on top of #1226 

## Introduces new config variables : 

`METAFLOW_AIRFLOW_KUBERNETES_KUBECONFIG_CONTEXT`: Context to use in the kubeconfig file. 

`METAFLOW_AIRFLOW_KUBERNETES_KUBECONFIG_FILE`: Path to the kubeconfig file to provide the Airflow KubernetesOperator

Setting Either of these or `METAFLOW_AIRFLOW_KUBERNETES_CONN_ID` will set `in_cluster=True` in the Airflow `KubernetesOperator`

Changes done to support AWS MWAA

PR #1250, opened 2023-01-27
support timezone when scheduling with Argo workflows

amerberg:
This PR adds support for specifying a timezone when scheduling Argo workflows.

The relevant Argo documentation is here: https://argoproj.github.io/argo-workflows/cron-workflows/#cronworkflow-options

### DIFF
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -120,6 +120,7 @@ class ArgoWorkflows(object):
         self.parameters = self._process_parameters()
         self._workflow_template = self._compile()
         self._cron = self._cron()
+        self._timezone = self._timezone()
 
     def __str__(self):
         return str(self._workflow_template)
@@ -179,10 +180,14 @@ class ArgoWorkflows(object):
             return " ".join(schedule.schedule.split()[:5])
         return None
 
+    def _timezone(self):
+        schedule = self.flow._flow_decorators.get("schedule")
+        return schedule.timezone if schedule else None
+
     def schedule(self):
         try:
             ArgoClient(namespace=KUBERNETES_NAMESPACE).schedule_workflow_template(
-                self.name, self._cron
+                self.name, self._cron, self._timezone
             )
         except Exception as e:
             raise ArgoWorkflowsSchedulingException(str(e))
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -120,7 +120,6 @@ class ArgoWorkflows(object):
         self.parameters = self._process_parameters()
         self._workflow_template = self._compile()
         self._cron = self._cron()
-        self._timezone = self._timezone()
 
     def __str__(self):
         return str(self._workflow_template)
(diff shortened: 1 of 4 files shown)


metaflow/plugins/argo/argo_workflows.py
```diff
@@ -179,10 +180,14 @@ def _cron(self):
             return " ".join(schedule.schedule.split()[:5])
         return None
 
+    def _timezone(self):
```
savingoyal, line 183: small nit - can we have `_cron(...)` return a tuple that includes the timezone as well. that will negate the need for creating this function.

amerberg: Done.

amerberg:
How do we want to handle AWS Batch, which as far as I can tell, doesn't support a timezone with the cron schedule? It seems like we should at least issue a warning, perhaps even fail, if the user specifies a timezone and we can't use it with it.

shrinandj:
Thanks a lot for adding this!

Other than addressing the review comment above, can you also add a note about the testing you did for this change? The basic tests to do here would include running a flow providing a timezone and running a flow without a timezone (current behavior).

amerberg:
Yes, I tested by adding a schedule decorator to a simple flow. I created workflows with two different timezones, and then one with no timezone, and confirmed that they all started at the correct time.

savingoyal:
> How do we want to handle AWS Batch, which as far as I can tell, doesn't support a timezone with the cron schedule? It seems like we should at least issue a warning, perhaps even fail, if the user specifies a timezone and we can't use it with it.

Yep - within AWS Step Functions we would want to check if the user has specified timezone and fail hard if that is the case. Can you also add a note in `@schedule` re: the format that the timezone should be specified in.

savingoyal:
cc @valayDave re: impact on Airflow integration

amerberg:
Added documentation for timezone format (IANA standard) and an assertion to check that a timezone isn't provided with Step Functions.

shrinandj (dismissed):
LGTM! Thanks for your contributions!

metaflow/plugins/aws/step_functions/step_functions.py
```diff
@@ -329,6 +329,9 @@ def _visit(node, workflow, exit_node=None):
     def _cron(self):
         schedule = self.flow._flow_decorators.get("schedule")
         if schedule:
+            assert (
```
savingoyal, line 332: rather than the assert, can you raise `StepFunctionsException` so that Metaflow can pretty print the exception?

amerberg: Done.

savingoyal (approved):
LGTM! (pending a review from @valayDave regarding Airflow compatibility)

amerberg:
Based on the [Airflow docs](https://airflow.apache.org/docs/apache-airflow/stable/authoring-and-scheduling/timezone.html#time-zone-aware-dags), I think the way to make this work with Airflow would be to use a timezone-aware `pendulum.datetime` for the `start_date` here: https://github.com/Netflix/metaflow/blob/f6c632022d94f11313453084264a6fb9ae16ef28/metaflow/plugins/airflow/airflow.py#L649. However, I don't have a working Airflow setup to test it.

metaflow/plugins/aws/step_functions/schedule_decorator.py
```diff
@@ -21,7 +21,13 @@ class ScheduleDecorator(FlowDecorator):
     """
 
     name = "schedule"
-    defaults = {"cron": None, "weekly": False, "daily": True, "hourly": False}
+    defaults = {
+        "cron": None,
+        "weekly": False,
+        "daily": True,
+        "hourly": False,
+        "timezone": None,
```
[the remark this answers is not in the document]
amerberg: Thanks, done.

valayDave:
@amerberg ; Thanks for scoping the Airflow change. For now we can leave airflow with out the change. I already have some pending stacked PRs like #1234 where i can incorporate necessary changes.

amerberg:
@valayDave Should we at least add a check, analogous to the Step Functions check, to fail if the user specifies a timezone with Airflow?

valayDave:
@amerberg : I already have something in the works with #1252 ; For the moment we can merge this PR after which I will rebase and update my stacked PRs so that we do something analogous to the Step functions check.

PR #1251, opened 2023-01-28
Implement @secrets, with AWS support

jackie-ob:
RFC doc: https://docs.google.com/document/d/1vhU7lLZo1znF6of_cE8KewNae99_fUBsaoFHhW8FJew/edit#heading=h.tt3aazlsd22b 

This PR implements the general @secrets decorator interface, as well as AWS Secrets Manager backend.

How to use `@secrets(sources=[...])`?  `sources` is a list of strings or dictionaries.

String-style will cover most users with vanilla requirements (secret payload are JSON objects, to be unrolled as env vars).

Dictionary-style will allow specification of provider specific `options`.  For AWS here, the `json` option allows disabling of JSON parsing, in case the secret payload string is not the right kind of JSON (or not JSON at all).  `region` also allows control of which region's Secret Manager to use.

`inline` secrets provider type is included for testing purposes ("internal use").

This PR has been manually tested with local runs, `--with=kubernetes`, `--with=batch`, `step-functions` and `argo-workflows` (all on AWS).

Examples:
```
    @secrets(
        sources=[
            "RandomString-7Ld1",
            "arn:aws:secretsmanager:us-west-2:006988687827:secret:config_oleg2_codemate.json",
            "jackie-binary",
            {
                "id": "jackie-json-list",
                "options": {
                    "json": False,
                },
            },
            {
                "id": "jackie-non-json-str",
                "options": {
                    "json": False,
                },
            },
            {
                "type": "_inline",
                "id": "the_id",
                "options": {
                    "env_vars": {"FAKE_SECRET": "asdffgdfg", "PATH2": "party"},
                },
            },
        ]
    )
```

### DIFF
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -24,6 +24,8 @@ from metaflow.metaflow_config import (
     DATASTORE_SYSROOT_GS,
     CARD_AZUREROOT,
     CARD_GSROOT,
+    DEFAULT_SECRETS_BACKEND_TYPE,
+    AWS_SECRETS_MANAGER_DEFAULT_REGION,
 )
 from metaflow.mflog import BASH_SAVE_LOGS, bash_capture_logs, export_mflog_env_vars
 from metaflow.parameters import deploy_time_eval
@@ -801,6 +803,12 @@ class ArgoWorkflows(object):
 
             # support Metaflow sandboxes
             env["METAFLOW_INIT_SCRIPT"] = KUBERNETES_SANDBOX_INIT_SCRIPT
+
+            # secrets stuff
+            env["METAFLOW_DEFAULT_SECRETS_BACKEND_TYPE"] = DEFAULT_SECRETS_BACKEND_TYPE
+            env[
+                "METAFLOW_AWS_SECRETS_MANAGER_DEFAULT_REGION"
+            ] = AWS_SECRETS_MANAGER_DEFAULT_REGION
 
             # Azure stuff
             env[
--- a/metaflow/plugins/aws/secrets_manager/aws_secrets_manager_secrets_provider.py
+++ b/metaflow/plugins/aws/secrets_manager/aws_secrets_manager_secrets_provider.py
@@ -1,0 +1,163 @@
+import base64
+import json
+from json import JSONDecodeError
+
+
+from metaflow.exception import MetaflowException
+from metaflow.metaflow_config import AWS_SECRETS_MANAGER_DEFAULT_REGION
+from metaflow.plugins.secrets import SecretsProvider
+import re
+
+
+class MetaflowAWSSecretsManagerBadResponse(MetaflowException):
+    """Raised when the response from AWS Secrets Manager is not valid in some way"""
+
+
+class MetaflowAWSSecretsManagerDuplicateKey(MetaflowException):
+    """Raised when the response from AWS Secrets Manager contains duplicate keys"""
+
+
+class MetaflowAWSSecretsManagerJSONParseError(MetaflowException):
+    """Raised when the SecretString response from AWS Secrets Manager is not valid JSON"""
+
+
+class MetaflowAWSSecretsManagerNotJSONObject(MetaflowException):
+    """Raised when the SecretString response from AWS Secrets Manager is not valid JSON object (dictionary)"""
+
+
+def _sanitize_key_as_env_var(key):
+    """
+    Sanitize a key as an environment variable name.
+    This is purely a convenience trade-off to cover common cases well, vs. introducing
+    ambiguities (e.g. did the final '_' come from '.', or '-' or is original?).
+
+    1/27/2023(jackie):
+
+    We start with few rules and should *sparingly* add more over time.
+    Also, it's TBD whether all possible providers will share the same sanitization logic.
+    Therefore we will keep this function private for now
+    """
+    return key.replace("-", "_").replace(".", "_")
+
+
+class AwsSecretsManagerSecretsProvider(SecretsProvider):
+    TYPE = "aws-secrets-manager"
+
+    def get_secret_as_dict(self, secret_id, options={}):
+        """
+        Reads a secret from AWS Secrets Manager and returns it as a dictionary of environment variables.
+
+        The secret payload from AWS is EITHER a string OR a binary blob.
+
+        If the secret contains a string payload ("SecretString"):
+        - if the `parse_secret_string_as_json` option is True (default):
+            {SecretString} will be parsed as a JSON. If successfully parsed, AND the JSON contains a
+            top-level object, each entry K/V in the object will also be converted to an entry in the result. V will
+            always be casted to a string (if not already a string).
+        - If `parse_secret_string_as_json` option is False:
+            {SecretString} will be returned as a single entry in the result, with the key being the secret_id.
+
+        Otherwise, the secret contains a binary blob payload ("SecretBinary"). In this case
+        - The result dic contains '{SecretName}': '{SecretBinary}', where {SecretBinary} is a base64-encoded string
+
+        All keys in the result are sanitized to be more valid environment variable names. This is done on a best effort
+        basis. Further validation is expected to be done by the invoking @secrets decorator itself.
+
+        :param secret_id: ARN or friendly name of the secret
+        :param options: unused
+        :return: dict of environment variables. All keys and values are strings.
+        """
+        import botocore
+        from metaflow.plugins.aws.aws_client import get_aws_client
+
+        effective_aws_region = None
+        # arn:aws:secretsmanager:<Region>:<AccountId>:secret:SecretName-6RandomCharacters
+        m = re.match("arn:aws:secretsmanager:([^:]+):", secret_id)
+        if m:
+            effective_aws_region = m.group(1)
+        elif "region" in options:
+            effective_aws_region = options["region"]
+        else:
+            effective_aws_region = AWS_SECRETS_MANAGER_DEFAULT_REGION
+        # At the end of all that, `effective_aws_region` may still be None.
+        # This might still be OK, if there is fallback AWS region info in environment like:
+        # .aws/config or AWS_REGION env var or AWS_DEFAULT_REGION env var, etc.
+        try:
+            secrets_manager_client = get_aws_client(
+                "secretsmanager", client_params={"region_name": effective_aws_region}
+            )
+        except botocore.exceptions.NoRegionError:
+            # We try our best with a nice error message.
+            # When run in Kubernetes or Argo Workflows, the traceback is still monstrous.
+            # TODO: Find a way to show a concise error in logs
+            raise MetaflowException(
+                "Default region is not specified for AWS Secrets Manager. Please set METAFLOW_AWS_SECRETS_MANAGER_DEFAULT_REGION"
+            )
+        result = {}
+
+        def _sanitize_and_add_entry_to_result(k, v):
+            # Two jobs - sanitize, and check for dupes
+            sanitized_k = _sanitize_key_as_env_var(k)
+            if sanitized_k in result:
+                raise MetaflowAWSSecretsManagerDuplicateKey(
+                    "Duplicate key in secret: '%s' (sanitizes to '%s')"
+                    % (k, sanitized_k)
+                )
+            result[sanitized_k] = v
+
+        """
+        These are the exceptions that can be raised by the AWS SDK:
+        
+        SecretsManager.Client.exceptions.ResourceNotFoundException
+        SecretsManager.Client.exceptions.InvalidParameterException
+        SecretsManager.Client.exceptions.InvalidRequestException
+        SecretsManager.Client.exceptions.DecryptionFailure
+        SecretsManager.Client.exceptions.InternalServiceError
+        
+        Looks pretty informative already, so we won't catch here directly.
+        
+        1/27/2023(jackie) - We will evolve this over time as we learn more.
+        """
+        response = secrets_manager_client.get_secret_value(SecretId=secret_id)
+        if "Name" not in response:
+            raise MetaflowAWSSecretsManagerBadResponse(
+                "Secret 'Name' is missing in response"
+            )
+        secret_name = response["Name"]
+        if "SecretString" in response:
+            secret_str = response["SecretString"]
+            if options.get("json", True):
+                try:
+                    obj = json.loads(secret_str)
+                    if type(obj) == dict:
+                        for k, v in obj.items():
+                            # We try to make it work here - cast to string always
+                            _sanitize_and_add_entry_to_result(k, str(v))
+                    else:
+                        raise MetaflowAWSSecretsManagerNotJSONObject(
+                            "Secret string is a JSON, but not an object (dict-like) - actual type %s."
+                            % type(obj)
+                        )
+                except JSONDecodeError:
+                    raise MetaflowAWSSecretsManagerJSONParseError(
+                        "Secret string could not be parsed as JSON"
+                    )
+            else:
+                _sanitize_and_add_entry_to_result(secret_name, secret_str)
+
+        elif "SecretBinary" in response:
+            # boto3 docs say response gives base64 encoded, but it's wrong.
+            # See https://github.com/boto/boto3/issues/2735
+            # In reality, we get raw bytes.  We will encode it ourselves to become env var ready.
+            # Note env vars values may not contain null bytes.... therefore we cannot leave it as
+            # bytes.
+            #
+            # The trailing decode gives us a final UTF-8 string.
+            _sanitize_and_add_entry_to_result(
+                secret_name, base64.b64encode(response["SecretBinary"]).decode()
+            )
+        else:
+            raise MetaflowAWSSecretsManagerBadResponse(
+                "Secret response is missing both 'SecretString' and 'SecretBinary'"
+            )
+        return result
--- a/metaflow/plugins/secrets/__init__.py
+++ b/metaflow/plugins/secrets/__init__.py
@@ -1,0 +1,10 @@
+import abc
+from typing import Dict
+
+
+class SecretsProvider(abc.ABC):
+    TYPE = None
+
+    @abc.abstractmethod
+    def get_secret_as_dict(self, secret_id, options={}) -> Dict[str, str]:
+        """Retrieve the secret from secrets backend, and return a env var"""
--- a/metaflow/plugins/secrets/__init__.py
+++ b/metaflow/plugins/secrets/__init__.py
@@ -7,4 +7,5 @@ class SecretsProvider(abc.ABC):
 
     @abc.abstractmethod
     def get_secret_as_dict(self, secret_id, options={}) -> Dict[str, str]:
-        """Retrieve the secret from secrets backend, and return a env var"""
+        """Retrieve the secret from secrets backend, and return a dictionary of
+        environment variables."""
--- a/metaflow/plugins/secrets/secrets_decorator.py
+++ b/metaflow/plugins/secrets/secrets_decorator.py
@@ -1,0 +1,234 @@
+import os
+import re
+
+from metaflow.exception import MetaflowException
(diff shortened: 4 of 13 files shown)


metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -336,6 +336,7 @@ def make_flow(
 
     # Attach @kubernetes and @environment decorator to the flow to
     # ensure that the related decorator hooks are invoked.
+    # TODO why is this needed?  Do I need to add secrets deco to Kubernetes as well?
```
savingoyal, line 339: This is needed so that we have the default (if needed) decorator values initialized and ready to be attached to the yaml spec for argo workflows.

metaflow/plugins/secrets/secrets_decorator.py
```diff
@@ -0,0 +1,197 @@
...
+        return SecretSpec(secrets_backend_type, secret_id=secret_id, options=options)
+
+
+def validate_env_vars_across_secrets(all_secrets_env_vars):
+    vars_injected_by = defaultdict(str)
+    for secret_spec, env_vars in all_secrets_env_vars:
+        for k, v in env_vars.items():
+            if k in vars_injected_by:
+                raise MetaflowException(
+                    "Secret '%s' will inject '%s' as env var, but it also by '%s'"
```
savingoyal, line 98: minor language nit - `it is also added by ...`

metaflow/tutorials/00-helloworld/helloworld.py
```diff
@@ -19,13 +19,43 @@ def start(self):
         print("HelloFlow is starting.")
         self.next(self.hello)
 
+    @secrets(
```
savingoyal, line 22: we can scrub this from the tutorial

metaflow/plugins/secrets/secrets_decorator.py
```diff
@@ -0,0 +1,197 @@
...
+    Parameters
+    ----------
+    vars : List[Union[str,dict]]
+        List of secret specs, defining how the secrets are to be retrieved
+    """
+
+    name = "secrets"
+    defaults = {"vars": []}
+
+    def runtime_step_cli(
```
savingoyal, line 146: shouldn't this be `task_pre_step` so that secret access happens within the task execution context. Currently secrets are accessed before the task is launched.

metaflow/plugins/secrets/aws_secrets_manager_secrets_provider.py
```diff
@@ -0,0 +1,165 @@
...
+        try:
+            secrets_manager_client = get_aws_client(
+                "secretsmanager", client_params={"region_name": effective_aws_region}
+            )
+        except botocore.exceptions.NoRegionError:
+            # We try our best with a nice error message.
+            # When run in Kubernetes or Argo Workflows, the traceback is still monstrous.
+            # TODO: Find a way to show a concise error in logs
+            raise MetaflowException(
+                "AWS region not specified. Please specify region in METAFLOW_AWS_SECRETS_MANAGER_DEFAULT_REGION"
```
savingoyal, line 96: minor nit - `Default AWS region is not specified for AWS Secrets Manager...`

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -800,6 +802,12 @@ def _container_templates(self):
             # support Metaflow sandboxes
             env["METAFLOW_INIT_SCRIPT"] = KUBERNETES_SANDBOX_INIT_SCRIPT
 
+            # secrets stuff
+            env["METAFLOW_DEFAULT_SECRETS_BACKEND_TYPE"] = DEFAULT_SECRETS_BACKEND_TYPE
+            env[
+                "METAFLOW_AWS_SECRETS_MANAGER_DEFAULT_REGION"
+            ] = AWS_SECRETS_MANAGER_DEFAULT_REGION
```
savingoyal, line 809: We would need a similar change for Airflow too

jackie-ob: Done!  @valayDave  would you be able to lend a hand in testing this on Airflow?  It should be a case of just setting a single @secrets and running this on oleg2.

metaflow/plugins/secrets/secrets_decorator.py
```diff
@@ -0,0 +1,228 @@
...
+        """Mainly used for testing... not the same as the input dict in secret_spec_from_dict()!"""
+        return {
+            "secrets_backend_type": self.secrets_backend_type,
+            "secret_id": self.secret_id,
+            "options": self.options,
+        }
+
+    @staticmethod
+    def secret_spec_from_str(secret_spec_str):
+        # "." may be used in secret_id one day (provider specific). HOWEVER, it provides the best UX for
```
jackie-ob, line 57: cc @savingoyal per out convo... I think `.` is the right way to go after all.  Some jotted down the reasoning as a long comment for posterity.

metaflow/plugins/secrets/secrets_decorator.py
```diff
@@ -0,0 +1,228 @@
...
+        retry_count,
+        max_user_code_retries,
+        ubf_context,
+        inputs,
+    ):
+        # List of pairs (secret_spec, env_vars_from_this_spec)
+        all_secrets_env_vars = []
+        secret_specs = []
+
+        for secret_spec_str_or_dict in self.attributes["vars"]:
```
jackie-ob, line 183: I think `@secrets(from=[...])` would have been awesome - though Python doesn't allow this.  Good thing is `vars` matches, `@environment(vars={})` - BUT `@secrets` is doing something a bit different.

We will roll with `@secrets(vars=[])`, unless we come up with something better quickly (`@secrets(of=...)`?).

savingoyal (approved):
lgtm! closes #814

romain-intel (changes requested):
Before merging, we should discuss at least once? It was briefly brought up last week but we didn't really discuss it. Please wait to merge.

metaflow/plugins/argo/argo_workflows.py
valayDave: Changes work with Airflow.

metaflow/plugins/secrets/secrets_decorator.py
```diff
@@ -0,0 +1,228 @@
...
+from metaflow.exception import MetaflowException
+from metaflow.decorators import StepDecorator
+from metaflow.plugins.secrets.aws_secrets_manager_secrets_provider import (
+    AwsSecretsManagerSecretsProvider,
+)
+from metaflow.plugins.secrets.inline_secrets_provider import InlineSecretsProvider
+
+SECRETS_BACKEND_PROVIDERS = {
+    "aws-secrets-manager": AwsSecretsManagerSecretsProvider(),
+    "_inline": InlineSecretsProvider(),
```
valayDave, line 15: We should document this somewhere that the name of the `DEFAULT_SECRETS_BACKEND_TYPE ` is `aws-secrets-manager`. Had to siff through code to dig this out.

jackie-ob: There is no default value for `DEFAULT_SECRETS_BACKEND_TYPE`.  When that config is unset, each secret spec must mention its backend type explicitly.

Your comment is a good reminder to enumerate the possible backend types in certain error messages though.  I will add that.

test/unit/test_secrets_decorator.py
```diff
@@ -0,0 +1,151 @@
+import os
```
romain-intel, line 1: are these run?

metaflow/plugins/secrets/secrets_decorator.py
```diff
@@ -0,0 +1,241 @@
...
+from collections import defaultdict
+
+from metaflow.exception import MetaflowException
+from metaflow.decorators import StepDecorator
+from metaflow.plugins.secrets.aws_secrets_manager_secrets_provider import (
+    AwsSecretsManagerSecretsProvider,
+)
+from metaflow.plugins.secrets.inline_secrets_provider import InlineSecretsProvider
+
+SECRETS_BACKEND_PROVIDERS = {
```
romain-intel, line 13: this is what i am saying to move to __init__.py for plugins. Create a `SECRETS_BACKEND_PROVIDERS_SPEC` and add those in there like:
```
("inline", ".secrets.inline_secrets_provider.InlineSecretsProvider")
...
```
and then add a TYPE (or whatever you want to call it to all the secrets providers and then add it here: https://github.com/Netflix/metaflow/blob/38bcd81800f19478324337c17c836665877bce8e/metaflow/extension_support/plugins.py#L175. 

It seems like a minuscule change that won't impact you much but would make this a lot cleaner and mean that I can develop things without having to do much change.

metaflow/plugins/secrets/aws_secrets_manager_secrets_provider.py
```diff
@@ -0,0 +1,165 @@
+import base64
```
romain-intel, line 1: I didn't review this file. I would just move it. I don't think it belongs in the `secrets` directory.

metaflow/plugins/secrets/__init__.py
```diff
@@ -0,0 +1,14 @@
+import abc
+
+
+class SecretsProvider(abc.ABC):
```
romain-intel, line 4: nit: we have not used this pattern in the code base. am fine with it just pointing out that it is not consistent.

metaflow/plugins/secrets/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+    @abc.abstractmethod
+    def get_secret_as_env_vars(self, secret_id, options={}):
```
romain-intel, line 6: nit: document the return type (Dict[str, str]). Consider renaming the method: `get_secret_as_dict` which is a bit more accurate.

metaflow/plugins/secrets/secrets_decorator.py
```diff
@@ -0,0 +1,241 @@
...
+                raise MetaflowException(
+                    "Secret %s will inject %s as env var, but it already exists in env"
+                    % (secret_spec.secret_id, k)
+                )
+
+
+def validate_env_vars(env_vars):
+    for k, v in env_vars.items():
+        if type(k) != str:
+            raise ValueError("Found non-str key %s (%s)" % (str(k), type(k)))
```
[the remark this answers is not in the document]
romain-intel: nvm; I see you convert later. We could maybe just raise here too.

metaflow/plugins/secrets/secrets_decorator.py
```diff
@@ -0,0 +1,241 @@
...
+        max_user_code_retries,
+        ubf_context,
+        inputs,
+    ):
+        # List of pairs (secret_spec, env_vars_from_this_spec)
+        all_secrets_env_vars = []
+        secret_specs = []
+
+        for secret_spec_str_or_dict in self.attributes["vars"]:
+            if type(secret_spec_str_or_dict) == str:
```
romain-intel, line 197: for all these, consider `isinstance`

metaflow/plugins/secrets/secrets_decorator.py
```diff
@@ -0,0 +1,241 @@
...
+
+        validate_env_vars_across_secrets(all_secrets_env_vars)
+        validate_env_vars_vs_existing_env(all_secrets_env_vars)
+
+        # By this point
+        # all_secrets_env_vars contains a list of dictionaries... env maps.
+        # - env maps must be disjoint from each other
+        # - env maps must be disjoint from existing current process os.environ
+
+        for secrets_anv_vars in all_secrets_env_vars:
```
romain-intel, line 239: nit: secrets_env_vars

romain-intel (changes requested):
First quick pass. I'll do another round tomorrow and still thinking of names.

metaflow/plugins/secrets/secrets_decorator.py
jackie-ob: That's fair.  To the uninitiated, (like me) - not quite minuscule (may be just small)... but it's done!  Please help take a look to make sure I did everything right re: plugin / extension format during your second pass.

metaflow/plugins/secrets/__init__.py
romain-intel: oh ya. Absolutely not with compatibility just with code style. I am fine either way. Just mentioning it. I try to stay within the "feel" of the code but there is an argument to be made that we can evolve it to make it better.

metaflow/plugins/secrets/__init__.py
```diff
@@ -1,14 +1,10 @@
 import abc
+from typing import Dict
 
 
 class SecretsProvider(abc.ABC):
+    TYPE = None
+
     @abc.abstractmethod
-    def get_secret_as_env_vars(self, secret_id, options={}):
```
romain-intel: Am fine with the typing but was mostly just asking to put it in the comment like the rest so that it was clear what was expected.

test/unit/test_secrets_decorator.py
romain-intel: absolutely. Internally we also trigger some unit tests so was wondering if you had a similar functionality as well. It's fine if it is later. Agree that unit tests are useful for dev at least.

metaflow/plugins/secrets/secrets_decorator.py
```diff
@@ -0,0 +1,234 @@
...
+            secret_id = parts[1]
+        return SecretSpec(secrets_backend_type, secret_id=secret_id)
+
+    @staticmethod
+    def secret_spec_from_dict(secret_spec_dict):
+        if "type" not in secret_spec_dict:
+            secrets_backend_type = get_default_secrets_backend_type()
+        else:
+            secrets_backend_type = secret_spec_dict["type"]
+            if type(secrets_backend_type) != str:
```
romain-intel, line 81: nit: isinstance; and below too.

metaflow/plugins/secrets/secrets_decorator.py
```diff
@@ -0,0 +1,234 @@
...
+                "Bad @secrets specification - 'option' must be a dict - found %s"
+                % type(options)
+            )
+        return SecretSpec(secrets_backend_type, secret_id=secret_id, options=options)
+
+
+def validate_env_vars_across_secrets(all_secrets_env_vars):
+    vars_injected_by = {}
+    for secret_spec, env_vars in all_secrets_env_vars:
+        for k in env_vars:
```
romain-intel, line 104: I don't really mind either way but just thinking that maybe we could give all the errors at once. For this function and the one below. For this one, build vars_inject_by and then check if any len is > 1 and report all those (or something like that). For the one below, something like a set intersection. The only reason would be to give as much information to the user as possible in case of a conflict so that they don't have to iterate over and over. But it's a nit so fine either ways.

metaflow/plugins/secrets/secrets_decorator.py
```diff
@@ -0,0 +1,234 @@
...
+    Parameters
+    ----------
+    vars : List[Union[str,dict]]
+        List of secret specs, defining how the secrets are to be retrieved
+    """
+
+    name = "secrets"
+    defaults = {"vars": []}
+
+    def task_pre_step(
```
romain-intel, line 172: something for the docs. This may be obvious but I have seen ppl make this mistake in other contexts: anything that needs to use the secrets (including importing libraries that need the secrets) needs to happen *inside* the step for this to work properly. It's fairly obvious when you understand the mechanism but may not be obvious when users are writing the code :).

metaflow/plugins/aws/secrets_manager/aws_secrets_manager_secrets_provider.py
```diff
@@ -0,0 +1,167 @@
...
+    pass
+
+
+class MetaflowAWSSecretsManagerDuplicateKey(MetaflowException):
+    """Raised when the response from AWS Secrets Manager contains duplicate keys"""
+
+    pass
+
+
+class MetaflowAWSSecretsManagerJSONParseError(MetaflowException):
```
romain-intel, line 24: I am surprised you don't need "pass" here and below. May be safer to add though.

test/unit/test_secrets_decorator.py
jackie-ob: We do have some OB-side unit test triggers (manually), but only for the azure / gcp storage stuff (unit tests that actually exercise real cloud APIs).

I would advocate for running OSS unit tests like these automatically (add another GH actions workflow for it...).  Won't be part of this PR though.

metaflow/plugins/secrets/secrets_decorator.py
jackie-ob: cc @tuulos: when we eventually document `@secrets` in public docs.

metaflow/plugins/secrets/secrets_decorator.py
```diff
@@ -0,0 +1,238 @@
...
+
+DISALLOWED_SECRETS_ENV_VAR_PREFIXES = ["METAFLOW_"]
+
+
+def get_default_secrets_backend_type():
+    from metaflow.metaflow_config import DEFAULT_SECRETS_BACKEND_TYPE
+
+    if DEFAULT_SECRETS_BACKEND_TYPE is None:
+        raise MetaflowException(
+            "No default secrets backend type configured, but needed by @secrets.  Set METAFLOW_DEFAULT_SECRETS_BACKEND_TYPE."
```
romain-intel, line 16: nit: long line.

metaflow/plugins/secrets/secrets_decorator.py
romain-intel: ok, so I would maybe use something like this instead of `vars`:
  - `sources`: I actually kind of like that one -- it indicates where the secrets are coming from through either a path (the "." separated one) or through a descriptor block (the dictionary)
  - `secrets`: that one may be a bit weird and doing `@secrets(secrets=[])` may look weird
  - we could not name it anything at all and just have something like `@secrets('secret1', 'secret2')`. This would require changing a few things in the way we process decorators but mentioning as a possibility. It would solve this whole thing and possibly be cleaner. The change shouldn't be that much -- we just need to check if the first argument of the decorator is a function and if not, do the same thing that we do when we just have kwargs. Then, we package all the args into a _args attribute in the dictionary and tada, we are off. This change would be fully backward compatible and would just enable decorators to now just take arguments as opposed to just kwargs.

You could alternatively also use something like:
  - `names` for simple string secret names and `descriptors` for the dictionary descriptors to split them out. That may be a bit more complicated and doesn't really buy much but throwing it out there.

romain-intel (commented):
Looks good. The only concern remaining is the naming. I made a few suggestions. Let me know what you think.

metaflow/plugins/secrets/__init__.py
```diff
@@ -0,0 +1,10 @@
+import abc
+from typing import Dict
+
+
+class SecretsProvider(abc.ABC):
+    TYPE = None
+
+    @abc.abstractmethod
+    def get_secret_as_dict(self, secret_id, options={}) -> Dict[str, str]:
+        """Retrieve the secret from secrets backend, and return a env var"""
```
romain-intel, line 10: nit: return a dict

metaflow/plugins/secrets/secrets_decorator.py
romain-intel: nit: also two spaces after .

metaflow/plugins/secrets/secrets_decorator.py
```diff
@@ -0,0 +1,238 @@
...
+                )
+            vars_injected_by[k] = secret_spec
+
+
+def validate_env_vars_vs_existing_env(all_secrets_env_vars):
+    for secret_spec, env_vars in all_secrets_env_vars:
+        for k in env_vars:
+            if k in os.environ:
+                raise MetaflowException(
+                    "Secret %s will inject %s as env var, but it already exists in env"
```
romain-intel, line 119: nit: consistency between '%s' and %s

romain-intel (approved):
minor nits but nothing blocking. Am fine with merging. I did not run internal tests since I am out of date slightly with the multiple flow decos.

romain-intel (approved):

Issue #355, opened 2020-10-20
Allow flows to be defined any Python modules


tuulos:
Currently flow scripts must reside at the top of the directory hierarchy which makes it impossible to have package structures like

```
mypkg/flows/flow1.py
mypkg/flows/flow2.py
mypkg/util/common.py
```

We could relax the top-level requirement. Context https://gitter.im/metaflow_org/community?at=5f8e2eb96c8d484be29647e0

derfred:
I have a similar requirement. Looking through the code I wonder if the following approach would be viable:

The list of files to be put into the package is defined in the class `package.MetaflowPackage`. A quick way to customize this selection for a particular flow would be to allow it to define a custom class for this purpose. Something like this:

```
class CustomPackageFlow(FlowSpec):
  def package_class(self):
    return MyCustomPackage

  @step
  def start(self):
    self.next(self.do_something)
```

Then in the metaflow code, like for example https://github.com/Netflix/metaflow/blob/master/metaflow/cli.py#L886, this would be used like this:
```
print(f"Packaging {obj} {obj.flow}")
klass = getattr(obj.flow, "package_class", lambda: MetaflowPackage)()
obj.package = klass(
        obj.flow, obj.environment, obj.echo, obj.package_suffixes
)
```

What do you think?

Issue #124, opened 2020-02-12
Package code in more cases


romain-intel:
Currently, the code is packaged for Batch runs only. To be able to fully replicate and version things, it would be nicer if code was always stored as an artifact of each step.

benjaminbluhm:
We are also missing this feature. We regularly make runs in local mode and would like to store our code as an artifact. Are there any plans to address this feature?

Kylecrif:
I mentioned in the community conversation (https://gitter.im/metaflow_org/community) that it would be super valuable to be able to rerun a previously successful `Run`, but with different `Parameter`(s) or different input data. Basically the `resume` command but on successful runs with the ability to change `Parameter` values

@tuulos mentioned that there is effort on providing a top-level command for bootstrapping a project based on an existing run. This would be very useful for us 

Our particular use case right now for `metaflow` is to use it in creating datasets with proprietary feature engineering, and also have the ability to "replay" that same data set with additional/different input data points at any time in the future. Since our feature engineering logic may change over time, its important we can reproduce the same features from a previous run with new data

tuulos:
another +1 here https://gitter.im/metaflow_org/community?at=6075e54e06e2e024e84f645c

sappier:
+1.  Teams we're speaking with want to pack all code into the docker images and avoid downloading it from S3.  "Everything in docker images" is aligned with corp. rules (central vulnerabilities scanning, licenses check, etc.) and becoming a problem otherwise.

mikejmills:
another +1 here.

bsridatta:
Are there any subtask tickets to help with this?

rileyhun:
+1 from me as well

yudhiesh:
+1 from me as well

jixianyihao:
this issue was too long，did teams give a decision about this issue?

GunaPalanivel:
Hey, I'd like to take a shot at this!

I traced through the codebase and the gap is pretty clear — `MetaflowPackage._package_and_upload()` in `metaflow/package/__init__.py` already builds the tar archive for all backends, but only uploads it when `flow_datastore` is passed in. For local/k8s/Argo runs that path is skipped and `blob_url`/`blob_sha` are set to empty strings, so `run.code` silently returns nothing.

My plan:
- Ensure `MetaflowPackage` always creates (and stores) the snapshot regardless of backend
- Persist it as a reserved artifact (e.g. `_code_package`) from `task.py`, similarly to how Batch handles it today
[...]

PR #1260, opened 2023-02-03
Allow a flow to customize the content of the package

derfred:
This PR is intended to address #355 and to some extend #124.

We run Metaflow jobs as part of our CI/CD on a k8s cluster which requires more control of the version of the dependencies to be bundled in the package. This PR adds a `@packager` flow level decorator to define the content of the package.

The decorator takes a `content=` parameter which can either be a list-like object or a dict. The following example illustrates the function for a list-like object:
```
@packager(content=["subdirectory", "../peer-level-directory", "dataset.txt"])
```

- the directory `subdirectory` is walked and all files matching the default suffixes are placed at `subdirectory`
- the same applies to the directory `../peer-level-directory` is placed at `peer-level-directory`
- if a file is specified directly it is placed in the appropriate directory (at the top-level in this case) regardless of suffix

Using a dict as an argument this can be customized as follows:
```
@packager(content={ "../peer-level-directory": "package1", "subdirectory": "package2" })
```

### DIFF
--- a/metaflow/package.py
+++ b/metaflow/package.py
@@ -132,9 +132,16 @@ class MetaflowPackage(object):
         for path_tuple in self.environment.add_to_package():
             yield path_tuple
 
+        flowdir = os.path.dirname(os.path.abspath(sys.argv[0])) + "/"
         if "packager" in self._flow._flow_decorators:
-            for path_tuple in self._flow._flow_decorators["packager"][0].path_tuples():
-                yield path_tuple
+            packager = self._flow._flow_decorators["packager"][0]
+            for suffixes, source, target in packager.path_tuples(flowdir):
+                if suffixes is None:
+                    yield source, target
+                else:
+                    for fname, tname in self._walk(source, suffixes=suffixes):
+                        yield fname, os.path.join(target, os.path.relpath(tname, source))
+
             flow_file = os.path.abspath(sys.argv[0])
             yield flow_file, os.path.basename(flow_file)
         elif R.use_r():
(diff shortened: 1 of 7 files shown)


romain-intel:
Hello.

Thanks for the contribution. I think it would be nice to expand what is packaged but I am a little confused as to the way you accomplish this here. In particular, this seems to allow you to change the relative paths of the files so that on your local machine you may have:
```
parent/util/common.py
parent/flows/flow.py
```
and you would possibly have in the package:
```
flow.py
util/common.py
```

Or maybe I am missing something.

A possibly simpler solution (but definitely less flexible way but maybe enough) to doing this would be allow the packaging of a *parent* of the flow (instead of packaging stuff in the directory of the flow) or maybe the current working directory (provided it is a parent). It would require some changes to the way we call the script but may solve most issues in a way that is possibly more natural for the user.

What do you think?

derfred:
Hi @romain-intel,

yes you are right that this PR allows you to change the relative paths between the files. In my case the layout of our repo is:
```
library1/library1/util.py
library2/library2/util.py
application/flows/my_flow.py
```
We then package them to this layout:
```
library1/util.py
library2/util.py
my_flow.py
```

In production the libraries are installed using `pip` and in the flow they can be imported using `import library1`. So for our use-case this is a natural, if somewhat hackish, solution.

I do see your point however that this could be confusing to the user. I also like your idea of packaging the either the parent of the flow, or the parent of the current directory.

I'll look into that!

derfred:
Hi @romain-intel,

I have taken another shot at the packaging question based on your points. Would you mind looking at #1270?

For our use-case both implementations work well. I think the behaviour is more intuitive in this new PR, however the implementation touches more parts of the codebase. What do you think?

PR #1263, opened 2023-02-07
Expose AWS instance metadata for @kubernetes

savingoyal:
Introduces optional metadata for tasks that execute on Kubernetes on EKS

### DIFF
--- a/metaflow/plugins/kubernetes/kubernetes_decorator.py
+++ b/metaflow/plugins/kubernetes/kubernetes_decorator.py
@@ -1,6 +1,7 @@
 import json
 import os
 import platform
+import requests
 import sys
 
 from metaflow.decorators import StepDecorator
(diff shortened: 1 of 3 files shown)


metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -329,6 +330,23 @@ def task_pre_step(
...
+
+            # Capture AWS instance identity metadata. This is best-effort only since
+            # access to this end-point might be blocked on AWS and not available
+            # for non-AWS deployments.
+            # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html
+            # TODO (savin): Introduce equivalent support for Microsoft Azure and
+            #               Google Cloud Platform
+            try:
+                instance_meta = requests.get(
+                    url="http://169.254.169.254/latest/dynamic/instance-identity/document"
```
obgibson, line 342: Is this IP significant?

savingoyal: Yes, it's a magic IP that hosts instance metadata for AWS, Azure and GCP

obgibson:
To prevent uninformed questions like this, I suggest assigned the IP to a variable called magic_ip_that_hosts_metadata_for_aws_gcp_azure (is that pythonic?)

savingoyal:
That's fair - I can push a different PR that addresses this comment.

PR #1267, opened 2023-02-08
New MF configs for Argo Workflows

jackie-ob:
These new variables offer paths to fine tune Argo Workflows workloads.  E.g. use `ARGO_WORKFLOWS_KUBERNETES_SECRETS` to source in some env vars, instead of propagating them from local workstation where `argo-workflows create` is run.

- `ARGO_WORKFLOWS_ENV_VARS_TO_SKIP`. CSV format, default to empty str
- `ARGO_WORKFLOWS_KUBERNETES_SECRETS`. Set secrets exclusive to Argo Workflows orchestrated runs


metaflow/plugins/argo/argo_workflows.py
```diff
@@ -858,6 +860,15 @@ def _container_templates(self):
             env_without_none_values = {k: v for k, v in env.items() if v is not None}
             del env
 
+            # Now lets skip some env vars (e.g. in case we want to pull them from KUBERNETES_SECRETS)
```
savingoyal, line 863: minor nit 0 can we collapse L:860-871?
```
env = {
  k: v
  for k, v in env.items()
  if v is not none and if k not in env_vars_to_skip
}
```

savingoyal (approved):

PR #1271, opened 2023-02-13
Introduce support for event-triggered workflows

savingoyal:
Items remaining
- [x] auto-emit events from flows
- [x] record execution metadata
- [x] expose client
- [x] polish up UX
- [x] support flow dependency
- [ ] support NATS
- [x] support event spoofing
- [x] support mapping context and default maps
- [x] support finer grained @project attributes
- [x] doc strings & types

draft docs - https://github.com/Netflix/metaflow-docs/pull/86/files


romain-intel:
Let me know when this is ready for review. Happy to look and provide feedback whenever you want (I am out on the week of the 19th so responses that week may be delayed).

savingoyal:
> Let me know when this is ready for review. Happy to look and provide feedback whenever you want (I am out on the week of the 19th so responses that week may be delayed).

There shouldn't be much to review but I will give you a heads up.

romain-intel:
Sounds good. And yes, I will probably defer to you for any and all argo stuff :). It's just the common decorator(s)/parameter changes that I will review so yes, shouldn't take long. If you get that part done first, I can try to implement internally and see if I hit any snags as well (so we have a bit more confidence that we are all good).

metaflow/plugins/argo/argo_events_decorator.py
```diff
@@ -0,0 +1,173 @@
...
+        "event": None,
+        "events": [],
+        "options": {},  # TODO: introduce support for options
+    }
+
+    def flow_init(
+        self, flow, graph, environment, flow_datastore, metadata, logger, echo, options
+    ):
+        self.triggers = []
+        if sum(map(bool, (self.attributes["event"], self.attributes["events"]))) > 1:
```
romain-intel, line 23: I am still very very unclear as to why we want to have both, I don't think the syntax `events=[myevent]` is much worse than `event=myevent`. If you really just want something like `events=myevent`, that can be done too but we have never had this pattern before and I am not sure of the value of it. Am I missing something?

metaflow/plugins/argo/argo_events_decorator.py
```diff
@@ -0,0 +1,173 @@
...
+            raise MetaflowException(
+                "Specify only one of *event* or *events* "
+                "attributes in *@trigger* decorator."
+            )
+        elif self.attributes["event"]:
+            # event attribute supports the following formats -
+            #     1. event='table.prod_db.members'
+            #     2. event={'name': 'table.prod_db.members',
+            #               'parameters': {'alpha': 'member_weight'}}
```
romain-intel, line 32: what is the meaning of the dict syntax? Is this for the mapping of parameters?

metaflow/plugins/argo/argo_events_decorator.py
```diff
@@ -0,0 +1,173 @@
...
+            else:
+                raise MetaflowException(
+                    "Incorrect format for *event* attribute in *@trigger* decorator. "
+                    "Supported formats are string and dictionary - \n"
+                    "@trigger(event='foo') or @trigger(event={'name': 'foo', "
+                    "'parameters': {'alpha': 'beta'}})"
+                )
+        elif self.attributes["events"]:
+            # events attribute supports the following formats -
+            #     1. events='table.prod_db.members AND table.prod_db.metadata'
```
romain-intel, line 51: I think we have previously talked about not wanting to do some complex syntax like this? I actually liked what I thought we had talked about last time:
  - things within `events` are "AND"
  - you can have multiple decorators to do "OR"

romain-intel (changes requested):
I have not reviewed any of the airflow/argo/step_functions code. I also need to review more the decorators (wrong computer right now to review based on what we currently have). This is therefore an initial high-level review. Happy to talk more.

metaflow/plugins/argo/argo_events_decorator.py
savingoyal: it is common for a flow to depend on a singular event. As you can see from the implementation, it is just syntactic sugar and quite a bit helpful for the majority of use cases that we will run into.

metaflow/plugins/__init__.py
```diff
@@ -56,6 +56,8 @@
     ("conda_base", ".conda.conda_flow_decorator.CondaFlowDecorator"),
     ("schedule", ".aws.step_functions.schedule_decorator.ScheduleDecorator"),
     ("project", ".project_decorator.ProjectDecorator"),
+    ("trigger", ".argo.argo_events_decorator.ArgoEventsDecorator"),
```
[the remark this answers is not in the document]
savingoyal: Yes, I agree, but I would want to move them elsewhere when we have another use for them. There is already a comment in ArgoEventsDecorator noting this fact. The internal movement of files has no impact on user-visible UX.

metaflow/plugins/__init__.py
```diff
@@ -137,28 +139,30 @@ def get_plugin_cli():
 AWS_CLIENT_PROVIDERS = resolve_plugins("aws_client_provider")
 SECRETS_PROVIDERS = resolve_plugins("secrets_provider")
 
+from .argo.argo_events import ArgoEvent
```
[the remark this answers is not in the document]
savingoyal: I am iterating on the ArgoEvent UX, but this would be exposed to the user so that they can directly invoke the client to emit events either from within a flow or a different Python process.

metaflow/events.py
```diff
@@ -0,0 +1,138 @@
...
+        elif self.type in ("EVENT", "MANY-EVENTS", "ONE-OF-EVENTS"):
+            for event in self.events:
+                if event.name == key:
+                    return event
+        else:
+            return None
+
+
+# TODO: can we do away with this?
+class DataArtifactProxy(Mapping):
```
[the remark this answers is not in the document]
savingoyal: It's not quite `[artifact]`. There is no explicit `.data` call from the user.

metaflow/events.py
```diff
@@ -0,0 +1,138 @@
+from collections import namedtuple
+from collections.abc import Mapping
+from datetime import datetime
+from itertools import groupby
+
+MetaflowEvent = namedtuple("MetaflowEvent", ["name", "id", "timestamp", "type"])
+
+
+class MetaflowTrigger(object):
```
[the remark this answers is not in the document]
savingoyal: Being triggered by a task is something that can be introduced easily later. At this point, I have focused on a minimum set of functionality that doesn't close any doors. This proposed doc site for events will throw more color at the implementation choices here - https://github.com/Netflix/metaflow-docs/pull/86/files

metaflow/events.py
```diff
@@ -0,0 +1,138 @@
...
+    def runs(self):
+        """
+        If the triggering events correspond to Metaflow runs, returns a list of `Run` objects. Otherwise returns `None`.
+        """
+        # to avoid circular import
+        from metaflow import Run
+
+        return [
+            Run(
+                obj["id"][: obj["id"].index("/", obj["id"].index("/") + 1)],
```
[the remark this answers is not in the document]
savingoyal: will do

metaflow/events.py
```diff
@@ -0,0 +1,138 @@
...
+                _namespace_check=False,
```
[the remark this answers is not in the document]
savingoyal: `namespace(None)` changes the namespace for any user space calls as well, which is something we would want to avoid.

metaflow/events.py
```diff
@@ -0,0 +1,138 @@
...
+from collections.abc import Mapping
+from datetime import datetime
+from itertools import groupby
+
+MetaflowEvent = namedtuple("MetaflowEvent", ["name", "id", "timestamp", "type"])
+
+
+class MetaflowTrigger(object):
+    def __init__(self, _meta=[]):
+        self._meta = _meta
```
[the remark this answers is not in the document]
savingoyal: agreed - I will address this in one of the later commits to this PR.

metaflow/events.py
```diff
@@ -0,0 +1,138 @@
...
+        if _counts.get("event"):
+            if _counts.get("event") > 1:
+                return "MANY-EVENTS"
+            return "EVENT"
+
+    def __getitem__(self, key):
+        """
+        If triggering events are runs, `key` corresponds to the flow name of the triggering run. Returns a triggering `Run` object corresponding to the key. If triggering events are not runs, `key` corresponds to the event name and a `MetaflowEvent` object is returned.
+        """
+        if self.type in ("RUN", "MANY-RUNS", "ONE-OF-RUNS"):
```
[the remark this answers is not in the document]
savingoyal: to support the "OR" use case given that this is in the client (easy to ship the decorator-focused changes compared to client-focused changes). However, given that we don't have support for "OR" today, I am happy to yank this out.

metaflow/events.py
```diff
@@ -0,0 +1,138 @@
...
+        # to avoid circular import
+        from metaflow import Run
+
+        return [
+            Run(
+                obj["id"][: obj["id"].index("/", obj["id"].index("/") + 1)],
+                _namespace_check=False,
+            )
+            for obj in filter(
+                lambda x: x.get("type") in ["run"],
```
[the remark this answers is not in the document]
savingoyal: yeah - an artifact from when I had multiple elements in the list. I can change it but either approaches shouldn't be any different(?)

metaflow/client/core.py
```diff
@@ -1868,6 +1867,13 @@ def replace_tags(self, tags_to_remove: Iterable[str], tags_to_add: Iterable[str]
         self._user_tags = frozenset(final_user_tags)
         self._tags = frozenset([*self._user_tags, *self._system_tags])
 
+    @property
+    def trigger(self):
```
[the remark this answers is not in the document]
savingoyal: will add closer to merging the PR to make sure that public doc strings reflect the correct information

metaflow/client/core.py
```diff
@@ -338,7 +333,8 @@ def __iter__(self) -> Iterable["MetaflowObject"]:
             Iterator over all children
         """
         query_filter = {}
-        if current_namespace:
+        # skip namespace filtering if _namespace_check is unset.
+        if self._namespace_check and current_namespace:
```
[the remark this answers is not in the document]
savingoyal: good catch - I will change the comment to reflect that. This change is to ensure that MetaflowTrigger.run shows up within flows in Argo

metaflow/client/core.py
```diff
@@ -444,7 +440,10 @@ def __getitem__(self, id: str) -> "MetaflowObject":
         obj = self._get_child(id)
         if obj:
             return _CLASSES[self._CHILD_CLASS](
-                attempt=self._attempt, _object=obj, _parent=self
+                attempt=self._attempt,
+                _object=obj,
+                _parent=self,
+                _namespace_check=self._namespace_check,
```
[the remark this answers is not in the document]
savingoyal: see comment above

metaflow/events.py
```diff
@@ -0,0 +1,138 @@
+from collections import namedtuple
+from collections.abc import Mapping
+from datetime import datetime
+from itertools import groupby
+
+MetaflowEvent = namedtuple("MetaflowEvent", ["name", "id", "timestamp", "type"])
```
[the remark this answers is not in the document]
savingoyal: Will add types and public docs closer to merging the PR

metaflow/events.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+class MetaflowTrigger(object):
+    def __init__(self, _meta=[]):
+        self._meta = _meta
+
+    @property
+    def event(self):
```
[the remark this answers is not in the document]
savingoyal: will be added closer to the finish line

metaflow/events.py
```diff
@@ -0,0 +1,138 @@
...
+        """
+        return self.events[0] if self.events and len(self.events) == 1 else None
+
+    @property
+    def events(self):
+        """
+        Returns a list of `MetaflowEvent` objects correspondings to all the triggering events.
+        """
+        return [
+            MetaflowEvent(
```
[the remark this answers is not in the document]
savingoyal: bunch of accesses?

metaflow/plugins/argo/argo_events_decorator.py
savingoyal: correct. as specified in https://github.com/Netflix/metaflow-docs/pull/86/files

metaflow/plugins/argo/argo_events_decorator.py
savingoyal: that is still the case - things within `events` are `AND`. This is just an explicit flavor of that syntax. Multiple decorators is not supported at this point for this release. Can easily add it as a follow-up when the need presents itself.

metaflow/metaflow_config.py
```diff
@@ -262,6 +262,20 @@
 ARGO_WORKFLOWS_KUBERNETES_SECRETS = from_conf("ARGO_WORKFLOWS_KUBERNETES_SECRETS", "")
 ARGO_WORKFLOWS_ENV_VARS_TO_SKIP = from_conf("ARGO_WORKFLOWS_ENV_VARS_TO_SKIP", "")
 
+##
+# Argo Events Configuration
+##
+ARGO_EVENTS_SERVICE_ACCOUNT = from_conf(
```
savingoyal, line 268: add sensor here

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -1424,3 +1865,253 @@ def to_json(self):
...
+    def __init__(self):
+        tree = lambda: defaultdict(tree)
+        self.payload = tree()
+
+    def src(self, dependency_name, data_key, value):
+        self.payload["src"] = {
+            "dependencyName": dependency_name,
+            "dataKey": data_key,
+            "value": value,
+            "useRawData": True,
```
jackie-ob, line 2105: In my tests... this needed to be set to `False` in order for the payload to render correctly as an env var in the triggered workflow.

Not sure what we are trying to test - if the final sink for this data is always an env var, we should always stringify it (i.e. not raw).

cc @savingoyal

metaflow/client/core.py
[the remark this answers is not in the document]
valayDave: why not `triggered_by` ?

metaflow/plugins/argo/argo_workflows.py
jackie-ob: To add - we can also leave this unset (let it default to False).

metaflow/events.py
```diff
@@ -0,0 +1,83 @@
...
+
+    @property
+    def runs(self):
+        """
+        If the triggering events correspond to Metaflow runs, returns a list of `Run` objects. Otherwise returns `None`.
+        """
+        # to avoid circular import
+        from metaflow import Run
+
+        return [
```
romain-intel, line 59: nit: compute only once -- this is actually not a cheap operation.

metaflow/events.py
```diff
@@ -0,0 +1,83 @@
...
+            Run(
+                obj.id[: obj.id.index("/", obj.id.index("/") + 1)],
+                _namespace_check=False,
+            )
+            for obj in filter(
+                lambda x: x.type in ["run"],
```
romain-intel, line 65: lambda x: x.type == "run"

metaflow/events.py
```diff
@@ -0,0 +1,83 @@
...
+                self._meta.values(),
+            )
+        ] or None
+
+    def __getitem__(self, key):
+        """
+        If triggering events are runs, `key` corresponds to the flow name of the triggering run. Returns a triggering `Run` object corresponding to the key. If triggering events are not runs, `key` corresponds to the event name and a `MetaflowEvent` object is returned.
+        """
+        if self.runs:
```
romain-intel, line 74: for clarity, `return None` should not be in `else` but just there so it can serve as a clear fallback if the name doesn't exist. It's the current behavior but not clear if intended (it is in this case).

metaflow/metaflow_config.py
```diff
@@ -262,6 +262,20 @@
...
+    "ARGO_EVENTS_SERVICE_ACCOUNT", "operate-workflow-sa"
+)
+ARGO_EVENTS_EVENT_BUS = from_conf("ARGO_EVENTS_EVENT_BUS", "default")
+ARGO_EVENTS_EVENT_SOURCE = from_conf("ARGO_EVENTS_EVENT_SOURCE", "metaflow-webhook")
+ARGO_EVENTS_EVENT = from_conf("ARGO_EVENTS_EVENT", "event")
+ARGO_EVENTS_WEBHOOK_URL = from_conf(
+    "ARGO_EVENTS_WEBHOOK_URL", "http://10.10.29.11:12000/event"
```
romain-intel, line 275: nit: is this a standard URL or left over from testing?

metaflow/plugins/__init__.py
romain-intel, line 142: Need to provide the alias but this needs to go before the merge.

metaflow/plugins/argo/argo_events_decorator.py
```diff
@@ -0,0 +1,154 @@
...
+from metaflow.exception import MetaflowException
+from metaflow.util import is_stringish
+
+# TODO: Support dynamic parameter mapping through a context object that exposes
+#       flow name and user name similar to parameter context
+
+
+# TODO: This decorator interface can be a top-level decorator to support similar
+#       implementations for Netflix Maestro, AWS Step Functions, Apache Airflow...
+class ArgoEventsDecorator(FlowDecorator):
```
romain-intel, line 11: Name should be something like `TriggerOnDecorator`. Would also suggest the name `trigger_on` as a decorator name instead of just `trigger`. This would also be consistent with the `TriggerOnFinishDecorator` below (ditto with the name).

metaflow/client/core.py
```diff
@@ -1868,6 +1868,13 @@ def replace_tags(self, tags_to_remove: Iterable[str], tags_to_add: Iterable[str]
         self._user_tags = frozenset(final_user_tags)
         self._tags = frozenset([*self._user_tags, *self._system_tags])
 
+    @property
+    def trigger(self):
```
romain-intel, line 1872: would suggest `triggered_by` as the name here

metaflow/plugins/argo/argo_events_decorator.py
```diff
@@ -0,0 +1,154 @@
...
+        names = [x["name"] for x in self.triggers]
+        if len(names) != len(set(names)):
+            raise MetaflowException(
+                "Duplicate event names defined in *@trigger* decorator."
+            )
+
+
+class TriggerOnFinishDecorator(FlowDecorator):
+    name = "trigger_on_finish"
+    defaults = {
```
romain-intel, line 97: This needs to change to be consistent with way of specifying branch/project.

metaflow/plugins/argo/argo_events_decorator.py
```diff
@@ -0,0 +1,154 @@
...
+        elif self.attributes["flows"]:
+            # flows attribute supports the following formats -
+            #     1. flows=['FooFlow', 'BarFlow']
+            if isinstance(self.attributes["flows"], list):
+                for flow in self.attributes["flows"]:
+                    if is_stringish(flow):
+                        self.triggers.append(
+                            {
+                                "flow": flow,
+                                "project": self.attributes["project"],
```
romain-intel, line 133: This seems incorrect -- it should use the default project/branch. It makes sense to make this change here as that is the behavior across implementations.

metaflow/plugins/argo/argo_events_decorator.py
```diff
@@ -0,0 +1,154 @@
+from metaflow.decorators import FlowDecorator
+from metaflow.exception import MetaflowException
+from metaflow.util import is_stringish
+
+# TODO: Support dynamic parameter mapping through a context object that exposes
```
romain-intel, line 5: As discussed, we should enunciate some of the rules here (the ones we agreed to during the meeting) around parameter mapping.

romain-intel (commented):
More comments but converging :). I'll try to have the alias thing shortly. Also, please add typing and documentation to all public facing APIs and classes in a manner consistent with the rest of the documentation.

metaflow/plugins/argo/argo_events_decorator.py
savingoyal: we don't have access to the values from current here. `triggers` only records what the user specifies and the default value handling is taken care of by the implementation where we have access to current.

metaflow/plugins/argo/argo_events_decorator.py
savingoyal: it is listed as a TODO at the moment in the description of this PR.

metaflow/plugins/argo/argo_events_decorator.py
savingoyal: Let's chat about this. I created a version with `prod`, `user` etc. but thinking through what to specify as the value would be quite confusing for the user. This notion of specifying project and branch already maps to the values in system tags and current object.

metaflow/plugins/argo/argo_events_decorator.py
```diff
@@ -0,0 +1,154 @@
+from metaflow.decorators import FlowDecorator
```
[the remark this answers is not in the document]
savingoyal: that can happen after the PR is ready to be merged.

metaflow/metaflow_config.py
savingoyal: The PR is *not* yet ready to be merged. There are pending changes for `metaflow configure` etc.

metaflow/plugins/__init__.py
savingoyal: correct - although this seems to work without any issues.

metaflow/plugins/argo/argo_events_decorator.py
savingoyal: I kept it `@trigger` to keep it consistent with `@schedule`. Moving this file and changing class names is trivial to do and doesn't block this PR.

metaflow/events.py
```diff
@@ -0,0 +1,83 @@
+from collections import namedtuple, OrderedDict
+from datetime import datetime
+
+
+MetaflowEvent = namedtuple("MetaflowEvent", ["name", "id", "timestamp", "type"])
+
+
+class MetaflowTrigger(object):
```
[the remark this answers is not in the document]
savingoyal: Ah yes - I didn't have that in my notes. Will add that.

metaflow/events.py
savingoyal: will do

metaflow/plugins/argo/argo_client.py
```diff
@@ -181,3 +180,75 @@ def schedule_workflow_template(self, name, schedule=None, timezone=None):
             raise ArgoClientException(
                 json.loads(e.body)["message"] if e.body is not None else e.reason
             )
+
+    def register_sensor(self, name, sensor={}):
+        # Unfortunately, Kubernetes client does not handle optimistic
```
jackie-ob, line 185: Indeed... busy work here...

metaflow/plugins/argo/argo_events.py
```diff
@@ -0,0 +1,95 @@
...
+                    ).encode("utf-8"),
+                )
+                retries = 3
+                backoff_factor = 2
+
+                for i in range(retries):
+                    try:
+                        urllib.request.urlopen(request, timeout=10.0)
+                        print(
+                            "Argo Event (%s) published." % self._name, file=sys.stderr
```
jackie-ob, line 70: Publish notices (success or failure): should we be printing these by default.  They were very useful in my testing / infra setup work.  But doing this routinely for happy path is not ideal.

metaflow/plugins/argo/argo_events.py
```diff
@@ -0,0 +1,95 @@
...
+                        )
+                        break
+                    except urllib.error.HTTPError as e:
```
jackie-ob, line 73: Should we retry retryable status codes?

metaflow/plugins/argo/argo_events_decorator.py
```diff
@@ -0,0 +1,154 @@
...
+        "event": None,
+        "events": [],
+        "options": {},  # TODO: introduce support for options
+    }
+
+    def flow_init(
+        self, flow, graph, environment, flow_datastore, metadata, logger, echo, options
+    ):
+        self.triggers = []
+        if sum(map(bool, (self.attributes["event"], self.attributes["events"]))) > 1:
```
jackie-ob, line 23: Nit:

`if self.attributes["event"] and self.attributes["events']`

Also should we be checking with an XOR logic instead?  I.e. we want exactly one of the two to be defined right?

`if bool(self.attributes["event"]) == bool(self.attributes["events']):`

metaflow/plugins/argo/argo_events_decorator.py
```diff
@@ -0,0 +1,154 @@
...
+            # event attribute supports the following formats -
+            #     1. event='table.prod_db.members'
+            #     2. event={'name': 'table.prod_db.members',
+            #               'parameters': {'alpha': 'member_weight'}}
+            if is_stringish(self.attributes["event"]):
+                self.triggers.append({"name": str(self.attributes["event"])})
+            elif isinstance(self.attributes["event"], dict):
+                if "name" not in dict(self.attributes["event"]):
+                    raise MetaflowException(
+                        "The *event* attribute for *@trigger* is missing the "
```
jackie-ob, line 38: Should we pull the common parse/validation logic out?  Duplicated across "event" and "events" right now.

metaflow/plugins/argo/argo_events_decorator.py
```diff
@@ -0,0 +1,154 @@
...
+                raise MetaflowException(
+                    "Incorrect format for *events* attribute in *@trigger* decorator. "
+                    "Supported format is list - \n"
+                    "@trigger(events=[{'name': 'foo', 'parameters': {'alpha': "
+                    "'beta'}}, {'name': 'bar', 'parameters': "
+                    "{'gamma': 'kappa'}}])"
+                )
+
+        if not self.triggers:
+            raise MetaflowException("No event(s) specified in *@trigger* decorator.")
```
jackie-ob, line 85: I see in the end... XOR check is effectively enforced here.  Should we hint to user to set `event` or `events` inline, here?

metaflow/plugins/argo/argo_workflows_decorator.py
```diff
@@ -23,10 +31,47 @@ def task_pre_step(
...
+
+        triggers = []
+        # Expose event triggering metadata through current singleton
+        for key, payload in os.environ.items():
+            if key.startswith("METAFLOW_ARGO_EVENT_PAYLOAD_"):
+                if payload != "null":  # Argo-Workflow's None
+                    try:
+                        payload = json.loads(payload)
+                    except (TypeError, ValueError) as e:
+                        # There could be arbitrary events that Metaflow doesn't know of
```
jackie-ob, line 44: But this payload definitely came in through `METAFLOW_ARGO_EVENT_PAYLOAD_` right?  I wonder if we really should be swallowing this.  I think it's more likely that this is a bug in the setup (either in code or in supporting infra).

The end result is downstream flow gets run, but without the right payload - confusing to debug (first hand experience!).

Instead this COULD just throw an error indicating that the event payload was invalid (and possibly print it).

metaflow/plugins/argo/argo_workflows_decorator.py
```diff
@@ -61,3 +106,40 @@ def task_finished(
...
+                "project_flow_name",
+            ):
+                if current.get(key):
+                    event.add_to_payload(key, current.get(key))
+            # Add more fields here...
+            event.add_to_payload("auto-generated-by-metaflow", True)
+            # Keep in mind that any errors raised here will fail the run but the task
+            # will still be marked as success. That's why we explicitly swallow any
+            # errors and instead print them to std.err.
+            event.publish(ignore_errors=True)
```
jackie-ob, line 145: Discussion only.

I guess this is a difficulty with distributed systems.  So publish fails here.  How does the user find out?

Event source flow owner won't know, everything is OK (Run was successful!).

Event target flow owner can only know when they realize the source flow ran fine, but the target flow did not get triggered.

That is to say events plumbing has to be 100% reliable (or supporting infra must have failsafe alerting mechanisms).

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -176,32 +194,63 @@ def trigger(cls, name, parameters=None):
...
+                "is/are published."
+                % self.list_to_prose(
+                    [event["name"] for event in self.triggers], "event"
+                )
+            )
+
+        # Trigger explanation for @trigger_on_finish
+        elif self.flow._flow_decorators.get("trigger_on_finish"):
+            msg = (
+                "This workflow triggers automatically when the upstream %s succeed(s)"
```
saikonen, line 241: Due to triggers relying on the webhook, this is true if the upstream flow has also been deployed with events enabled, and is an argo-workflow (so it auto-emits).

This can be bit of a gotcha for users that have existing deployed flows that they try to use downstream with `trigger_on_finish`, but they never emit the necessary event.

I would either add a mention here that the upstream also needs to have been deployed with an event-capable version of metaflow, or make note of the requirement in the release notes

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -247,24 +296,145 @@ def _process_parameters(self):
...
+                    )
+                # Throw an error if we are unable to impute project for specified
+                # branch.
+                if (
+                    not event.get("project")
+                    and not current.get("project_name")
+                    and event.get("branch")
+                ):
+                    raise ArgoWorkflowsException(
+                        "No project specified for branch in *@trigger_on_finish* "
```
saikonen, line 395: Another case is that project can be specified even though branch is left empty. This will listen to `project/None` for completion. Is this a valid use-case or should there be a check that only both values are available at once?

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -247,24 +296,145 @@ def _process_parameters(self):
...
+        if self.flow._flow_decorators.get("trigger_on_finish"):
+            for event in self.flow._flow_decorators.get("trigger_on_finish")[
+                0
+            ].triggers:
+                # TODO: Also sanity check project and branch names
+                if not re.match(r"^[A-Za-z0-9_]+$", event["flow"]):
+                    raise ArgoWorkflowsException(
+                        "Invalid flow name *%s* in *@trigger_on_finish* "
+                        "decorator. Only alphanumeric characters and "
+                        "underscores(_) are allowed." % (value, event["flow"])
```
saikonen, line 385: `value` is unassigned here and possibly not meant to be used either?

metaflow/events.py
```diff
@@ -0,0 +1,83 @@
...
+            )
+            for obj in filter(
+                lambda x: x.type in ["run"],
+                self._meta.values(),
+            )
+        ] or None
+
+    def __getitem__(self, key):
+        """
+        If triggering events are runs, `key` corresponds to the flow name of the triggering run. Returns a triggering `Run` object corresponding to the key. If triggering events are not runs, `key` corresponds to the event name and a `MetaflowEvent` object is returned.
```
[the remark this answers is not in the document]
savingoyal: Currently we only support either events or runs

metaflow/events.py
```diff
@@ -0,0 +1,83 @@
...
+                obj.id[: obj.id.index("/", obj.id.index("/") + 1)],
+                _namespace_check=False,
+            )
+            for obj in filter(
+                lambda x: x.type in ["run"],
+                self._meta.values(),
+            )
+        ] or None
+
+    def __getitem__(self, key):
```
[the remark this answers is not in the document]
savingoyal: will do

metaflow/events.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+MetaflowEvent = namedtuple("MetaflowEvent", ["name", "id", "timestamp", "type"])
+
+
+class MetaflowTrigger(object):
+    def __init__(self, _meta=[]):
+        _meta.sort(key=lambda x: x.get("timestamp") or float("-inf"), reverse=True)
+        self._meta = OrderedDict(
+            {
```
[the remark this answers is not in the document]
savingoyal: Unfortunately, the extra set of brace is to ensure that we are able to do type conversion for timestamp and override the value

metaflow/plugins/argo/argo_workflows.py
savingoyal: +1

metaflow/plugins/argo/argo_workflows_decorator.py
savingoyal: It may come in through `METAFLOW_ARGO_EVENT_PAYLOAD_ ` but might still result from an event we don't control (hence outside the purview of the event triggering). If the event is a valid JSON then this will work as is.

metaflow/plugins/argo/argo_events_decorator.py
savingoyal: I tried doing that. Given the subtle changes in the error messages, it would have been quite a bit more complicated than simply repeating it. Definitely scope for improvement though. We can do another pass post merge.

metaflow/plugins/argo/argo_events_decorator.py
savingoyal: Yes - you are correct. This is a remnant of when I had more options. I can simplify this.

metaflow/plugins/argo/argo_events.py
savingoyal: Yeah - we can turn them off and only print the failures later. these go to stderr so hopefully less cloggy.

metaflow/plugins/argo/argo_events.py
savingoyal: yes, I will introduce that

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -172,10 +179,12 @@ def create(
         max_workers,
         workflow_timeout,
         workflow_priority,
+        auto_emit_argo_events,
     )
 
     if only_json:
         obj.echo_always(str(flow), err=False, no_bold=True)
+        # TODO: Support echo-ing Argo Events Sensor template
```
talebzeghmi, line 187: also suggest suporting echo-ing the @schedule CronWorkflow

metaflow/plugins/argo/argo_client.py
```diff
@@ -119,7 +118,7 @@ def trigger_workflow_template(self, name, parameters={}):
     def schedule_workflow_template(self, name, schedule=None, timezone=None):
```
talebzeghmi, line 118: what if schedule is None and the schedule exists in the cluster? Does this client delete the schedule?

talebzeghmi: ah you always create schedule, and set `suspend=None` if `schedule==None`... why not delete the `CronWorkflow` instead?

metaflow/client/core.py
```diff
@@ -1901,6 +1896,23 @@ def replace_tags(self, tags_to_remove: Iterable[str], tags_to_add: Iterable[str]
...
+        Returns
+        -------
+        Trigger, optional
+            Container of triggering events
+        """
+        if "start" in self:
+            meta = self["start"].task.metadata_dict.get("execution-triggers")
+            if meta:
+                return Trigger(json.loads(meta))
+
```
romain-intel, line 1915: nit: explicit: return None

metaflow/events.py
```diff
@@ -0,0 +1,136 @@
...
+    Container of metadata that identifies the event that triggered
+    the `Run` under consideration.
+
+    Attributes
+    ----------
+    name : str
+        name of the event.
+    id : str
+        unique identifier for the event.
+    timestamp : datetime
```
romain-intel, line 15: this seems wrong as you are treating it as an int later on

metaflow/plugins/events_decorator.py
```diff
@@ -0,0 +1,298 @@
...
+    ```
+    or
+    ```
+    @trigger(events=[{'name':'foo', 'parameters':{'my_param_1': 'event_field_1'},
+                     {'name':'bar', 'parameters':{'my_param_2': 'event_field_2'}])
+    ```
+
+    Parameters
+    ----------
+    event : str or dict, optional
```
romain-intel, line 39: To keep standard, please use Union[]

metaflow/events.py
```diff
@@ -0,0 +1,136 @@
...
+            # to avoid circular import
+            from metaflow import Run
+
+            self._runs = [
+                Run(
+                    # object id is the task pathspec for events that map to run
+                    obj.id[: obj.id.index("/", obj.id.index("/") + 1)],
+                    _namespace_check=False,
+                )
+                for obj in filter(
```
romain-intel, line 105: simpler: `for obj in self.events if obj.type == "run"`

romain-intel: To avoid a copy (cheap but why), just do `for obj in self._events`. It also will work if empty events (this code won't since filter over None will break)

metaflow/events.py
```diff
@@ -0,0 +1,136 @@
...
+                    lambda x: x.type in ["run"],
+                    self.events,
+                )
+            ] or None
```
romain-intel, line 109: nit: I would actually do `self._runs = [...]` (without the `or None`) and then do the same as for events which is: `return list(self._runs) or None` which is cleaner and avoids the recompute.

metaflow/events.py
```diff
@@ -0,0 +1,136 @@
...
+                    "id": run_obj.end_task.pathspec,
+                }
+                for run_obj in run_objs
+            ]
+        )
+        trigger._runs = run_objs
+        return trigger
+
+    @property
+    def event(self):
```
romain-intel, line 70: If we are exposing this object, we should have the full API doc (including the "Return" section)

romain-intel (commented):
Not done reviewing yet but initial comments to get them to you as fast as possible.

metaflow/plugins/argo/argo_client.py
```diff
@@ -119,7 +118,7 @@ def trigger_workflow_template(self, name, parameters={}):
...
         # Unfortunately, Kubernetes client does not handle optimistic
         # concurrency control by itself unlike kubectl
-        client = self._kubernetes_client.get()
+        client = self._client.get()
         body = {
             "apiVersion": "argoproj.io/v1alpha1",
             "kind": "CronWorkflow",
```
talebzeghmi, line 124: As soon as CronWorkflow is introduced, the need for backfill will be requested.
How do you plan on supporting backfill? With `python flow.py argo backfill --start --end --format` 
- https://argoproj.github.io/argo-workflows/cron-backfill/
- https://raw.githubusercontent.com/argoproj/argo-workflows/master/examples/cron-backfill.yaml

metaflow/plugins/argo/argo_workflows_decorator.py
```diff
@@ -113,8 +113,15 @@ def task_finished(
...
         if self.attributes["auto-emit-argo-events"]:
-            # Event name is set to metaflow.flow.step
+            # Event name is set to metaflow.project.branch.step so that users can
+            # place explicit dependencies on namespaced events. Also, argo events
+            # sensors don't allow for filtering against absent fields - which limits
+            # our ability to subset non-project namespaced events.
             # TODO: Check length limits for fields in Argo Events
+            event = ArgoEvent(
+                name="metaflow.%s.%s"
+                % (current.get("project_flow_name", flow.name), step_name)
```
saikonen, line 123: something is not working with this part:
- during task execution, `current.project_flow_name` returns correctly `sa_test.user.saikonen.EventTriggeringTest` for my test
- the event that is auto-emitted is `metaflow.EventTriggeringTest.start` though.

metaflow/plugins/argo/argo_client.py
savingoyal: @talebzeghmi - cron workflows have been supported with Metaflow-on-Argo Workflows for a while. Supporting backfills is on our roadmap too

metaflow/plugins/events_decorator.py
```diff
@@ -0,0 +1,298 @@
...
+            raise MetaflowException(
+                "No flow(s) specified in *@trigger_on_finish* decorator."
+            )
+
+        # Make triggers @project aware
+        for trigger in self.triggers:
+            if trigger["fq_name"].count(".") == 0:
+                # fully qualified name is just the flow name
+                trigger["flow"] = trigger["fq_name"]
+            elif trigger["fq_name"].count(".") == 3:
```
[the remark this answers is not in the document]
savingoyal: good catch. fixed.

metaflow/events.py
romain-intel: line 44 -- you have `isinstance(obj.get("timestamp"), int)`

metaflow/events.py
```diff
@@ -0,0 +1,134 @@
...
+        """
+        Returns a `MetaflowEvent` object corresponding to the triggering event. If multiple events triggered the run, returns the latest event.
+        """
+        return next(iter(self.events), None)
+
+    @property
+    def events(self):
+        """
+        Returns a list of `MetaflowEvent` objects correspondings to all the triggering events.
```
romain-intel, line 79: nit: would clarify: runs or non runs to make it clear that `runs` is a subset and not a disjoint set.

metaflow/events.py
```diff
@@ -0,0 +1,134 @@
...
+        )
+        trigger._runs = run_objs
+        return trigger
+
+    @property
+    def event(self):
+        """
+        Returns a `MetaflowEvent` object corresponding to the triggering event. If multiple events triggered the run, returns the latest event.
+        """
+        return next(iter(self.events), None)
```
romain-intel, line 74: sorry, I ddin't catch this earlier -- for all these (iterating in some way over events), you need to use `self._events` as `self.events` can be None which is not iterable.

metaflow/events.py
```diff
@@ -0,0 +1,134 @@
...
+        Returns a list of `MetaflowEvent` objects correspondings to all the triggering events.
+        """
+        return list(self._events) or None
+
+    @property
+    def run(self):
+        """
+        If the triggering event is a Metaflow run, returns the corresponding `Run` object. If multiple runs triggered the run, returns the latest run. `None` if the event is not a `Run`.
+        """
+        return next(iter(self.runs), None)
```
romain-intel, line 88: here you may need something a little different:
```
if self._runs is None:
  self.runs
return next(iter(self._runs), None)
```
or something like that. Same deal -- avoiding case where self.runs is None

metaflow/plugins/events_decorator.py
```diff
@@ -0,0 +1,298 @@
...
+class TriggerDecorator(FlowDecorator):
+    """
+    Specifies the event(s) that this flow depends on.
+
+    ```
+    @trigger(event='foo')
+    ```
+    or
+    ```
+    @trigger(events=['foo', 'bar'])
```
romain-intel, line 23: clarify: `in which case the flow will be triggered when *all* events are satisfied` or something like that to clearly state that this is an and. iirc, we were going to maybe later extend to `or` (if needed) using multiple decorators.

metaflow/plugins/events_decorator.py
```diff
@@ -0,0 +1,298 @@
...
+
+    Additionally, you can specify the parameter mappings
+    to map event payload to Metaflow parameters for the flow.
+    ```
+    @trigger(event={'name':'foo', 'parameters':{'my_param': 'event_field'})
+    ```
+    or
+    ```
+    @trigger(events=[{'name':'foo', 'parameters':{'my_param_1': 'event_field_1'},
+                     {'name':'bar', 'parameters':{'my_param_2': 'event_field_2'}])
```
romain-intel, line 34: As a tiny tiny nit: black imposes " instead of ' so we could also have that in our doc.

metaflow/plugins/events_decorator.py
```diff
@@ -0,0 +1,298 @@
...
+            #               'parameters': {'alpha': 'member_weight'}}
+            if is_stringish(self.attributes["event"]):
+                self.triggers.append({"name": str(self.attributes["event"])})
+            elif isinstance(self.attributes["event"], dict):
+                if "name" not in dict(self.attributes["event"]):
+                    raise MetaflowException(
+                        "The *event* attribute for *@trigger* is missing the "
+                        "*name* key."
+                    )
+                self.triggers.append(self.attributes["event"])
```
romain-intel, line 84: we could check the other fields here too. The error message hints at the format needed but we don't actually check.

metaflow/plugins/events_decorator.py
```diff
@@ -0,0 +1,298 @@
...
+            )
+
+        # Make triggers @project aware
+        for trigger in self.triggers:
+            if trigger["fq_name"].count(".") == 0:
+                # fully qualified name is just the flow name
+                trigger["flow"] = trigger["fq_name"]
+            elif trigger["fq_name"].count(".") in (2, 3):
+                # fully qualified name is of the format - project.branch.flow_name
+                trigger["project"], _ = trigger["fq_name"].split(".", maxsplit=1)
```
romain-intel, line 258: nit, `_` is not meant to be used as an actual variable :)

metaflow/plugins/events_decorator.py
```diff
@@ -0,0 +1,298 @@
...
+        if not self.triggers:
+            raise MetaflowException(
+                "No flow(s) specified in *@trigger_on_finish* decorator."
+            )
+
+        # Make triggers @project aware
+        for trigger in self.triggers:
+            if trigger["fq_name"].count(".") == 0:
+                # fully qualified name is just the flow name
+                trigger["flow"] = trigger["fq_name"]
```
romain-intel, line 255: shouldn't you add the current project and branch info here? Ie: if we have a branch/project info, the default should be "current project/branch", right?

metaflow/plugins/events_decorator.py
```diff
@@ -0,0 +1,298 @@
...
+            elif trigger["fq_name"].count(".") in (2, 3):
+                # fully qualified name is of the format - project.branch.flow_name
```
romain-intel, line 257: clarify the two formats for clarity:
`project.prod.flow_name` or `project.[user|prod|test].branch_name.flow_name`.

metaflow/plugins/events_decorator.py
```diff
@@ -0,0 +1,298 @@
...
+    @trigger_on_finish(flow='FooFlow')
+    ```
+    or
+    ```
+    @trigger_on_finish(flows=['FooFlow', 'BarFlow'])
+    ```
+    This decorator respects the @project decorator and triggers the flow
+    when upstream runs within the same namespace complete successfully
+
+    Additionally, you can specify project aware upstream flow dependencies
```
romain-intel, line 156: this is a new concept not documented so would add examples:
`project.prod.flow_name` for the production branch or `project.[prod|test].branch_name.flow_name` or `project.user.user_name.flow_name` or something like that.

metaflow/plugins/events_decorator.py
```diff
@@ -0,0 +1,298 @@
...
+                trigger["project"], _ = trigger["fq_name"].split(".", maxsplit=1)
+                trigger["branch"], trigger["flow"] = _.rsplit(".", maxsplit=1)
+            else:
+                raise MetaflowException(
+                    "Incorrect format for *flow* in *@trigger_on_finish* "
+                    "decorator. Specify either just the *flow_name* or a fully "
+                    "qualified name like *project_name.branch_name.flow_name*."
+                )
+            # TODO: Also sanity check project and branch names
+            if not re.match(r"^[A-Za-z0-9_]+$", trigger["flow"]):
```
romain-intel, line 267: project and branch names have a different constraint (just noting here for future). We could/should import the re constants from there and check.

romain-intel (commented):
Ok. I think I am done. I have not reviewed any of the airflow, argo, sfn or kube portions instead focusing on the common changes.

metaflow/plugins/events_decorator.py
savingoyal: we don't have access to the current object here - it's taken care of in the implementation.

metaflow/plugins/events_decorator.py
savingoyal: We will add examples in the documentation.

metaflow/plugins/events_decorator.py
savingoyal: Yes - already noted as a future TODO.

metaflow/plugins/events_decorator.py
savingoyal: The definition of "satisified" will vary from event backend to event backend - hence it is best left as-is and can be clarified in the documentation.

metaflow/events.py
savingoyal: can you expand on this?

romain-intel: I mean in the doc. It was initially unclear to me that `events` is a *superset* of `runs`. I initially thought they were disjoint and only one would return something. I guess the confusion comes from the fact that event/events returns a MetaflowEvent which is either an *event* or a run (notice that an event is an event or a run so there is a name overload). It's more a doc thing than anything else.

metaflow/plugins/events_decorator.py
romain-intel: fine. it's just inconsistent with the error message but not a big deal.

metaflow/plugins/events_decorator.py
saikonen: this would be standard for typings, but I don't see Union[] being widely used in docstrings in the codebase so far. 
The "param: a or b" style seems much more prevalent at this point in the docstrings. Should we start steering away from it?

metaflow/events.py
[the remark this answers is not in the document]
romain-intel: type hints still missing?

PR #1272, opened 2023-02-13
Introduce codeql

savingoyal:
quick experiment to see it's utility


saikonen (approved):

saikonen:
looked into the issues that CodeQL was reporting

- the two critical ones are related to the vendored `click`, they seem like non-issues due to the way click is being used. Possibly fixed in a later version as well.
- the javascript prototype pollution issues might be resolvable by upgrading the Chart.js version

PR #1279, opened 2023-02-21
feat: delete, delete_many, delete_recursive to s3 wrapper

mSounak:
Reference to Issue https://github.com/Netflix/metaflow/issues/1010

1. Added support to delete one object using s3 client
2. Added support to delete multiple objects.
3. Added support to delete objects recursively.


metaflow/plugins/datatools/s3/s3.py
```diff
@@ -819,6 +819,75 @@ def _head():
...
+        urls = [self._url(key) for key in keys]
+        src = [urlparse(u) for u in urls]
+
+        Bucket_names = set([s.netloc for s in src])
+        if len(Bucket_names) > 1:
+            raise Exception("Cannot delete from multiple buckets at once")
+        else:
+            bucket_name = Bucket_names.pop()
+
+            resp = self._s3_client.delete_objects(
```
romain-intel, line 860: I will check if there are any restrictions on delete_objects (failure modes) etc but o/w, this looks reasonable to do a single operation to delete a bunch of things.

metaflow/plugins/datatools/s3/s3.py
```diff
@@ -819,6 +819,75 @@ def _head():
...
+        keys: List[str]
+            List of keys to delete.
+        """
+
+        urls = [self._url(key) for key in keys]
+        src = [urlparse(u) for u in urls]
+
+        Bucket_names = set([s.netloc for s in src])
+        if len(Bucket_names) > 1:
+            raise Exception("Cannot delete from multiple buckets at once")
```
romain-intel, line 856: We may want to support this to be consistent with the rest. In that case, maybe we can do this operation over buckets using s3op?

metaflow/plugins/datatools/s3/s3.py
```diff
@@ -819,6 +819,75 @@ def _head():
...
+
+        url = self._url(key)
+        src = urlparse(url)
+
+        def _delete(s3, tmp):
+            bucket = s3.Bucket(src.netloc)
+            if not src.path.lstrip('/'):
+                raise TypeError("'None' type object cannot be deleted.")
+            else:
+                for key_ in bucket.list(prefix=src.path.lstrip("/")):
```
romain-intel, line 886: For this one though, I think it may be preferable to go the s3op route (ie: use a bunch of subprocesses) to parallelize it a bit. We could also potentially do a hybrid approach where we use s3op to get the listing using s3op (like in `list_recursive`) and then fallback on the `delete_many` implementation above where we delete all files in one fell swoop (modulo any restrictions on that as mentioned earlier).

romain-intel (commented):
Apologies for the delay. I added a few comments and will do a bit more research on the delete_objects operation in S3. I also confirmed with other developers that we are all good shepherding this through and merging it. There is no urgency on our end but it is a legitimate feature that would be cool to add.

PR #1284, opened 2023-02-27
feat(sfn): delete a workflow

stevenhoelscher:
Adds a new CLI command to the Step Functions plugin that enables deleting a deployed SFN State Machine.

### DIFF
--- a/metaflow/plugins/aws/step_functions/event_bridge_client.py
+++ b/metaflow/plugins/aws/step_functions/event_bridge_client.py
@@ -60,6 +60,20 @@ class EventBridgeClient(object):
             ],
         )
 
+    def delete(self):
+        try:
+            response = self._client.remove_targets(
+                Rule=self.name,
+                Ids=[self.name],
+            )
+            if response.get("FailedEntryCount", 0) > 0:
+                raise RuntimeError("Failed to remove targets from rule %s" % self.name)
+            self._client.delete_rule(Name=self.name)
+        except self._client.exceptions.ResourceNotFoundException:
+            # Ignore if the rule does not exist.
+            pass
+
+
 
 def format(name):
     # AWS Event Bridge has a limit of 64 chars for rule names.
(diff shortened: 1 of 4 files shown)


savingoyal:
@stevenhoelscher Wouldn't this also involve tearing down the event-bridge rule if any?

stevenhoelscher:
> @stevenhoelscher Wouldn't this also involve tearing down the event-bridge rule if any?

That's a good point. Let me take a stab at that.

saikonen:
Some further notes and suggestions on the behavior of this feature.

1. running the delete command exits immediately without any indication for the user that deletion was successful. Some notification at the end of the process would give assurance that things worked out as intended
2. also noticed that after issuing the delete, the actual state machine still stays in a "status: deleting" state for a while, even though the CLI exited already. Could consider waiting until the delete completes before exiting, if this does not take too long
3. Somewhat expected, but in-flight executions will fail when the state machine gets deleted mid execution. Should there some notice on this?
4. a quick delete, followed by a create will lead to a `StateMachineDeleting("An error occurred (StateMachineDeleting) when calling the CreateStateMachine operation: State Machine is being deleted`, which might be unavoidable as other people can affect the state machines with shared production tokens.
5. the delete is not currently respecting production tokens, instead allowing any state machine to be deleted if the user role has enough permissions to perform it. Preferably the delete would require the `--authorize` flag with a production token shared by the original author.

saikonen:
some further thoughts after the weekend:
After thinking about it for a while, making the CLI wait for the deletion will not be a good experience for the user, as the time for deletion is most likely directly affected by the size of the state machine. So for **2** it would be enough to output a confirmation to the user that the command succeeded, and resources are being deleted.

also small finding, issuing consecutive `delete` commands simply succeeds. Not sure what the AWS Api returns in this case, but was surprising that this did not fail to the same 'state machine is being deleted' exception

saikonen:
There's been some back and forth regarding an identical feature for argo-workflows which now has the UX more or less done. If you have time, could you could check out the PR #1307 and implement something similar to how the `--authorize` is handled, and the output for the user when operations are performed. These should address all of my previous comments.

savingoyal:
#1379 supercedes this PR

PR #1287, opened 2023-03-01
Introduce tmpfs config options for @batch

savingoyal:


metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -282,6 +287,29 @@ def _register_job_definition(
                     {"sourceVolume": name, "containerPath": host_path}
                 )
 
+        if use_tmpfs or tmpfs_size:
+            if tmpfs_size:
```
saikonen, line 291: there should probably be a check that the final value for tmpfs_size is less than `memory=`, because otherwise tasks run into
```
Error: [Errno 12] Cannot allocate memory
```

saikonen: also a bit unsure on the exact constraints of `tmpfs_size`, as setting this to 1000 and trying to S3 get a 1gb file results in a
```
Error: [Errno 28] No space left on device
```
but somewhat expected that redlining the attached storage would run out of space due to possible overhead

metaflow/plugins/aws/batch/batch_decorator.py
```diff
@@ -194,6 +208,11 @@ def task_pre_step(
         self.metadata = metadata
         self.task_datastore = task_datastore
 
+        # current.tempdir reflects the value of METAFLOW_TEMPDIR (the current working
+        # directory by default), or the value of tmpfs_path if tmpfs_tempdir=False.
+        if not self.attributes["tmpfs_tempdir"]:
+            current._update_env("tempdir", self.attributes["tmpfs_path"])
```
tuulos, line 214: this line fails with `@batch( tmpfs_tempdir=False, use_tmpfs=True)`

metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -149,7 +149,12 @@ def _register_job_definition(
         max_swap,
         swappiness,
         inferentia,
+        memory,
         host_volumes,
+        use_tmpfs,
+        tmpfs_tempdir,
+        tmpfs_size,
+        tmpfs_path,
```
tuulos, line 157: `tmpfs_path` needs to be an absolute path (start with `/`) or Batch barfs:
```
botocore.errorfactory.ClientException: An error occurred (ClientException) when calling the RegisterJobDefinition operation: Error executing request, Exception : The containerPath (foobar) specified for tmpfs is invalid. It must contain an absolute path., RequestId: de9d0a90-358c-48b3-b104-99d040595ed9
```
we could have a static check that ensuring
```
if tmpfs_path and tmpfs_path[0] != '/':
    raise BatchJobException("tmpfs_path needs to be an absolute path")
```
so the user doesn't have to wait until the task executes to see the error

metaflow/plugins/aws/batch/batch_client.py
tuulos: I tried setting `@batch(memory=10000, tmpfs_size=120000)` and surprisingly it creates a tmpfs volume of the requested size. It'd probably fail allocating more than 10GB though. I think it's ok to leave it as-is for now.

Storing more data than what's set by `tmpfs_size` results to `No space left` as you said, which is expected behavior

metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -282,6 +287,29 @@ def _register_job_definition(
                     {"sourceVolume": name, "containerPath": host_path}
                 )
 
+        if use_tmpfs or tmpfs_size:
```
tuulos, line 290: maybe we could drop `or tmpfs_size` here.  `@batch(use_tmpfs=False, tmpfs_size=10000)` leads to this error:
```
 FileNotFoundError: [Errno 2] No such file or directory: '/metaflow_temp/metaflow.s3.evqtjhbe'
 ```

we can interpret that `use_tmpfs=False` means disabling all `tmpfs` behavior altogether

File: metaflow/plugins/aws/batch/batch.py [hunk 5]
```diff
@@ -248,6 +257,9 @@ def create_job(
                 AWS_SECRETS_MANAGER_DEFAULT_REGION,
             )
 
+        if tmpfs_tempdir:
```
tuulos, line 260: this should be `if use_tmpfs and tmpfs_tempdir` - currently it always sets `METAFLOW_TEMPDIR`, breaking `metaflow.S3` when `tmpfs` is not enabled

metaflow/plugins/aws/batch/batch_client.py
tuulos: (this error is probably related to the other comment related to `if tmpfs_tempdir` that I made below)

tuulos, line 290: we could change this to
```
if use_tmpfs or (tmpfs_size and use_tmpfs != False):
```
it is confusing that `@batch(use_tmpfs=False, tmpfs_size=10000)` uses `tmpfs` even though it says `False`

metaflow/plugins/aws/batch/batch.py
tuulos: maybe it's easier to have one conditional at the top a'la
```
tmpfs_enabled = use_tmpfs or (tmpfs_size and use_tmpfs is None)
```
and have
```
if tmpfs_enabled and tmpfs_tempdir
```
here

metaflow/plugins/aws/batch/batch_client.py
saikonen: diff doesn't show "outdated" for this file but this is now covered after the latest commit. Check was added to the `step_init` in `batch_decorator.py`. I tried raising the exception inside the `batch_client.py` `_register_job_definition()` at first, but this was too late in the lifecycle and allowed the flow to start executing before encountering the exception.

PR #1289, opened 2023-03-06
[make-metaflow-token-home] make metaflow token home configurable

jackie-ob:


metaflow/plugins/aws/step_functions/production_token.py
```diff
@@ -38,7 +38,11 @@ def _load_config(path):
 
 
 def _path(token_prefix):
-    home = os.environ.get("METAFLOW_HOME", "~/.metaflowconfig")
+    # TODO make this a MF config variable
```
romain-intel, line 41: No strong opposition to this (it only touches aws sfn) but yes, I would agree that it should be in metaflow config. I am not sure thought of why we would want to move it (wouldn't it be possibly more confusing)?

PR #1290, opened 2023-03-06
[mfgui-url] show mfgui url in console

jackie-ob:


savingoyal:
Can we file an issue for a follow up PR that prints the UI URL for `step-functions trigger` and `argo-workflows trigger`?

metaflow/runtime.py
```diff
@@ -198,8 +198,19 @@ def persist_constants(self, task_id=None):
...
 
     def execute(self):
-
-        self._logger("Workflow starting (run-id %s):" % self._run_id, system_msg=True)
+        if MFGUI_URL:
+            self._logger(
+                "Workflow starting (run-id %s), see it in the UI at %s"
+                % (
+                    self._run_id,
+                    os.path.join(MFGUI_URL, self._flow.name, self._run_id),
```
shrinandj, line 206: Should this be `urlib.parse.urljoin` (or a similar function) that does url joins? os.path.join would result in the wrong url on non-unix envs. Not sure how if Metaflow is used on non-envs though.

romain-intel: the current assumption is mf only works on linux like systems. Your point remains valid but I think the assumption is pretty pervasive.

romain-intel (approved):
Apart from a minor comment on naming, lgtm.

jackie-ob:
Noted.  We will switch to `urlib.parse.urljoin` and `METAFLOW_UI_URL`

savingoyal (dismissed):
good to go after pending changes are shipped!

metaflow/runtime.py
jackie-ob: Note: we switched to just using manual string construction.  `urljoin` does not do what we want (e.g. `urljoin("http://A/B", "C") == "http://A/C"`.

PR #1291, opened 2023-03-06
[resource-limits-config] max cpu and max memory setting in config

jackie-ob:


jackie-ob:
Cancelling this one for now, as possibly less applicable to OSS.

PR #1297, opened 2023-03-09
fix: stdout k8s and batch formatting

saikonen:
Strips styling from stdout text wrapped in asterisks from kubernetes_cli and batch_cli in order for the terminal output to be a 1-to-1 match with what is stored in the log files, and what is visible in MFGUI.


saikonen:
not sure why Python 3.7 tests failed on macos-latest when no changes to dependencies have been made, and all other python version checks passed.

maybe a one-off failure and the check could be retried?

shrinandj (approved):

PR #1301, opened 2023-03-10
[k8s/argo support] Add tunables for TTL behavior of kubernetes jobs and argo workflows

jackie-ob:
Two new config vars to control k8s job (`METAFLOW_KUBERNETES_JOB_TTL_SECONDS_AFTER_FINISHED`) or workflow (`METAFLOW_ARGO_WORKFLOWS_TTL_STRATEGY`)TTLs.  If unset, default behavior matches current.

### DIFF
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -28,6 +28,7 @@ from metaflow.metaflow_config import (
     AWS_SECRETS_MANAGER_DEFAULT_REGION,
     ARGO_WORKFLOWS_KUBERNETES_SECRETS,
     ARGO_WORKFLOWS_ENV_VARS_TO_SKIP,
+    ARGO_WORKFLOWS_TTL_STRATEGY,
 )
 from metaflow.mflog import BASH_SAVE_LOGS, bash_capture_logs, export_mflog_env_vars
 from metaflow.parameters import deploy_time_eval
@@ -341,6 +342,8 @@ class ArgoWorkflows(object):
                 # .image_pull_secrets(...)
                 # Limit workflow parallelism
                 .parallelism(self.max_workers)
+                # Possibly have Argo Workflows controller clean up old Workflows
+                .ttl_strategy(ARGO_WORKFLOWS_TTL_STRATEGY)
                 # TODO: Support Prometheus metrics for Argo
                 # .metrics(...)
                 # TODO: Support PodGC and DisruptionBudgets
@@ -1111,6 +1114,11 @@ class WorkflowSpec(object):
     def parallelism(self, parallelism):
         # Set parallelism at Workflow level
         self.payload["parallelism"] = int(parallelism)
+        return self
+
+    def ttl_strategy(self, ttl_strategy={}):
+        if ttl_strategy != {}:
+            self.payload["ttlStrategy"] = ttl_strategy
         return self
 
     def pod_metadata(self, metadata):
--- a/metaflow/metaflow_config.py
+++ b/metaflow/metaflow_config.py
@@ -258,7 +258,14 @@ KUBERNETES_CONTAINER_IMAGE = from_conf(
 KUBERNETES_CONTAINER_REGISTRY = from_conf(
     "KUBERNETES_CONTAINER_REGISTRY", DEFAULT_CONTAINER_REGISTRY
 )
-
+# Remove job after a week by default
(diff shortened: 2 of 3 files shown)


shrinandj:
Thank you so much for doing this!

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -1113,6 +1116,11 @@ def parallelism(self, parallelism):
         self.payload["parallelism"] = int(parallelism)
         return self
 
+    def ttl_strategy(self, ttl_strategy={}):
+        if ttl_strategy != {}:
+            self.payload["ttlStrategy"] = ttl_strategy
+        return self
```
saikonen, line 1122: more generic and possibly not in scope of this PR, but how is the situation handled where a user sets a malformed ttlStrategy, and Argo (I would assume?) rejects the workflow. Is the raised error explicit enough?

saikonen: related to this feature as I believe the current setup allows to set arbitrary values for the ttlStrategy

saikonen (approved):

metaflow/plugins/argo/argo_workflows.py
jackie-ob: Good point.  I will try it out so we know.  I have a feeling the error from Argo is readable and we propagate it to the user in a good way.

savingoyal (changes requested):
@jackie-ob - it would be better to expose a TTL config key and apply it to all GC'able objects rather than treat these separately.

jackie-ob:
> @jackie-ob - it would be better to expose a TTL config key and apply it to all GC'able objects rather than treat these separately.

This means we would be trading off simplicity of a single "k8s gc config", vs having full control over Argo Workflows behavior, right?  Argo Workflows supports much more detailed GC configuration.  We are saying we will expose a lowest common denominator, right?  It would be `METAFLOW_KUBERNETES_OBJECTS_TTL_SECONDS`.

For Kubernetes it maps directly, but for Argo Workflows, we would have to give up disambiguation between successful workflows vs failed workflows.

tylerpotts:
@savingoyal I want to give a little bump on this

We had a job fail because of name collisions. In k8s we had accumulated ~3000 jobs

```
2023-05-01 16:32:20.968 [992/eval/10860 (pid 792139)] Task is starting.                                                                                                  
2023-05-01 16:32:23.113 [992/eval/10860 (pid 792139)] Traceback (most recent call last):                                                                                 
2023-05-01 16:32:23.197 [992/eval/10860 (pid 792139)] File "/opt/envs/shared/ds/lib/python3.10/site-packages/metaflow/plugins/kubernetes/kubernetes_cli.py", line 197, in
 step                                                                                                                                                                    
2023-05-01 16:32:23.197 [992/eval/10860 (pid 792139)] kubernetes.launch_job(                                                                                             
2023-05-01 16:32:23.198 [992/eval/10860 (pid 792139)] File "/opt/envs/shared/ds/lib/python3.10/site-packages/metaflow/plugins/kubernetes/kubernetes.py", line 128, in lau
nch_job
2023-05-01 16:32:23.198 [992/eval/10860 (pid 792139)] self._job = self.create_job(**kwargs).execute()
2023-05-01 16:32:23.198 [992/eval/10860 (pid 792139)] File "/opt/envs/shared/ds/lib/python3.10/site-packages/metaflow/plugins/kubernetes/kubernetes_job.py", line 215, in
 execute
2023-05-01 16:32:23.198 [992/eval/10860 (pid 792139)] raise KubernetesJobException(
2023-05-01 16:32:23.198 [992/eval/10860 (pid 792139)] metaflow.plugins.kubernetes.kubernetes_job.KubernetesJobException: Unable to launch Kubernetes job.
2023-05-01 16:32:23.198 [992/eval/10860 (pid 792139)] jobs.batch "t-d5jts" already exists, the server was not able to generate a unique name for the object
2023-05-01 16:32:23.245 [992/eval/10860 (pid 792139)] Task failed.
2023-05-01 16:32:23.396 This failed task will not be retried.
```

Because of the way that the job failed, despite having the @retry decorator the job didn't retry

savingoyal:
@tylerpotts would it be helpful to have a shorter TTL which is kubernetes wide at the cluster level instead? That will ensure that all pods are accounted for with a short TTL and not just jobs launched through a specific version of Metaflow.

savingoyal:
We can close out this PR and create a new one that only takes care of pod-level TTLs for a variety of objects (k8s jobs, workflow executions, sensor pods etc.)

Issue #1303, opened 2023-03-14
introduce `terminate` and `disable` commands for `argo-workflows`


savingoyal:

savingoyal:
while at it, maybe do delete as well

saikonen:
the `disable` feature will require some further thought around the UX, as it opens up a lot of questions on the desired behavior. A non-exhaustive list of open questions

- with the upcoming Sensors for workflows, what does it mean to `disable` a workflow?
- should a workflow that is disabled still be able to be triggered manually? can we even prevent this without deleting the workflow?
- should we have a symmetric command to re-enable a workflow? having to rely on `create` can be a bit confusing
- wording of the command in case of offering an enable as well is up for debate. Some ideas include separate `enable/disable` commands, or affecting the status with a flag

talebzeghmi:
It would be great to also have this feature in the Metaflow UI!
However the Metaflow UI is only cognizant of the running Flow once the Argo step has started...

saikonen:
`disable` was superseded by `suspend/unsuspend`. These are now available since https://github.com/Netflix/metaflow/releases/tag/2.9.7

PR #1307, opened 2023-03-20
feature: add delete and terminate for argo workflows

saikonen:
- adds a delete command for argo-workflows
- outputs the outcome of the delete and possible errors
- checks the validity of the production token before trying a delete

relates to #1303

- also adds terminate command for argo-workflows


metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -511,3 +501,81 @@ def _convert_value(param):
...
+    "--authorize",
+    default=None,
+    type=str,
+    help="Authorize the deletion with a production token",
+)
+@click.pass_obj
+def delete(obj, authorize=None):
+    validate_token(obj.workflow_name, obj.token_prefix, obj, authorize, "delete")
+    obj.echo("Deleting workflow *{name}*...".format(name=obj.workflow_name))
+    schedule_deleted = ArgoWorkflows.remove_schedule(obj.workflow_name)
```
jackie-ob, line 517: Might need to account for any `@trigger` or `@trigger_on_finish` resources (Argo Events `Sensor` objects in k8s).  Not sure if this PR merges first or last vs that [PR](https://github.com/Netflix/metaflow/pull/1271).

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -511,3 +501,81 @@ def _convert_value(param):
...
+    if response:
+        obj.echo(
+            "   Workflow is being deleted\n"
+            "Deletion of resources might take a while. "
+            "Deploying a flow with the same name during this time "
+            "will fail until the deletion has completed."
+        )
+
+
+def validate_token(name, token_prefix, obj, authorize, cmd_name):
```
jackie-ob, line 531: Is there any commonality to be had between this function and existing `resolve_token()` [here](https://github.com/Netflix/metaflow/blob/09a2821cb143bbe283f6b70c6d807afdfad48ec5/metaflow/plugins/argo/argo_workflows_cli.py#L378)?

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -143,6 +143,37 @@ def _sanitize(name):
         # allowed by Metaflow - guaranteeing uniqueness.
         return name.replace("_", "-")
 
+    @classmethod
+    def remove_schedule(cls, name):
```
jackie-ob, line 147: Nit: I guess it's an existing pattern - could be `@staticmethod` instead - we don't use `cls` here (but we don't use `cls` for other existing methods either.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -143,6 +143,37 @@ def _sanitize(name):
...
+        client = ArgoClient(namespace=KUBERNETES_NAMESPACE)
+
+        try:
+            response = client.delete_workflow_schedule(name)
+        except Exception:
```
jackie-ob, line 152: `as e`

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -143,6 +143,37 @@ def _sanitize(name):
...
+    def delete(cls, name):
+        client = ArgoClient(namespace=KUBERNETES_NAMESPACE)
+
+        try:
+            response = client.delete_workflow_template(name)
+        except Exception as e:
+            raise ArgoWorkflowsException(repr(e))
+        if response is None:
+            raise ArgoWorkflowsException(
+                "The workflow *%s* doesn't exist "
```
jackie-ob, line 171: This messaging feels a bit weird.  I don't think users want to fix the deletion error by first deploying it :)

metaflow/plugins/argo/argo_client.py
```diff
@@ -87,6 +87,44 @@ def register_workflow_template(self, name, workflow_template):
...
+            return client.CustomObjectsApi().delete_namespaced_custom_object(
+                group=self._group,
+                version=self._version,
+                namespace=self._namespace,
+                plural="workflowtemplates",
+                name=name,
+            )
+        except client.rest.ApiException as e:
+            if e.status == 404:
+                return None
```
jackie-ob, line 122: I can see that later, `None` shall be interpreted as "not found", and handled by caller.  May be we should document this somewhat at the top of the function, to make it clear to future developers?

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -511,3 +501,81 @@ def _convert_value(param):
...
+    default=None,
+    type=str,
+    help="Authorize the deletion with a production token",
+)
+@click.pass_obj
+def delete(obj, authorize=None):
+    validate_token(obj.workflow_name, obj.token_prefix, obj, authorize, "delete")
+    obj.echo("Deleting workflow *{name}*...".format(name=obj.workflow_name))
+    schedule_deleted = ArgoWorkflows.remove_schedule(obj.workflow_name)
+    if schedule_deleted:
```
jackie-ob, line 518: Should we add a short comment explaining why it's OK to just swallow schedule deletion failure?  I am guessing for these reasons:
* We want to proceed to template deletion, even if `CronWorkflow` already got deleted some other way (or by someone else).
* Users might run this twice (after seeing message that it might take a while, etc.)
* Even if `CronWorkflow` deletion fails and the object sticks around, it is not a big deal, as long as WorkflowTemplate gets deleted below, anyway.

In both cases we don't want to worry the end user.  I used up some brain juice to consider it... can spell it out for future developers!

metaflow/plugins/argo/argo_workflows_cli.py
saikonen: if the other PR gets merged before this, I'll account for the teardown of the `Sensor` as well. Any thoughts on if it can be treated the same way as the cronworkflow deletion, or should errors on sensor deletion be blocking?

metaflow/plugins/argo/argo_workflows.py
saikonen: good call. I went with the existing pattern but see no reason not to break the habit here, so changed to staticmethods instead :)

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -511,3 +501,90 @@ def _convert_value(param):
...
+        obj.echo("   Removed schedule")
+
+    # After cleaning up related resources, delete the workflow in question.
+    # Failure in deleting is treated as critical and will be made visible to the user
+    # for further action.
+    workflow_deleted = ArgoWorkflows.delete(obj.workflow_name)
+    if workflow_deleted:
+        obj.echo(
+            "   Workflow is being deleted\n"
```
shrinandj, line 533: A running workflow object would be present in the cluster if a cronworkflow is running when the delete command is issued. That running workflow will neither be interrupted nor deleted with this `delete` command.

Should we add that info to this message? Something like:

```
Deletion of resource might take a while.
Currently running workflows and previously run workflows may have to be deleted manually.
Deploying a flow ...
```

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -143,6 +143,34 @@ def _sanitize(name):
...
+            return False
+
+        return True
+
+    @staticmethod
+    def delete(name):
+        client = ArgoClient(namespace=KUBERNETES_NAMESPACE)
+
+        try:
+            response = client.delete_workflow_template(name)
```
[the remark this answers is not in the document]
saikonen: from testing it out, the workflow seems to finish with no problem even if the template is deleted mid-flight.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -143,6 +143,34 @@ def _sanitize(name):
         # allowed by Metaflow - guaranteeing uniqueness.
         return name.replace("_", "-")
 
+    @staticmethod
+    def remove_schedule(name):
+        client = ArgoClient(namespace=KUBERNETES_NAMESPACE)
+
+        try:
+            response = client.delete_cronworkflow(name)
```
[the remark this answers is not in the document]
saikonen: also cleaned up some of the try/except from this file as unnecessary, as the important errors are handled inside `argo_client.py` anyway, and I found out that any error inheriting from `MetaflowException` does not actually display meaningful information other than the classname when passed into `repr(e)`

metaflow/plugins/argo/argo_workflows_cli.py
saikonen: added a line to the output describing the effect that delete has on running workflows so the outcome should be more clear now.

metaflow/plugins/argo/argo_workflows.py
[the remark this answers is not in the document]
shrinandj: 👍

shrinandj (approved):

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -521,3 +511,94 @@ def _convert_value(param):
...
+    obj.echo("Deleting workflow *{name}*...".format(name=obj.workflow_name))
+
+    # Always try to delete the schedule. Failure in deleting the schedule should not
+    # be treated as an error, due to any of the following reasons
+    # - there might not have been a schedule, or it was deleted by some other means
+    # - retaining these resources should have no consequences as long as the workflow deletion succeeds.
+    # - regarding cost and compute, the significant resources are part of the workflow teardown, not the schedule.
+    schedule_deleted = ArgoWorkflows.remove_schedule(obj.workflow_name)
+    if schedule_deleted:
+        obj.echo("   Removed schedule")
```
savingoyal, line 535: Can we instead mention which cronworkflow was deleted here. Maybe just `Deleting cronworkflow foo as well...` without the tabbed indent

savingoyal:
Screen Shot 2023-03-28 at 10 04 45 AM

Note the - `to None this flow`

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -143,6 +143,50 @@ def _sanitize(name):
         # allowed by Metaflow - guaranteeing uniqueness.
         return name.replace("_", "-")
 
+    @staticmethod
+    def remove_schedule(name):
```
savingoyal, line 147: Can we move this inside the delete method in L:158 and move the comments about the behavior in cli.py to here?

PR #1308, opened 2023-03-21
feature: add disable for argo-workflows

saikonen:
- adds ability to suspend CronWorkflow schedule for argo-workflows

relates to #1303

### DIFF
--- a/metaflow/plugins/argo/argo_client.py
+++ b/metaflow/plugins/argo/argo_client.py
@@ -35,6 +35,23 @@ class ArgoClient(object):
                 json.loads(e.body)["message"] if e.body is not None else e.reason
             )
 
+    def get_cronworkflow(self, name):
+        client = self._kubernetes_client.get()
+        try:
+            return client.CustomObjectsApi().get_namespaced_custom_object(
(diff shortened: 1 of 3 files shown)


saikonen:
> What is the corresponding `enable` UX? I guess user would just do `create` again?

current approach would be that simply redeploying with `create` would re-enable the schedule for the workflow. A toggleable command for enabling/disabling the schedule would also be an option, maybe this warrants some further discussion on the approach.

jackie-ob:
> > What is the corresponding `enable` UX? I guess user would just do `create` again?
> 
> current approach would be that simply redeploying with `create` would re-enable the schedule for the workflow. A toggleable command for enabling/disabling the schedule would also be an option, maybe this warrants some further discussion on the approach.

Agree on further consideration.  Another one: enable/disable - does that apply to `@schedule` only?  AND/OR `@trigger` AND/OR `@trigger_on_finish`?  Is enable/disable state persisted across "create" ops?  This feature seems to be a different beast to all-in delete, and terminate.

savingoyal:
@saikonen can you close this PR for now.

PR #1309, opened 2023-03-21
feature: add terminate for argo workflows

saikonen:
- adds `argo-workflows terminate` as a command, requires a run_id for an argo workflow

TODO:

- [ ] add checks before issuing a terminate to a run to verify the run can be terminated.


metaflow/plugins/argo/argo_client.py
```diff
@@ -152,6 +152,37 @@ def delete_workflow_template(self, name):
...
+                name=name,
+            )
+        except client.rest.ApiException as e:
+            raise ArgoClientException(
+                json.loads(e.body)["message"] if e.body is not None else e.reason
+            )
+        # TODO: perform check that the found workflow is able to be terminated. We do not want to patch failed / completed workflows.
+        try:
+            body = {"spec": workflow["spec"]}
+            body["spec"]["shutdown"] = "Terminate"
```
shrinandj, line 172: Can you confirm that `shutdown` is a required part of the spec?

shrinandj (approved):
Thanks for doing this!

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -194,6 +194,24 @@ def delete(name):
...
         return True
 
+    @staticmethod
+    def terminate(name, run_id):
+        client = ArgoClient(namespace=KUBERNETES_NAMESPACE)
+
+        # Verify that user is trying to terminate an Argo workflow
+        if not run_id.startswith("argo-"):
+            raise ArgoWorkflowsException(
+                "The run *%s* is not an Argo workflow." % run_id
```
savingoyal, line 204: nit - No active workflow for ${flow_name}/${run_id} found on Argo Workflows.

metaflow/plugins/argo/argo_client.py
```diff
@@ -152,6 +152,42 @@ def delete_workflow_template(self, name):
...
+                name=run_id,
+            )
+        except client.rest.ApiException as e:
+            raise ArgoClientException(
+                json.loads(e.body)["message"] if e.body is not None else e.reason
+            )
+
+        if workflow["status"]["finishedAt"] is not None:
+            raise ArgoClientException(
+                "Can not terminate a workflow that has already finished."
```
savingoyal, line 172: nit - `Cannot terminate an execution that has already finished.`

savingoyal (changes requested):
Added some minor nits. 

One major question - Can you verify that terminate will only ensure that no new tasks are launched, but existing tasks still continue? We may want to introduce a `kill` command if it is feasible to kill all the in-flight tasks.

metaflow/plugins/argo/argo_client.py
```diff
@@ -152,6 +152,42 @@ def delete_workflow_template(self, name):
...
+            )
+
+        try:
+            body = {"spec": workflow["spec"]}
```
jackie-ob, line 176: Minor thing: should we skip the patching if we see that shutdown is already set to Terminate as we wish?  This code path probably will get exercised - when users rerun stuff (expected delays / inflight tasks etc).

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -605,3 +605,26 @@ def disable(obj, authorize=None):
...
+    help="Specify the ID of the run to be terminated.",
+)
+@click.pass_obj
+def terminate(obj, run_id, authorize=None):
+    validate_token(obj.workflow_name, obj.token_prefix, obj, authorize, "terminate")
+    obj.echo("Terminating workflow *{run_id}*...".format(run_id=run_id))
+
+    terminated = ArgoWorkflows.terminate(obj.workflow_name, run_id)
+    if terminated:
+        obj.echo("   Terminated.")
```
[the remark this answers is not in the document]
jackie-ob: This is clearly a great feature as is.  What about tasks that already started?  I'd think that is the UX users actually want - though probably much more involved to implement.  Imagine a wide foreach that is inflight...

Wish-list for future (`--abort-running-tasks`)!

saikonen:
> Added some minor nits.
> 
> One major question - Can you verify that terminate will only ensure that no new tasks are launched, but existing tasks still continue? We may want to introduce a `kill` command if it is feasible to kill all the in-flight tasks.

the behavior from the Task perspective is more or less the same whether we use `Terminate` or `Stop` as the shutdown plan. All tasks are stopped and not allowed to finish executing.

metaflow/plugins/argo/argo_workflows_cli.py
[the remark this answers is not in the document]
saikonen: note that after issuing the terminate command, even currently running tasks will not be able to finish, as they are part of the workflow to be terminated.

saikonen:
rebased off the `delete` feature branch so we can move forward with those two features, and leave `disable` for further discussion later.

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -602,3 +602,30 @@ def echo_token_instructions(obj, name, prev_user, cmd_name, cmd_description=None
...
+    required=True,
+    type=str,
+    help="Specify the ID of the run to be terminated.",
+)
+@click.pass_obj
+def terminate(obj, run_id, authorize=None):
+    validate_token(obj.workflow_name, obj.token_prefix, obj, authorize, "terminate")
+    obj.echo(
+        "Terminating run *{run_id}* for {flow_name} ...".format(
+            run_id=run_id, flow_name=obj.flow.name
```
savingoyal, line 625: bold=True

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -602,3 +602,30 @@ def echo_token_instructions(obj, name, prev_user, cmd_name, cmd_description=None
...
+
+
+@argo_workflows.command(help="Terminate the Argo workflow")
+@click.option(
+    "--authorize",
+    default=None,
+    type=str,
+    help="Authorize the termination with a production token",
+)
+@click.option(
```
savingoyal, line 614: we can take an arg run id instead of an option run id here

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -165,6 +165,28 @@ def delete(name):
...
         return True
 
+    @staticmethod
+    def terminate(flow_name, run_id):
+        client = ArgoClient(namespace=KUBERNETES_NAMESPACE)
+
+        # Verify that user is trying to terminate an Argo workflow
+        if not run_id.startswith("argo-"):
+            raise ArgoWorkflowsException(
+                "No active workflow for {flow_name}/{run_id} found in Argo Workflows.".format(
```
savingoyal, line 175: No execution ... to keep in sync with L:184?

metaflow/plugins/argo/argo_client.py
savingoyal: +1

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -602,3 +602,30 @@ def echo_token_instructions(obj, name, prev_user, cmd_name, cmd_description=None
         'See "Organizing Results" at docs.metaflow.org for more information '
         "about production tokens."
     )
+
+
+@argo_workflows.command(help="Terminate the Argo workflow")
```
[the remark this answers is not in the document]
savingoyal: correct

PR #1314, opened 2023-03-23
fix: update cards dependencies

saikonen:
fixes security vulnerability with one of the dependencies of the `cards` plugin. See https://osv.dev/vulnerability/GHSA-f8q6-p94x-37v3 for details.

simply ran `yarn upgrade` and `yarn build`on the package. Still needs verifying that the plugin functions as expected.


valayDave:
I vaguely remember @seethroughdev telling me about the tests for the JS framework. @seethroughdev do you know where the tests are present so we can just run them once to verify if the PR doesn't break anything?

seethroughdev:
thanks for looking into this @saikonen . I'm not seeing a change to the yarn file, only yarn.lock, I think we should update both, that way we know we can regenerate a lockfile, and it will work.

also, I'm a bit surprised by changes in the css and js outputs. do you know change is causing them?

saikonen:
> I'm not seeing a change to the yarn file, only yarn.lock

I only ran `yarn upgrade` which updates dependencies to the latest versions allowed by the pinned versions found in package.json, didn't want to start off with minor/major version upgrades in fear of breaking changes :)

> also, I'm a bit surprised by changes in the css and js outputs. do you know change is causing them?

My best guess is that between versions the parser has had some changes and might just be ordering the output differently, hence the massive diff. the files are generated with `yarn build` so its all up to the build process

seethroughdev:
Yeah, `npm ci` is nice for that as your guaranteed packages won't shift.  Both things make sense. As long as you're not concerned with the diff change, then works for me!

Thanks.

seethroughdev (approved):
as noted in the discussion, the build output change is worth keeping an eye out for, but it sounds like @saikonen is on top of it!

PR #1320, opened 2023-03-25
[tracing] opentelemetry support

jackie-ob:
Refreshed version of https://github.com/jackie-ob/metaflow/pull/2 which was a stacked one.  This PR directly targets nflx master.

### DIFF
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -3,6 +3,7 @@ import sys
 import traceback
 from datetime import datetime
 from functools import wraps
+import metaflow.tracing as tracing
 
 from metaflow._vendor import click
 
@@ -796,6 +797,7 @@ def resume(
     runtime.execute()
 
 
+@tracing.cli_entrypoint("cli/run")
 @parameters.add_custom_parameters(deploy_mode=True)
 @cli.command(help="Run the workflow locally.")
 @common_run_options
@@ -894,6 +896,7 @@ def version(obj):
     echo_always(obj.version)
 
 
+@tracing.cli_entrypoint("cli/start")
 @decorators.add_decorator_options
 @click.command(
     cls=click.CommandCollection,
--- a/metaflow/metaflow_config.py
+++ b/metaflow/metaflow_config.py
@@ -347,6 +347,8 @@ if AWS_SANDBOX_ENABLED:
 
 KUBERNETES_SANDBOX_INIT_SCRIPT = from_conf("KUBERNETES_SANDBOX_INIT_SCRIPT")
 
+OTEL_ENDPOINT = from_conf("OTEL_ENDPOINT")
+
 # MAX_ATTEMPTS is the maximum number of attempts, including the first
 # task, retries, and the final fallback task and its retries.
 #
--- a/metaflow/plugins/metadata/service.py
+++ b/metaflow/plugins/metadata/service.py
@@ -59,7 +59,7 @@ class ServiceMetadataProvider(MetadataProvider):
     def compute_info(cls, val):
         v = val.rstrip("/")
         try:
-            resp = requests.get(os.path.join(v, "ping"), headers=SERVICE_HEADERS)
+            resp = requests.get(os.path.join(v, "ping"), headers=SERVICE_HEADERS.copy())
             resp.raise_for_status()
         except:  # noqa E722
             raise ValueError("Metaflow service [%s] unreachable." % v)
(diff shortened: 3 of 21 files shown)


jackie-ob:
Created this non-stacked PR to make it clearer.  Also @oavdeev  is working on some minor additional improvements, which we would like to combine into this PR.

metaflow/plugins/metadata/service.py
```diff
@@ -59,7 +59,7 @@ def __init__(self, environment, flow, event_logger, monitor):
     def compute_info(cls, val):
         v = val.rstrip("/")
         try:
-            resp = requests.get(os.path.join(v, "ping"), headers=SERVICE_HEADERS)
+            resp = requests.get(os.path.join(v, "ping"), headers=SERVICE_HEADERS.copy())
```
oavdeev, line 62: this and other header .copy() calls are a workaround for https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1729 , we can prob remove this once its fixed upstream

metaflow/tracing.py
saikonen, line 1: would it make sense to restructure this as `metaflow/tracing/__init__.py` and moving the noop/otel under it?

metaflow/tracing_propagator.py
saikonen, line 1: unsure if this is a vendor file or not due to the inline license. If its a vendor file then would it make sense to store it in `metaflow/_vendor/` ?

metaflow/tracing.py
```diff
@@ -0,0 +1,28 @@
+import sys
+from metaflow.metaflow_config import (
+    OTEL_ENDPOINT,
+)
+
+import metaflow.tracing_noop
```
saikonen, line 6: I believe this could be cleaned up a bit with import aliases
```
import metaflow.tracing_noop as tracing
...
import metaflow.tracing_otel as tracing
...
init_tracing = tracing.init_tracing
cli_entrypoint = tracing.cli_entrypoint
...
```
it has the added benefit of acting similar to an interface, throwing errors if the latest imported `tracing` does not implement all the functions that are being assigned later on.

metaflow/tracing_propagator.py
oavdeev: Its from an unmerged PR on upstream OTEL libraries, so its a little diffferent situation from `_vendor`. My interpretation of `_vendor` folder as a code reader is "ignore this, all this code could as well be a pip dependency but we vendored it for some logistical reason". Not the case here because this snippet is not from some released version of a public library, and I may have tweaked it a bit

metaflow/tracing.py
```diff
@@ -0,0 +1,28 @@
...
+
+        init_tracing = metaflow.tracing_otel.init_tracing
+        cli_entrypoint = metaflow.tracing_otel.cli_entrypoint
+        inject_tracing_vars = metaflow.tracing_otel.inject_tracing_vars
+        get_trace_id = metaflow.tracing_otel.get_trace_id
+        traced = metaflow.tracing_otel.traced
+        tracing = metaflow.tracing_otel.tracing
+        post_fork = metaflow.tracing_otel.post_fork
+    except ImportError:
+        pass
```
saikonen, line 28: no objections at catching the `ImportError` per se, but this bit me while testing as passing the error made it hard to debug why data was not being emitted when `OTEL_ENDPOINT` was clearly set.

Reason: I was missing relevant opentelemetry libraries:

- `opentelemetry-instrumentation-otlp`
- `opentelemetry-exporters-otlp`
- `opentelemetry-sdk`
- `opentelemetry-api`

metaflow/tracing.py
wangchy27: Agree that it would be better to design an interface for all tracing endpoint. In the current design, we need to copy/paste line 16 to 26 to add another tracing plugin (e.g. ZIPKIN_ENDPOINT)

metaflow/tracing.py
wangchy27: Feel like refactor all tracing related files to `metaflow/tracing/` directory could make the code easier to organize, especially if we support multiple endpoint. I'm open to keep the current  `tracing.py` or make it an `__init__.py` as Sakari suggested

metaflow/tracing_otel.py
```diff
@@ -0,0 +1,170 @@
...
+        print("Tracing already initialized", file=sys.stderr)
+        return
+    # print("initializing tracing", os.environ.get("traceparent"))
+
+    from metaflow.tracing_propagator import EnvPropagator
+
+    set_global_textmap(EnvPropagator(None))
+
+    if SERVICE_AUTH_KEY:
+        span_exporter = OTLPSpanExporter(
```
wangchy27, line 39: Open discussion: Is it possible to make the propogator and exporter configurable?
For example, if the user want to use ZIPKIN, they're supposed to use `opentelemetry.propagators.b3.B3MultiFormat` propagator, and `opentelemetry.exporter.zipkin.proto.http.ZipkinExporter`, which may add multiple if/else in the script, something like:
```
if otel:
    use propagator A, exporter A
if zipkin:
    use propagator B, exporter B
if xxx:
    use propagotor C, exporter C
```
Just wondering if we can offer a cleaner way to configure propagator and exporter

metaflow/metaflow_config.py
```diff
@@ -347,6 +347,8 @@
 
 KUBERNETES_SANDBOX_INIT_SCRIPT = from_conf("KUBERNETES_SANDBOX_INIT_SCRIPT")
 
+OTEL_ENDPOINT = from_conf("OTEL_ENDPOINT")
```
saikonen, line 350: Setting this to an invalid OTEL URL will lead to a lot of the following error
`Failed to export batch code: 404, reason: 404 page not found`

and if the URL is completely unreachable, full stack traces from `requests` will be flooding the logs.

this is sort of expected behavior, but the errors could be cleaned up a bit.

metaflow/tracing_otel.py
oavdeev: Generally OTEL SDK supports [doing this through environment variables](https://opentelemetry.io/docs/concepts/sdk-configuration/general-sdk-configuration/). Usually when instrumenting services, like web backends, I'd just rely on those. 

However in Metaflow case, its hard to expect end users to set all those correctly and making sure those are consistently set across all subproces/task boundaries is pretty error prone. So we opted for not using this mechanism to simplify the implementation. 

Maybe initially we could support a couple combinations that are most practically useful (lets say OTEL HTTP & Zipkin) via a metaflow config var, something like METAFLOW_OTEL_MODE=zipkin and come back to this later if more flexibility is necessary.

metaflow/cli.py
```diff
@@ -894,6 +896,7 @@ def version(obj):
     echo_always(obj.version)
 
 
+@tracing.cli_entrypoint("cli/start")
```
savingoyal, line 899: we should also enable `--enable-tracing` or some such flag at the top level

metaflow/tracing.py
savingoyal: we could also make `tracing` a plugin like datastore and metadata

PR #1323, opened 2023-03-27
allow disabling dependency installation 

amerberg:
Currently Metaflow installs some dependencies at run time, which can increase latency. For production use cases, it will usually be preferable to have these dependencies pre-installed in a Docker image. This PR adds an environment variable `METAFLOW_INSTALL_DEPENDENCIES` to allow disabling the dependency installation.


savingoyal:
@amerberg it will be useful to understand if we can make the pip check quick. That direction will reduce the odds of a misconfigured config failing the workflow when the dependencies are not installed.

PR #1328, opened 2023-03-29
 Add docstring style to contribution code style guide

jimbudarz:
Since it appears the docstring standard has been chosen (correct me if I'm wrong), it would help contributors to have this information available alongside the specification of black formatting.


savingoyal:
@jimbudarz would you be open to moving the entire `code style` section to contributing.md?

jimbudarz:
I think that's a great idea, tbh. As long as it's all in one place, I'm happy.

savingoyal:
Great! LGTM, except we don't use sphinx but nbdoc for generating docs for metaflow. Here is an example - https://github.com/Netflix/metaflow-docs/blob/master/docs/api/current.ipynb. If you change the ref to nbdoc - we should be good to merge.

savingoyal:
@jimbudarz can you push a null commit to the repo to tickle the pre-commit check? GH actions were down today and it is likely that's the reason we are not seeing the checks make progress at this point.

Issue #1318, opened 2023-03-24
2 minute pause when running a kubernetes step


ianmoss:
When running a step on kubernetes there is a ~2 minute pause during the step execution. It is very apparent in the output on the screen as the log time has a consistent 2 minutes 11 seconds gap during the kubernetes step.

Having investigated the issue its much easier to highlight the line that is causing the pause than describe how to reproduce.

https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/kubernetes/kubernetes_decorator.py#L330

This is waiting for a response from the ec2 instance identity api

https://github.com/Netflix/metaflow/blob/230fc73eb04553cf08d3177cd14ab49b9b4e2859/metaflow/plugins/aws/aws_utils.py#L24

I am using a local kubernetes node so its no real surprise that this api is not responding. I am not totally sure why it retries for 2 minutes but I could likely fix this with some internal network configs - for now I have just commented out getting the ec2 metadata.

The comments say "end-point might be blocked on AWS and not available for non-AWS deployments" so seem to accept that this line maybe run on a non AWS instance.

Either I am missing something fundamental that I have got wrong in the configuration or I need a means to disable off get_ec2_instance_metadata(). Ideally it would be automatic if the node is not on an ec2 instance, or may be it only executes if it is on a AWS instance?

I'll update here if I find an elegant solution, I'd like to hear others thoughts.

saikonen:
Could you check out the quick fix PR I opened regarding this issue. The only change is that it reduces the timeout for the requests significantly, which should get the overhead of running kubernetes steps on non-AWS environments down to 1 sec.

Admittedly this is only half of the solution, as noted in the comments, the instance identity service might be inaccessible even when running on AWS. For the time being we need to try the EC2 metadata request on every kubernetes run as we do not know which cloud provider (if any) is being used.

Going forward we would want to add some way to detect the cloud provider in order to defer from performing unnecessary requests like this, especially since there are plans to fetch instance metadata on GCP and Azure as well.

ianmoss:
I have checked it out, it works a treat when using with local kubernetes I'll certainly vote for it to be merged.

From what I have read, this is the most reliable way of determining if a server is an EC2 instance, remembering to set a timeout.

saikonen:
closing this issue as completed for now, and opening another for future improvements regarding fetching cloud-agnostic instance metadata in a kubernetes native way.

PR #1335, opened 2023-03-30
fix: kubernetes ec2 instance metadata timeout 

saikonen:
- add a more aggressive timeout to fetching ec2 instance metadata to the helper. This should alleviate the experienced delays running on kubernetes in a non-AWS environment

relates to #1318


metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -327,6 +327,8 @@ def task_pre_step(
 
             # TODO (savin): Introduce equivalent support for Microsoft Azure and
             #               Google Cloud Platform
+            # TODO: Introduce a way to detect Cloud Provider, so unnecessary requests (and delays)
```
shrinandj, line 330: Two other options to consider:
1. The information obtained from the metadata service here is also available in K8s. One option could be to get that info from K8s and avoid making this call. It will also make the implementation cloud agnostic.
2. Is it possible to make this call async? Since the info here is not directly needed for the actual task function to execute, it could be called in async.

WDYT?

oavdeev: To add to this, there is a decent chance that you won't be able to call this endpoint on EKS at all, depending on node group configuration, specifically IMDS settings (IMDSv1 vs IMDSv2 and the hop limit). 

But its a known issue. When this call was introduced it was considered a "best effort" thing, longer term would indeed be better to get this info in more reliable, K8S-native way.

ianmoss: This is great for now. 

Call should probably be async, it is a completely independent process. It should still need a relatively aggressive timeout though.

Ideally K8s should be responsible for the metadata of the node the step is run on to create a consistent way of working across all infrastructure setups.

Perhaps then making a flag to optionally get AWS EC2 host metadata directly.

savingoyal (approved):

PR #1338, opened 2023-04-01
Fix: Check all steps for MetaflowCode and return if any

### LINKED ISSUE #1322: run.code should return the code package from any step, not just start step
currently `run.code` returns `None` if the `start` step executes locally if even if other steps executed remotely. The code package is the same for all steps, so `run.code` could return the code package from any step where it is available

bsridatta:
Loops over steps of the run and return step.task.code if present.

Test flow with intermediate non-local step.

```
from metaflow import step, FlowSpec, conda_base, kubernetes


class HelloFlow(FlowSpec):
    @step
    def start(self):
        print("Starting 👋")
        self.next(self.eat)

    @kubernetes
    @step
    def eat(self):
        print("Eating 🍜")
        self.next(self.end)

    @step
    def end(self):
        print("Done! 🏁")


if __name__ == "__main__":
    HelloFlow()
```

Checking code with metadata

```
In [1]: from metaflow import Flow

In [2]: run = Flow("HelloFlow").latest_successful_run

In [3]: run.code
Out[3]: <metaflow.client.core.MetaflowCode at 0x7f28273010a0>
```

closes #1322

### DIFF
--- a/metaflow/client/core.py
+++ b/metaflow/client/core.py
@@ -669,7 +669,7 @@ class MetaflowData(object):
 class MetaflowCode(object):
     """
     Snapshot of the code used to execute this `Run`. Instantiate the object through
-    `Run(...).code` (if all steps are executed remotely) or `Task(...).code` for an
+    `Run(...).code` (if any step is executed remotely) or `Task(...).code` for an
     individual task. The code package is the same for all steps of a `Run`.
 
     `MetaflowCode` includes a package of the user-defined `FlowSpec` class and supporting
@@ -1648,16 +1648,17 @@ class Run(MetaflowObject):
     def code(self) -> Optional[MetaflowCode]:
         """
         Returns the MetaflowCode object for this run, if present.
-
-        Not all runs save their code so this call may return None in those cases.
+        Code is packed if atleast one `Step` runs remotely, else None is returned.
 
         Returns
         -------
         MetaflowCode, optional
             Code package for this run
         """
-        if "start" in self:
-            return self["start"].task.code
+        for step in self:
+            code = step.task.code
+            if code:
+                return code
 
     @property
     def data(self) -> Optional[MetaflowData]:


bsridatta:
I have updated the docstrings for `MetaflowCode` and `Run`. I searched the repo and did not find other mentions of the requirement for `Run(...).code` to work.  Not sure about test cases given that steps need to run remotely.

saikonen (dismissed):

metaflow/client/core.py
```diff
@@ -1648,16 +1648,17 @@ def steps(self, *tags: str) -> Iterable[Step]:
...
+        Code is packed if atleast one `Step` runs remotely, else None is returned.
 
         Returns
         -------
         MetaflowCode, optional
             Code package for this run
         """
-        if "start" in self:
-            return self["start"].task.code
+        for step in self:
```
saikonen, line 1658: Just a testing note: was wondering if this can be extremely slow with very tall Flows where the remote step is at the very end, but seems to work fine possibly due to `task.code` being a cached property.

tried this out with a flow with over a 100 local tasks before the first remote one and `run.code` still returns within 3 seconds, though it did not get any faster with subsequent calls.

saikonen: I assume we do not want the run level `code` property to be cached to avoid users accessing this for in-flight flows and getting `None` cached in some cases?

bsridatta: I havent thought about that and honestly dont understand it to that point. Is this something that should be explicitly handled?

saikonen (approved):

metaflow/client/core.py
bsridatta: That would be slow indeed. What are your thoughts making the MetaflowTask to add the code package info to the Run metadata, since its the same for all steps. And you wouldnt have to iterate over all steps. 

I have also seen discussions about packing the code for local only runs too, then this probably not be required?

bsridatta:
@saikonen Any idea how to fix these "transient s3" errors, is it just rerunning them?

saikonen:
> @saikonen Any idea how to fix these "transient s3" errors, is it just rerunning them?

usually rerunning them works, as long as the failure is not code-related. Let's see if this time we get the run to pass, but I've seen a lot of failures regarding the 5gb s3 test lately so that might need some looking into if it keeps being an issue. Not in the scope of this PR though

romain-intel:
As a quick comment, the transient failures are by design (we "simulate" transient failures to test the retry mechanism). They should not be causing permanent failures typically and looking at the error here, it looks to be something else (not sure what though but it says Invalid Argument). A transient failure is retried up to 10 times. Here you can see that the previous test with transient failures seems to pass but the next one does not for some reason.

Issue #1114, opened 2022-08-18
Support auto-completion of Metaflow objects/API in notebooks


romain-intel:
We could support some form of auto-completion for Metaflow in iPython. Some examples:
```
def _ipython_key_completions_(self):
    return [s.id for s in self.steps()]
```
on a run to complete something like `my_run['<tab>`

Details on the possibilities here: https://ipython.readthedocs.io/en/stable/config/integrating.html#tab-completion

bsridatta:
Hey @romain-intel can I take this up?

I think this would apply flow, all the flow components from Flow to task right?

romain-intel:
Of course -- I am not currently working on it. But yes, it would apply to the objects in the client/core.py file so flows to artifacts.

bsridatta:
Got it, will send a PR soon!

bsridatta:
@romain-intel Should this be closed or kept open for adding autocomplete during object creation?

avinabneogy23:
Hi @romain-intel and @bsridatta ,
Is this issue still open
Thanks!

KrishnaKapoor612:
It looks like instance-level autocomplete support was merged in #1348.

Is there any remaining scope for this issue,
perhaps around constructor-level completion?
If not, should this be closed?

Ronit-Raj9:
Hi @kKAPS22, @avinab-neogy, and @romain-intel!

The core functionality here was merged in PR #1348, which added `_ipython_key_completions_` to `MetaflowObject` — so `flow["<tab>`, `run["<tab>`, `step["<tab>`, and `task["<tab>` all autocomplete children IDs. Constructor-level completion (`Flow("<tab>`) isn't feasible in IPython, as discussed in the PR.

I did notice two minor remaining gaps:

- **`Metaflow` class** — doesn't inherit from `MetaflowObject`, so `Metaflow()["<tab>` won't suggest flow names. Adding `_ipython_key_completions_` there is a 3-line fix.
- **`MetaflowData` class** — `task.data.<tab>` won't suggest artifact names because `MetaflowData` doesn't override `__dir__()`. Adding it would make `task.data.my_artifact` discoverable via tab completion.

Both are trivial fixes. Happy to submit a quick PR if there's interest!

KrishnaKapoor612:
Thanks @Ronit-Raj9 for outlining the remaining gaps clearly!

The fixes sound straightforward and valuable for improving the notebook experience.
[...]

PR #1348, opened 2023-04-06
feat: add ipython autocomplete support for MetaflowObject instances

bsridatta:
relates to #1114 

The autocomplete for objects works with this simple addition of returning children ids. Suggestions when initiating classes like `Flow("` could be helpful too but I couldn't find a way to do this. But for the class instances of Flow, Run, Step, Task, the autocomplete works and for reference the options are printed by calling the function that ipython would use.

Any thoughts @romain-intel ?

```
from metaflow import Flow, Metaflow

Metaflow().flows
>>> [Flow('HelloFlow'), Flow('MovieStatsFlow')]

flow = Flow('HelloFlow') # No autocomplete here
flow._ipython_key_completions_()
>>> 
['1680815181013681',
 '1680815178214737',
 '1680432265121345',
 '1680430310127401']

run = flow["1680815178214737"]
run._ipython_key_completions_()
>>> ['end', 'hello', 'start']

step = run["hello"]
step._ipython_key_completions_()
>>> ['2']

task = step["2"]
task._ipython_key_completions_()
>>> ['name']

data = task["name"]
data._ipython_key_completions_()
>>> Error (since no childern), so autcomplete just list unrelated things
```

### DIFF
--- a/metaflow/client/core.py
+++ b/metaflow/client/core.py
@@ -384,6 +384,10 @@ class MetaflowObject(object):
             if all(tag in child.tags for tag in tags):
                 yield child
 
+    def _ipython_key_completions_(self):
+        """Returns available options for ipython auto-complete."""
+        return [child.id for child in self._filtered_children()]
+
     @classmethod
     def _url_token(cls):
         return "%ss" % cls._NAME


savingoyal (approved):
lgtm! @romain-intel any thoughts here?

romain-intel:
This looks great. I did do some research about adding in constructor but yes, it's not straightforward so we can merge this. Thanks for the contribution.

Issue #948, opened 2022-02-13
Pathspec to create Flow/Run/Step/Task/DataArtifact is not validated


romain-intel:
Currently, you can do something like this: `Task(Flow/RunID/StepName)` and this will not result in an error but then the resulting `Task` object behaves in a bizarre manner where things like `t.data` will work but `t.data.my_artifact` will not for example.

We should validate the format of the pathspec passed in to each object and verify that the following are the only possible cases:
- Metaflow()
- Flow('MyFlow')
-  Run('MyFlow/1')
- Step('MyFlow/1/start')
- Task('MyFlow/1/start/1')
- DataArtifact('MyFlow/1/start/1/name')

We should make sure that `MyFlow` and `MyFlow/` behave similarly for example though.

soma2000-lang:
Working on this@romain-intel

romain-intel:
Awesome! Thanks so much.

bsridatta:
@romain-intel Is this issue still valid? Was trying to have a look and 

``` Task("MovieStatsFlow/1680379703087709/start")```

resulted in an error

```KeyError: 'task_id'```


Please let me know. Thanks!

**Comment by romain-intel** [comment]
I haven't checked recently. Did you try things like:
```
Run("MovieStatsFlow/123/start/")
Step("MovieStatsFlow/123/start/456")
Task("MovieStatsFlow/123/start/")
```
and other combos (basically with and without the trailing / and too short or too long. I suspect some combinations that are invalid are still working.

**Comment by bsridatta** [comment]
Got it, Passing a step to a run works, task to a step works, which is not expected and should be disabled. Thanks for the clarification, this is indeed confusing for new developers. I am picking up another ticket and will get to this after that, thank you!

```
In [19]: Run("HelloFlow/1/start/")
Out[19]: Run('HelloFlow/1/start/')

In [20]: Step("HelloFlow/1/start/2")
Out[20]: Step('HelloFlow/1/start/2')

In [21]: Task("HelloFlow/1/start/")
---------------------------------------------------------------------------
MetaflowNotFound                          Traceback (most recent call last)
...
MetaflowNotFound: Task('HelloFlow/1/start/') does not exist
```

KrishnaKapoor612:
Thanks for the clarification!

It sounds like enforcing numeric-only Run IDs and Task IDs might be too restrictive,especially if non-numeric IDs are valid in some backends.

Perhaps the validation could focus on:
- Ensuring the correct number of path components
[...]

PR #1350, opened 2023-04-09
Fix: Validate pathspec argument for MetaflowObject

bsridatta:
related to #948 

- Adds `MetaflowInvalidPathspec` exception
- Validates `pathspec` for all 5 MetaflowObjects - Flow to Artifact
- Changes unnecessary id re-assignment for MetaflowObjects created using `pathspec`

I have looked into writing test case but couldn't see that there are any specifically for testing metadata. Please let me know. Thanks

### DIFF
--- a/metaflow/client/core.py
+++ b/metaflow/client/core.py
@@ -22,6 +22,7 @@ from metaflow.exception import (
     MetaflowNotFound,
     MetaflowNamespaceMismatch,
     MetaflowInternalError,
+    MetaflowInvalidPathspec,
 )
 from metaflow.includefile import IncludedFile
 from metaflow.metaflow_config import DEFAULT_METADATA, MAX_ATTEMPTS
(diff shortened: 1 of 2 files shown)


bsridatta:
The MetaflowObject class has the parameters `_object` and `_parent` with the type `Optional[MetaflowObject]`. But from the tests it seems to take a dictionary and not an object, causing them to fail.

I have removed this check, but the typehint seems a bit misleading. Do you think that should be changed?

metaflow/client/core.py
```diff
@@ -290,23 +291,40 @@ def __init__(
         if pathspec:
             ids = pathspec.split("/")
 
+            if self._NAME == "flow" and len(ids) != 1:
+                raise MetaflowInvalidPathspec("Expects Flow('FlowName')")
+            elif self._NAME == "run" and len(ids) != 2:
+                raise MetaflowInvalidPathspec("Expects Run('FlowName/RunID')")
+            elif self._NAME == "step" and len(ids) != 3:
+                raise MetaflowInvalidPathspec(
+                    " Expects Step('FlowName/RunID/StepName')"
```
romain-intel, line 300: nit: addl space in front of `Expects` -- same in the next 2 instances.

romain-intel (commented):
LGTM but I want to verify something with the id and object particularly with the notion of run id and run number. Let me check a few things and get back to you shortly.

metaflow/client/core.py
bsridatta: Fixed, thanks!

metaflow/client/core.py
```diff
@@ -290,23 +291,38 @@ def __init__(
...
+                raise MetaflowInvalidPathspec("Expects Step('FlowName/RunID/StepName')")
+            elif self._NAME == "task" and len(ids) != 4:
+                raise MetaflowInvalidPathspec(
+                    "Expects Task('FlowName/RunID/StepName/TaskID')"
+                )
+            elif self._NAME == "artifact" and len(ids) != 5:
+                raise MetaflowInvalidPathspec(
+                    "Expects DataArtifact('FlowName/RunID/StepName/TaskID/ArtifactName')"
+                )
+
```
saikonen, line 308: Tried this out and seems to be working with the happy path. Original issue mentioned that pathspec should work with a trailing slash as well unless there's been a decision not to support this.

With the current implementation, a trailing `/` in the pathspec will raise a `MetaflowInvalidPathspec` even though the pathspec is otherwise valid

savingoyal: we can punt adding support for trailing slash for now (to account for the scenario where we may want to attach a special meaning to trailing slash - highly unlikely)

metaflow/exception.py
```diff
@@ -91,6 +91,13 @@ def __init__(self, namespace):
         super(MetaflowNamespaceMismatch, self).__init__(msg)
 
 
+class MetaflowInvalidPathspec(MetaflowException):
+    headline = "Invalid pathspec for the MetaflowObject type"
```
savingoyal, line 95: Can we tweak this to just read - `Invalid pathspec`? Otherwise the PR is ready to be shipped.

PR #1353, opened 2023-04-11
Kubernetes running job cancellation to fallback to patching parallelism

jackie-ob:


jackie-ob:
Tested cancellation by hand in environment that does not support the old killall5 path (websocket connection not allowed).

savingoyal (approved):

PR #1360, opened 2023-04-18
Add labels and fix argo

dhpollack:
Revises: #1236

Fixes error related to null labels trying to get added to a dictionary in argo workflows.

Refactored based on comments from @saikonen 

Now one can only add labels via the environmental variable `METAFLOW_KUBERNETES_LABELS` with the format `label=val,label2=val2`.  Json is no longer supported.  And the decorator does not allow labels either.


metaflow/plugins/argo/argo_workflows.py
```diff
@@ -1062,7 +1064,7 @@ def label(self, key, value):
     def labels(self, labels):
         if "labels" not in self.payload:
             self.payload["labels"] = {}
-        self.payload["labels"].update(labels)
+        self.payload["labels"].update(labels or {})
```
dhpollack, line 1067: This should solve the issue in the previous PR.  Do you have an example of a command that can reproduce the error and then I can check to see if this works?

dhpollack:
fyi, I test this with a vanilla `argo-workflows create` command and everything was fine.

saikonen (commented):
As an additional note, the newly added Sensors for argo-workflows from #1271 would also require labeling. If you have time, this could be done in the scope of this PR.

dhpollack:
> As an additional note, the newly added Sensors for argo-workflows from #1271 would also require labeling. If you have time, this could be done in the scope of this PR.

I personally would like to get this merged as quickly as possible.  I can look into this but I haven't used Sensors so I would also need to look into that.  If it's easy, I can do it, but otherwise I would like to create a new PR.

metaflow/plugins/argo/argo_workflows.py
saikonen: seems I'm late to the party, but as you noted in another comment, the problematic case was `argo-workflows create` without passing any labels. This seems to be working correctly now

saikonen:
We can do another PR for the sensors, they are just another template that is being compiled as part of `argo-workflows create` so should be straightforward to add the labels support for those. I might have time today to look into it.

Could you resolve the conflicts with this branch so it is ready for merging?

saikonen (dismissed):

dhpollack:
> We can do another PR for the sensors, they are just another template that is being compiled as part of `argo-workflows create` so should be straightforward to add the labels support for those. I might have time today to look into it.
> 
> Could you resolve the conflicts with this branch so it is ready for merging?

Conflicts resolved.  Thanks for your help.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -469,6 +474,11 @@ def _compile_workflow_template(self):
                 }
             )
 
+        # get labels from env vars
```
dhpollack, line 477: If you want this to be a method of some sort let me know.  This seems to add the labels to pods.  If you think it should be somewhere else, let me know.  I am not that familiar with the argo plugin since I was previously relying on the fact that the kubernetes decorator would automatically add labels for us.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -1102,15 +1113,16 @@ def _container_templates(self):
...
+                ).metadata(
                     ObjectMeta().annotation("metaflow/step_name", node.name)
                     # Unfortunately, we can't set the task_id since it is generated
                     # inside the pod. However, it can be inferred from the annotation
                     # set by argo-workflows - `workflows.argoproj.io/outputs` - refer
                     # the field 'task-id' in 'parameters'
                     # .annotation("metaflow/task_id", ...)
                     .annotation("metaflow/attempt", retry_count)
+                    # Set labels
+                    .labels(resources.get("labels"))
```
dhpollack, line 1125: Should this be removed?  I think it was getting the labels from the kubernetes resource.  Perhaps this is still useful tho or will be useful in the future?  If not, I am find with removing it.

dhpollack:
Tested the latest with:

```
METAFLOW_KUBERNETES_LABEL="" python hello.py run --with kubernetes
METAFLOW_KUBERNETES_LABEL="label1=val1" python hello.py run --with kubernetes
METAFLOW_KUBERNETES_LABEL="label1=val1,label2=val2" python hello.py run --with kubernetes
METAFLOW_KUBERNETES_LABEL="label=creation" python hello.py argo-workflows create
METAFLOW_KUBERNETES_LABEL="label=trigger" python hello.py argo-workflows trigger  # label has no effect here, the creation label is used
METAFLOW_KUBERNETES_LABEL="" python hello.py argo-workflows create
python hello.py run --with kubernetes:labels=l1=v1  # this also fails as expected
```

dhpollack:
Sensor Labels: dhpollack/metaflow#1

PR #1361, opened 2023-04-18
Feature/tmpfs for kubernetes and argo

saikonen:
- adds tmpfs support for kubernetes and argo-workflows via kube native `emptyDir` volume mount

Note: enforcing the `tmpfs_size` is heavily dependent on the available kubernetes version, as we require support for the `size_limit=` argument for the `emptyDir` to have an effect.

### DIFF
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -36,6 +36,12 @@ from metaflow.util import compress_list, dict_to_cli_options, to_camelcase
 
 from .argo_client import ArgoClient
 
+try:
+    unicode
+except NameError:
+    unicode = str
+    basestring = str
+
 
 class ArgoWorkflowsException(MetaflowException):
     headline = "Argo Workflows error"
@@ -867,11 +873,35 @@ class ArgoWorkflows(object):
                 if v is not None and k not in env_vars_to_skip
             }
 
+            # Tmpfs variables
+            use_tmpfs = resources["use_tmpfs"]
+            tmpfs_size = resources["tmpfs_size"]
+            tmpfs_path = resources["tmpfs_path"]
+            tmpfs_tempdir = resources["tmpfs_tempdir"]
+
+            tmpfs_enabled = use_tmpfs or (tmpfs_size and not use_tmpfs)
+            if tmpfs_enabled:
+                if tmpfs_size:
+                    if not (isinstance(tmpfs_size, (int, unicode, basestring))):
+                        raise ArgoWorkflowsException(
+                            "Invalid tmpfs value: ({}) (should be 0 or greater)".format(
+                                tmpfs_size
+                            )
+                        )
+                else:
+                    # default tmpfs behavior - https://man7.org/linux/man-pages/man5/tmpfs.5.html
+                    tmpfs_size = int(resources["memory"]) / 2
+                # Add default unit as ours differs from Kubernetes default.
+                tmpfs_size = "{}Mi".format(tmpfs_size)
+
+            if tmpfs_enabled and tmpfs_tempdir:
+                env["METAFLOW_TEMPDIR"] = tmpfs_path
+
             # Create a ContainerTemplate for this node. Ideally, we would have
             # liked to inline this ContainerTemplate and avoid scanning the workflow
             # twice, but due to issues with variable substitution, we will have to
             # live with this routine.
-            yield (
+            template = (
                 Template(self._sanitize(node.name))
                 # Set @timeout values
                 .active_deadline_seconds(run_time_limit)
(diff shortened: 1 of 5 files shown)


metaflow/plugins/argo/argo_workflows.py
```diff
@@ -867,11 +873,35 @@ def _container_templates(self):
                 if v is not None and k not in env_vars_to_skip
             }
 
+            # Tmpfs variables
+            use_tmpfs = resources["use_tmpfs"]
```
savingoyal, line 877: we can move these checks from here and in kubernetes_job.py into the kubernetes decorator.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -977,11 +1007,23 @@ def _container_templates(self):
...
+                                ]
+                                if tmpfs_enabled
+                                else []
+                            ),
                         ).to_dict()
                     )
                 )
             )
+            if tmpfs_enabled:
+                template.empty_dir_volume("argo-tmpfs-volume", size_limit=tmpfs_size)
```
savingoyal, line 1025: what happens if the tmpfs_size is 0?

savingoyal: if tmpfs_size = 0 implies no volume would be attached - we can move that logic in `def empty_dir_volume (self...` and avoid this if-else conditional here. - 

```
.empty_dir_volume("argo-tmpfs-volume", size_limit=tmpfs_size if tmpfs_enabled else 0)
```

savingoyal (changes requested):
lgtm! minor comments. I haven't tested this PR but it looks like we should be able to ship this out soon

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -867,11 +873,35 @@ def _container_templates(self):
                 if v is not None and k not in env_vars_to_skip
             }
 
+            # Tmpfs variables
+            use_tmpfs = resources["use_tmpfs"]
+            tmpfs_size = resources["tmpfs_size"]
```
shrinandj, line 878: I'm guess the memory backing `tmpfs` is part of the total memory assigned to the pod, right? 

What happens if a user accidentally sets tmpfs_size >= total memory of the pod? Does K8s fail with an appropriate error message?

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -899,6 +910,10 @@ def _container_templates(self):
                 )
                 # Set emptyDir volume for state management
                 .empty_dir_volume("out")
+                # Set tmpfs emptyDir volume if enabled
+                .empty_dir_volume(
+                    "argo-tmpfs-volume", size_limit=tmpfs_size if tmpfs_enabled else 0
```
savingoyal, line 915: should we call this - `tmpfs-ephemeral-volume` to make this consistent with @kubernetes

metaflow/plugins/argo/argo_workflows.py
saikonen: great point, this was a neat way to keep the method chaining tidy :)

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -1244,12 +1286,17 @@ def retry_strategy(self, times, minutes_between_retries):
...
+    def empty_dir_volume(self, name, size_limit=None):
         # Attach an emptyDir volume
         # https://argoproj.github.io/argo-workflows/empty-dir/
         if "volumes" not in self.payload:
             self.payload["volumes"] = []
-        self.payload["volumes"].append({"name": name, "emptyDir": {}})
+        self.payload["volumes"].append(
+            {
+                "name": name,
+                "emptyDir": {"sizeLimit": size_limit} if size_limit else {},
```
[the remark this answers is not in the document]
saikonen: good catch, indeed it needs to be set to memory to be on par with the batch tmpfs implementation.

metaflow/plugins/argo/argo_workflows.py
saikonen: with the cluster that is available for my testing, nothing fails even when the emptyDir size is set to exceed the memory available for the pod. free disk space reported on the mount stays at `memory // 2` though, which seems to be a hard cap for that.

metaflow/plugins/kubernetes/kubernetes.py
```diff
@@ -149,6 +149,10 @@ def create_job(
         gpu_vendor=None,
         disk=None,
         memory=None,
+        use_tmpfs=None,
```
shrinandj, line 152: I should've asked this earlier. Why do we need this config option? If the other `tmpfs` config options are specified, should't it automatically mean that `use_tmpfs` is True?

metaflow/plugins/kubernetes/kubernetes_cli.py
```diff
@@ -80,6 +80,12 @@ def kubernetes():
 @click.option(
     "--max-user-code-retries", default=0, help="Passed to the top-level 'step'."
 )
+@click.option("--use-tmpfs", is_flag=True, help="tmpfs requirement for Kubernetes pod.")
+@click.option(
```
shrinandj, line 84: `help` needs to be updated for these options.

metaflow/plugins/argo/argo_workflows.py
shrinandj: Excellent! So, looks like internally Linux takes care that `tmpfs` never exceeds half of main memory.

metaflow/plugins/kubernetes/kubernetes.py
shrinandj: I think I know why.. The simplest way to run this would be to simply use `--use-tmpfs`. The other config otpions have default values that will get used if tmpfs is enabled. Is that correct?

saikonen: correct, and it's the way the batch tmpfs feature works as well.

shrinandj (dismissed):
Looks good!

savingoyal (approved):

PR #1363, opened 2023-04-20
Fixing a MetaflowObject backward compatibility issue

pjoshi30:
The newly added `_namespace_check` attribute in the `MetaflowObject` broke backward compatibility in cases where the object is deserialized from an older version of Metaflow. 

Steps to reproduce this issue:

- Launch a flow with OSS version `2.8.0` that saves the Run object as a flow artifact
- Try to access the artifact (which a Run object) from a new flow that is running on version `2.8.3`
- You will see an error like this `AttributeError: 'Run' object has no attribute '_namespace_check'`

This is because the Run object is pickled in the original flow when it is assigned to an artifact. In the other flow, it is unpickled to the old Run object that does not contain the newly added attribute.

This PR adds a custom implementation of the `__getstate__` and `__setstate__` functions to help with backward compatibility.

### DIFF
--- a/metaflow/client/core.py
+++ b/metaflow/client/core.py
@@ -326,16 +326,8 @@ class MetaflowObject(object):
         self._user_tags = frozenset(self._object.get("tags") or [])
         self._system_tags = frozenset(self._object.get("system_tags") or [])
 
-        if self._get_attribute("_namespace_check", True) and not self.is_in_namespace():
+        if self._namespace_check and not self.is_in_namespace():
             raise MetaflowNamespaceMismatch(current_namespace)
-
-    def _get_attribute(self, name, default=None):
-        """
-        Returns the attribute if it exists in this object else returns a default
-        """
-        if name in self.__dict__:
-            return self.__dict__[name]
-        return default
 
     def _get_object(self, *path_components):
         result = self._metaflow.metadata.get_object(
@@ -360,7 +352,7 @@ class MetaflowObject(object):
         query_filter = {}
 
         # skip namespace filtering if _namespace_check is unset.
-        if self._get_attribute("_namespace_check", True) and current_namespace:
+        if self._namespace_check and current_namespace:
             query_filter = {"any_tags": current_namespace}
 
         unfiltered_children = self._metaflow.metadata.get_object(
@@ -473,7 +465,7 @@ class MetaflowObject(object):
                 attempt=self._attempt,
                 _object=obj,
                 _parent=self,
-                _namespace_check=self._get_attribute("_namespace_check", True),
+                _namespace_check=self._namespace_check,
             )
         else:
             raise KeyError(id)
@@ -493,6 +485,11 @@ class MetaflowObject(object):
             True if the child exists or False otherwise
         """
         return bool(self._get_child(id))
+
+    def __setstate__(self, state):
+        self.__dict__ = state
+        if "_namespace_check" not in state:
+            self._namespace_check = True
 
     @property
     def tags(self) -> FrozenSet[str]:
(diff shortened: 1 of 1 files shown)


metaflow/client/core.py
```diff
@@ -475,6 +475,44 @@ def __contains__(self, id: str):
...
+        """
+        This function is used during the unpickling operation.
+        More info here https://docs.python.org/3/library/pickle.html#object.__setstate__
+        """
+        if isinstance(state, dict):
+            # For backward compatibility: handles pickled objects that were serialized as a `dict`
+            self.__init__(
+                pathspec=state.get("_pathspec", None),
+                attempt=state.get("_attempt", None),
+                _namespace_check=state.get("_namespace_check", None),
```
romain-intel, line 488: this should default to `True` as this is what it defaults to in init and what objects had before we introduced this namespace_check thing.

metaflow/client/core.py
```diff
@@ -475,6 +475,44 @@ def __contains__(self, id: str):
...
+                "Type not supported: {}".format(state.__class__)
+            )
+
+    def __getstate__(self):
+        """
+        This function is used during the pickling operation.
+        More info here https://docs.python.org/3/library/pickle.html#object.__getstate__
+
+        This function is not forward compatible i.e., if this object (or any of the objects deriving
+        from this object) are pickled (serialized) in a later version of Metaflow, it will not be possible
```
romain-intel, line 511: I would say: "it may not be possible" instead of "it will not be possible".

metaflow/client/core.py
```diff
@@ -475,6 +475,44 @@ def __contains__(self, id: str):
...
+        More info here https://docs.python.org/3/library/pickle.html#object.__setstate__
+        """
+        if isinstance(state, dict):
+            # For backward compatibility: handles pickled objects that were serialized as a `dict`
+            self.__init__(
+                pathspec=state.get("_pathspec", None),
+                attempt=state.get("_attempt", None),
+                _namespace_check=state.get("_namespace_check", None),
+            )
+        elif isinstance(state, tuple):
```
romain-intel, line 490: to simplify our lives a little bit, I would actually pickle something like this:
```
{
  "version": 1,
  "data": [self.pathspec, self._attempt, self._namespace_check]
}
```

This would replace the check on the type (and instead you can check if state contains "version" -- it won't unless it's the new format). You now have two cases for now:
  - no version (it's the case you have above)
  - a version and you can write a "unpickle_v1" function that does what you do below passing it data

If we ever want to change it again, it will be easy to just add a new "unpickle_v2" function and maintain backward compatibility that way. It's a bit more data in terms of number of bytes for the object but I don't think it makes that big of a difference and makes it a little easier to reason about going forward.

As an example, if we had had this, we would have changed to version 2 and added an additional field easily that way.

romain-intel: Actually, we can also maybe do something like this: `"version": "2.8.4"` (or whatever but basically the version of Metaflow from which we have this format). That way, we could give a very nice error message like: "Unpickling this object requires Metaflow version 2.8.4 or greater". Newer versions would have the proper if/else clauses and the default else clause would print out this message.

romain-intel (commented):
Changes look fine. I had a suggestion in terms of how we could make pickling/unpickling a little more backward compatible but I think we can ship this soon.

pjoshi30:
I have addressed the comments and updated this PR. We now create a dictionary of `_unpickle_<version>` functions with the version string as the key. This prints a nicer error for forward compatibility errors (object pickled using a newer version being un-pickled by an older version). 

Tests:
* :white_check_mark: write an artifact which is a MetaflowObject from an older version of Metaflow (2.8.0) and read from this new version (from the PR)
* :white_check_mark: write from this new version and read from the new version
* ❌  [forward compatibility not supported] write from the new version and read from an older version

savingoyal (approved):

romain-intel (dismissed):
LGTM (post rebase/merge of master and testing labels).

PR #1364, opened 2023-04-20
Enable NFS volume mount

DonIvanCorleone:
This PR enables Kubernetes pods to mount NFS shares (only).
At the moment there is no support for Airflow or Argo  - Airflow I can do later, as I have access to an Airflow instance. Argo I cannot test properly hence someone else has to do that job. 

If somebody would have the time a review would be highly appreciated. Many thanks

### DIFF
--- a/metaflow/plugins/kubernetes/kubernetes.py
+++ b/metaflow/plugins/kubernetes/kubernetes.py
@@ -152,9 +152,8 @@ class Kubernetes(object):
         run_time_limit=None,
         env=None,
         tolerations=None,
-        volumes=None
+        volumes=None,
     ):
-
         if env is None:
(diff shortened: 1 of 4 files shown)


DonIvanCorleone:
Hi there.

Has somebody the time to review this PR? @romain-intel or @savingoyal
Any review would be highly appreciated. Many thank!

DonIvanCorleone:
Hi there,

I removed f-strings as they are obviously not supported in Python 3.5

Unfortunately I do not understand the error message for "Python 3.8 on mac-os:latest" any ideas why this task failed? 

Happy if the workflow could be started once again :) 
Thanks

@romain-intel @savingoyal

PR #1369, opened 2023-04-25
Job completion check based on container status.

shrinandj:
This change modifies the way a Metaflow step running as a Kubernetes job is marked completed. Previously, Metaflow would wait until the job itself is marked as completed. However, this can take much longer than the time the container running in the job is marked completed.

There don't seem to be any obvious tunables in K8s for making it faster to mark a job as completed.

Given that a Metaflow job runs a single container, this change relies on checking that containers' status.

Testing Done:

- Basic linear flows run to completion.
- Foreach based flows run to completion.
- Flows that have failed steps show up as failed. Logs were shown on stderr and the metaflow ui.
  - Tested with a step that explicitly did a sys.exit()
  - Tested with a step that had it's pod explicitly terminated (kubectl delete pod)
- Terminated one step in a foreach. Verified that the run shows up as failed
- Tested that flows run just fine with Argo workflows

### DIFF
--- a/metaflow/plugins/kubernetes/kubernetes_job.py
+++ b/metaflow/plugins/kubernetes/kubernetes_job.py
@@ -482,17 +482,15 @@ class RunningJob(object):
 
     @property
     def is_done(self):
-        # Check if the job is done. As a side effect, also refreshes self._job and
+        # Check if the container is done. As a side effect, also refreshes self._job and
         # self._pod with the latest state
         def done():
-            # Either the job succeeds or fails naturally or else we may have
+            # Either the container succeeds or fails naturally or else we may have
             # forced the pod termination causing the job to still be in an
             # active state but for all intents and purposes dead to us.
-            return (
-                bool(self._job["status"].get("succeeded"))
-                or bool(self._job["status"].get("failed"))
-                or (self._job["spec"]["parallelism"] == 0)
-            )
+            if self.is_container_running or (self._job["spec"]["parallelism"] == 0):
+                return True
+
 
         if not done():
             # If not done, fetch newer status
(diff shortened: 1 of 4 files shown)


metaflow/plugins/kubernetes/kubernetes_job.py
```diff
@@ -339,15 +339,16 @@ def __init__(self, client, name, uid, namespace):
         self._job = self._fetch_job()
         self._pod = self._fetch_pod()
 
-        import atexit
+        import signal
```
savingoyal, line 342: what's the motivation for this change?

metaflow/plugins/kubernetes/kubernetes_job.py
```diff
@@ -563,9 +564,24 @@ def has_failed(self):
...
+        if self.is_done:
+            return False
+
+        container_statuses = self._pod.get("status", {}).get("container_statuses", [])
+        if not container_statuses:
+            # Yet to start, most likely
+            return True
+        if len(container_statuses) < 1:
+            # Yet to start, most likely
+            return True
```
savingoyal, line 576: should we return from pod phase here and above?

metaflow/plugins/kubernetes/kubernetes_job.py
```diff
@@ -563,9 +564,24 @@ def has_failed(self):
...
+        # There is guaranteed to be only 1 container
+        cstatus = container_statuses[0]
```
savingoyal, line 578: can we fetch the container status using the name of the container? that will guard against the scenario if/when we have more than one container.

savingoyal (changes requested):

metaflow/plugins/kubernetes/kubernetes_job.py
shrinandj: Without this change, the code would try to fetch the job object and then patch it to set parallelism to 0. This even when knowing that the container has actually completed execution. So, it's only a matter of time when the job is also marked completed. Having the original code using `atexit` as is reduces the latency improvements introduced in this change.

AFAICT, the original goal of this code was to cleanup the pods (best-effort) in the event of the user-created process running Metaflow is terminated. The change continues to do that.

Let me know if there was any additional motivation for having this and I can test that scenario out.

metaflow/plugins/kubernetes/kubernetes_job.py
shrinandj: Done

metaflow/plugins/kubernetes/kubernetes_job.py
shrinandj: name of the container is derived from the step name and is not available in the `RunningJob`. 

I'll change the code to iterate over all containers and if any one of them is NOT terminated, return True for `is_running`. That should protect against the case where there are multiple containers.

metaflow/plugins/kubernetes/kubernetes_job.py
```diff
@@ -563,9 +564,29 @@ def has_failed(self):
     @property
     def is_running(self):
         # Returns true if the pod is running.
-        return not self.is_done and (
-            self._pod.get("status", {}).get("phase") == "Running"
-        )
+        if self.is_done:
+            return False
+
+        container_statuses = self._pod.get("status", {}).get("container_statuses", [])
```
savingoyal, line 570: we can add a quick comment here re: the motivation for this change

```
# Container status updates before pod phase, often by multiple seconds.
        container_statuses = self._pod.get("status", {}).get("container_statuses", [])
        if not container_statuses or len(container_statuses) < 1:
            # Yet to start, most likely
            return self._pod.get("status", {}).get("phase") == "Running"

        for cstatus in container_statuses:
            # If the terminated field is not set, the pod is still running.
            terminated = cstatus.get("state", {}).get("terminated", {})
            if not terminated or not terminated.get("finished_at"):
                # If the terminated field is set but the `finished_at` field is not 
                # set, the pod is still considered as running.
                return True
```

metaflow/plugins/kubernetes/kubernetes_job.py
```diff
@@ -563,9 +564,29 @@ def has_failed(self):
     @property
     def is_running(self):
         # Returns true if the pod is running.
-        return not self.is_done and (
-            self._pod.get("status", {}).get("phase") == "Running"
-        )
+        if self.is_done:
```
savingoyal, line 567: the `is_done`, `has_failed` and `has_succeeded` methods need a corresponding update as well, otherwise they will be out of sync and communicate wrong information about the state of the object.

metaflow/plugins/kubernetes/kubernetes_job.py
savingoyal: `atexit` was also supposed to handle [this scenario](https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/kubernetes/kubernetes_cli.py#L239) - which would break after this change.

savingoyal (changes requested):
Few comments. Let's ship this as soon as these are addressed.

metaflow/plugins/kubernetes/kubernetes_job.py
```diff
@@ -544,28 +539,128 @@ def status(self):
...
+                return False
+
+            # If finished_at is set AND reason is Completed
+            if terminated.get("reason", "").lower() != "completed":
+                return False
+
+        return True
+
+    @property
+    def have_pod_containers_succeeded(self):
```
saikonen, line 580: unused property, duplicate logic to `has_container_succeeded`

saikonen (changes requested):

Issue #1345, opened 2023-04-05
AWS batch job not submitted due to invalid AWS tag


wrighting:
Given I am using AWS batch
and the configuration BATCH_EMIT_TAGS is true
and I am using custom extensions
and the version is set in the custom extension
then the AWS tag metaflow.version has an invalid value

It doesn't like the brackets e.g. `'2.8.1+mc(0.10.0)'`

`botocore.exceptions.ClientError: An error occurred (BadRequestException) when calling the SubmitJob operation: '2.8.1+mc(0.10.0)' is not a valid value for TagValue - it contains illegal characters`

working around it by setting `__version__` in `metaflow_extensions/mc/toplevel/mc_toplevel.py` to `None`

https://github.com/Netflix/metaflow/blob/f4b47cc7d0f99594e5322d9b891e4885d5ee300d/metaflow/plugins/aws/batch/batch.py#L293

The fix could be either/both of changing the version format or sanitizing it before adding as a tag.

(My operations people would be happy if there was a way to define our own AWS tags)

PR #1376, opened 2023-04-26
feature: add sanitization for batch tags

saikonen:
add sanitization for batch tags so they are guaranteed to adhere to AWS requirements.

addresses #1345

### DIFF
--- a/metaflow/plugins/aws/aws_utils.py
+++ b/metaflow/plugins/aws/aws_utils.py
@@ -140,3 +140,15 @@ def compute_resource_attributes(decos, compute_deco, resource_defaults):
             result[k] = str(compute_deco.attributes[k] or "0")
 
     return result
+
+
+def sanitize_batch_tag(key, value):
+    """
+    Sanitize a key and value for use as a Batch tag.
+    """
+    # https://docs.aws.amazon.com/batch/latest/userguide/using-tags.html#tag-restrictions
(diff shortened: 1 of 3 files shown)


metaflow/plugins/aws/aws_utils.py
```diff
@@ -140,3 +140,15 @@ def compute_resource_attributes(decos, compute_deco, resource_defaults):
...
     return result
+
+
+def sanitize_batch_tag(key, value):
+    """
+    Sanitize a key and value for use as a Batch tag.
+    """
+    # https://docs.aws.amazon.com/batch/latest/userguide/using-tags.html#tag-restrictions
+    RE_NOT_PERMITTED = r"[^A-Za-z0-9\s\+\-\=\.\_\:\/\@]"
+    _key = re.sub(RE_NOT_PERMITTED, "", key)[:128]
```
romain-intel, line 151: should we error if the size is too much?

saikonen: we could, but as the original issue was not being able to post batch jobs due to tags not adhering to AWS restrictions, this would lead to the same issue if a company is using their own mf_ext with a very long version/name that exceeds the limit.

main point is that we're sanitizing system specified tags that the user has very little control of to begin with, so an error here would not be actionable in many cases, completely blocking them instead.

romain-intel (approved):
LGTM

saikonen:
> Can you verify with step functions too -
Yep, tested and working as intended with SFN as well.

Issue #1191, opened 2022-11-11
Ability to add custom headers to S3 calls


oavdeev:
One practical use case is having org policy requiring encryption headers on all PutObject calls via [SCP](https://docs.aws.amazon.com/organizations/latest/userguide/orgs_manage_policies_scps.html). Could be an "expert" level config option to specify the headers.

juliuskittler:
This would be very helpful! 

As @oavdeev says, the current implementation of MetaFlow does not work for AWS accounts that have a service control policy, which requires all PutObject calls to specify SSE-S3 encryption in their header.

PR #1382, opened 2023-05-01
feat: ability to add custom headers to S3 upload calls

zendesk-klross:
https://github.com/Netflix/metaflow/issues/1191

### DIFF
--- a/metaflow/metaflow_config.py
+++ b/metaflow/metaflow_config.py
@@ -277,6 +277,8 @@ KUBERNETES_PERSISTENT_VOLUME_CLAIMS = from_conf(
     "KUBERNETES_PERSISTENT_VOLUME_CLAIMS", ""
 )
 KUBERNETES_SECRETS = from_conf("KUBERNETES_SECRETS", "")
+# Default labels for kubernetes pods
+KUBERNETES_LABELS = from_conf("KUBERNETES_LABELS", "")
 # Default GPU vendor to use by K8S jobs created by Metaflow (supports nvidia, amd)
 KUBERNETES_GPU_VENDOR = from_conf("KUBERNETES_GPU_VENDOR", "nvidia")
 # Default container image for K8S
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -30,6 +30,7 @@ from metaflow.metaflow_config import (
     DEFAULT_METADATA,
     DEFAULT_SECRETS_BACKEND_TYPE,
     KUBERNETES_FETCH_EC2_METADATA,
+    KUBERNETES_LABELS,
     KUBERNETES_NAMESPACE,
     KUBERNETES_NODE_SELECTOR,
     KUBERNETES_SANDBOX_INIT_SCRIPT,
@@ -40,6 +41,10 @@ from metaflow.metaflow_config import (
 )
 from metaflow.mflog import BASH_SAVE_LOGS, bash_capture_logs, export_mflog_env_vars
 from metaflow.parameters import deploy_time_eval
+from metaflow.plugins.kubernetes.kubernetes import (
+    parse_kube_keyvalue_list,
+    validate_kube_labels,
+)
 from metaflow.util import (
     compress_list,
     dict_to_cli_options,
@@ -146,6 +151,7 @@ class ArgoWorkflows(object):
         self.triggers, self.trigger_options = self._process_triggers()
         self._schedule, self._timezone = self._get_schedule()
 
+        self.kubernetes_labels = self._get_kubernetes_labels()
         self._workflow_template = self._compile_workflow_template()
         self._sensor = self._compile_sensor()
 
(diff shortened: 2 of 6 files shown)


metaflow/metaflow_config_funcs.py
```diff
@@ -49,7 +49,7 @@ def config_values(include=0):
             yield name, config_value.serializer(config_value.value)
 
 
-def from_conf(name, default=None, validate_fn=None):
+def from_conf(name, default=None, validate_fn=None, prefix=False):
```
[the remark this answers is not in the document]
saikonen: I'm also not entirely sure if we want to task the `from_conf` with this kind of dict-building, or if a separate function would make more sense for this use case. 
The prefix-parsing is a bit of a niche for the time being, and only works when the `key` can be used as an environment variable, so reusability for f.ex. building a dict with hyphenated keys will not work.

metaflow/plugins/aws/batch/batch.py
```diff
@@ -262,6 +263,11 @@ def create_job(
         if tmpfs_enabled and tmpfs_tempdir:
             job.environment_variable("METAFLOW_TEMPDIR", tmpfs_path)
 
+        if S3_UPLOAD_ARGS is not None:
+            for key, value in S3_UPLOAD_ARGS.items():
+                env_var = "METAFLOW_S3_UPLOAD_ARGS_" + key
+                job.environment_variable(env_var, value)
+
```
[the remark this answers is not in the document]
zendesk-klross: Thanks for your feedback! We've tested this with a simple HelloAWS flow and it worked fine, but we haven't tried it with a foreach loop yet.

metaflow/metaflow_config_funcs.py
```diff
@@ -106,6 +106,28 @@ def from_conf(name, default=None, validate_fn=None):
     return value
 
 
+def from_conf_dict(prefix, validate_fn=None):
```
savingoyal, line 109: @zendesk-klross Any specific reason for this change? Given that this PR focuses on just the encryption bits - I wonder if we need to support arbitrary headers?

savingoyal: Happy to jump on a quick call if that would be useful to get to a quick conclusion.

zendesk-klross: We added the encryption attribute to S3 objects mainly to test a use-case for the new custom headers (this way, we can make sure the object actually gets updated using these headers). But we also figured it's a good idea to let users pick their own headers for their specific needs, as mentioned [here](https://github.com/Netflix/metaflow/issues/1191).

savingoyal: We need to ensure that we take care of all scenarios with arbitrary headers, including any considerations that would affect the data being read back. If you have a specific use case, we are happy to tackle that as part of this PR.

zendesk-klross: We also thought that this parsing function might be useful for other configurable arguments that we don't want to explicitly define in Metaflow, but let us know if you can't think of any other applications for this.

zendesk-klross: Right now the PR starts the groundwork to support all these options: `ALLOWED_UPLOAD_ARGS = ['ACL', 'CacheControl', 'ChecksumAlgorithm', 'ContentDisposition', 'ContentEncoding', 'ContentLanguage', 'ContentType', 'ExpectedBucketOwner', 'Expires', 'GrantFullControl', 'GrantRead', 'GrantReadACP', 'GrantWriteACP', 'Metadata', 'ObjectLockLegalHoldStatus', 'ObjectLockMode', 'ObjectLockRetainUntilDate', 'RequestPayer', 'ServerSideEncryption', 'StorageClass', 'SSECustomerAlgorithm', 'SSECustomerKey', 'SSECustomerKeyMD5', 'SSEKMSKeyId', 'SSEKMSEncryptionContext', 'Tagging', 'WebsiteRedirectLocation'].` The list is pretty long and testing for all combinations of headers might be complicated, so if you feel that an untested setup isn't a good idea we can definitely refocus the PR to support ServerSideEncryption only.

zendesk-klross: We're also happy to expand this PR to include support for these `ALLOWED_DOWNLOAD_ARGS = ['ChecksumMode', 'VersionId', 'SSECustomerAlgorithm', 'SSECustomerKey', 'SSECustomerKeyMD5', 'RequestPayer', 'ExpectedBucketOwner']`if you think that could be useful.

savingoyal: @zendesk-klross We can make this PR specific to encryption to keep the testing burden low for now if there aren't any other motivating use cases.

savingoyal: Would you be able to create a version that just handles that so that we can kick off the review process?

zendesk-klross: @savingoyal I've just opened a new PR [https://github.com/Netflix/metaflow/pull/1436](url) that limits the scope of this to server-side-encryption, let me know what you think!

PR #1383, opened 2023-05-01
added missing return statement

felipeGarciaDiaz:
added a return statement to one of the methods in the guage class.


romain-intel (approved):
LGTM. Thanks for the fix.

PR #1384, opened 2023-05-02
Update current object adding `run` and `task` object.

romain-intel:
Update documentation to make it more accurate.

### DIFF
--- a/metaflow/current.py
+++ b/metaflow/current.py
@@ -2,10 +2,13 @@ from collections import namedtuple
 import os
 from typing import Any, Optional
 
-from metaflow import Run, Task
 from metaflow.metaflow_config import TEMPDIR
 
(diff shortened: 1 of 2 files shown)


metaflow/current.py
```diff
@@ -83,49 +87,53 @@ def is_running_flow(self) -> bool:
         return self._is_running
 
     @property
-    def flow_name(self) -> str:
+    def flow_name(self) -> Optional[str]:
         """
-        The name of the currently executing flow.
+        The name of the currently executing flow. Will be None if not inside
+        a currently executing flow.
```
savingoyal, line 93: The current object is only used inside an executing flow from a user POV. This comment would indicate otherwise.

romain-intel: I mean it is accessible outside and the value is sometimes None so wanted to capture that. We can remove if you want though.

PR #1392, opened 2023-05-05
Fix an issue introduced in 2.8.4 that prevented pickled MetaflowObjec…

romain-intel:
…t from being accessed after pickling in a different namespace

The core issue is that we now call the constructor when un-pickling which will recheck the namespace. This broke previous behavior where the constructor was not called and no namespace check was done

### DIFF
--- a/metaflow/client/core.py
+++ b/metaflow/client/core.py
@@ -516,10 +516,12 @@ class MetaflowObject(object):
             self._UNPICKLE_FUNC[version](self, state["data"])
         else:
             # For backward compatibility: handles pickled objects that were serialized without a __getstate__ override
+            # We set namespace_check to False if it doesn't exist for the same
+            # reason as the one listed in __getstate__
             self.__init__(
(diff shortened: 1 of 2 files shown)


metaflow/client/core.py
```diff
@@ -531,12 +533,16 @@ def __getstate__(self):
         from this object) are pickled (serialized) in a later version of Metaflow, it may not be possible
         to unpickle (deserialize) them in a previous version of Metaflow.
         """
+        # Note that we set _namespace_check to False because we want the user to
+        # be able to access this object even after unpickling it. If we set it to
+        # True, it would check the namespace again at the time of unpickling even
+        # if the user properly got the object in the first place and pickled it.
         return {
             "version": "2.8.4",
```
pjoshi30, line 541: this will still be a problem with older versions that do not understand this new dictionary format. how do we handle that case? One option is to not use this custom dictionary format using a version

PR #1400, opened 2023-05-09
Add configuration of username from metaflow_config.py

tfurmston:
This PR does the following:

* Allows for the username, as determined from the function `get_username`, to be configured through standard configuration functionality in `metaflow_config.py`. In particular, it allows for the username to be configured through Metaflow configuration files.


test/unit/test_util.py
```diff
@@ -0,0 +1,62 @@
+import os
```
savingoyal, line 1: @tfurmston I don't think that these unit tests are executed by the CI harness. We already have tests that check if the user name is getting correctly passed around - https://github.com/Netflix/metaflow/blob/master/test/core/contexts.json#L7 and https://github.com/Netflix/metaflow/blob/master/test/core/metaflow_test/metadata_check.py#L32

savingoyal (changes requested):
quick comments. otherwise lgtm!

test/unit/test_util.py
tfurmston: Ah, ok. Good to know. I did wonder about this point, but there were some other tests in the `unit` test folder that were structured in the same way. 

I should remove these new tests completely then?

savingoyal: Yes - that will be super!

tfurmston: cool. All done.

savingoyal (approved):

PR #1407, opened 2023-05-13
Introduce slack notifications for argo workflows

savingoyal:


savingoyal:
> Do we need to add a check about the Argo workflows version when the notifications functionality was added?

Good point. Is there a way client side for us to even check what version of Argo Workflows is running currently? Potentially that will help simplify code longer term as we support more Argo releases.

PR #1416, opened 2023-05-16
feature: add argo workflows list-runs command

saikonen:
- add command for listing runs (and filtering by status) for argo-workflows

Note: due to API limitations, this feature required adding a custom label to workflows with the workflow-template as a value, as labels are the only available filter for custom fields on objects. This makes the feature not backwards compatible, as the necessary label does not exist on older runs.

### DIFF
--- a/metaflow/plugins/argo/argo_client.py
+++ b/metaflow/plugins/argo/argo_client.py
@@ -50,7 +50,6 @@ class ArgoClient(object):
                 namespace=self._namespace,
                 plural="workflows",
                 label_selector=",".join(filters),
-                limit=10,
             )["items"]
         except client.rest.ApiException as e:
             raise ArgoClientException(
--- a/metaflow/plugins/argo/argo_client.py
+++ b/metaflow/plugins/argo/argo_client.py
@@ -34,6 +34,32 @@ class ArgoClient(object):
                 json.loads(e.body)["message"] if e.body is not None else e.reason
             )
 
+    def list_executions(self, name, states):
+        client = self._client.get()
+        # Problems with the API capabilities:
+        # https://github.com/argoproj/argo-workflows/discussions/7123
+        # https://github.com/argoproj/argo-workflows/discussions/8500
+        # kubernetes/argo api does not support label-selectors for the following
+        # spec.workflowTemplateRef.name = name
+        # status.phase = states
+        #
+        # but status can be filtered through labels.
+        # it would seem that the only way to filter with workflow template name is to do it in-memory
+        filters = ["status.phase=%s" % state for state in states]
+        try:
+            return client.CustomObjectsApi().list_namespaced_custom_object(
+                group=self._group,
+                version=self._version,
+                namespace=self._namespace,
+                plural="workflows",
+                # field_selector=",".join(filters),
+                limit=10,
+            )["items"]
+        except client.rest.ApiException as e:
+            raise ArgoClientException(
+                json.loads(e.body)["message"] if e.body is not None else e.reason
+            )
+
     def register_workflow_template(self, name, workflow_template):
         # Unfortunately, Kubernetes client does not handle optimistic
         # concurrency control by itself unlike kubectl
--- a/metaflow/plugins/argo/argo_client.py
+++ b/metaflow/plugins/argo/argo_client.py
@@ -36,16 +36,13 @@ class ArgoClient(object):
 
     def list_executions(self, name, states):
         client = self._client.get()
-        # Problems with the API capabilities:
-        # https://github.com/argoproj/argo-workflows/discussions/7123
-        # https://github.com/argoproj/argo-workflows/discussions/8500
-        # kubernetes/argo api does not support label-selectors for the following
-        # spec.workflowTemplateRef.name = name
-        # status.phase = states
-        #
-        # but status can be filtered through labels.
-        # it would seem that the only way to filter with workflow template name is to do it in-memory
-        filters = ["workflows.argoproj.io/phase=%s" % state for state in states]
+        # NOTE: At the moment Kubernetes API does not support field-selectors for most of the custom fields,
+        # including the following: spec.workflowTemplateRef.name, status.phase
+        # Therefore we use labels for filtering the runs, but this has required us to add a label for the workflow-template name
+        # making the solution not backwards compatible.
+        filters = ["workflows.argoproj.io/workflow-template=%s" % name] + [
+            "workflows.argoproj.io/phase=%s" % state for state in states
+        ]
         try:
             return client.CustomObjectsApi().list_namespaced_custom_object(
                 group=self._group,
--- a/metaflow/plugins/argo/argo_client.py
+++ b/metaflow/plugins/argo/argo_client.py
@@ -45,14 +45,14 @@ class ArgoClient(object):
         #
         # but status can be filtered through labels.
         # it would seem that the only way to filter with workflow template name is to do it in-memory
-        filters = ["status.phase=%s" % state for state in states]
+        filters = ["workflows.argoproj.io/phase=%s" % state for state in states]
         try:
             return client.CustomObjectsApi().list_namespaced_custom_object(
                 group=self._group,
                 version=self._version,
                 namespace=self._namespace,
                 plural="workflows",
-                # field_selector=",".join(filters),
+                label_selector=",".join(filters),
                 limit=10,
             )["items"]
         except client.rest.ApiException as e:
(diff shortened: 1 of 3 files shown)


metaflow/plugins/argo/argo_client.py
```diff
@@ -34,6 +34,28 @@ def get_workflow_template(self, name):
                 json.loads(e.body)["message"] if e.body is not None else e.reason
             )
 
+    def list_executions(self, name, states):
+        client = self._client.get()
+        # NOTE: At the moment Kubernetes API does not support field-selectors for most of the custom fields,
+        # including the following: spec.workflowTemplateRef.name, status.phase
+        # Therefore we use labels for filtering the runs, but this has required us to add a label for the workflow-template name
+        # making the solution not backwards compatible.
+        filters = ["workflows.argoproj.io/workflow-template=%s" % name] + [
```
savingoyal, line 43: can you verify that this solution works when the "name" is rather large?

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -276,6 +276,23 @@ def get_existing_deployment(cls, name):
                 )
         return None
 
+    @classmethod
+    def list(cls, name, states):
+        client = ArgoClient(namespace=KUBERNETES_NAMESPACE)
+        try:
+            workflow_template = client.get_workflow_template(name)
```
savingoyal, line 283: Is this needed? what happens if I delete the workflow template but still want to list executions for the deleted template?

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -541,3 +542,100 @@ def _convert_value(param):
...
+    is_flag=True,
+    help="List all runs of the workflow in Running state on " "Argo Workflows.",
+)
+@click.option(
+    "--succeeded",
+    default=False,
+    is_flag=True,
+    help="List all runs of the workflow in Succeeded state on " "Argo Workflows.",
+)
+@click.option(
```
savingoyal, line 566: We can club failed and error into a single state - failed for now.

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -541,3 +542,100 @@ def _convert_value(param):
             "See the run in the UI at %s" % run_url,
             bold=True,
         )
+
+
+@argo_workflows.command(help="List all runs of the workflow on Argo Workflows.")
```
savingoyal, line 547: nit - here and below - we can specify flow instead of workflow since workflow has a very different connotation in Argo Workflows.

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -541,3 +542,100 @@ def _convert_value(param):
...
+    for execution in executions:
+        found = True
+        obj.echo(
+            "*argo-{id}* "
+            "startedAt: '{startedAt}' "
+            "finishedAt: '{finishedAt}' "
+            "*{status}*".format(
+                id=execution["metadata"]["name"],
+                status=execution["status"]["phase"],
+                startedAt=format_timestamp(execution["status"]["startedAt"]),
```
savingoyal, line 616: will we have `startedAt` always (in case the run is in an unknown state)?

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -541,3 +542,100 @@ def _convert_value(param):
...
+
+    for execution in executions:
+        found = True
+        obj.echo(
+            "*argo-{id}* "
+            "startedAt: '{startedAt}' "
+            "finishedAt: '{finishedAt}' "
+            "*{status}*".format(
+                id=execution["metadata"]["name"],
+                status=execution["status"]["phase"],
```
savingoyal, line 615: If we club `failed` and `error` in one state (`failed`), it might help to still display `status` as just `failed`

savingoyal (commented):
lgtm! few minor comments

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -513,6 +530,7 @@ def _compile_workflow_template(self):
                     Metadata()
                     .label("app.kubernetes.io/name", "metaflow-run")
                     .label("app.kubernetes.io/part-of", "metaflow")
+                    .label("workflows.argoproj.io/workflow-template", self.name)
```
savingoyal, line 533: custom label name to avoid clashes with argo

metaflow/plugins/argo/argo_client.py
saikonen: I'll change this to a hash of the name as discussed. Kubernetes has a max length of 63 for labels, but argo workflow names can be 253 characters long

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -541,3 +542,100 @@ def _convert_value(param):
...
+    "--error",
+    default=False,
+    is_flag=True,
+    help="List all runs of the workflow in Error state on " "Argo Workflows.",
+)
+@click.option(
+    "--unknown",
+    default=False,
+    is_flag=True,
+    help="List all runs of the workflow in Unknown state on " "Argo Workflows.",
```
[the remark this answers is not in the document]
saikonen: added a comment to the help string about this. I originally exposed all available statuses as filters, but if this is too much granularity for the CLI then we can also reconsider some of them.

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -541,3 +542,100 @@ def _convert_value(param):
...
+
+@argo_workflows.command(help="List all runs of the workflow on Argo Workflows.")
+@click.option(
+    "--pending",
+    default=False,
+    is_flag=True,
+    help="List all runs of the workflow in Pending state on " "Argo Workflows.",
+)
+@click.option(
+    "--running",
```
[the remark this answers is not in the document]
saikonen: my personal preference would be to keep the default return as unfiltered. This is also the way that `step-functions list-runs` works at the moment.

metaflow/plugins/argo/argo_workflows.py
saikonen: Done. changed to custom label and using hash of name as a value.

metaflow/plugins/argo/argo_workflows_cli.py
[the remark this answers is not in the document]
savingoyal: makes sense

saikonen:
Some thoughts and observations on this feature before shipping.
The way we currently truncate our deployed Argo workflows names to the maximum that Kubernetes allows for its metadata (253 characters) has some problems, the main one being that these workflows can not be launched through Argo's web UI as it tries to populate a label with the workflow template name.

Creating workflows through Argo web UI has a validation that checks that the metadata name does not exceed 63 characters (so it fits in a label). So Argo is much stricter with naming than what we are with submitted workflows. For the Metaflow use case, we can not start truncating workflow names all the way down to 63 characters though, mainly because of how project branches are named. As an example, these two flows would result in a collision with their workflow template names if truncated to 63 characters:
```
averylongprojectname.user.verylongfirstname.verylonglastnamecompanyemailaddress.com.FirstFlow
averylongprojectname.user.verylongfirstname.verylonglastnamecompanyemailaddress.com.SecondFlow
```
which is a very real possibility with the latest changes of allowing email addresses as usernames for deployed flows.

For the scope of this PR, the only solution for adding discoverability to workflows via the Kubernetes API is to introduce a custom label that uses a hash of the workflow template name, as was implemented.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -175,6 +175,14 @@ def _sanitize(name):
         # allowed by Metaflow - guaranteeing uniqueness.
         return name.replace("_", "-")
 
+    @staticmethod
+    def _label_hash(name):
+        # Hash a name for use as a Kubernetes label.
+        # Use the maximum allowed 63 characters for the hash to minimize collisions.
+        # Preserve part of the name for legibility purposes.
```
saikonen, line 182: after some thought, preserving legibility is probably not a priority for a label, as these are not meant for human consumption to begin with. If we want to store the template name as well, this can be done in the annotations which do not have the same length limitations. I don't see a use case for storing the full template name at this point though

saikonen:
this feature is somewhat blocked until #1521 is solved. After that we have a canonical way of generating the workflow-template name label, and can use that for finding the runs.

PR #1420, opened 2023-05-17
feature: add argo workflows suspend command

saikonen:
- adds suspend/unsuspend commands to argo workflows.

addresses another part of #1387

TODO: 

- [x] add production token authorization?

Note: Argo wording for the unsuspend is 'resume', but unfortunately that is already a reserved keyword in our case.


metaflow/plugins/argo/argo_client.py
```diff
@@ -86,6 +104,54 @@ def register_workflow_template(self, name, workflow_template):
                 json.loads(e.body)["message"] if e.body is not None else e.reason
             )
 
+    def suspend_workflow(self, run_id):
```
savingoyal, line 107: nit - rather than using `run_id`, we can use `name`

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -594,3 +594,37 @@ def _convert_value(param):
...
+@click.option(
+    "--authorize",
+    default=None,
+    type=str,
+    help="Authorize the suspension with a production token",
+)
+@click.argument("run-id", required=True, type=str)
+@click.pass_obj
+def suspend(obj, run_id, authorize=None):
+    # TODO: implement production token authorize
```
savingoyal, line 609: Let's implement this before shipping. The logic can be the same as the one that is in-play for deploys

savingoyal (changes requested):
lgtm! minor changes needed

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -168,6 +168,32 @@ def _sanitize(name):
         # allowed by Metaflow - guaranteeing uniqueness.
         return name.replace("_", "-")
 
+    @staticmethod
+    def pause(flow_name, run_id, suspend=True):
```
savingoyal, line 172: separate suspend and unsuspend methods

PR #1421, opened 2023-05-17
feature: use block styles for slack messages

saikonen:
- use block styles for slack messages
- add argo UI url and metaflow UI url's to the message if environment variable is configured


savingoyal:
Thanks! Let's chat async about the contents and the structure of the message before shipping this PR. It would be good to embed link to the UI in the message (if one can be ascertained)

PR #1425, opened 2023-05-18
More clearly indicate @project name needs to be explicit

emattia:


metaflow/plugins/project_decorator.py
```diff
@@ -17,7 +17,7 @@ class ProjectDecorator(FlowDecorator):
     Specifies what flows belong to the same project.
 
     A project-specific namespace is created for all flows that
-    use the same `@project(name)`.
```
romain-intel, line 20: nit: `my-project` is actually not a valid project name :)

emattia: Thanks, nice catch! Definitely important thing in this case. Updated.

savingoyal:
@emattia Ideally `decorators.py` has no notion of what all decorators are implemented inside the plugins directory.

PR #1435, opened 2023-05-26
Update KubernetesDecorator class docstring to include persistent_volume_claims

tfurmston:
This PR adds the missing docstring for the new `persistent_volume_claims` argument in the KubernetesDecorator class.

It also removes some unused imports.

### DIFF
--- a/metaflow/plugins/kubernetes/kubernetes_decorator.py
+++ b/metaflow/plugins/kubernetes/kubernetes_decorator.py
@@ -1,10 +1,7 @@
-import hashlib
 import json
 import os
 import platform
-import re
 import sys
-from typing import Dict, List, Optional, Union
 
 from metaflow import current
 from metaflow.decorators import StepDecorator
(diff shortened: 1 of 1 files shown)


metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -80,6 +76,8 @@ class KubernetesDecorator(StepDecorator):
         memory allocated for this step.
     tmpfs_path: string, optional
         Path to tmpfs mount for this step. Defaults to /metaflow_temp.
+    persistent_volume_claims: Dict[str, str], optional
+        A dictionary of persistent volumes to be mounted to the pod for this step.
```
saikonen, line 80: Great addition. Could maybe be more specific in what the values of the dict are, maybe with an example? Something along the lines of
```
example: { "pvc-name": "/path/to/mount/on" }
```

tfurmston: yep, good point. 

I've made the docstring more explicit now. Not sure if examples have to follow a particular format. Happy to update if needed. 

Thanks

saikonen (approved):

PR #1436, opened 2023-05-29
feat: ability to use ServerSideEncryption for S3 uploads

zendesk-klross:
This PR allows users to enable Server-Side Encryption (SSE) for S3 uploads through the use of the  `METAFLOW_S3_SERVER_SIDE_ENCRYPTION` configuration setting.

See [this PR](https://github.com/Netflix/metaflow/pull/1382) for prior conversations regarding the topic.


saikonen (approved):

saikonen (changes requested):
looks good and works as intended. Could you run the PR through Black formatting so the pre-commit step passes as well?

https://github.com/Netflix/metaflow/blob/master/CONTRIBUTING.md#code-style

saikonen (approved):

zendesk-klross:
@saikonen just wondering if anything else is needed from our side before we can get this merged, not sure if those failing checks are a blocker? Let us know if you need anything!

Issue #1231, opened 2023-01-13
Argo workflows created by metaflow do not have ttl configured


shrinandj:
When metaflow creates Kubernetes jobs for running steps in a flow, the jobs have a ttl of 7 days. See [this](https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/kubernetes/kubernetes_job.py#L90)

```
                 ttl_seconds_after_finished=7
                 * 60
                 * 60  # Remove job after a week. TODO: Make this configurable
                 * 24,
```

But, when an argo workflow is created, there are no ttl settings set for it. Argo supports a `ttlStrategy` section in the workflow spec as shown [here](https://github.com/argoproj/argo-workflows/blob/master/examples/gc-ttl.yaml)

PR #1438, opened 2023-05-30
feature: add TTL for argo workflows

saikonen:
- adds a TTL of one week for argo workflow specs, matching the TTL present in Kubernetes job specs

addresses #1231

### DIFF
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -32,7 +32,6 @@ from metaflow.metaflow_config import (
     KUBERNETES_FETCH_EC2_METADATA,
     KUBERNETES_LABELS,
     KUBERNETES_NAMESPACE,
-    KUBERNETES_NODE_SELECTOR,
     KUBERNETES_SANDBOX_INIT_SCRIPT,
     KUBERNETES_SECRETS,
     S3_ENDPOINT_URL,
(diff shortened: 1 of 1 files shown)


metaflow/plugins/argo/argo_workflows.py
```diff
@@ -513,6 +512,12 @@ def _compile_workflow_template(self):
                 WorkflowSpec()
                 # Set overall workflow timeout.
                 .active_deadline_seconds(self.workflow_timeout)
+                # Remove workflow after a week.
+                .ttl_strategy(
+                    {
```
savingoyal, line 517: This would only remove the workflow after a week, but leave the pod object untouched. I think we should leave the ttl_strategy for workflows as is but introduce ttl strategy within the container template and the sensor object.

savingoyal:
> If there any chance this functionality could be made configurable? In particular, would it be possible to turn it on or off?
> 
> In our organisation we set the TTL via workflow defaults on the Argo workflow controller. We would like to manage the TTL for all workflows, including those running via Metaflow, in the same manner. As such, we would like to be able to turn off this functionality and rely on our existing approach.

Great point! We can reorient this PR to address pod TTLs so that your current strategy still works.

metaflow/plugins/argo/argo_workflows.py
roofurmston: I think if you delete the workflow the corresponding resources will be deleted too. 

Don't have a reference at hand, but pretty confident this is the case.

roofurmston:
> > If there any chance this functionality could be made configurable? In particular, would it be possible to turn it on or off?
> > In our organisation we set the TTL via workflow defaults on the Argo workflow controller. We would like to manage the TTL for all workflows, including those running via Metaflow, in the same manner. As such, we would like to be able to turn off this functionality and rely on our existing approach.
> 
> Great point! We can reorient this PR to address pod TTLs so that your current strategy still works.

We could adapt things on our side to handle this PR in its current format, e.g., remove the `ttlStrategy` fields from the workflows when we create the manifests, so it is not a massive issue either way. 

It would definitely be some additional work though, so would prefer something else if possible. 

Pod TTLs would work. Also something like having a `METAFLOW_ARGO_TLL` setting in `metaflow_config.py` that defaults to a week. If the TTL was not set when this piece of config is set to zero/none, then it would also be trivial for us to handle.

PR #1441, opened 2023-06-03
fix two docstrings that make API docs unhappy

tuulos:


metaflow/plugins/cards/card_modules/card.py
```diff
@@ -49,7 +49,7 @@ def _get_mustache(self):
         except ImportError:
             return None
 
-    def render(self, task: "Task") -> str:
```
tuulos, line 52: @romain-intel this `"Task"` broke the API docs - given that it is a string, it is probably not useful/necessary as a type annotation

PR #1442, opened 2023-06-06
Adds custom annotations via env variables

tylerpotts:
Implementation of the work described here: https://github.com/Netflix/metaflow/pull/1370

Adds the ability to add custom annotations to the kubernetes jobs via a environment variable

### DIFF
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -43,7 +43,7 @@ from metaflow.mflog import BASH_SAVE_LOGS, bash_capture_logs, export_mflog_env_v
 from metaflow.parameters import deploy_time_eval
 from metaflow.plugins.kubernetes.kubernetes import (
     parse_kube_keyvalue_list,
-    validate_kube_labels,
+    validate_kube_labels_or_annotations,
 )
 from metaflow.util import (
     compress_list,
@@ -217,7 +217,7 @@ class ArgoWorkflows(object):
             return {}
         env_labels = KUBERNETES_LABELS.split(",")
         env_labels = parse_kube_keyvalue_list(env_labels, False)
-        validate_kube_labels(env_labels)
+        validate_kube_labels_or_annotations(env_labels)
         return env_labels
 
     def _get_schedule(self):
--- a/metaflow/plugins/kubernetes/kubernetes.py
+++ b/metaflow/plugins/kubernetes/kubernetes.py
@@ -282,10 +282,14 @@ class Kubernetes(object):
         for name, value in env.items():
             job.environment_variable(name, value)
 
-        annotations = {
-            "metaflow/user": user,
-            "metaflow/flow_name": flow_name,
-        }
+        annotations = self._get_annotations(annotations)
+        annotations.update(
+            {
+                "metaflow/user": user,
+                "metaflow/flow_name": flow_name,
+            }
+        )
+
         if current.get("project_name"):
             annotations.update(
                 {
@@ -294,10 +298,6 @@ class Kubernetes(object):
                     "metaflow/project_flow_name": current.project_flow_name,
                 }
             )
-
-        extra_annotations = self._get_annotations(annotations)
-        if extra_annotations:
-            annotations.update(extra_annotations)
 
         for name, value in annotations.items():
             job.annotation(name, value)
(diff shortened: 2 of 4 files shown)


tylerpotts:
@ savingoyal or @shrinandj could you take a look when you have a moment?

saikonen:
Took this for a spin and it is working as intended for the Kubernetes side. Also applies custom annotations first so does not make it possible to override system annotations (f.ex. `metaflow/*`) as expected 👍

Small question though, as the original discussion was focused on the  `safe-to-evict` control annotation, but now we're introducing arbitrary annotations, do we want this to work with Argo Workflows as well? Currently there is no setup for the custom annotations for Argo Workflows, so that could still be added (similar to how labels are treated in `argo_workflows.py`).

tylerpotts:
@saikonen Agreed that it makes sense to have argo included in this PR. I just added it in the recent commit and it's ready for your review

metaflow/plugins/kubernetes/kubernetes.py
```diff
@@ -167,6 +168,7 @@ def create_job(
         persistent_volume_claims=None,
         tolerations=None,
         labels=None,
+        annotations=None,
```
saikonen, line 171: Not an issue; I think this is unnecessary, as the job will never receive annotations passed into it? The only source for custom annotations being from the environment variable

metaflow/plugins/kubernetes/kubernetes.py
```diff
@@ -280,10 +282,14 @@ def create_job(
...
-        }
+        annotations = self._get_annotations(annotations)
+        annotations.update(
+            {
+                "metaflow/user": user,
+                "metaflow/flow_name": flow_name,
+            }
+        )
+
         if current.get("project_name"):
```
saikonen, line 293: `_get_annotations()` is always being passed an implicit `None` here as `kubernetes.launch_job()` in `kubernetes_cli.py` does not pass annotations. Could just call without an input (seems to be an oversight with the labels as well).

consider if it would be cleaner to lump the system annotations to be defined inside `_get_annotations` as well

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -479,13 +494,15 @@ def _compile_workflow_template(self):
...
         # generate container templates at the top level (in WorkflowSpec) and maintain
         # references to them within the DAGTask.
-
-        annotations = {
-            "metaflow/production_token": self.production_token,
-            "metaflow/owner": self.username,
-            "metaflow/user": "argo-workflows",
-            "metaflow/flow_name": self.flow.name,
-        }
+        annotations = self.kubernetes_annotations
```
saikonen, line 497: we would probably also want to add the same annotations to the Sensor, so similar change to `_compile_sensor` should be enough.

tylerpotts: done

metaflow/plugins/kubernetes/kubernetes.py
tylerpotts: done

metaflow/plugins/kubernetes/kubernetes.py
tylerpotts: remove this for annotations as well as for labels since they are both unused

metaflow/plugins/kubernetes/kubernetes.py
```diff
@@ -280,18 +280,7 @@ def create_job(
...
-        }
-        if current.get("project_name"):
-            annotations.update(
-                {
-                    "metaflow/project_name": current.project_name,
-                    "metaflow/branch_name": current.branch_name,
-                    "metaflow/project_flow_name": current.project_flow_name,
-                }
-            )
+        annotations = self._get_annotations(user, flow_name, current)
```
saikonen, line 283: Backpedaling on my previous comments, as I'm a bit unsure whether this is cleaner than the original approach where `_get_annotations` only parses the env. Now we need to pass in a bunch of additional details to be able to generate the system annotations as well.

Not a dealbreaker for shipping this, I leave it up to you which approach you find better :)

saikonen (changes requested):

metaflow/plugins/kubernetes/kubernetes.py
tylerpotts: I think I do prefer the original approach. It makes it more difficult to follow the logic of the annotations with the system level annotations being put in the `_get_annotations` function. 

I reverted the changes to the function

saikonen (dismissed):

saikonen (approved):

tylerpotts:
@savingoyal Could you merge this when you have a moment?

savingoyal (approved):
approved! i will make this part of the next release.

PR #1452, opened 2023-06-13
feature: add validation for what flow the run_id belongs to

saikonen:
The new CLI commands for argo-workflows that accept a `run_id` have had a pending issue regarding production token authorization, and lacking any guard against passing an arbitrary run_id to a flow file that the user has a valid production token for.

This PR introduces checks for
- the run_id is of the same flow_name as the CLI command was issued on
- the run_id is of the same project and branch as the CLI command was issued on
- the user issuing the command has a production token for the workflow with the run_id
- the run_id is for an Argo Workflow (existing prefix check, simply moved it to the same validation function)

### DIFF
--- a/metaflow/plugins/argo/argo_workflows_cli.py
+++ b/metaflow/plugins/argo/argo_workflows_cli.py
@@ -37,6 +37,10 @@ class IncorrectProductionToken(MetaflowException):
     headline = "Incorrect production token"
 
 
+class RunIdMismatch(MetaflowException):
+    headline = "Run ID mismatch"
+
+
 class IncorrectMetadataServiceVersion(MetaflowException):
     headline = "Incorrect version for metaflow service"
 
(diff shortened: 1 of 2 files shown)


metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -721,3 +717,40 @@ def validate_token(name, token_prefix, authorize, instructions_fn=None):
...
+
+
+def validate_run_id(workflow_name, run_id):
+    """
+    Validates that a run_id adheres to the Argo Workflows naming rules, and
+    that it belongs to the current flow (accounting for project branch as well).
+    """
+    # Verify that user is trying to change an Argo workflow
+    if not run_id.startswith("argo-"):
+        raise MetaflowException("Argo workflow execution id's start with 'argo-'")
```
savingoyal, line 729: Nit - Run IDs for flows executed through Argo Workflows begin with `argo-`

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -721,3 +717,40 @@ def validate_token(name, token_prefix, authorize, instructions_fn=None):
...
+        raise MetaflowException(
+            "The workflow with the run_id *%s* is not for the flow *%s*.\n"
+            "Please use the correct Flow file." % (run_id, current.flow_name)
+        )
+
+    # Verify that we are operating on the correct project and branch.
+    # Project and branch are stored in the workflow template name so this is the easiest comparison to perform.
+    if workflow_name != deployed_workflow_template_name:
+        raise MetaflowException(
+            "The run_id *%s* belongs to the workflow *%s*, not to the workflow *%s*\n"
```
savingoyal, line 751: We don't communicate to the user how `deployed_workflow_template_name` is imputed. Can we check for `project_name` and `branch_name` and compare them with that?

saikonen: Added some more robust checks that separately verify that the user is targeting the correct branch and project, and notifying which one is not matching the deployed workflow. Do the changes seem adequate?

PR #1460, opened 2023-06-19
[bugfix]: Optional check for encryption in s3op response

valayDave:


metaflow/plugins/datatools/s3/s3.py
```diff
@@ -759,7 +759,9 @@ def _info(s3, tmp):
                 "metadata": resp["Metadata"],
                 "size": resp["ContentLength"],
                 "last_modified": get_timestamp(resp["LastModified"]),
-                "encryption": resp["ServerSideEncryption"],
+                "encryption": resp["ServerSideEncryption"]
```
savingoyal, line 762: can you do `resp.get("ServerSideEncryption")` instead?

savingoyal: Can you also add a comment here re: why `ServerSideEncryption` might be missing and under what circumstances?

saikonen:
Complete oversight on my part while testing the original feature. Now verified that this fixes the issue with MinIO. 
I second the suggestion to use `resp.get("ServerSideEncryption")` for a bit of a cleaner setup. Otherwise this is good to go👌

metaflow/plugins/datatools/s3/s3.py
valayDave: Done.

ashtuchkin:
Hey, I'm getting an exception here when working with moto S3 mocking server, seems like this should be changed to `.get` as well?
https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/datatools/s3/s3op.py#L282

saikonen:
> Hey, I'm getting an exception here when working with moto S3 mocking server, seems like this should be changed to `.get` as well? https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/datatools/s3/s3op.py#L282

Thank you for the report, this should now be fixed with the latest 2.9.9 release.

PR #1462, opened 2023-06-22
Otel oss integration

wangchy27:


metaflow/cmd/configure_cmd.py
```diff
@@ -7,6 +7,7 @@
 from metaflow.util import to_unicode
 from metaflow._vendor import click
 from metaflow.util import to_unicode
+import metaflow.tracing as tracing
```
saikonen, line 10: unused import?

metaflow/metadata/heartbeat.py
```diff
@@ -56,7 +56,9 @@ def _ping(self):
 
     def _heartbeat(self):
         if self.hb_url is not None:
-            response = requests.post(url=self.hb_url, data="{}", headers=self.headers)
+            response = requests.post(
+                url=self.hb_url, data="{}", headers=self.headers.copy()
```
saikonen, line 60: is this related to the Otel feature, or some other fix that carried over?

metaflow/mflog/save_logs.py
```diff
@@ -8,9 +8,12 @@
 from metaflow.util import Path
 from . import TASK_LOG_SOURCE
 
+from metaflow.tracing import cli_entrypoint
+
 SMALL_FILE_LIMIT = 1024 * 1024
 
 
+@cli_entrypoint("save_logs")
```
saikonen, line 16: should there be a prefix for the entrypoint or what is the naming convention for these?

metaflow/plugins/datatools/s3/s3.py
```diff
@@ -1601,10 +1607,13 @@ def try_s3_op(last_ok_count, pending_retries, out_lines, inject_failures):
                     try:
                         debug.s3client_exec(cmdline + addl_cmdline)
                         # Run the operation.
+                        env = os.environ.copy()
+                        tracing.inject_tracing_vars(env)
```
saikonen, line 1611: ok to inject the whole env here? no concern of exposing secrets?

metaflow/plugins/metadata/service.py
```diff
@@ -59,7 +59,7 @@ def __init__(self, environment, flow, event_logger, monitor):
     def compute_info(cls, val):
         v = val.rstrip("/")
         try:
-            resp = requests.get(os.path.join(v, "ping"), headers=SERVICE_HEADERS)
+            resp = requests.get(os.path.join(v, "ping"), headers=SERVICE_HEADERS.copy())
```
saikonen, line 62: multiple instances of `.copy()` in this file. Same thoughts, if this is related to the Otel feature or remains of some other fix?

metaflow/metadata/heartbeat.py
wangchy27: Oleg mentioned this in the previous [discussion](https://github.com/Netflix/metaflow/pull/1320#discussion_r1155375173), quote here:
> this and other header .copy() calls are a workaround for https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1729 , we can prob remove this once its fixed upstream

metaflow/plugins/metadata/service.py
wangchy27: Same as above

metaflow/mflog/save_logs.py
wangchy27: For now there's no naming convention. This is basically giving a name to the related trace so it could be used for search/filtering

metaflow/plugins/datatools/s3/s3.py
wangchy27: Do we foreseen any risk here?
cc @oavdeev as this is copied from the initial commit

metaflow/cmd/configure_cmd.py
wangchy27: Good catch! Would removed in next commit

metaflow/plugins/datatools/s3/s3.py
oavdeev: inject_tracing_vars adds tracing vars into dictionary passed to it

metaflow/plugins/metadata/service.py
oavdeev: https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1729

metaflow/tracing/span_exporter.py
```diff
@@ -0,0 +1,62 @@
...
+from metaflow.metaflow_config import (
+    OTEL_ENDPOINT,
+    ZIPKIN_ENDPOINT,
+    CONSOLE_TRACE_ENABLED,
+)
+
+from opentelemetry.exporter.otlp.proto.http.trace_exporter import (
+    OTLPSpanExporter,
+)
+from opentelemetry.exporter.zipkin.proto.http import ZipkinExporter
```
saikonen, line 11: These imports could possibly be made conditional.  With the current implementation I have to install `opentelemetry.exporter.zipkin` as a dependency even if I am not using Zipkin for telemetry, or my stdout fills with import errors due to missing dependencies. Likely same story for Zipkin users getting errors for `opentelemetry.exporter.otlp`

saikonen (dismissed):

metaflow/plugins/kubernetes/kubernetes_job.py
```diff
@@ -135,6 +138,10 @@ def create(self):
                                         "METAFLOW_KUBERNETES_SERVICE_ACCOUNT_NAME": "spec.serviceAccountName",
                                         "METAFLOW_KUBERNETES_NODE_IP": "status.hostIP",
                                     }.items()
+                                ]
+                                + [
+                                    client.V1EnvVar(name=k, value=str(v))
+                                    for k, v in inject_tracing_vars({}).items()
```
saikonen, line 144: this is a bit of a headscratcher, as there are two implementations of this function, but the one imported here seems effectively a no-op?

metaflow/tracing/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+    @wraps(func)
+    def wrapper_func(*args, **kwargs):
+        return func(*args, **kwargs)
+
+    return wrapper_func
+
+
+if CONSOLE_TRACE_ENABLED or OTEL_ENDPOINT or ZIPKIN_ENDPOINT:
+    try:
+        from .tracing_modules import *
```
oavdeev, line 55: Can we change this to an explicit import of those 7 functions? Arguably it would make it more clear for the reader that above dummy implementations get overridden.

saikonen: Pondering on this as well, as it caught me off guard while reviewing even though I was pushing for this approach a while ago :D There are definite benefits to this approach though.

metaflow/plugins/kubernetes/kubernetes_job.py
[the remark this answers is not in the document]
saikonen: only happens when trying to enable telemetry, but not including required libraries with either conda / a custom docker image. I think this is a good notification as is so no need to change things.

metaflow/tracing/__init__.py
saikonen: changed these to explicit imports for clarity.

wangchy27:
The changes are straightforward and make a lot of sense. Thanks @saikonen for handling this!

romain-intel:
Internal tests are fine (the 2 failures are flaky). The R test I don't know.

romain-intel (dismissed):

metaflow/tracing/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+
+def tracing(func):
+    @wraps(func)
+    def wrapper_func(*args, **kwargs):
+        return func(*args, **kwargs)
+
+    return wrapper_func
+
+
+if not BOOTSTRAPPING_CONDA_ENVIRONMENT and (
```
savingoyal, line 53: just `TRACE_DISABLED`

PR #1463, opened 2023-06-22
Make argo events sensor namespace configurable

### LINKED ISSUE #1455: Allow configuration of the namespace argo event sensors
The current implementation of the workflow triggering allows one to specify the webhook url, which can theoretically be in any namespace.  However, the sensors always get generated into the `KUBERNETES_NAMESPACE`, which is the same namespace as the workflow.  We run our workflows in multiple namespaces, but have our eventbus, event sources, and sensors in a single namespace.  The changes required to allow for this would be minimal:

1. [argo_client.py](https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/argo/argo_client.py#L184) - change `._namespace` in this function to `._sensor_namespace` and set that to some config value or fall back to `KUBERNETES_NAMESPACE`
2. [argo_workflows.py - argo client](https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/argo/argo_workflows.py#L239) - initialize this client into the desired sensor namespace
3. [argo_workflows.py - _compile_sensors](https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/argo/argo_workflows.py#L1475) - Change this valued to the sensor namespace

This would create a argo events sensor in one namespace (i.e. `argo-events`) and then trigger a workflow in the namespace that the workflow was ran in.

I could put together a PR pretty quickly if you are open to these changes.  I successfully sent the argo event to a webhook in a separate namespace and triggered a workflow in the original namespace with these three changes.

dhpollack:
- Add ` ARGO_EVENTS_SENSOR_NAMESPACE`  to config (defaults to ` KUBERNETES_NAMESPACE` )
- Add ` sensor_namespace`  parameter to argo client
- Inititalize argo client with sensor namespace parameter when creating sensor

Resolves #1455


dhpollack:
In our set up at work, we don't have event buses in the namespaces where the workflows run.  Thus the "normal" behavior is that if one tries to deploy a workflow (`argo-workflows create`) with the `@trigger` or `@triggered_on_finish` decorators then I receive an error from kubernetes that the eventbus does not exist.

With this PR, I was able to test specifying a namespace and not specifying a namespace.  When I did not specify a namespace, I got the same behavior as the master branch (i.e. kubernetes error).  When I specified the argo events namespace, the argo workflow and sensor successfully get created.  When triggered, the triggered workflow runs in the same workspace as the triggering workflow (not the namespace of the sensor) which is the desired behavior.

metaflow/metaflow_config.py
```diff
@@ -312,6 +312,9 @@
 ARGO_EVENTS_INTERNAL_WEBHOOK_URL = from_conf(
     "ARGO_EVENTS_INTERNAL_WEBHOOK_URL", ARGO_EVENTS_WEBHOOK_URL
 )
+ARGO_EVENTS_SENSOR_NAMESPACE = from_conf(
+    "ARGO_EVENTS_SENSOR_NAMESPACE", KUBERNETES_NAMESPACE
```
chenxiujin10, line 316: Isn't it confusing to put KUBERNETES_NAMESPACE value here?

dhpollack: This basically ensures that the default behavior is the current behavior.  You wouldn't want to set the `KUBERNETES_NAMESPACE` but then have this be some sort of default namespace.  Why do you think it is confusing?

metaflow/metaflow_config.py
```diff
@@ -312,6 +312,9 @@
 ARGO_EVENTS_INTERNAL_WEBHOOK_URL = from_conf(
     "ARGO_EVENTS_INTERNAL_WEBHOOK_URL", ARGO_EVENTS_WEBHOOK_URL
 )
+ARGO_EVENTS_SENSOR_NAMESPACE = from_conf(
```
saikonen, line 315: Is the use case you have such that all deployed sensors would end up in the same namespace, or are there a bunch of different ones?

I'm wondering whether to add the new namespace as part of the CLI configure command as well, or is it expected that users set the relevant env var when deploying their flows?

saikonen (commented):
`argo-workflows delete` (was merged in recently) and `argo-workflows create` both try to delete a sensor as part of the command. Can you make sure that these work across the configured namespaces as well, so users are not left with dangling sensors due to the delete targeting the wrong namespace

metaflow/metaflow_config.py
dhpollack: Yes, our use case is that all sensors live in the same namespace and that we would always set the env var along with other metaflow config env vars when running on kubernetes.  We generally put these config things into a .env file for each repo that uses metaflow so there would be no need for a CLI option in our case.  

Generally, we run workflows in one of two kubernetes namespaces and all of our metaflow + argo related services (including the argo-events eventbus) live in a third namespace.  While we have two namespaces now this could grow and I didn't want to put an eventbus and webhook into each namespace.

dhpollack: I just pushed a commit to use this namespace in the delete command.

saikonen (commented):
To revisit my earlier point on tracking where resources get created: One addition that could be done in scope of this PR is that we could add the sensor name, and the sensor namespace to the workflow template we deploy, either as annotations or as labels.

The main point for this is to provide visibility on where resources are deployed as part of the workflow template, but as an added benefit these could be used as another check for create(redeploy)/delete operations when looking for existing sensors

metaflow/plugins/argo/argo_client.py
```diff
@@ -11,9 +11,10 @@ class ArgoClientException(MetaflowException):
 
 
 class ArgoClient(object):
-    def __init__(self, namespace=None):
+    def __init__(self, namespace=None, sensor_namespace=None):
         self._client = KubernetesClient()
         self._namespace = namespace or "default"
+        self._sensor_namespace = sensor_namespace or self._namespace
```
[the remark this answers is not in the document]
dhpollack: I am not sure if that would work. The `namespace` is also used for knowing where to trigger the workflow.  So I have workflow 1 in namespace A that communicates with a sensor in namespace B that triggers workflow 2 in namespace A again.  And the way the argo client works is `namespace = A` in this case.  If you were to set `namespace = B`, I would have to make sure that workflow 2 does not run in namespace B.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -187,6 +232,12 @@ def _sanitize(name):
     def delete(name):
         client = ArgoClient(namespace=KUBERNETES_NAMESPACE)
 
+        workflow_template = argo_client.get_workflow_template(name)
+        sensor_name = name.replace(".", "-")
```
saikonen, line 236: sensor name is also recorded in the template, but we cannot use that here because of the legacy case where it was not recorded yet, correct?

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -225,6 +225,7 @@ def create(
         obj.echo_always(str(flow), err=False, no_bold=True)
         # TODO: Support echo-ing Argo Events Sensor template
     else:
+        flow.cleanup_previous()
```
saikonen, line 228: I believe there is an issue with running the cleanup only as part of the deploy command. Consider the following scenario:

- user has deployed a sensor flow to namespace `dev-A`, sensor exists in the same namespace
- this feature gets released and the Argo sensor namespace env_var gets configured to `dev-B`
- the user decides that they want to get rid of their flow, and issue an `argo-workflows delete` directly.

Wouldn't this only check for the currently configured sensor namespace, leaving the original sensor in `dev-A` untouched?

saikonen: At least in the case of `delete` it seems sensible to me to keep the concern of cleaning up resources as part of the method.

metaflow/plugins/argo/argo_client.py
```diff
@@ -333,12 +333,15 @@ def schedule_workflow_template(self, name, schedule=None, timezone=None):
                 json.loads(e.body)["message"] if e.body is not None else e.reason
             )
 
-    def register_sensor(self, name, sensor=None):
+    def register_sensor(self, name, sensor=None, sensor_namespace=None):
         if sensor is None:
             sensor = {}
+        if sensor_namespace is None:
+            sensor_namespace = ARGO_EVENTS_SENSOR_NAMESPACE
```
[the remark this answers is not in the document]
dhpollack: yea, I can do that.  I think that I have it this way because the original namespace parameter in the `__init__` method is a null value.  But I can very easily make this the default.

metaflow/plugins/argo/argo_workflows.py
dhpollack: Good point, I will get the actual sensor name from the annotations and then use this as the fallback.

metaflow/plugins/argo/argo_workflows_cli.py
dhpollack: For the legacy case, we need to do cleanup before creating a new argo workflow.  This can only happen if we clean up before calling `flow.deploy()`.

One reason I did this is that there is some inconsistency in the API.  The current implementation already does some sort of "cleanup".  For schedules, we create/suspend them within the `schedule_workflow_template` method which is in the `argo_client.py`, which is called within the `schedule` method in `argo_workflows.py`.  However the sensors are created/deleted within the `schedule` method in `argo_workflows.py`.  So I wanted to move this all into one place, but I couldn't put my code into this `schedule` method.  The previous annotations for the triggers would be overwritten before the `schedule`method.

Currently, the `flow.schedule` function is used to create both the schedule and the sensor with the assumption that the sensor will get overwritten because it always exists in the same namespace.  So in your scenario, if you configured you local environment with `dev-B` then you would have to do an `argo-workflows create` in order to create a sensor in `dev-B`.  Once you did that, it would delete the sensor in `dev-A` due to this line:

https://github.com/dhpollack/metaflow/blob/feature/configure-sensor-namespace/metaflow/plugins/argo/argo_workflows.py#L195

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -205,6 +248,14 @@ def list_templates(flow_name, all=False):
     def delete(name):
         client = ArgoClient(namespace=KUBERNETES_NAMESPACE)
 
+        workflow_template = client.get_workflow_template(name)
+        sensor_name = workflow_template["metadata"]["annotations"].get(
+            "metaflow/sensor_name", name.replace(".", "-")
+        )
+        sensor_namespace = workflow_template["metadata"]["annotations"].get(
```
saikonen, line 255: There's a slight problem with using `ARGO_EVENTS_SENSOR_NAMESPACE` as a fallback here. It should probably be `KUBERNETES_NAMESPACE` instead as:
- if the sensor_namespace annotation is missing, the flow was deployed before the custom sensor namespace feature (so it will only exist in the default kubernetes namespace)
- if `ARGO_EVENTS_SENSOR_NAMESPACE` is set to a custom one, and the flow was deployed before this feature, then `argo-workflows delete` will fail to delete the sensor.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -175,6 +180,44 @@ def deploy(self):
...
+                if (
+                    not self._sensor
+                    or old_sensor_namespace != ARGO_EVENTS_SENSOR_NAMESPACE
+                ):
+                    client.delete_sensor(old_sensor_name, old_sensor_namespace)
+            # Clean up old schedules/cronworkflow
+            old_has_schedule = old_template["metadata"]["annotations"].get(
+                "metaflow/has_schedule"
+            )
+            if not self._schedule and (old_has_schedule is None or old_has_schedule):
```
saikonen, line 216: Only an observation: Previous implementation would simply set the schedule to None I believe, leaving a cron_workflow dangling. This should be a welcome improvement that cleans up the schedule correctly.

metaflow/plugins/argo/argo_workflows.py
dhpollack: Good catch.  Changing it and adding a code comment.

saikonen (approved):

jmmclean:
bump on this please; looks to be approved, just maybe some formalities around getting it merged/deployed

metaflow/plugins/argo/argo_workflows.py
dhpollack: I reverted this

saikonen (approved):

jmmclean:
Bueller? Anyone looking into this PR - its approved but likely need some maintainers to merge/release it

dhpollack:
While this continues to work as of today, I recently changed our setup to deploy the required event infra into all of our namespaces so I probably will not update this again.

saikonen (approved):

PR #1471, opened 2023-06-30
chore: refactor Kubernetes label support

saikonen:
Move all label defining logic to happen inside the Kubernetes decorator instead. The main motivation for this change is to remove duplicate logic between `argo_workflows.py` and `kubernetes.py`. This change also has the benefit of running validators before job submission.

As a side-effect of moving to the decorator, we can also support passing labels as a dict through the deco attributes.


saikonen:
superseded by #1442 after these changes were merged into that branch.

Issue #1458, opened 2023-06-15
Issue with Argo event payload parameters serialization


saikonen:
Due to the latest release containing a rollback to restore existing functionality and fixing https://github.com/Netflix/metaflow/issues/1453, it re-introduces an issue with passing parameters in an Argo Event payload. The problem is with payload values with spaces being used as parameters as follows:

*This one fails*
```python
payload = {
  "first": "test A"
}
event = ArgoEvent(name='test-event', payload=payload)
```

*This one succeeds due to the double-quotes*
```python
payload = {
  "first": '"test A"'
}
event = ArgoEvent(name='test-event', payload=payload)
```

Preferably we would not require the manual double-quotes in order to use values as parameters from the payload

PR #1475, opened 2023-07-07
fix: Argo events parameters with spaces

saikonen:
Fix handling spaces in argo events payload when mapping payload values to parameters.

Tested that this does not re-introduce the double quoting bug for regular parameter usage. Tried out different Sprig functions for escaping, including `squote` and `quote`, but settled with `toJson` in order to escape errant quotes in the middle of the inputs.

Encountered some funky behavior with Argo events parameter handling: when using `dataTemplate`, if the template execution fails (for example due to providing a non-existant jsonpath), the workflow will not fail, instead it will fallback to the default parameter values provided in the sensor. This was a bit of a head scratcher during testing.

Addresses #1458

### DIFF
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -1654,7 +1654,7 @@ class ArgoWorkflows(object):
                                                 # Technically, we don't need to create
                                                 # a payload carry-on and can stuff
                                                 # everything within the body.
-                                                data_template="{{body.payload.%s | squote}}"
+                                                data_template="{{ .Input.body.payload.%s | squote }}"
                                                 % v,
                                                 # Unfortunately the sensor needs to
                                                 # record the default values for
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -1018,7 +1018,9 @@ class ArgoWorkflows(object):
                     + [
                         # Parameter names can be hyphenated, hence we use
                         # {{foo.bar['param_name']}}.
-                        "--%s={{workflow.parameters.%s}}"
+                        # https://argoproj.github.io/argo-events/tutorials/02-parameterization/
+                        # http://masterminds.github.io/sprig/strings.html
+                        "--%s={{workflow.parameters.%s | squote}}"
                         % (parameter["name"], parameter["name"])
                         for parameter in self.parameters.values()
                     ]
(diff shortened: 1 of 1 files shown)


saikonen:
Flows used for testing:

Event emitter
```python
from metaflow import step, FlowSpec, project, current
from metaflow.plugins.argo.argo_events import ArgoEvent

@project(name="sa_test")
class EventTriggeringTest(FlowSpec):
    @step
    def start(self):
        print("Triggering an event with name 'sa-test3'")
        payload = {
            "first": "test A",
            "second": "'testing spaces B'",
            "third": 'single quotes',
            "fourth": "single quote ' in the middle",
            "fifth": 'double " quote in the middle'
        }
        event = ArgoEvent(name='sa-test3', payload=payload)
        event.publish({"data": "with start task content"})

        print(f"current branch name is: {current.branch_name}")
        print(f"current project flow name: {current.project_flow_name}")
        self.next(self.end)

    @step
    def end(self):
        print("Done! 🏁")


if __name__ == "__main__":
    EventTriggeringTest()
```

Triggered flow
```python
import os
from metaflow import step, FlowSpec, current, Parameter

from metaflow import trigger

@trigger(
    events=[
        {
            "name": "sa-test3",
            "parameters": {
                "param_a": "first",
                "param_dupe": "first",
                "param_b": "second",
                "param_c": "third",
                "param_d": "fourth",
                "param_e": "fifth",
            }
        }
    ]
)
class EventWithQuotes(FlowSpec):
    param_a = Parameter(
        name="param_a",
        default="default value A",
        type=str,
        required=True
    )

    param_dupe = Parameter(
        name="param_dupe",
        default="default value A duplicate",
        type=str,
        required=True
    )

    param_b = Parameter(
        name="param_b",
        default="default value B",
        type=str,
        required=True
    )

    param_c = Parameter(
        name="param_c",
        default="default value C",
        type=str,
        required=True
    )

    param_d = Parameter(
        name="param_d",
        default="default value D",
        type=str,
        required=True
    )

    param_e = Parameter(
        name="param_e",
        default="default value E",
        type=str,
        required=True
    )

    @step
    def start(self):
        print("Starting 👋")
        print(f"current trigger: {current.trigger}")

        for p in ["a", "dupe", "b", "c", "d", "e"]:
            param = getattr(self, f"param_{p}", None)
            print(f"value for param_{p}: {param}")

        argo_envs = dict([(key, val) for key, val in os.environ.items() if key.startswith('METAFLOW_ARGO_EVENT_PAYLOAD_')])
        print(f"Argo Environment payload vars: {argo_envs}")

        self.next(self.end)

    @step
    def end(self):
        print("Done! 🏁")


if __name__ == "__main__":
    EventWithQuotes()
```

PR #1478, opened 2023-07-10
feature: add PagerDuty support for Argo Workflows

saikonen:
Adds support for PagerDuty notifications for Argo Workflows.

- [x] Agree upon the default contents that should be passed in `custom_details` and other optional fields.
- [ ] Possibly refactor notification providers in this PR to more easily support future additions?

### DIFF
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -700,7 +700,7 @@ class ArgoWorkflows(object):
                                 .expression("workflow.status == 'Error'")
                                 .template("slack-notify-on-error"),
                             }
-                            if self.notify_on_error
+                            if self.notify_on_error and self.notify_slack_webhook_url
                             else {}
                         ),
                         **(
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -1393,35 +1393,9 @@ class ArgoWorkflows(object):
         # TODO: Add details to slack message
         templates = []
         if self.notify_on_error:
-            templates.append(
-                Template("notify-on-error").http(
-                    Http("POST")
-                    .url(self.notify_slack_webhook_url)
-                    .body(
-                        json.dumps(
-                            {
-                                "text": ":rotating_light: _%s/argo-{{workflow.name}}_ failed!"
-                                % self.flow.name
-                            }
-                        )
-                    )
-                )
-            )
+            templates.append(self._slack_error_template)
         if self.notify_on_success:
-            templates.append(
-                Template("notify-on-success").http(
-                    Http("POST")
-                    .url(self.notify_slack_webhook_url)
-                    .body(
-                        json.dumps(
-                            {
-                                "text": ":white_check_mark: _%s/argo-{{workflow.name}}_ succeeded!"
-                                % self.flow.name
-                            }
-                        )
-                    )
-                )
-            )
+            templates.append(self._slack_success_template)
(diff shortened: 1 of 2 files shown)


metaflow/plugins/argo/argo_workflows.py
```diff
@@ -1393,49 +1422,149 @@ def _exit_hook_templates(self):
...
+                    .url(
+                        self.notify_slack_webhook_url
+                        or "https://events.pagerduty.com/v2/enqueue"
+                    )
                     .success_condition("true == true")
                 )
             )
         return templates
 
+    @property
```
savingoyal, line 1449: any reason why this needs to be a property?

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -670,24 +674,49 @@ def _compile_workflow_template(self):
...
+                        ),
+                        **(
+                            {
+                                # workflow status maps to Failed or Error
+                                "pagerduty-notify-on-failure": LifecycleHook()
+                                .expression("workflow.status == 'Failed'")
+                                .template("pagerduty-notify-on-error"),
+                                "pagerduty-notify-on-error": LifecycleHook()
+                                .expression("workflow.status == 'Error'")
+                                .template("pagerduty-notify-on-error"),
```
savingoyal, line 716: minor nit - `notify-pagerduty-on-error` etc. to be consistent with `notify-pagerduty-integration-key`

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -391,12 +399,18 @@ def make_flow(
             "Argo Workflows requires --datastore=s3 or --datastore=azure or --datastore=gs"
         )
 
-    if (notify_on_error or notify_on_success) and not notify_slack_webhook_url:
+    if (notify_on_error or notify_on_success) and not (
+        notify_slack_webhook_url or notify_pagerduty_integration_key
+    ):
         raise MetaflowException(
-            "Slack notifications require specifying an incoming Slack "
+            "Workflow notifications require specifying a notification provider.\n\n"
```
savingoyal, line 406: `Notifications require specifying an incoming Slack webhook url via --notify-slack-webhook-url or Pager Duty v3 webhook url via --notify-pager-duty-webhook-url.\n
If you would like to set up notifications for your Slack workspace, follow the instructions at https://api.slack.com/messaging/webhooks to generate a webhook url.\n For notifications through Pager Duty, generate an integration key by following the instructions at https://support.pagerduty.com/docs/services-and-integrations#create-a-generic-events-api-integration"`

savingoyal (approved):
LGTM! Minor comments. Let's address those and cut a release.

We can create a follow-up PR to refactor the notifications bit to make introducing new providers easier.

metaflow/plugins/argo/argo_workflows_cli.py
saikonen: Second line should probably instead be
`Pager Duty Events API V2 integration key via --notify-pager-duty-integration-key.\n If you would like to set up `

From what I glanced, the PagerDuty webhook support is for providing a webhook subscription (own endpoint for Pagerduty to send data to), which is why it is not used here. 

`--notify-pager-duty-integration-key` is the only required value, as the Events API V2 is a static endpoint.

metaflow/plugins/argo/argo_workflows.py
saikonen: just a bit cleaner usage, but it does introduce unnecessary setters/delete to the object, so I can revert to a simple instance methods instead.

PR #1479, opened 2023-07-10
fix: s3 op bug with ServerSideEncryption

saikonen:
fixes bug with S3 ServerSideEncryption headers affecting S3 operations (`get_many / get_recursive / get_all / put_many`)

### DIFF
--- a/metaflow/plugins/datatools/s3/s3.py
+++ b/metaflow/plugins/datatools/s3/s3.py
@@ -1004,9 +1004,9 @@ class S3(object):
                                 "metadata"
                             ], range_info, info[
                                 "last_modified"
-                            ], info[
+                            ], info.get(
                                 "encryption"
-                            ]
+                            )
                     else:
                         yield self._s3root, s3prefix, None
                 else:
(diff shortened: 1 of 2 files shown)


saikonen:
minimum tests I ran for coverage on the reported issues:

```python
from metaflow import S3

S3_BASE="s3://sakari-tests"

def test_s3(filename: str):
    with S3(s3root=S3_BASE) as s3:
            s3.put(filename, "super secret content")

            # Additional test for multi-mode operations.
            s3.put(filename, "super secret content", overwrite=True)

            s3.put_many([(filename, "super secret content")], overwrite=True)

    with S3(s3root=S3_BASE) as s3:
        file = s3.get(filename)
        info = s3.info(filename)
        print(info.encryption)
        print(f"{file.text}")

        files = s3.get_many([filename])
        info_many = s3.info_many([filename])
        print(info_many[0].encryption)
        print(f"{files[0].text}")
        
        
        files = s3.get_all(return_info=True)
        s3.info(filename)
        print(f"{files[0].text}")

        files = s3.get_recursive([filename], return_info=True)
        s3.info(filename)
        print(f"{files[0].text}")
```

PR #1483, opened 2023-07-13
fix: validate required configuration for Batch

saikonen:
add validation for the Batch decorator that checks that an `iam_role` is configured (either through the decorator, or with an environment variable).

Currently if `METAFLOW_ECS_S3_ACCESS_IAM_ROLE` is not configured, trying to run a flow with `--with batch` or deploying with step-functions will fail to an error
```
botocore.exceptions.ParamValidationError: Parameter validation failed:
Invalid type for parameter containerProperties.jobRoleArn, value: None, type: <class 'NoneType'>, valid types: <class 'str'>
```

### DIFF
--- a/metaflow/plugins/aws/batch/batch_decorator.py
+++ b/metaflow/plugins/aws/batch/batch_decorator.py
@@ -91,8 +91,8 @@ class BatchDecorator(StepDecorator):
         "gpu": None,
         "memory": None,
         "image": None,
-        "queue": BATCH_JOB_QUEUE,
-        "iam_role": ECS_S3_ACCESS_IAM_ROLE,
+        "queue": BATCH_JOB_QUEUE,  # Will be replaced with an available queue if not provided.
+        "iam_role": ECS_S3_ACCESS_IAM_ROLE,  # Required
         "execution_role": ECS_FARGATE_EXECUTION_ROLE,
         "shared_memory": None,
         "max_swap": None,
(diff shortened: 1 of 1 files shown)


metaflow/plugins/aws/batch/batch_decorator.py
```diff
@@ -149,6 +149,13 @@ def step_init(self, flow, graph, step, decos, environment, flow_datastore, logge
         if flow_datastore.TYPE != "s3":
             raise BatchException("The *@batch* decorator requires --datastore=s3.")
 
+        # Require iam_role
+        if self.attributes["iam_role"] is None:
+            raise BatchException(
+                "Executions with AWS Batch require an IAM Role.\n"
```
savingoyal, line 155: Minor nit - `The *@batch* decorator requires @batch(iam_role="role-name") that allows AWS Batch job to communicate with Amazon S3 datastore.`

savingoyal (approved):
Minor nit - otherwise looks good to go

metaflow/plugins/aws/batch/batch_decorator.py
saikonen: Any reason to remove the mention of configuring this with the environment variable instead? I would suggest
`The *@batch* decorator requires an IAM Role that allows AWS Batch job to communicate with Amazon S3 datastore.\n`
and keeping the other two lines intact

savingoyal: You can specify it either with *@batch(iam_role="role-name")* or by setting *METAFLOW_ECS_S3_ACCESS_IAM_ROLE* in your configuration

Issue #906, opened 2022-01-16
`metaflow_environment` dependencies can override or conflict with those set by the Batch docker image, breaking user code


ryan-williams:
Pasting the README from [runsascoded/mf-pip-issue](https://github.com/runsascoded/mf-pip-issue), where I have some repro files as well:

# Metaflow/`pip`/Batch issue
Metaflow runs `pip install awscli … boto3` [while setting up task environements in Batch](https://github.com/Netflix/metaflow/blob/2.4.8/metaflow/metaflow_environment.py#L85), which can break `aiobotocore<2.1.0`.

## Repro
Docker image [runsascoded/mf-pip-issue-batch](https://hub.docker.com/repository/docker/runsascoded/mf-pip-issue-batch) ([`batch.dockerfile`](https://github.com/runsascoded/mf-pip-issue/blob/main/batch.dockerfile)) pins recent versions of `botocore` and `aiobotocore`:
- [`aiobotocore==1.4.2`](https://pypi.org/project/aiobotocore/1.4.2/) (October 5, 2021)
- [`botocore==1.20.106`](https://pypi.org/project/botocore/1.20.106/) (July 6, 2021, [required by `aiobotocore==1.4.2`](https://github.com/aio-libs/aiobotocore/blob/1.4.2/setup.py#L10))

### Local mode: ✅
They work fine together normally; [runsascoded/mf-pip-issue-local](https://hub.docker.com/repository/docker/runsascoded/mf-pip-issue-local) ([`local.dockerfile`](https://github.com/runsascoded/mf-pip-issue/blob/main/local.dockerfile)) runs [`s3_flow_test.py`](https://github.com/runsascoded/mf-pip-issue/blob/main/s3_flow_test.py) successfully (in "local" mode):
```bash
docker run -it --rm runsascoded/mf-pip-issue-local
# Metaflow 2.4.8 executing S3FlowTest for user:user
# …
# 2022-01-16 21:21:59.162 Done!
```

### Batch mode: ❌
However, with a Metaflow Batch queue configured:
```bash
python s3_flow_test.py run --with batch:image=runsascoded/mf-pip-issue-batch
```
fails with:
```
AttributeError: 'AioClientCreator' object has no attribute '_register_lazy_block_unknown_fips_pseudo_regions'
```
due to a version mismatch (`botocore>=1.23.0`, `aiobotocore<2.1.0`).

### Version mismatch
[...]

PR #1490, opened 2023-07-21
feat: config option to disable package installation in AWS Batch docker containers

zendesk-klross:
This resolves [issue 906](https://github.com/Netflix/metaflow/issues/906), where AWS Batch default package installation may supersede user-specified dependency versions. The new `AUTO_INSTALL_DEPENDENCIES` configuration flag enables users to opt out of this default install. Note that it becomes the user's responsibility to configure their docker images with the required dependencies for their selected datastore-type.

### DIFF
--- a/metaflow/metaflow_config.py
+++ b/metaflow/metaflow_config.py
@@ -248,6 +248,12 @@ SERVICE_INTERNAL_URL = from_conf("SERVICE_INTERNAL_URL", SERVICE_URL)
 # default to True.
 BATCH_EMIT_TAGS = from_conf("BATCH_EMIT_TAGS", False)
 
+# Set to False to disable default package installation in docker images during job launch. 
+# This allows better dependency control for your flow, but please ensure that your 
+# environment has all the required packages for your datastore type (any necessary 
+# dependencies should be explicitly installed in your custom container).
+AUTO_INSTALL_DEPENDENCIES = from_conf("AUTO_INSTALL_DEPENDENCIES", True)
+
 ###
 # AWS Step Functions configuration
 ###
--- a/metaflow/metaflow_environment.py
+++ b/metaflow/metaflow_environment.py
@@ -8,6 +8,7 @@ from metaflow.exception import MetaflowException
 from metaflow.extension_support import dump_module_info
 from metaflow.mflog import BASH_MFLOG
 from . import R
+from metaflow.metaflow_config import AUTO_INSTALL_DEPENDENCIES
 
 version_cache = None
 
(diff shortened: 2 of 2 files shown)


savingoyal:
@zendesk-klross would finding a better workaround for #906 help?

zendesk-klross:
> @zendesk-klross would finding a better workaround for #906 help?

We played around with the idea of activating a pre-defined virtual environment for the actual job execution (following the `metaflow_environment` dependency installs) and then deactivating it, but there was no obvious spot in the batch code to implement this. We're definitely open to discussing alternative ideas though!

saikonen (approved):

savingoyal:
@saikonen is there a way to only install these dependencies if they are not present? that will reduce the need for this PR.

saikonen:
I had a quick trial with another idea that seems to actually solve the original issue. The problem is that `pip install` does not respect the dependencies of already installed packages that are not part of the command. In the example it leads to an existing installation of `aiobotocore` breaking due to depending on an older version of `botocore`, which gets upgraded as part of `pip install awscli`.

One workaround for this is to freeze the existing environment and carry this along all of the commands, e.g.
`python -m pip freeze > environment_requirements.txt`
`python -m pip install -r environment_requirements.txt awscli boto3`

This seems to resolve the original issue, but I still need to verify that the `-r environment_requirements.txt` is effectively a no-op. It also carries with it a possible gotcha, where effectively constraining the version of `boto3` due to existing packages might at some point install a version not supporting required features.

Any thoughts on the approach? I'll try to get some further testing done today and open a draft for the other approach.

tylerpotts:
@saikonen @savingoyal Another alternative is to use bash and `pip show` to only invoke the pip install command if the pip package is not found.

Example for awscli: 

`if ! pip show awscli &> /dev/null; then pip install awscli; fi`

`pip show ..` returns `0` if a package is found, otherwise `1`

fsat:
@savingoyal @saikonen - thank you for considering and implementing alternate approach in #1502

After a discussion we've had with @savingoyal, I've been thinking about our requirements some more.

I think despite having #1502, we will still need a way to ensure `pip install` is not called during execution.

This is because our security team requires that during execution in our Production environment, we should bake all dependencies we need during Docker image build time. With the `pip install` in the `_get_install_dependencies_cmd`, the install will take place during execution time which will break this requirement.

Also on top of that, our security team also requires all Python libraries installed should come through our local internal mirror instead of public PyPi. The way `pip install` specified in the `_get_install_dependencies_cmd` will not allow us to install the Python libraries from our local internal mirror.

I think this PR can co-exist with https://github.com/Netflix/metaflow/pull/1502 since `AUTO_INSTALL_DEPENDENCIES` configuration falls back to `True`, which means the existing behaviour is still preserved by default.

Please let me know what you think, always happy to discuss if you have any objections or questions.

saikonen:
The environment variable approach of this PR is the most straightforward solution for introducing definitive control over whether pip install runs. The security considerations are very understandable and #1502 offers no solution to those so I would agree we go ahead with this PR, especially since it does not change the default behaviour.

zendesk-klross:
@saikonen Thanks so much for looking into this issue with us. Do you know if everyone is on board with this solution? If that's the case, is there anything else we should do from our end to get the PR over the line?

saikonen:
> @saikonen Thanks so much for looking into this issue with us. Do you know if everyone is on board with this solution? If that's the case, is there anything else we should do from our end to get the PR over the line?

Had a quick chat around this implementation and #1502 today and some questions came up still.

For the security aspect of getting rid of pip installs as part of bootstrapping. As you mentioned that you are already required to bake in all the dependencies into your Docker images, it should already lead to the pip install acting as a no-op. Is this not the case due to using an internal mirror or some other reason? Trying to get an understanding on what aspect is not working in the case of baked in dependencies.

We can also schedule a quick call and have a chat on this if it is more convenient.

fsat:
> For the security aspect of getting rid of pip installs as part of bootstrapping. As you mentioned that you are already required to bake in all the dependencies into your Docker images, it should already lead to the pip install acting as a no-op. Is this not the case due to using an internal mirror or some other reason? Trying to get an understanding on what aspect is not working in the case of baked in dependencies.

Since the `pip install` executes when the container runs, we may have a different library versions and its transitive dependencies between what is baked inside the container vs what is being fetched by `pip install`.

In other words, I don't believe we can provide guarantee that `pip install` will act as a no-op.

Let's suppose we have a container which bakes in the `boto3==1.28.30` as of 22 August 2023. And let's suppose there's a new release of `boto3` on 23 August 2023.

Since we don't specify the version of the libraries we install, i.e. for S3 `python -m pip install awscli boto3 -qqq` - both public PyPi and our internal mirror will resolve to the latest of `awscli` and `boto`.

This means, if we were to run the container on 23 August 2023 onwards, the `pip install` will silently upgrade the `boto3` inside the container whilst we need it to be running on `boto3==1.28.30`. Our security requirement mandates us to prevent these silent upgrade from taking place.

I suppose the only way `pip install` can truly become no-op is if we can supply the previously frozen the versions of the library we need and its transitive dependencies.

Hopefully this explanation makes sense.

Please do let me know if you'd like a follow up call or chat - always more than happy to discuss if you have objections or more questions.

zendesk-klross:
We've discussed this some more, and although the implementation in [#1502](https://github.com/Netflix/metaflow/pull/1502) would function as a no-op in situations where all required dependencies are already packaged inside the container, it would continue to silently install from the public PyPi when dependencies are missing, or potentially introduce some overhead time when dependencies are conflicting. If possible, we would still prefer to have the batch task fail in cases of missing/conflicting dependencies, and have our users resolve the gaps and/or conflicts themselves by building a new container. Do you think it would it be possible to combine both your solution and our `AUTO_INSTALL_DEPENDENCIES` toggle?

savingoyal:
@zendesk-klross Given that users can write arbitrary code within their steps, it is feasible for someone to execute `os.system(pip install foo)` and install arbitrary packages from PyPI. Our default recommendation is to turn off external network access to cloud workloads - that ensures no data goes out or gets in. 

Yet another workaround would be to set the global pip index to null in the image - which would ensure that metaflow cannot find any PyPI packages by itself.

Would either of these workaround work? Our default stance is to see if there are valid workarounds that don't require any code changes. Introducing new configurations locks us into specific patterns - hence the extended nature of this deliberation.

fsat:
@savingoyal - thank you for the suggestion, we'll try it on our end and get back to you.

zendesk-klross:
@savingoyal @saikonen After discussing, we've decided to try and disable public PyPi access from AWS Batch on our end, so it looks like we can close this PR. We know that adding an extra environment variable might entail additional maintenance from your end, which isn't ideal. In any case, thanks so much for looking into this with us!

PR #1491, opened 2023-07-25
Fix/863 - Improve error message in metaflow.S3 class when DATATOOLS_S3ROOT is not configured.

tfurmston:
I recently encountered the issue raised in I[ssue 863](https://github.com/Netflix/metaflow/issues/683). 

I note that there is an existing [PR](https://github.com/Netflix/metaflow/pull/853) on this issue. However, it is very old and I assume now dead. As a result, I have made a new PR that addresses the comments in the original PR.

### DIFF
--- a/metaflow/plugins/datatools/s3/s3.py
+++ b/metaflow/plugins/datatools/s3/s3.py
@@ -495,6 +495,10 @@ class S3(object):
 
     @classmethod
     def get_root_from_config(cls, echo, create_on_absent=True):
+        if DATATOOLS_S3ROOT is None:
+            raise MetaflowS3URLException(
+                "DATATOOLS_S3ROOT is not configured when trying to use S3 storage"
+            )
         return DATATOOLS_S3ROOT
 
     def __init__(
@@ -512,6 +516,10 @@ class S3(object):
 
         if run:
             # 1. use a (current) run ID with optional customizations
+            if DATATOOLS_S3ROOT is None:
+                raise MetaflowS3URLException(
+                    "DATATOOLS_S3ROOT is not configured when trying to use S3 storage"
+                )
             parsed = urlparse(DATATOOLS_S3ROOT)
             if not bucket:
                 bucket = parsed.netloc
--- a/metaflow/plugins/datatools/s3/s3.py
+++ b/metaflow/plugins/datatools/s3/s3.py
@@ -495,10 +495,6 @@ class S3(object):
 
     @classmethod
     def get_root_from_config(cls, echo, create_on_absent=True):
-        if DATATOOLS_S3ROOT is None:
-            raise MetaflowS3URLException(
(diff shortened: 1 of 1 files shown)


metaflow/plugins/datatools/s3/s3.py
```diff
@@ -495,6 +495,10 @@ class S3(object):
 
     @classmethod
     def get_root_from_config(cls, echo, create_on_absent=True):
+        if DATATOOLS_S3ROOT is None:
```
savingoyal, line 498: @tfurmston any insights into why we need to throw an error here?

tfurmston: Yes, I believe it is because this mwthod is called from `IncludeFile` class [here](https://github.com/Netflix/metaflow/blob/master/metaflow/includefile.py#L374). 

@romain-intel mentioned it in the old PR

tfurmston: @savingoyal @romain-intel Any thoughts on this one? Happy to remove this line, if you prefer?

saikonen:
could you give some more context on what the setup looks like where you ran into the issue? I'm trying to see how widely this issue might've been affecting users.

So far the two cases I've tested with seem to be the only plausible ones. The first one is from the original issue where a user has no config at all. The second is a case where the config is specifically set to **not** have `"METAFLOW_DEFAULT_DATASTORE": "s3"` and the flow tries S3 access.
In other cases as soon as the datastore is set to S3, the validation in `cli.py` ensures that the config is correct.

metaflow/plugins/datatools/s3/s3.py
tfurmston: OK, cool. 

Yeah, it makes sense. I check the `IncludeFile` class only, from which it wasn't obvious to me that any such previous checks would occur. 

I've removed this part from the PR now.

tfurmston:
> could you give some more context on what the setup looks like where you ran into the issue? I'm trying to see how widely this issue might've been affecting users.
> 
> So far the two cases I've tested with seem to be the only plausible ones. The first one is from the original issue where a user has no config at all. The second is a case where the config is specifically set to **not** have `"METAFLOW_DEFAULT_DATASTORE": "s3"` and the flow tries S3 access. In other cases as soon as the datastore is set to S3, the validation in `cli.py` ensures that the config is correct.

 When one runs a flow locally assignments of the form, `self.x = 1`, save the data artefacts to local files and not S3. So I was, perhaps naively given the name of the class, testing whether the S3 client has analogous behaviour. Such behaviour would be useful for quick local code development as it doesn't require the user to sign into AWS. (Within our infrastructure there is a bit of overhead to signing into AWS in order to use things such as S3.) Under your description my use case would fall under case 1, i.e., no config at all.

It turns out this functionality is not possible. However, instead of a clear error message I got the one mentioned in the issue, which was quite confusing.

PR #1497, opened 2023-08-03
Fix/doctring

wangchy27:
Internal netflix is using Griffe format for all documentation. This doc string failed because not all parameters listed in the comment are in the init() function. This PR is to remove unused parameters from the comments


saikonen (approved):
Approving as the change is in line with how Griffe suggest handlings kwargs in docstrings: https://mkdocstrings.github.io/griffe/docstrings/#other-parameters

For context, could you mention what the failure with Griffe was, as I wasn't able to get any errors even without the introduced change with `griffe dump metaflow.includefile` or `griffe check metaflow.includefile`

PR #1500, opened 2023-08-11
Adding 'disk' parameter to @resources decorator

jaypond:
We would like to use the @resources decorator instead of @kubernetes to specify resource requirements for our job so we can run flows locally or on Kubernetes by using `--with kubernetes` instead of modifying the code. The @resources decorator only supports cpu, gpu, memory and shared_memory at the moment.

This PR just adds the disk parameter. Default value (10240mb) is the same as the one in @kubernetes.

### DIFF
--- a/metaflow/plugins/resources_decorator.py
+++ b/metaflow/plugins/resources_decorator.py
@@ -25,7 +25,7 @@ class ResourcesDecorator(StepDecorator):
         Number of CPUs required for this step.
     gpu : int, default: 0
         Number of GPUs required for this step.
-    disk : int, default: 10240
+    disk : int, optional
         Disk size (in MB) required for this step. Only applies on Kubernetes.
     memory : int, default: 4096
         Memory size (in MB) required for this step.
@@ -38,7 +38,7 @@ class ResourcesDecorator(StepDecorator):
     defaults = {
         "cpu": "1",
         "gpu": "0",
-        "disk": "10240",
+        "disk": None,
         "memory": "4096",
         "shared_memory": None,
     }
--- a/metaflow/plugins/resources_decorator.py
+++ b/metaflow/plugins/resources_decorator.py
@@ -25,6 +25,8 @@ class ResourcesDecorator(StepDecorator):
         Number of CPUs required for this step.
     gpu : int, default: 0
         Number of GPUs required for this step.
+    disk : int, default: 10240
+        Disk size (in MB) required for this step. Only applies on Kubernetes.
     memory : int, default: 4096
         Memory size (in MB) required for this step.
     shared_memory : int, optional
@@ -36,7 +38,7 @@ class ResourcesDecorator(StepDecorator):
     defaults = {
         "cpu": "1",
         "gpu": "0",
+        "disk": "10240",
         "memory": "4096",
         "shared_memory": None,
-        "disk": "10240",
     }
--- a/metaflow/plugins/resources_decorator.py
+++ b/metaflow/plugins/resources_decorator.py
@@ -33,4 +33,10 @@ class ResourcesDecorator(StepDecorator):
     """
 
     name = "resources"
-    defaults = {"cpu": "1", "gpu": "0", "memory": "4096", "shared_memory": None}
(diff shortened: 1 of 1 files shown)


metaflow/plugins/resources_decorator.py
```diff
@@ -33,4 +33,10 @@ class ResourcesDecorator(StepDecorator):
     """
```
saikonen, line 33:
```
disk : int, default: 10240
   Disk size (in MB) required for this step. Only applies on Kubernetes.
```

metaflow/plugins/resources_decorator.py
```diff
@@ -33,4 +33,10 @@ class ResourcesDecorator(StepDecorator):
...
 
     name = "resources"
-    defaults = {"cpu": "1", "gpu": "0", "memory": "4096", "shared_memory": None}
+    defaults = {
+        "cpu": "1",
+        "gpu": "0",
+        "memory": "4096",
+        "shared_memory": None,
```
saikonen, line 40: #1504 opened for this, currently not yet supported for Kubernetes.

saikonen: https://outerbounds-community.slack.com/archives/C020U025QJK/p1691751137791999 earlier discussions with context on why `disk` has been missing.

saikonen:
Added some context and an addition to the docstring. I'll give this a more thorough spin tomorrow morning :)

metaflow/plugins/resources_decorator.py
saikonen: could you add this to the docstring for the decorator for clarity? otherwise the PR is looking great :)

saikonen (approved):

saikonen:
As one last discussion point, we do need to consider the impact of this change as it is a breaking one in the following scenario due to how attribute overriding works between the resources and kubernetes decorators. Consider someone who has an existing flow as such:

```python
from metaflow import step, FlowSpec, resources, kubernetes


class HelloFlow(FlowSpec):
    @resources
    @kubernetes(disk=4096)
    @step
    def start(self):
        print("Starting 👋")
        self.next(self.end)

    @step
    def end(self):
        print("Done! 🏁")


if __name__ == "__main__":
    HelloFlow()
```

prior to this change they would allocate only 4096 of disk, after this PR is introduced the default 10240 from resources will override this value due to being larger.

jaypond:
Good catch. Just tested and pushed the update you recommended.

No particular reason for using the same default as the one in the Kubernetes decorator aside from having them match.

saikonen (dismissed):

saikonen:
Encountered one last issue with this feature while testing on other platforms. Specifically with running on batch while using the resource deco, it now fails with an error of
```
Decorator 'batch' does not support the attribute 'disk'. These attributes are supported: cpu, gpu, memory, image, queue, iam_role, execution_role, shared_memory, max_swap, swappiness, inferentia, host_volumes, use_tmpfs, tmpfs_tempdir, tmpfs_size, tmpfs_path.
```

This is due to the way we inject attributes for the batch deco via `aws_utils.py#compute_resource_attributes()` which does not have a distinction between supported or unsupported attributes.

jaypond:
Would it be feasible if we just remove attributes which aren't supported by a decorator instead of throwing an error? Maybe throw a warning instead?

Something like this:
```
class Decorator(object):
    """
    Base class for all decorators.
    """

    name = "NONAME"
    defaults = {}
    # `allow_multiple` allows setting many decorators of the same type to a step/flow.
    allow_multiple = False

    def __init__(self, attributes=None, statically_defined=False):
        self.attributes = self.defaults.copy()
        self.statically_defined = statically_defined

        if attributes:
            for k, v in attributes.items():
                if k in self.defaults:
                    self.attributes[k] = v
                else:
                    #raise InvalidDecoratorAttribute(self.name, k, self.defaults)
                    del self.attributes[k]
```

saikonen (approved):
should be good to go with the merged fixes to batch/sfn resource decorator handling.

PR #1501, opened 2023-08-14
Enabling Ray Cluster

rileyhun:
# Overview
We require the ability to activate a Ray cluster using AWS Batch multi-node parallel jobs

This PR accomplishes the following: 
- Creates a `@ray_parallel` decorator that a user can use to effectively spin up a Ray cluster from AWS Batch multi-node parallel job. This decorator follows the official Ray documentation on setting up Ray on-premise. 

### What has changed
- `@ray_parallel` decorator modeled from base `@parallel` decorator

## Testing evidence


savingoyal:
This PR is encapsulated in https://github.com/outerbounds/metaflow-ray and we can close this for now.

PR #1502, opened 2023-08-15
fix: freeze environment packages before bootstrapping

saikonen:
This aims to resolve issue #906 where installing code package bootstrapping dependencies inadvertently upgrades dependencies, possibly breaking existing libraries that depend on older versions.
e.g. 
```
aiobotocore==1.4.2
botocore==1.20.106

$ pip install boto3
# observe that botocore was upgraded, breaking the installed version of aiobotocore
```
Alternative approach to #1490 

Testing notes:

- by freezing the environment packages and appending them to the `pip install` commands, it successfully guards all existing packages from upgrades, instead trying to resolve the minimum viable version for the requested packages, e.g. `awscli, boto3`
- because this relies on pip to resolve the dependencies, it can be quite slow in cases where environment packages could introduce conflicts. With a test case I observed a ~4min overhead with the pip install.
- the regular use case where awscli and boto3 can be installed without concerns remains fast.

### DIFF
--- a/metaflow/metaflow_environment.py
+++ b/metaflow/metaflow_environment.py
@@ -119,24 +119,29 @@ class MetaflowEnvironment(object):
             )
 
     def _get_install_dependencies_cmd(self, datastore_type):
-        cmds = ["%s -m pip install requests -qqq" % self._python()]
+        dependencies = ""
         if datastore_type == "s3":
-            cmds.append("%s -m pip install awscli boto3 -qqq" % self._python())
+            dependencies = "awscli boto3"
         elif datastore_type == "azure":
-            cmds.append(
-                "%s -m pip install azure-identity azure-storage-blob simple-azure-blob-downloader -qqq"
-                % self._python()
+            dependencies = (
+                "azure-identity azure-storage-blob simple-azure-blob-downloader"
             )
         elif datastore_type == "gs":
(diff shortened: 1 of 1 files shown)


tylerpotts:
While this bash line is a little lengthy, it would get around doing any pip install operations if `boto3` and `awscli` are present

`pip show awscli &> /dev/null; awscli_missing=$?; pip show boto3 &> /dev/null; boto3_missing=$?; if ! (( $(($awscli_missing + $boto3_missing)) == 0 )); then pip install awscli boto3; fi`

saikonen:
> While this bash line is a little lengthy, it would get around doing any pip install operations if `boto3` and `awscli` are present
> 
> `pip show awscli &> /dev/null; x1=$?; pip show boto3 &> /dev/null; x2=$?; if ! (( $(($x1 + $x2)) == 0 )); then pip install awscli boto3; fi`

To my understanding the regular `pip install` behavior would match that of the bash script, as trying to install a package without version pinning should result in a no-op.

This approach would still break with the initial example, as in it boto3 is not present, so the pip install would kick off, resulting in botocore being upgraded, breaking the existing version of aiobotocore.

savingoyal:
postponing this PR for now since #1490 has been closed.

PR #1505, opened 2023-08-23
@pypi and faster @conda

savingoyal:
To try it out
```
python linear_flow.py --environment=conda run --with kubernetes
```

PR #1506, opened 2023-08-23
vend packaging

savingoyal:
#1505 LOC is bloated because of a vendored package. Pulling it out as a separate PR since it seems `vendor.py` might need some love.


saikonen:
`zipp` has had a breaking change as they've dropped support for Python versions that are EOL, introducing unsupported syntax. As this is a dependency for importlib-metadata, we need to start pinning a specific version for the time being (TODO: figure out what is the latest supported version).

typing-extensions is also required for importlib-metadata for python3.6, which was not being bundled unless running the vendoring script with py36 specifically. Had to pin the version for this as well in order to install a compatible one (TODO: figure out latest supported version).

saikonen (approved):

romain-intel (approved):
Tests are fine on Netflix's side.

jixianyihao:
why not just use specific package as dependency as plugin， all source to vender will cause souce code more large

PR #1508, opened 2023-08-23
add security policy

saikonen:
- adding a security policy to detail how to safely disclose vulnerabilities.

Discussion:
- Supported versions, are all reports of interest, or should we already have a cut off at some version number?
- As an alternative method we do have the option of reaching out to developers through Slack, but I would rather have us suggest one preferred method for reporting for simplicity.


romain-intel (commented):
Apart from the nit, it looks fine. Not sure about the supported version thing.

PR #1510, opened 2023-08-27
Enable Elastic Fabric Adapter Support

rileyhun:
# Overview
We require the ability to activate elastic fabric adapter for low latency, high throughput inter-node connectivity in HPC Clusters. 

This PR accomplishes the following: 
- When using the `@parallel` decorator to launch a gang-scheduled multi-node cluster, allows the container to invoke Elastic Fabric Adapter network devices attached to the EC2 instance

### What has changed
- Added field called `efa` in `@batch` decorator that receives numerical input value specifying number of efa network devices to attach

### Testing Evidence
Screenshot 2023-09-14 at 11 50 05 AM

## Testing evidence

### DIFF
--- a/metaflow/plugins/aws/batch/batch_client.py
+++ b/metaflow/plugins/aws/batch/batch_client.py
@@ -320,14 +320,23 @@ class BatchJob(object):
                 )
             else:
                 job_definition["containerProperties"]["linuxParameters"]["devices"] = []
-                for i in range(int(efa)):
+                if (num_parallel or 0) > 1:
+                    for i in range(int(efa)):
+                        job_definition["containerProperties"]["linuxParameters"][
+                            "devices"
+                        ].append(
+                            {
+                                "hostPath": "/dev/infiniband/uverbs{}".format(i),
+                                "containerPath": "/dev/infiniband/uverbs{}".format(i),
+                                "permissions": ["READ", "WRITE", "MKNOD"]
+                            }
+                        )
+                else:
                     job_definition["containerProperties"]["linuxParameters"][
                         "devices"
                     ].append(
                         {
-                            "hostPath": "/dev/infiniband/uverbs{}".format(i),
-                            "containerPath": "/dev/infiniband/uverbs{}".format(i),
-                            "permissions": ["READ", "WRITE", "MKNOD"]
+                            "hostPath": "/dev/infiniband/uverbs0"
                         }
                     )
 
--- a/metaflow/plugins/aws/batch/batch_client.py
+++ b/metaflow/plugins/aws/batch/batch_client.py
@@ -321,6 +321,7 @@ class BatchJob(object):
             else:
                 job_definition["containerProperties"]["linuxParameters"]["devices"] = []
                 if (num_parallel or 0) > 1:
+                    # Multi-node parallel jobs require the container path and permissions explicitly specified in Job definition
                     for i in range(int(efa)):
                         job_definition["containerProperties"]["linuxParameters"][
                             "devices"
@@ -332,6 +333,7 @@ class BatchJob(object):
                             }
                         )
                 else:
+                    # Single-node container jobs only require host path in job definition
                     job_definition["containerProperties"]["linuxParameters"][
                         "devices"
                     ].append(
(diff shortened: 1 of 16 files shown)


metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -310,6 +311,24 @@ def _register_job_definition(
...
+            if not (isinstance(efa, (int, unicode, basestring))):
+                raise BatchJobException(
+                    "Invalid efa value: ({}) (should be 0 or greater)".format(
+                        efa
+                    )
+                )
+            else:
+                job_definition["containerProperties"]["linuxParameters"]["devices"] = []
+                for i in range(int(efa)):
+                    job_definition["containerProperties"]["linuxParameters"]["devices"] = [
```
emattia, line 324: This job definition seems to follow the single node case in these [docs](https://docs.aws.amazon.com/batch/latest/userguide/efa.html). 

Please correct if I'm misunderstanding, but to function as expected with `@parallel` shouldn't it need a conditional here?
- If single node leave same `job_definition` path and only include the `hostPath` key.
- If multi-node the same vars are set in `job_definition["nodeProperties"]["nodeRangeProperties"]["container"]["linuxParameters"]["devices"]`.

rileyhun: @emattia - I may be interpreting this the wrong way, but wouldn't `@parallel` inherit the `job_definition["containerProperties"]` set by the `efa` conditional?

So the user will want to use `@parallel` in conjunction with `efa=1` in the `@batch` decorator

https://github.com/rileyhun/metaflow/blob/cf0cc2f2f66f83a6743a052299a2430b491d695d/metaflow/plugins/aws/batch/batch_client.py#L344C23-L344C23

emattia: I agree with this end goal:
> So the user will want to use `@parallel` in conjunction with efa=N in the `@batch` decorator

Could you elaborate on what you mean by "`@parallel` inherit the `job_definition["containerProperties"]`"?  

The `@parallel` with `@batch` triggers AWS multinode batch, which ultimately is one job submission through boto3. My comment is because [EFA docs](https://docs.aws.amazon.com/batch/latest/userguide/efa.html) say:

> The location of the [LinuxParameters](https://docs.aws.amazon.com/batch/latest/APIReference/API_LinuxParameters.html) members are different for multi-node parallel jobs and single-node container jobs.

rileyhun: I actually missed an `append`, so just made a small fix. Took a look at the job definition, and this looks correct to me
Screenshot 2023-09-15 at 10 39 39 AM

I think I understand what you're trying to say. I basically need to account for single node cases. 

I was confused and more just pointing out that `@parallel` does most of the correct formatting for the job definition already so the job_definition for efa will be embedded w/in

emattia: Yep, looks that @parallel is doing the thing I refer to itself, as you point out. Nice!

rileyhun: I think it would make sense to disregard the single node use case? If this is meant to be used with `@parallel`, can we assume number of nodes > 1?

metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -310,6 +311,35 @@ def _register_job_definition(
...
 
+        if efa:
+            if not (isinstance(efa, (int, unicode, basestring))):
+                raise BatchJobException(
+                    "Invalid efa value: ({}) (should be 0 or greater)".format(
+                        efa
+                    )
+                )
+            else:
+                job_definition["containerProperties"]["linuxParameters"]["devices"] = []
```
savingoyal, line 322: @rileyhun @emattia can you add a comment here on why there is a different treatment for num_parallel for future readers.

savingoyal (approved):

savingoyal:
@rileyhun can you run black on your PR?

savingoyal:
@rileyhun `black` is still failing the pre-commit check

emattia (approved):

savingoyal:
#1584 supersedes this PR

PR #1511, opened 2023-08-29
Fix an issue where 0 was not considered False for extension debug opt…

romain-intel:
…ions

### DIFF
--- a/metaflow/metaflow_config.py
+++ b/metaflow/metaflow_config.py
@@ -467,7 +467,7 @@ try:
                 DEBUG_OPTIONS.extend(o)
                 for typ in o:
                     vars()["DEBUG_%s" % typ.upper()] = from_conf(
-                        "DEBUG_%s" % typ.upper()
+                        "DEBUG_%s" % typ.upper(), False
                     )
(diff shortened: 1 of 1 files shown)


saikonen (approved):

metaflow/metaflow_config.py
```diff
@@ -467,7 +467,7 @@ def get_pinned_conda_libs(python_version, datastore_type):
                 DEBUG_OPTIONS.extend(o)
                 for typ in o:
                     vars()["DEBUG_%s" % typ.upper()] = from_conf(
-                        "DEBUG_%s" % typ.upper()
+                        "DEBUG_%s" % typ.upper(), False
```
savingoyal, line 470: can this be just `bool(from_conf("DEBUG_%s" % typ.upper()))`?

savingoyal (approved):

metaflow/metaflow_config.py
romain-intel: No -- there is special handling in the from_conf function that handles a boolean default value (ie: it does the conversion from "0", "false" to False so it needs to be a boolean default.

PR #1520, opened 2023-09-04
fix: GPU scheduling issue

saikonen:
Kubernetes jobs are getting preferred for scheduling on gpu nodes in cases where kubernetes or resources decorator sets `gpu=0` when this should not be the case.

Changes so that gpu vendor is not added to limits if the value is None or zero.


saikonen:
alternative approach would've been to check for the zero value in `kubernetes_job.py` and `argo_workflows.py`, but rewriting in the kube decorator seemed to be simpler.

shrinandj (approved):

PR #1522, opened 2023-09-05
return the AWS error message rather than just the URL closes #545

### LINKED ISSUE #545: Helper functions / verbose debugging to triage IAM (S3) issues

rsmith013:


savingoyal:
@rsmith013 can you help me with the before and after for the error message

PR #1526, opened 2023-09-07
chore: fix micromamba detection

saikonen:
fixes micromamba binary detection and adds a download step for the platform if micromamba is not installed.

### DIFF
--- a/metaflow/plugins/pypi/micromamba.py
+++ b/metaflow/plugins/pypi/micromamba.py
@@ -246,6 +246,12 @@ def _install_micromamba():
         return
 
     print("Installing micromamba.")
-    download = subprocess.Popen(download_cmds, stdout=subprocess.PIPE)
-    subprocess.check_output(extract_cmds, stdin=download.stdout, stderr=subprocess.PIPE)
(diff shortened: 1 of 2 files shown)


metaflow/plugins/pypi/micromamba.py
```diff
@@ -26,10 +27,15 @@ def __init__(self):
...
-        # TODO: Fix micromamba deteection before shipping
         self.bin = which(
-            os.environ.get("METAFLOW_PATH_TO_MICROMAMBA") or "./micromamba"
-        )
+            os.environ.get("METAFLOW_PATH_TO_MICROMAMBA") or "micromamba"
+        ) or which("./bin/micromamba")
+        if self.bin is None:
+            # Try installing micromamba
+            print("Could not find micromamba installation.")
+            _install_micromamba()
```
saikonen, line 36: one big question here is do we need to install micromamba, or can we simply provide users an error and some direction on how to install it?

Thinking that downloading binaries as part of a setup process might not jive well with some organisation policies, even though it is gated by the `which` checks.

PR #1530, opened 2023-09-12
fix: vendor older version of packaging for python 3.5

saikonen:
vendored version of `packaging` does not work for python 3.5 as it contains types. Vendored an older version for the python 3.5 use case

### DIFF
--- a/metaflow/_vendor/v3_5/packaging.LICENSE
+++ b/metaflow/_vendor/v3_5/packaging.LICENSE
@@ -1,0 +1,23 @@
+Copyright (c) Donald Stufft and individual contributors.
+All rights reserved.
+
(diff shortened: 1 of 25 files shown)


saikonen:
quite an enormous diff considering the very narrow use of the library in the upcoming pip decorator. Should we instead consider hoisting the implementation for the following, if these are all we need in the future?

```python
from metaflow._vendor.packaging.tags import (
    _cpython_abis,
    compatible_tags,
    cpython_tags,
    mac_platforms,
)
```

savingoyal (approved):

savingoyal:
unfortunately getting these methods and the transitive methods cleanly out of packaging is not a straightforward procedure

PR #1532, opened 2023-09-12
chore: support disable for new conda deco

saikonen:
add support for `disabled=` for the new conda flow and step decorators. Also support `disabled=` for PyPI decorator.


metaflow/plugins/pypi/conda_environment.py
```diff
@@ -68,14 +68,16 @@ def environments(type_):
...
+                    continue
                 if type_ in environment and environment["id_"] not in seen:
                     seen.add(environment["id_"])
                     for platform in environment[type_]["platforms"]:
                         yield environment["id_"], {
                             **{
                                 k: v
                                 for k, v in environment[type_].items()
-                                if k != "platforms"
+                                if k in ["python", "packages"]
```
saikonen, line 80: can rethink this change, but was necessary for now as not being explicit about the attributes here meant that we end up passing unsupported ones to the solver, in this case `disabled`

metaflow/plugins/pypi/conda_environment.py
```diff
@@ -68,14 +68,19 @@ def environments(type_):
...
                     for platform in environment[type_]["platforms"]:
+                        # Only yield parameters that solve() can handle.
+                        # This is because the environment can contain items that the solver does not care about,
+                        # such as default values (e.g. disabled=None)
                         yield environment["id_"], {
                             **{
                                 k: v
                                 for k, v in environment[type_].items()
-                                if k != "platforms"
+                                if k in ["python", "packages"]
```
saikonen, line 83: went with the simplest way for restricting passing non-supported parameters to the solvers.

savingoyal:
fixed and merged to #1505

PR #1533, opened 2023-09-13
Fix the merging of borked lines with MFLOG

romain-intel:
In case there are borked lines in a log file, the current code would not properly merge the logs because borked lines returned a datetime string as a str and not as bytes like non-borked lines.

### DIFF
--- a/metaflow/mflog/mflog.py
+++ b/metaflow/mflog/mflog.py
@@ -137,7 +137,7 @@ def merge_logs(logs):
             res = MFLogline(
                 False, None, MISSING_TIMESTAMP_STR, None, None, line, MISSING_TIMESTAMP
             )
-            yield res.utc_tstamp_str, res
+            yield res.utc_tstamp_str.encode("utf-8"), res
 
     # note that sorted() below should be a very cheap, often a O(n) operation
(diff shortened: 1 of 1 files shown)


metaflow/mflog/mflog.py
```diff
@@ -137,7 +137,7 @@ def line_iter(logblob):
             res = MFLogline(
                 False, None, MISSING_TIMESTAMP_STR, None, None, line, MISSING_TIMESTAMP
             )
-            yield res.utc_tstamp_str, res
+            yield res.utc_tstamp_str.encode("utf-8"), res
```
saikonen, line 140: just some thoughts on the use of the timestamps, am I reading it correctly that the utc_tstamp_str is only used internally for sorting the log lines? I had a slight concern if encoding the tstamp_str only for the `yield` would lead to issues, as we're also yielding the `res`, which contains the non-bytestring version of the timestamp. 

This doesn't seem to be used anywhere else in the codebase for now, but it is somewhat of a sharp edge. Would it make sense to encode the tstamp_str already for the Tuple?

Not a blocker for merging :)

saikonen (approved):

saikonen (approved):

PR #1541, opened 2023-09-15
chore: support python version attribute for pip deco

saikonen:
TODO: Decide what we want to support in regards to Python versions. There is an issue with the way the solver for the pip support is implemented, as it relies on the `--dry-run` flag, which was implemented in [v22.2](https://discuss.python.org/t/announcement-pip-22-2-release/17543)

This version of pip is only available starting from Python 3.7, so the solver is broken for versions 3.5 and 3.6


savingoyal:
merged into #1505

Issue #1316, opened 2023-03-24
Optimize parallel_map by properly waiting on child pids


tuulos:
Currently `parallel_map` launches K subprocesses and then waits for the Kth child to finish before launching another subprocess. A more efficient approach would be to use epoll etc to wait for the first subprocess to finish.

PR #1546, opened 2023-09-23
Check for the first finished parallel proccess

maxzheng:
This fixes https://github.com/Netflix/metaflow/issues/1316 to optimize the parallel map runs by checking for the first finished process instead of waiting for the first process before checking others.

### DIFF
--- a/metaflow/multicore_utils.py
+++ b/metaflow/multicore_utils.py
@@ -4,6 +4,7 @@ import traceback
 from itertools import islice
 from multiprocessing import cpu_count
 from tempfile import NamedTemporaryFile
+import time
 
 try:
     # Python 2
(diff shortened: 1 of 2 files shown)


romain-intel:
Thanks for the contribution. This LGTM. We'll get it merged soon.

romain-intel (dismissed):

metaflow/multicore_utils.py
```diff
@@ -68,8 +69,17 @@ def parallel_imap_unordered(func, iterable, max_parallel=None, dir=None):
...
-        pid, output_file = pids.pop()
-        if os.waitpid(pid, 0)[1]:
+        for idx, pid_info in enumerate(pids):
+            pid, output_file = pid_info
+            pid, exit_code = os.waitpid(pid, os.WNOHANG)
+            if pid:
+                pids.pop(idx)
+                break
+        else:
+            time.sleep(1)  # Wait a bit before re-checking
```
romain-intel, line 79: After a quick discussion, someone brought up the fact that for very short tasks, this now makes it at least 1 second. Could we either skip the sleep (and just do tight polling) or do something a bit fancier (no sleep initially and then some sleep after a while)?

maxzheng: I did consider that as well but decided not to optimize as it might be premature and I am not sure if those very short tasks (less than 1 sec) would need to use this function?  If you are certain there are real scenarios where all tasks are done less than 1 sec and they would use this multi-process function for that, then I can optimize it.

maxzheng: Reduced sleep to 0.1 secs as that should be small enough for any short tasks while not affecting the CPU as it checks.

PR #1547, opened 2023-09-26
chore: add pip decorator ext support

saikonen:
add extension support for pip deco


savingoyal:
fixed and merged into #1505

PR #1548, opened 2023-09-26
chore: add pypi base decorator

saikonen:
add `pypi_base` decorator.

TODO can be handled in this PR if support for `disabled` and `python` attributes gets merged in first.


savingoyal:
fixed and merged into #1505

PR #1549, opened 2023-09-26
Add parameter info to argo workflow template annotations

rohanrebello:
This PR adds information on which parameters are required, to the annotations section of argo-workflow templates.  This can be used to validate the well-formed-ness of workflow submissions before submitting the workflow to argo-workflows.  Before this PR the only way to validate parameters was to submit the workflow and wait until metaflow runs within the argo-workflow, only to fail at parameter validation.  After this change, we can fail-fast and, in the process, save on compute as well as etcd storage.

Manually tested for the following parameter types both locally and in argo-workflows:
- [x] String
- [x] Int
- [x] Float
- [x] Boolean
- [x] JSON
- [x] IncludeFile
- [x] [Deploy-time parameter evaluation](https://docs.metaflow.org/production/scheduling-metaflow-flows/scheduling-with-argo-workflows#deploy-time-parameters)

### DIFF
--- a/metaflow/parameters.py
+++ b/metaflow/parameters.py
@@ -40,7 +40,7 @@ context_proto = None
 
 
 class JSONTypeClass(click.ParamType):
-    name = __name__ = "JSON"
+    name = "JSON"
 
     def convert(self, value, param, ctx):
         if not isinstance(value, strtype):
--- a/metaflow/parameters.py
+++ b/metaflow/parameters.py
@@ -40,7 +40,7 @@ context_proto = None
 
 
 class JSONTypeClass(click.ParamType):
-    name = "JSON"
+    name = __name__ = "JSON"
 
     def convert(self, value, param, ctx):
         if not isinstance(value, strtype):
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -574,6 +574,11 @@ class ArgoWorkflows(object):
             "metaflow/user": "argo-workflows",
             "metaflow/flow_name": self.flow.name,
         }
+
+        for name, parameter in self.parameters.items():
+            if parameter["is_required"]:
+                annotations[f"metaflow/parameter-required-{name}"] = "true"
+    
         if current.get("project_name"):
             annotations.update(
                 {
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -8,6 +8,7 @@ from collections import defaultdict
 from hashlib import sha1
 
 from metaflow import current, JSONType
+from metaflow.includefile import FilePathClass
 from metaflow.decorators import flow_decorators
 from metaflow.exception import MetaflowException
 from metaflow.metaflow_config import (
@@ -401,7 +402,9 @@ class ArgoWorkflows(object):
                 )
             seen.add(norm)
 
-            if param.kwargs.get("type") == JSONType:
+            if param.kwargs.get("type") == JSONType or isinstance(
+                param.kwargs.get("type"), FilePathClass
+            ):
                 # Special-case this to avoid touching core
                 param_type = str(param.kwargs.get("type").name)
             else:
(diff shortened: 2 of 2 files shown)


metaflow/plugins/argo/argo_workflows.py
```diff
@@ -574,6 +574,11 @@ def _compile_workflow_template(self):
             "metaflow/user": "argo-workflows",
             "metaflow/flow_name": self.flow.name,
         }
+
+        for name, parameter in self.parameters.items():
+            if parameter["is_required"]:
+                annotations[f"metaflow/parameter-required-{name}"] = "true"
```
saikonen, line 580: one thing to note is that there does not seem to be a max length to parameter names, but there is a hard limit on what is allowed in Kubernetes annotations: 63 characters for the 'name' segment (the key without prefix). The size constraint for annotation value is much more lenient, though not as well documented. Something like 
```
annotations["metaflow/required-parameters"] = "paramA,paramB, paramC,..."
```
should presumably fit in the annotations value though?

saikonen (changes requested):

metaflow/plugins/argo/argo_workflows.py
rohanrebello: Thanks for the feedback!  I'm going with your suggestion now - I had to switch strategies anyway because `/` is an allowed character in metaflow parameter names but not in annotations.

metaflow/parameters.py
```diff
@@ -40,7 +40,7 @@
 
 
 class JSONTypeClass(click.ParamType):
-    name = "JSON"
+    name = __name__ = "JSON"
```
rohanrebello, line 43: Needed to supply the type information in a standardized way across all types.  The rest of the types are builtins which already have this field.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -401,6 +401,8 @@ def _process_parameters(self):
                 )
             seen.add(norm)
 
+            param_type = str(param.kwargs.get("type").__name__)
```
rohanrebello, line 404: Print just the name of the type rather than the `<class '...'>` stuff.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -412,16 +414,17 @@ def _process_parameters(self):
...
             # has the side effect of casting the parameter value to string null during
             # execution - which needs to be fixed imminently.
-            if not is_required or value is not None:
-                value = json.dumps(value)
+            if not is_required or default_value is not None:
+                default_value = json.dumps(default_value)
             parameters[param.name] = dict(
                 name=param.name,
-                value=value,
+                value=default_value,
```
rohanrebello, line 426: It would be nice to rename this in the parameter map too but I didn't want to get too ambitious, especially since it's serialized in the argo workflow template as `value` again.

obgibson (dismissed):
Confirmed UI is getting
```
"metaflow/parameters": "{\"alpha\": {\"name\": \"alpha\", \"value\": \"0.01\", \"type\": \"float\", \"description\": \"Learning rate\", \"is_required\": false}, \"Booly\": {\"name\": \"Booly\", \"value\": null, \"type\": \"bool\", \"description\": null, \"is_required\": true}}",

```

saikonen (approved):

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -401,6 +402,14 @@ def _process_parameters(self):
                 )
             seen.add(norm)
 
+            if param.kwargs.get("type") == JSONType or isinstance(
+                param.kwargs.get("type"), FilePathClass
```
rohanrebello, line 406: @savingoyal should I use `isinstance()` for both comparisons?  For some reason I didn't fully understand, only the `FilePathClass` required the use of it and it failed with the `==` comparison.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -574,6 +584,10 @@ def _compile_workflow_template(self):
             "metaflow/user": "argo-workflows",
             "metaflow/flow_name": self.flow.name,
         }
+
+        if self.parameters:
+            annotations.update({"metaflow/parameters": json.dumps(self.parameters)})
```
rohanrebello, line 589: Serialize the parameters dictionary into the `metaflow/parameters` annotation for consumers to deserialize.

PR #1550, opened 2023-09-26
[@card] Realtime cards (1/N)

valayDave:
# Realtime Cards

## PR Stack 
1. Interface changes to `current.card` and machinery needed to render the cards / data-updates from the user code (#1550) [Review-able]
2. Metaflow Card UI Code changes to support real-time updates of components (#1551) [Review-able]
3. Make internal `MetaflowCard` and some `MetaflowCardComponent`s real-time updatable (#1552) [Review-able]
4. Introduce a `card server` command to launch a local card server that will allow viewing cards in "realtime" (#1553) [WIP]

## Metaflow UI Changes PRs

1. [x] Expose API in the Metaflow UI Service [Netflix/metaflow-service#402]
2. [ ] Make Client side changes in the Netflix/metaflow-ui 

## Key Changes in the PR

1. Introduce the `current.card.refresh` interface for rendering / pushing data updates
3. Introduce the `current.card.components` interface to update components 
4. Introduce the `MetaflowCardComponent.update` interface
5. Introduce the `MetaflowCard.render_runtime` and `MetaflowCard.refresh` interfaces to render cards during runtime and handle the logic for data updates
6. Lift card creation logic outside the card decorator so it can be called independently even if multiple decorators are given to the same card. We needed to change this because the `current.card` object is a singleton that is only instantiated once per task and is responsible for catching runtime rendering calls. Since all the card rendering logic was inside the instance of a decorator, the individual instance of a decorator was bound to only a single card. This PR lifts that logic into a separate abstraction called `CardCreator`


metaflow/plugins/cards/card_modules/card.py
```diff
@@ -68,8 +94,45 @@ def render(self, task) -> str:
         """
         return NotImplementedError()
 
+    # FIXME document
+    def render_runtime(self, task, data):
+        return
+
+    # FIXME document
+    def refresh(self, task, data):
```
valayDave, line 102: structure of `data` is given by the return value of `CardComponentManager._get_latest_data`

romain-intel (commented):
Not fully done reviewing. First pass. I'll circle back after I looked at all the PRs to get a better understanding instead of asking stupid questions first. SOme comments though.

metaflow/plugins/cards/card_cli.py
```diff
@@ -500,18 +580,20 @@ def create(
...
+                )
             except:
                 if render_error_card:
                     error_stack_trace = str(UnrenderableCardException(type, options))
                 else:
                     raise UnrenderableCardException(type, options)
         #
 
-    if error_stack_trace is not None:
+    if error_stack_trace is not None and mode != "refresh":
```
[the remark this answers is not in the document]
valayDave: No, It's not a typo. Currently, the card_cli command takes three modes: 
1. `refresh`: Involves passing JSON data used for data updates (JSON shipped to S3)
2. `render_runtime`: Involves rendering the card _during_ runtime (HTML shipped to S3)
3. `render`: Involves rendering the card after runtime has completed execution. (HTML shipped to S3)

If the `mode` is `refresh`, we need to ensure that we don't create error cards, which happens in the line below this line.

metaflow/plugins/cards/card_cli.py
```diff
@@ -375,14 +377,45 @@ def wrapper(*args, **kwargs):
...
+
+    def _call():
+        # compatibility with old render()-method that doesn't accept the data arg
+        new_render = "data" in inspect.getfullargspec(mf_card.render).args
+        if mode == "render":
+            if new_render:
+                return _add_token_html(mf_card.render(task, data))
+            else:
+                return _add_token_html(mf_card.render(task))
+        elif mode == "render_runtime":
```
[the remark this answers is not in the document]
valayDave: Will Fix!

metaflow/plugins/cards/card_cli.py
```diff
@@ -452,11 +517,26 @@ def create(
...
 
+    if card_uuid is None:
+        card_uuid = str(uuid.uuid4()).replace("-", "")
+
     # Components are rendered in a Step and added via `current.card.append` are added here.
     component_arr = []
     if component_file is not None:
         with open(component_file, "r") as f:
             component_arr = json.load(f)
+        # data is passed in as temporary files which can be deleted after use
```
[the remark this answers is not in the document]
valayDave: Will Fix!

metaflow/plugins/cards/card_modules/card.py
```diff
@@ -68,8 +94,45 @@ def render(self, task) -> str:
...
+        return
+
+    # FIXME document
+    def reload_content_token(self, task, data):
```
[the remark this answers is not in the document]
valayDave: Will Fix!

metaflow/plugins/cards/card_modules/card.py
```diff
@@ -68,8 +94,45 @@ def render(self, task) -> str:
...
+        return "content-token"
+
 
 class MetaflowCardComponent(object):
+
+    # Setting REALTIME_UPDATABLE as True will make the card component
+    # updatable via the `current.card.update` method for realtime updates
+    REALTIME_UPDATABLE = False
```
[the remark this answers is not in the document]
valayDave: Actually the entire comment block above it needs to be refactored.

metaflow/plugins/cards/card_modules/card.py
```diff
@@ -68,8 +94,45 @@ def render(self, task) -> str:
...
+
+    _component_id = None
+
+    @property
+    def id(self):
```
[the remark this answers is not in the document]
valayDave: Will Fix!

metaflow/plugins/cards/component_serializer.py
```diff
@@ -16,11 +22,267 @@ def get_card_class(card_type):
...
 
+class ComponentStore:
+    """
+    The `ComponentStore` class handles the in-memory storage of the components for a single card.
+    This class has combination of a array/dictionary like interface to access the components.
+
+    It exposes the `append` /`extend` methods like an array to add components.
+    It also exposes the `__getitem__`/`__setitem__` methods like a dictionary to access the components by thier Ids.
+
+    The reason this has dual behavior is because components cannot be stored entirely as a map/dictionary because
```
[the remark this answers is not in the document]
valayDave: That's great! It will certainly remove all the complexity of managing/updating two objects (a dict and list). We can keep the class as is but make its internal implementation much much simpler!

metaflow/plugins/cards/component_serializer.py
```diff
@@ -16,11 +22,267 @@ def get_card_class(card_type):
     return filtered_cards[0]
 
 
+def _component_is_valid(component):
+    """
+    Validates if the component is of the correct class.
+    """
+    if not issubclass(type(component), MetaflowCardComponent):
```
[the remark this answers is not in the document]
valayDave: Sure! renaming it to `RUNTIME_UPDATABLE`.

metaflow/plugins/cards/card_modules/card.py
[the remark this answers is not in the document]
valayDave: Sure; Renaming it in the MetaflowCard to `RUNTIME_UPDATABLE`

metaflow/plugins/cards/card_modules/card.py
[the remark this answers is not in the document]
valayDave: Replaced it with `component_id`

metaflow/plugins/cards/card_cli.py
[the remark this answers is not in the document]
valayDave: The comment is now changed to what it represents.

metaflow/plugins/cards/card_cli.py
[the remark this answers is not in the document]
valayDave: Done. This function now robustly handles many cases. I have also added a comment that explains what it is doing. I also added more comments in the `card create` function that helps understand the behavior for different `modes` and failure scenarios.

metaflow/plugins/cards/component_serializer.py
```diff
@@ -357,7 +716,50 @@ def extend(self, components):
...
+                    "Components list will not be accessible and `current.card.components` will not work for any call during this runtime execution.",
+                    "To fix this set the `id` argument in all @card when using multiple @card decorators over a single @step and reference `current.card[ID].components`",  # todo : Add Link to documentation
+                    "to update/access the appropriate card component.",
+                ]
+            if not self._warned_once["components"]:
+                self._warning(" ".join(_warning_msg))
+                self._warned_once["components"] = True
+            return
+
+        return self._card_component_store[self._default_editable_card].components
```
valayDave, line 744: @tuulos : Something interesting is happening in this case where we _may_ need to change the behavior. The current behavior of cards is that `current.card.append` will result in warnings when you try to add components to a card that doesn't exist or when `current.card` cannot figure out the default editable card. 

Now that we are supporting runtime updates, we need to have a clear story of how we deal with this scenario: 

```python
# `current.card` cannot disambiguate the default editable card
current.card.append(componentA, id="a") # this will throw a warning. 

current.card.components["a"].update(...) # This will crash since currently we are not returning anything
componentA.update(...) # this will work, but nothing will change on the UI
```

The main point is that if `append` and `extend` only throw warnings, it will also affect how the code behaves when `update` is called. Should we now throw exceptions when users try to `append` to cards that cannot be disambiguated?

metaflow/plugins/cards/component_serializer.py
```diff
@@ -16,11 +23,327 @@ def get_card_class(card_type):
...
+            return
+        self._last_refresh = nu
+
+        # This block of code will render the card in `render_runtime` mode when:
+        # 1. refresh is called with `force=True`
+        # 2. Layout of the components in the card has changed. i.e. The actual elements in the component array have changed.
+        # 3. The last time the card was rendered was more the minimum interval after which they should be rendered.
+        last_rendered_before_minimum_interval = (
+            nu - self._last_refresh
+        ) > RUNTIME_CARD_RENDER_INTERVAL
```
valayDave, line 296: @tuulos : Where should we provide users to set `RUNTIME_CARD_RENDER_INTERVAL` and `RUNTIME_CARD_MIN_REFRESH_INTERVAL`; Should we add it to : 
1. metaflow_config.py ?
2. `@card` parameters (what should be the name ?)
3. `MetaflowCard` (should we create a new attribute ?)

One recommendation : 
- Make `RUNTIME_CARD_RENDER_INTERVAL` configurable via `MetaflowCard` and `@card` decorator since this is card-specific. The key idea is that the developer of a card can provide sensible defaults for this value and Metaflow users can over-ride these values. Since this variable governs how often `card create` calls will happen, the best objects to provide such information are the card itself, and the user setting the @card decorator.  
- Make `RUNTIME_CARD_MIN_REFRESH_INTERVAL` a metaflow_config variable. It governs when how often we hit the logic that makes the decision of calling `render_runtime` or `refresh` and behaves like a rate-limiter . _It would also make sense in the MetaflowCard class._

valayDave (commented):
I addressed @romain-intel 's comments and added a little more code. Here is the summary of all the new changes added to the code : 
1. `card_cli.py`
    - Refactored the `update_card` method to handle many different cases gracefully. 
    - The logic of the function is the same, it just gracefully handles the cases where `render_runtime`/`refresh` are not implemented
    - The main logic responsible rendering error cards now robustly handles cases of timeout and missing implementation of the `render_runtime` method.
    - Added more comments that help with the readability of the code.
2. `card.py`
    - rename `MetaflowCard.IS_RUNTIME_CARD` to `MetaflowCard.RUNTIME_UPDATABLE`
    - rename `MetaflowCardComponent.id` to `MetaflowCardComponent.component_id`
3. `card_creator.py`:
    - fix bug for the edge case were async sidecars handling card creation may not have completed execution. 
4. `card_decorator.py`:
    - Refactor based on the rename of `MetaflowCard.IS_RUNTIME_CARD` to `MetaflowCard.RUNTIME_UPDATABLE`
5. `component_serializer.py`:
    - Refactor based on the rename of `MetaflowCardComponent.id` to `MetaflowCardComponent.component_id`
    - Remove the `FIXME` in the `__setitem__` function to not allow overwriting to card components via `current.card.components[ID] = NewComponent`
    - Add functionality to the `current.card.refresh` method that allows re-rendering cards when the layout of the card has changed. This mean that when new components are added to a card or when components are removed from a card then a re-render should be triggered. Modified the `ComponentStore` to expose a property that tells when the component array was last modified. The `CardComponentManager` uses this property in the `ComponentStore` to determine if a re-render should be triggered.
    - Add checks to ensure that objects passed to `current.card.refresh(data)` are JSON serializable. If these are not JSON serializable then warn the user that they are not JSON serializable. 
6. `exception.py`
    - Introduce a new exception for when users try to over-write components.

metaflow/plugins/cards/component_serializer.py
```diff
@@ -16,11 +23,327 @@ def get_card_class(card_type):
...
+
+    def refresh(self, data=None, force=False):
+        self._latest_user_data = data
+        nu = time.time()
+
+        # We set this so that when the self._last_layout_change is not the same
+        # as `self.components.layout_last_changed_on`, then the component array itself
+        # has been modified. So we should force a re-render of the card.
+        if self._last_layout_change is None:
+            self._last_layout_change = self.components.layout_last_changed_on
```
valayDave, line 282: TODO: This line introduces a bug so remove it!

metaflow/plugins/cards/component_serializer.py
```diff
@@ -16,11 +23,327 @@ def get_card_class(card_type):
...
+        # This block of code will render the card in `render_runtime` mode when:
+        # 1. refresh is called with `force=True`
+        # 2. Layout of the components in the card has changed. i.e. The actual elements in the component array have changed.
+        # 3. The last time the card was rendered was more the minimum interval after which they should be rendered.
+        last_rendered_before_minimum_interval = (
+            nu - self._last_refresh
+        ) > RUNTIME_CARD_RENDER_INTERVAL
+        layout_has_changed = (
+            self._last_layout_change != self.components.layout_last_changed_on
+        )
```
valayDave, line 299: TODO: Fix this bug
```suggestion
        layout_has_changed = (
            self._last_layout_change != self.components.layout_last_changed_on or self._last_layout_change is None
        )
```

metaflow/plugins/cards/component_serializer.py
```diff
@@ -16,11 +23,327 @@ def get_card_class(card_type):
...
+                "Data provided to `refresh` is not JSON serializable. It will be ignored."
+            )
+        else:
+            user_data = self._latest_user_data
+
+        return {
+            "user": user_data,
+            "components": component_dict,
+            "render_seq": seq,
+        }
```
valayDave, line 341:
```suggestion
        return {
            "user": user_data,
            "components": component_dict,
            "render_seq": seq,
            "component_update_ts" : self.components.layout_last_changed_on
        }
```

metaflow/plugins/cards/card_cli.py
```diff
@@ -703,6 +703,13 @@ def create(
                 )
                 rendered_content = rendered_info.data
             except:
+                rendered_info = CardRenderInfo(
+                    mode=mode,
+                    is_implemented=True,
+                    data=None,
+                    timed_out=False,
+                    timeout_stack_trace=None,
+                )
```
valayDave, line 712: Bug discovered because of Tests.

valayDave:
Tests are failing rn because they are timing out `get_card` method called in the testing code; We are calling `get_card`  to verify that cards are getting created during runtime. This happens when we run all tests at the same time and not when we individually run only the card tests in the same gh action runner.

test_runner
```diff
@@ -16,7 +16,15 @@ install_extensions() {
...
-install_deps && install_extensions && run_tests 
+# We run realtime cards tests separately because there these tests validate the asynchronous updates to the
+# information stored in the datastore. So if there are other processes starving resources then these tests will
+# surely fail since a lot of checks have timeouts. 
+run_runtime_card_tests() {
+  CARD_GRAPHS="small-foreach,small-parallel,nested-branches,single-linear-step,simple-foreach"
+  cd test/core && PYTHONPATH=`pwd`/../../ python3 run_tests.py --num-parallel 8 --contexts python3-all-local-cards-realtime --graphs $CARD_GRAPHS && cd ../../
+}
+
+install_deps && install_extensions && run_tests && run_runtime_card_tests
```
valayDave, line 30: These are the changes from the latest commit I pushed. This will ensure that the realtime cards tests run **after** the normal metaflow test suite and are not included as a part of the test suite. The card tests will fail if there are too many processes; When too many processes are running it starves the metaflow's card-update processes of resources. Due to starvation of resources metaflow's card-update processes are unable to ship card/data updates and the test time's out. When ran in the happy case(i.e. no resource starvation) tests test succeed. 

Nothing else has been changed in the latest rebase.

metaflow/plugins/cards/card_modules/card.py
```diff
@@ -49,7 +74,8 @@ def _get_mustache(self):
         except ImportError:
             return None
 
-    def render(self, task) -> str:
+    # FIXME document data
+    def render(self, task, data=None) -> str:
```
romain-intel, line 78: I think generally that we should:
- keep the same `render` API. We do need the `data` portion for runtime cards and could add a `get_data` which would return the latest data passed to `render_runtime`.
- the new methods `render_runtime`, `refresh` and `get_data` for example should be probably wrapped (or rather cards should implement a submethod). The wrapper should check whether the card is runtime render-able.

Basically, we should be able to say a card support runtime rendering in one place `RUNTIME_UPDATABLE` and the rest should come from there and we should catch errors that way (ie: the card says it is updatable but doens't provide a method or the other way around)

metaflow/plugins/cards/card_creator.py
```diff
@@ -0,0 +1,208 @@
...
+            components_file = tempfile.NamedTemporaryFile(
+                "w", suffix=".json", delete=False
+            )
+            json.dump(component_strings, components_file)
+            components_file.seek(0)
+        if data is not None:
+            data_file = tempfile.NamedTemporaryFile("w", suffix=".json", delete=False)
+            json.dump(data, data_file)
+            data_file.seek(0)
+
```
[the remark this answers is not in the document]
romain-intel: I would probably prefer a pattern like
```
with tempfile....:
  # do stuff
```
as that would ensure cleanup even if the subprocess has an issue and never gets to cleanup. Here it feels like there are more cases where files can leak and stick around. We could use a similar mechanism to the one in S3 if you wanted to keep them around for debugging.

metaflow/plugins/cards/card_creator.py
```diff
@@ -0,0 +1,213 @@
...
+        fail = False
+        timeout_args = {}
+        async_timeout = ASYNC_TIMEOUT
+        if timeout is not None:
+            async_timeout = int(timeout) + 10
+            timeout_args = dict(timeout=int(timeout) + 10)
+
+        if wait:
+            try:
+                rep = subprocess.check_output(
```
romain-intel, line 177: I don't know if this can happen but here we could have a async process and another non async one that run in parallel. Is this something we are ok with?

metaflow/plugins/cards/card_decorator.py
```diff
@@ -176,17 +199,20 @@ def task_pre_step(
...
+            card_uuid=self._card_uuid,
+            user_set_card_id=self._user_set_card_id,
+            runtime_card=self._is_runtime_card,
+            decorator_attributes=self.attributes,
+            card_options=self.card_options,
+            logger=self._logger,
+        )
+        if is_task_ok:
+            self.card_creator.create(mode="render", final=True, **create_options)
+            self.card_creator.create(mode="refresh", final=True, **create_options)
```
romain-intel, line 215: WHy do we refresh after a final render?

metaflow/plugins/cards/component_serializer.py
```diff
@@ -1,11 +1,18 @@
...
 from .card_modules.components import UserComponent
+from .exception import ComponentOverwriteNotSupportedException
+from functools import partial
 import uuid
 import json
+import time
 
 _TYPE = type
 
+# TODO move these to config
```
romain-intel, line 12: +1

metaflow/plugins/cards/component_serializer.py
```diff
@@ -1,11 +1,18 @@
 from .card_modules import MetaflowCardComponent
 from .card_modules.basic import ErrorComponent, SectionComponent
 from .card_modules.components import UserComponent
+from .exception import ComponentOverwriteNotSupportedException
+from functools import partial
```
romain-intel, line 5: not used

metaflow/plugins/cards/component_serializer.py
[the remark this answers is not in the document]
romain-intel: This hasn't changed yet right?

metaflow/plugins/cards/component_serializer.py
```diff
@@ -16,11 +23,368 @@ def get_card_class(card_type):
...
+    and exposes methods to add, remove and access the components.
+
+    It also exposes a `refresh` method that will allow refreshing a card with new data
+    for realtime(ish) updates.
+
+    The `CardComponentCollector` class helps manage interaction with individual cards. The `CardComponentManager`
+    class helps manage the components for a single card.
+    The `CardComponentCollector` class uses this class to manage the components for a single card.
+
+    The `CardComponentCollector` exposes convinience methods similar to this class for a default editable card.
```
romain-intel, line 184: nit spelling

metaflow/plugins/cards/component_serializer.py
```diff
@@ -16,11 +23,368 @@ def get_card_class(card_type):
...
+        # 3. The last time the card was rendered was more the minimum interval after which they should be rendered.
+        last_rendered_before_minimum_interval = (
+            nu - self._last_refresh
+        ) > RUNTIME_CARD_RENDER_INTERVAL
+        layout_has_changed = (
+            self._last_layout_change != self.components.layout_last_changed_on
+            or self._last_layout_change is None
+        )
+
+        if force or last_rendered_before_minimum_interval or layout_has_changed:
```
romain-intel, line 297: nit: can proably bypass stuff if force is true :)

metaflow/plugins/cards/component_serializer.py
```diff
@@ -16,11 +23,368 @@ def get_card_class(card_type):
...
+                # `component_update_ts` is the timestamp of the last time the component array was modified.
+                # `component_update_ts` can get used by the `reload_content_token` to make decisions on weather to
+                # reload the card on the UI when component array has changed.
+                "mode": mode,
+            }
+            ```
+        """
+        seq = "final" if final else self._render_seq
+        # Extract all the runtime-updatable components as a dictionary
+        component_dict = {}
```
romain-intel, line 359: I wonder if we need to similarly break down data to components so each component can have their own data section?

metaflow/plugins/cards/card_creator.py
```diff
@@ -0,0 +1,213 @@
...
+        if carduuid in cls.async_card_processes:
+            cls.async_card_processes[carduuid]["proc"].kill()
+            del cls.async_card_processes[carduuid]
+
+
+class CardCreator:
+    def __init__(self, top_level_options):
+        self._top_level_options = top_level_options
+
+    def create(
```
romain-intel, line 50: Given this is user-driven, it is possible that things get out of hand if, for example, the user doesn't call `refresh` for a while -- subprocesses can go crazy and have free rein to wreak havoc.

romain-intel (commented):
i have a few high level concerns and a bunch of small comments. At a high level:
- I think we can clean up the render/render_runtime methods a bit and make it clearer around what refreshable cards need.
- I am not sure if we wouldn't need "section" in the data file given the breaking up of a card into components
- I have some concerns around the launching of subprocesses and the possibility of having run-away ones.

metaflow/plugins/cards/card_decorator.py
valayDave: When we are viewing the card in the UI, the data update calls contain the information about what is the latest reload token; This helps the UI make a decision on weather it needs to re-fetch the cards or not. So during the final render, since we are consciously setting the reload token to `final`, we also want to ship a data update that can help information the UI that new cards need to be fetched again.

metaflow/plugins/cards/card_modules/card.py
```diff
@@ -68,8 +94,43 @@ def render(self, task) -> str:
         """
         return NotImplementedError()
 
+    # FIXME document
```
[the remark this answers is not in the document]
valayDave: cc @tuulos ;

metaflow/plugins/cards/component_serializer.py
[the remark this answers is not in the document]
valayDave: This hasn't changed; I think changing this will result in breaking functionality for anything < py37; Currently cards can work < py37

metaflow/plugins/cards/component_serializer.py
```diff
@@ -16,11 +23,368 @@ def get_card_class(card_type):
...
+
+    ```python
+    current.card["mycardid"].append(component, id="comp123")
+    current.card["mycardid"].extend([component])
+    current.card["mycardid"].refresh(data) # refreshes the card with new data
+    current.card["mycardid"].components["comp123"] # returns the component with id "comp123"
+    current.card["mycardid"].components["comp123"].update()
+    current.card["mycardid"].components.clear() # Wipe all the components
+    del current.card["mycardid"].components["mycomponentid"] # Delete a component
+    current.card["mycardid"].components["mynewcomponent"] = Markdown("## New Component") # Set a new component
```
valayDave, line 209: FIXME: Remove this from the docstring

metaflow/plugins/cards/component_serializer.py
valayDave: can you expand on that? I didn't quite get what meant

metaflow/plugins/cards/component_serializer.py
valayDave: I don't quite catch that. Can you expand on that?

metaflow/plugins/cards/component_serializer.py
```diff
@@ -16,11 +23,327 @@ def get_card_class(card_type):
...
+            # We don't support the replacement of individual components in the card.
+            # Instead we support rewriting the entire component array instead.
+            # So users can run `current.card[ID] = [FirstComponent, SecondComponent]` which will instantiate an entirely
+            # new ComponentStore.
+            # So we should throw an error over here, since it is clearly an operation which is not supported.
+            raise ComponentOverwriteNotSupportedException(
+                key, self._user_set_id, self._card_type
+            )
+        else:
+            self._store_component(value, component_id=key)
```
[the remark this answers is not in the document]
tuulos: makes sense

metaflow/plugins/cards/component_serializer.py
tuulos: the main case we want to handle is card updates hidden deep inside a library, referring to a card that you forgot to add `--with card`. To continue the policy that cards are best-effort and will never crash the user code, we could handle even
```
current.card.components["a"].update(...)
```
leniently and make a non-existing ID return a stub object that just shows a warning. I don't think this is a common case but being lenient sometimes and not other times is confusing too. We could just be always lenient - just show warnings, never crash

metaflow/plugins/cards/component_serializer.py
tuulos: we can easily start simply and add more knobs later. We could add `refresh_interval` to `@card` as a starting point. It shouldn't be too hard, right?

metaflow/plugins/cards/card_modules/card.py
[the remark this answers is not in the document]
tuulos: makes sense - `render` takes a `metaflow.Task` always

metaflow/plugins/cards/card_cli.py
```diff
@@ -375,14 +385,154 @@ def wrapper(*args, **kwargs):
...
+        NamedTuple which will contain:
+            - `mode`: The mode of rendering the card.
+            - `is_implemented`: weather the function was implemented or not.
+            - `data` : output from rendering the card (Can be string/dict)
+            - `timed_out` : weather the function timed out or not.
+            - `timeout_stack_trace` : stack trace of the function if it timed out.
+    """
+
+    def _reload_token():
+        if data is None or "render_seq" not in data:
```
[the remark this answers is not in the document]
valayDave: `card create` can be called outside a metaflow step process too. In those cases, it's important to handle the case of data being None.

metaflow/plugins/cards/card_cli.py
```diff
@@ -143,7 +152,7 @@ def resolve_card(
 
 
 @contextmanager
-def timeout(time):
+def timeout(time, on_timeout_return_callback):
```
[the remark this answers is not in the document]
valayDave: Fair point. Will refactor to something more simpler.

metaflow/plugins/cards/card_cli.py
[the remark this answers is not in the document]
romain-intel: right but line 650 sets data to {}.

metaflow/plugins/cards/card_creator.py
romain-intel: I mean you check for subprocess timeouts in a user-driven manner (ie: there is no separate thing that checks periodically). So if the user does one refresh but then nothing else, the subprocess timeout isn't checked. At least I think that's what I remember from my first review.

metaflow/plugins/cards/component_serializer.py
romain-intel: well, the parts that are refreshable are at the component level (not the card level if I got this right). Yet, the data file is at the card level. It feels like there is a mismatch there but maybe I missed something.

metaflow/plugins/cards/component_serializer.py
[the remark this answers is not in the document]
romain-intel: I think we are fine with new functionality not working for such old versions of python. This will actually work in python 3.6+ so it's really just 3.5. Up to you but it would simplify things a bit :) .

metaflow/plugins/cards/component_serializer.py
romain-intel: I think I menat that the part before this can probably be skipped since we know we are going to force it.

metaflow/plugins/cards/card_creator.py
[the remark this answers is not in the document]
valayDave: More context here. 

many of these commands will run in async mode, so if we use a context manager, then we may delete the file before:
1.  the async process is created 
or 
2. the async process has read the file. 

The card create cli command deletes the file right after reading it.

metaflow/plugins/cards/card_decorator.py
```diff
@@ -10,13 +11,16 @@
...
+from .card_creator import CardCreator
 
 
 # from metaflow import get_metadata
 import re
 
 from .exception import CARD_ID_PATTERN, TYPE_CHECK_REGEX
 
+ASYNC_TIMEOUT = 30
+
```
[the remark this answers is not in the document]
valayDave: This is configurable since it is already coming from the top level.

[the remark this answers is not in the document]
valayDave: Actually uncessary since this is set based on the timeout value set in the decorator

romain-intel (approved):
OK. I looked at the diff you sent me and quickly at this. I think it looks ok so we should be good for an initial merge. I haven't extensively played with it too much but I trust you ahve :).

PR #1551, opened 2023-09-26
[@card] Realtime cards (2/N)

valayDave:
# [card-ui] default card real-time updatable
- cleanup the ui component code and add naive progress-bar
- add Vega chart support in default card
- Remove Chartjs dependency from the code.


metaflow/plugins/cards/card_modules/base.html
```diff
@@ -18,6 +18,7 @@
 <body>
   <div class="card_app"></div>
   <script>
+    var METAFLOW_RELOAD_TOKEN = "[METAFLOW_RELOAD_TOKEN]"
```
seethroughdev, line 21: total nit, but we use let/const these days to be more explicit on if the value is immutable.

metaflow/plugins/cards/ui/src/components/bar-chart.svelte
```diff
@@ -17,11 +17,11 @@
     BarController,
     LinearScale,
     CategoryScale,
-    PointElement
+    PointElement,
   );
 
   export let componentData: types.BarChartComponent;
-  const { config, data, labels } = componentData;
+  $: ({ config, data, labels } = componentData);
```
seethroughdev, line 24: Fun sidenote, svelte just announced they're changing this api for the same confusion we got caught up in here. Glad its fixed for now.

metaflow/plugins/cards/ui/src/components/progress-bar.svelte
```diff
@@ -0,0 +1,14 @@
+<!-- A simple text component to render a paragraph wherever needed.App
```
seethroughdev, line 1: This comment should be updated, totally my fault.

metaflow/plugins/cards/ui/src/store.ts
```diff
@@ -5,11 +5,58 @@ import type { Writable } from "svelte/store";
 export const cardData: Writable<types.CardResponse | undefined> =
   writable(undefined);
 
+/**
+ * This function is used to update the cardData object from the window.
+ * It is used to update the cardData object with
+ * changes that have come from the parent window.
```
seethroughdev, line 11: If you're updating comments, probably helpful here to call out that we're intentionally assigning it to the window object as well.

seethroughdev (approved):
LGTM! Changes are optional and purely cosmetic. Really nice job @valayDave .

romain-intel:
Not reviewing this one :). I'll assume it works :).

romain-intel:
As mentioned -- am fine with this (didn't review). My only comment is somewhat orthogonal (I think I had mentioned before): it would be nice to remove things that are actually not packaged with Metaflow to outside the `metaflow` directory to be a bit cleaner.

romain-intel (dismissed):

PR #1553, opened 2023-09-26
[@card] Realtime cards (4/N)

valayDave:
## [card-server] cli command to expose a card server to view realtime updates

- Modified card datastore to accomodate retrieval of runtime data updates
- Added a card viewer html file
- Created a simple HTTP based card server that will help showcase the realtime cards from querying the server
- Card datastore's read and write path retrieval methods now explicitly are given the suffix they retrieve from. We do this because the suffix determines if we are extracting a card or a data update
- Added a private method in the `Card` (user-interface) to get the data related to a card.

### DIFF
--- a/metaflow/plugins/cards/card_server.py
+++ b/metaflow/plugins/cards/card_server.py
@@ -302,9 +302,9 @@ class CardViewerRoutes(BaseHTTPRequestHandler):
         if card_data is not None:
             self.log_message(
                 "Task Success: %s, Task Finished: %s"
-                % (task_object.successful, task_object.finished)
+                % (task_object.successful, is_complete)
             )
-            if not task_object.successful and task_object.finished:
+            if not task_object.successful and is_complete:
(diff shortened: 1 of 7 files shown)


romain-intel (commented):
LGTM but I'll make another pass

metaflow/plugins/cards/card_server.py
```diff
@@ -0,0 +1,179 @@
...
+        else:
+            self._response(CARD_VIEWER_HTML.encode("utf-8"))
+
+    def get_runinfo(self, suffix):
+        task_card_generator = cards_for_run(
+            self.card_options.flow_datastore,
+            self.card_options.run_object,
+            self.card_options.only_running,
+            max_cards=self.card_options.max_cards,
+        )
```
valayDave, line 102: TODO: Gracefully handle `CardNotPresentException`

romain-intel (approved):
Minor nits.

PR #1565, opened 2023-10-03
Remove vendored libs for Py 3.5 (not needed)

romain-intel:


savingoyal (changes requested):
this PR drops support for Py3.5 from Metaflow and not just `@conda`

romain-intel:
I am not sure why you are saying that? I just reverted the additions of the packages that were added to vendor_v3_5. See here: https://github.com/Netflix/metaflow/blame/d18be77c5f585307a7b31e6027a9605e43dc62a3/metaflow/_vendor/vendor_v3_5.txt

romain-intel:
Ah, now I see. it doesn't find zipp from v3.5

romain-intel:
Not sure what it did before. Let me check.

romain-intel:
OK, there we go. zipp was floating before and when I committed I didn't see the directory because on the floating version, it has more stuff. I re-pinned to zipp 3.6.0 and should be fine now.

savingoyal (approved):

Issue #1544, opened 2023-09-18
Support resource limits in Kubernetes. 


roofurmston:
Kubernetes provides support for both resource [requests & limits](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/). At present, Metaflow supports requests only. It would be good if Metaflow could also provide support for limits, as these can be beneficial in certain situations. 

After discussion on Slack, the general idea would be to have a flag (in the Metaflow config) that can be used to switch between the use of either requests or limits. Slack thread can be found [here](https://outerbounds-community.slack.com/archives/C020U025QJK/p1694515923600079).

elgehelge:
While this might solve a few people's problems, I think the suggested solution is still not sufficient.

Requests are “I need at least X”, limits are “I want this pod to use at most Y”. These should not really be exclusive. When using Kubernetes, often you want to specify both of these. So I am a little worried that the suggested solution, will move us to a worse position for making a good solution in the future, because then we end up having this flag which then needs to be deprecated.

tfurmston:
In my experience people tend to use either requests or limits and not both. 

I'm guessing the solution was proposed as adding functionality to specify both requests & limits would have required far more extensive changes to the code base?

savingoyal:
draft PR #2155

PR #1566, opened 2023-10-03
Issue 1544: Add support for Kubernetes limits

tfurmston:
This PR adds support for specifying limits (as opposed to requests) for workflows on Kubernetes. It addresses issue [1544](https://github.com/Netflix/metaflow/issues/1544).

Notes:
* I've added a single flag in the Metaflow config to cover CPU, RAM & disk space. Not sure if you would prefer different flags for each resource?
* I've tested by performing some manual runs. (I couldn't find any relevant tests, so assume this is the current practice for this part of the code base.)
* There is an open comment on the issue. Not sure how you would like to address that comment?

### DIFF
--- a/metaflow/metaflow_config.py
+++ b/metaflow/metaflow_config.py
@@ -295,6 +295,8 @@ KUBERNETES_IMAGE_PULL_POLICY = from_conf("KUBERNETES_IMAGE_PULL_POLICY", None)
 KUBERNETES_CONTAINER_REGISTRY = from_conf(
     "KUBERNETES_CONTAINER_REGISTRY", DEFAULT_CONTAINER_REGISTRY
 )
+# Specify Kubernetes resource requirements as requests. (If not True, then use limits.)
+KUBERNETES_RESOURCE_REQUESTS = from_conf("KUBERNETES_RESOURCE_REQUESTS", True)
 # Toggle for trying to fetch EC2 instance metadata
 KUBERNETES_FETCH_EC2_METADATA = from_conf("KUBERNETES_FETCH_EC2_METADATA", False)
 
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -299,7 +299,6 @@ class ArgoWorkflows(object):
             "limits": {**x, **y} if KUBERNETES_RESOURCE_REQUESTS else y,
         }
 
-
     def _get_schedule(self):
         schedule = self.flow._flow_decorators.get("schedule")
         if schedule:
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -33,6 +33,7 @@ from metaflow.metaflow_config import (
     KUBERNETES_LABELS,
     KUBERNETES_NAMESPACE,
     KUBERNETES_NODE_SELECTOR,
+    KUBERNETES_RESOURCE_REQUESTS,
     KUBERNETES_SANDBOX_INIT_SCRIPT,
     KUBERNETES_SECRETS,
     S3_ENDPOINT_URL,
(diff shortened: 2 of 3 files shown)


elgehelge:
Too me it seems odd to have _either_ requests or limits, but I am not a Kubernetes expert myself, so the only feedback I can really give is to seek advice in a Cloud Native community.

For my own specific use-case I need a couple of different requests to make sure the job can run while also limiting the memory to avoid all memory being captured which sometimes caused the the node to be killed because it could not free the memory fast enough.

shrinandj:
There is a genuine use-case of having requests and NOT limits. Consider the case where a user doesn't know exactly how much cpu/mem is going to be used for a given job. The user can specify some basic value so as to get started. The pod starts with that value guaranteed but can use more. When the pod completes, users can look at resource utilization metrics and can adjust their decorators/cli args, etc.

The above would still be possible with this PR. But the user will have to additionally factor in the value of KUBERNETES_RESOURCE_REQUESTS option.

There is also the case of actually needing to specify both requests AND limits which will have to be solved at some point.

The approach in https://github.com/Netflix/metaflow/pull/1612 might be better. WDYT?

tfurmston:
Hey @shrinandj , 

Thanks for the review!.

> The above would still be possible with this PR. But the user will have to additionally factor in the value of KUBERNETES_RESOURCE_REQUESTS option.

You mean the user will need to remember to switch this flag?

> There is also the case of actually needing to specify both requests AND limits which will have to be solved at some point.

You mean requests and limits with different values, right?

If there is a need to support the possibility of different requests & limits, then I agree the approach in this PR would be too restrictive. https://github.com/Netflix/metaflow/pull/1612 is definitely more flexible in that respect.

One thing from my perspective is that I would like to be able to provide an easy way to ensure  that all our internal users are specifying limits. For example, this would allow us to start using LimitRangers, which is something we would like to do internally. Expecting users to specify both requests & limits in all the steps of their different workflows themselves wouldn't be the best user experience. 

Perhaps a mix of the two approaches would work? So something like:

* The ability to specify limits via arguments in the Kubernetes decorator.
* Introduction of some new flag. If the flag is set and the limits are not specified in the kubernetes decorator, then the limits are set to match the requests. If the limits are specified in the kubernetes decorator, then this takes precendence.

In this way it would be easy to set limits on all workflows, but also provide flexibility if people want them to be different. 

What do you think?

cstano-attn:
Adding on some thoughts here - I think support for both requests and limits would be incredibly helpful. Right now on K8s, we see about 1/4 of our flows encounter the noisy neighbor problem when pods burst CPU or memory past the request and if they are scheduled on the same node (which in about 25% of the time they are) they exhaust the node's resources and kill the node. 

I would +1 on the need for both limits and requests as separate values that can be passable. Setting them as the same just requires teams to increase everything to account for bursting

elgehelge:
> > There is also the case of actually needing to specify both requests AND limits which will have to be solved at some point.
> 
> You mean requests and limits with different values, right?
> 
> If there is a need to support the possibility of different requests & limits, then I agree the approach in this PR would be too restrictive. #1612 is definitely more flexible in that respect.
> 
> One thing from my perspective is that I would like to be able to provide an easy way to ensure that all our internal users are specifying limits. For example, this would allow us to start using LimitRangers, which is something we would like to do internally. Expecting users to specify both requests & limits in all the steps of their different workflows themselves wouldn't be the best user experience.

I think I was one of the voices arguing for requests AND limits, however, I have become wiser (I think 🤔). As I understand it now, setting limits means that the requests are automatically set to the same on the pods. This was basically what I wanted in the first place.

But @cstano-attn seems to want the values to be able to differ:

> I would +1 on the need for both limits and requests as separate values that can be passable. Setting them as the same just requires teams to increase everything to account for bursting

Could you elaborate on why it is beneficial to have separate values? What is a scenario where this is preferred?

savingoyal:
given we now support `@kubernetes(qos=)`, this PR can be safely closed.

PR #1567, opened 2023-10-04
chore: enable extending conda deps in new implementation

saikonen:
re-enables the support for extending conda dependencies via overriding `get_pinned_conda_libs` in extensions.

Addresses pending review comment from @romain-intel


romain-intel (approved):
LGTM.

PR #1568, opened 2023-10-04
feature: custom annotations revisited

saikonen:
reintroduces the feature first merged in #1442 which had to be reverted in #1516 due to a regression that broke annotations for recording user and project info.

This feature keeps the 'system annotations' outside of the decorator scope, as in many cases the `current` singleton is not initialized to its full extent.

Also reintroduces the ability to set custom Kubernetes labels via the decorator, instead of only the environment variable.


metaflow/plugins/argo/argo_workflows.py
```diff
@@ -290,18 +286,99 @@ def trigger(cls, name, parameters=None):
...
+                "metaflow/production_token": self.production_token,
+                "metaflow/owner": self.username,
+                "metaflow/user": "argo-workflows",
+                "metaflow/flow_name": self.flow.name,
+            }
+        )
+
+        if self.parameters:
+            annotations.update({"metaflow/parameters": json.dumps(self.parameters)})
+
```
saikonen, line 335: Moved the new convenience annotations to this helper as well. One thing to note though is that this applies all of them to the generic `self.kubernetes_annotations`, which is applied generously as a baseline set of annotations (including for Sensors)

Do these annotations make sense for all objects, or should they be kept out of this helper? This simplifies the setup somewhat at least.

tylerpotts: @savingoyal Do you have an opinion on this?

ggutierrez545:
Anything further need to be done here to push this PR along?

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -173,7 +169,8 @@ def __init__(
         self.triggers, self.trigger_options = self._process_triggers()
         self._schedule, self._timezone = self._get_schedule()
 
-        self.kubernetes_labels = self._get_kubernetes_labels()
+        self.shared_kubernetes_labels = self._shared_kubernetes_labels()
```
savingoyal, line 172: nit - rename both functions to `_base_kubernetes_labels` and `_base_kubernetes_annotations`. Also maybe the variables become - `_base_labels` and `_base_annotations` - at the moment both the variables and functions have the same name which can lead to some confusion.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -755,9 +756,9 @@ def _compile_workflow_template(self):
                 # is released, we should be able to support multi-namespace /
                 # multi-cluster scheduling.
                 .namespace(KUBERNETES_NAMESPACE)
-                .label("app.kubernetes.io/name", "metaflow-flow")
-                .label("app.kubernetes.io/part-of", "metaflow")
                 .annotations(annotations)
+                .annotations(self.shared_kubernetes_annotations)
+                .label("app.kubernetes.io/name", "metaflow-flow")
```
savingoyal, line 761: why aren't the base labels applied here?

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -823,10 +828,10 @@ def _compile_workflow_template(self):
                 # Set common pod metadata.
                 .pod_metadata(
                     Metadata()
+                    .labels(self.shared_kubernetes_labels)
                     .label("app.kubernetes.io/name", "metaflow-task")
-                    .label("app.kubernetes.io/part-of", "metaflow")
+                    .annotations(self.shared_kubernetes_annotations)
                     .annotations(annotations)
```
savingoyal, line 834: the order for annotations is the reverse of order of annotations in L795:796

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -1989,27 +2000,25 @@ def _container_templates(self):
...
+
+                jobset.annotations(
+                    {
+                        **resources["annotations"],
+                        **self.shared_kubernetes_annotations,
+                        **annotations,
+                    }
+                )
+                # Set task specific labels
+                jobset.labels(resources["labels"])
```
savingoyal, line 2021: this conflicts with L:1968?

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -2066,13 +2075,16 @@ def _container_templates(self):
...
+                        ObjectMeta()
+                        .annotation("metaflow/step_name", node.name)
                         # Unfortunately, we can't set the task_id since it is generated
                         # inside the pod. However, it can be inferred from the annotation
                         # set by argo-workflows - `workflows.argoproj.io/outputs` - refer
                         # the field 'task-id' in 'parameters'
                         # .annotation("metaflow/task_id", ...)
                         .annotation("metaflow/attempt", retry_count)
+                        .annotations(resources["annotations"])
+                        .labels(resources["labels"])
```
savingoyal, line 2087: any reason not to set base annotations/labels here? it might be helpful if at the top of the file, we can add a comment regarding where annotations/labels apply and where they don't - would be helpful in preventing future regressions.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -2840,44 +2852,16 @@ def _compile_sensor(self):
...
-            annotations.update(
-                {
-                    "metaflow/project_name": current.project_name,
-                    "metaflow/branch_name": current.branch_name,
-                    "metaflow/project_flow_name": current.project_flow_name,
-                }
-            )
-
-        # Useful to paint the UI
-        trigger_annotations = {
```
savingoyal, line 2861: where is this set going forward?

metaflow/plugins/kubernetes/kube_utils.py
```diff
@@ -52,3 +59,50 @@ def qos_requests_and_limits(qos: str, cpu: int, memory: int, storage: int):
     # TODO: Add support for BestEffort once there is a use case for it.
     # BestEffort - no limit or requests for cpu/memory
     return qos_requests, qos_limits
+
+
+def validate_kube_labels_or_annotations(
```
savingoyal, line 64: this only validates labels and not annotations?

savingoyal (dismissed):

metaflow/plugins/argo/argo_workflows.py
saikonen: seems like some unintentional duplicate logic, will remove. L1968 should take care of this by itself already

metaflow/plugins/argo/argo_workflows.py
saikonen: this variable has always been unused. The annotation is set separately on L2921

metaflow/plugins/argo/argo_workflows.py
saikonen: these are already set on the workflow template level so they should be inherited. I'll do a quick verify on this still

saikonen: confirmed that these are inherited from the workflow template to the workflow, and workflow pods as expected.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -823,10 +829,10 @@ def _compile_workflow_template(self):
                 # Set common pod metadata.
                 .pod_metadata(
                     Metadata()
+                    .labels(self._base_labels)
                     .label("app.kubernetes.io/name", "metaflow-task")
-                    .label("app.kubernetes.io/part-of", "metaflow")
                     .annotations(annotations)
-                    .labels(self.kubernetes_labels)
```
dhpollack, line 829: @saikonen is there a reason that you remove these labels from the podMetadata section and instead repeat them in each task?

saikonen: the podMetadata applying to all pods from the template wasn't a good fit for setting task-pod-specific labels, e.g.
```
@kubernetes(labels={"test-label": "first-step"})
@step
def first_step(self)
   ...

@kubernetes(labels={"test-label": "second-step"})
@step
def second_step(self)
   ...
```
as on the template level we can't make a choice which deco to pull the 'common' labels from.

The previous implementation did treat the common labels from the environment variable as shared, and set them in the podMetadata. Functionally there shouldn't be any difference though whether those are set here, or individually per-pod?

saikonen: Is there an issue with repeating common labels on the pod level?

We can extend on this, as there is a todo for the future to provide separate label/annotation global env configuration for argo-workflows which would enable labeling all relevant resources (templates, sensors etc.)

PR #1575, opened 2023-10-06
fix timezone in default card time format string

amerberg:
The default card formats timezones in dataframes with `time_format = "%Y-%m-%dT%H:%M:%SZ"`. This should be `time_format = "%Y-%m-%dT%H:%M:%S%Z"`; without the % before the Z the formatted time will end with the literal Z, which means UTC time.

### DIFF
--- a/metaflow/plugins/cards/card_modules/convert_to_native_type.py
+++ b/metaflow/plugins/cards/card_modules/convert_to_native_type.py
@@ -307,7 +307,7 @@ class TaskToDict:
         # and truncates it to 30 characters.
         # If there is any form of TypeError or ValueError we set the column value to "Unsupported Type"
         # We also set columns which are have null values to "null" strings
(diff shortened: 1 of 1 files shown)


metaflow/plugins/cards/card_modules/convert_to_native_type.py
```diff
@@ -365,7 +365,7 @@ def _parse_pandas_dataframe(self, data_object, truncate=True):
         if truncate:
             data = data_object.head()
         index_column = data.index
-        time_format = "%Y-%m-%dT%H:%M:%SZ"
+        time_format = "%Y-%m-%dT%H:%M:%S%Z"
```
valayDave, line 368: It is safe to remove this since #1574 already handles the index case.

valayDave (dismissed):
LGTM! Will require a rebase if we merge #1574 first. Thanks for doing this!

Issue #1216, opened 2022-12-14
Support DistributedMap within AWS Step Functions for higher throughput


savingoyal:
While at it - move over from deprecated APIs

PR #1576, opened 2023-10-09
Enable`Distributed` mode for Step Function maps

bhiles:
This PR introduces a way to utilize the "Distributed" type of Map (vs "Inline") to enable Step Functions to scale to larger sizes (issue: https://github.com/Netflix/metaflow/issues/1216).

### How it works

* Add in a new flag `--use-distributed-map` when creating a step function (ie `python helloworld.py step-functions create --use-distributed-map`)
* All map steps are then turned to use `Distributed` mode (vs the `Inline`) mode
* Distributed Map values are written to S3 and then we need to run some additional steps (fetch the S3 manifest and run a second Distributed Map step to extract the value)
* Lastly, the Run ID of the initial step function is passed down to maps so that it can be accessed on child Step Functions so data can be found in the S3 bucket. This is achieved by passing it into the Dynamo table.

### Additional Notes

This PR works for my test flows (one that maps to 1k values to test the scale, and another that does a map -> map -> join -> join).

### DIFF
--- a/metaflow/plugins/aws/step_functions/step_functions.py
+++ b/metaflow/plugins/aws/step_functions/step_functions.py
@@ -173,6 +173,7 @@ class StepFunctions(object):
             )
 
         return schedule_deleted, sfn_deleted
+
     @classmethod
     def trigger(cls, name, parameters):
         try:
--- a/metaflow/plugins/aws/step_functions/step_functions.py
+++ b/metaflow/plugins/aws/step_functions/step_functions.py
@@ -308,8 +308,7 @@ class StepFunctions(object):
                     .parameter("Parameters.$", "$.Parameters")
                     .parameter("Index.$", "$$.Map.Item.Value")
                     .parameter("RootRunId.$", "$.Result.Item.root_run_id.S")
-                    #.next(node.matching_join)
-                    .next(f"{iterator_name}_GetManifest")
+                    .next(f"{iterator_name}_GetManifest" if workflow.use_distributed_map else node.matching_join)
                     .iterator(
                         _visit(
                             self.graph[node.out_funcs[0]],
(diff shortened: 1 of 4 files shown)


metaflow/plugins/aws/step_functions/step_functions.py
```diff
@@ -307,19 +310,20 @@ def _visit(node, workflow, exit_node=None):
...
                 workflow.add_state(
-                    Map(iterator_name)
+                    Map(iterator_name, self.use_distributed_map)
                     .items_path("$.Result.Item.for_each_cardinality.NS")
                     .parameter("JobId.$", "$.JobId")
                     .parameter("SplitParentTaskId.$", "$.JobId")
                     .parameter("Parameters.$", "$.Parameters")
                     .parameter("Index.$", "$$.Map.Item.Value")
-                    .next(node.matching_join)
+                    .parameter("RootRunId.$", "$.Result.Item.root_run_id.S")
```
bhiles, line 325: Here we are extracting the root Run ID from dynamo and putting into the parameters as `RootRunId` so it can be used.

metaflow/plugins/aws/step_functions/step_functions.py
```diff
@@ -307,19 +310,20 @@ def _visit(node, workflow, exit_node=None):
...
+                    .next(f"{iterator_name}_GetManifest" if self.use_distributed_map else node.matching_join)
```
bhiles, line 326: If we are using the distributed map, the next step won't be the `node.matching_join`, but some additional steps to get the parameters from one of the distributed map's S3 data. To achieve that we add:

* Write the data to S3 using `ResultWriter`
* Add a `S3.GetObject` step to get the `manifest.json` file which gives us the list of files our results are saved in
* Add a `distributed map` step (with a limit of 1 value) to read just the first entry of the first file
* Add a `pass` step to just transform the data to the expected format to be passed to the join

metaflow/plugins/aws/step_functions/step_functions.py
```diff
@@ -328,8 +332,62 @@ def _visit(node, workflow, exit_node=None):
                         )
                     )
                     .max_concurrency(self.max_workers)
-                    .output_path("$.[0]")
+                    .output_path("$" if self.use_distributed_map else "$.[0]")
```
bhiles, line 335: This allows us to get the S3 result writer destination when using the distributed map mode.

metaflow/plugins/aws/step_functions/step_functions.py
```diff
@@ -328,8 +332,62 @@ def _visit(node, workflow, exit_node=None):
...
                 )
+
+                # If we are using DistributedMap we need to obtain the Parameters
+                # one of the batch jobs. Distributed Map jobs can be so big that
+                # we can't just pull the value from the OutputPath, instead
+                # we need to write the results to S3 and then get the data from S3.
+                # These additional states are how we pull the data from S3.
+                if self.use_distributed_map:
+                    workflow.add_state_hack(f"{iterator_name}_GetManifest", {
+                        "Type": "Task",
```
bhiles, line 345: We potentially can skip this step if we just transform the `$bucket/$prefix/$UUID/manifest.json` to be `$bucket/$prefix/$UUID/SUCCEED_0.json`, but wanted to share this running flow before I optimize more.

metaflow/plugins/aws/step_functions/step_functions.py
```diff
@@ -328,8 +332,62 @@ def _visit(node, workflow, exit_node=None):
...
+                    .output_path("$" if self.use_distributed_map else "$.[0]")
                 )
+
+                # If we are using DistributedMap we need to obtain the Parameters
+                # one of the batch jobs. Distributed Map jobs can be so big that
+                # we can't just pull the value from the OutputPath, instead
+                # we need to write the results to S3 and then get the data from S3.
+                # These additional states are how we pull the data from S3.
+                if self.use_distributed_map:
+                    workflow.add_state_hack(f"{iterator_name}_GetManifest", {
```
bhiles, line 344: `add_state_hack` just allowed me to add states that weren't objects but just dictionaries. This isn't something I'd check in, just wanted to get the flow function and get a cursory review before I optimized that functionality.

metaflow/plugins/aws/step_functions/step_functions.py
```diff
@@ -538,11 +596,13 @@ def _batch(self, node):
                     attrs[
                         "split_parent_task_id_%s.$" % node.split_parents[-1]
                     ] = "$.SplitParentTaskId"
+                    attrs["root_run_id.$"] = "$.RootRunId"
```
bhiles, line 599: I tried to copy the logic for `$.SplitParentTaskId` to pass the `$.RootRunId` through steps.

bhiles:
Closing this in favor of #1683

PR #1577, opened 2023-10-09
feature: add list-templates command for argo workflows

saikonen:
Adds a command for listing Argo workflow templates

`python testflow.py argo-workflows list-templates`
will list all workflow templates that have the same flow name as testflow.py. These will include all projects using the flow, production deployments, non-project deployments etc.

`python testflow.py argo-workflows list-templates --all`
will remove scoping to the flow file from the command and return a complete list of workflow templates

TODO: the API call for listing the objects needs consideration. Presumably the amount of objects even for large organisations shouldn't be extremely massive, as this is fetching workflow templates, and not workflows. Nevertheless some caution here, possibly requiring pagination in the future.

### DIFF
--- a/metaflow/plugins/argo/argo_client.py
+++ b/metaflow/plugins/argo/argo_client.py
@@ -52,6 +52,35 @@ class ArgoClient(object):
                 json.loads(e.body)["message"] if e.body is not None else e.reason
             )
 
+    def get_workflow_templates(self, flow_name=None):
+        """
+        Return a list of workflow template names for a given Flow name.
+        Returns all workflow template names if no name is provided
+        """
+        client = self._client.get()
+        try:
+            results = client.CustomObjectsApi().list_namespaced_custom_object(
+                group=self._group,
+                version=self._version,
+                namespace=self._namespace,
+                plural="workflowtemplates",
+            )["items"]
+            return [
+                result["metadata"]["name"]
+                for result in results
+                if flow_name is None
+                or flow_name
+                == result["metadata"]
+                .get("annotations", {})
+                .get("metaflow/flow_name", None)
+            ]
+        except client.rest.ApiException as e:
+            if e.status == 404:
+                return None
+            raise ArgoClientException(
+                json.loads(e.body)["message"] if e.body is not None else e.reason
+            )
+
     def register_workflow_template(self, name, workflow_template):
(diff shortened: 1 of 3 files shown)


metaflow/plugins/argo/argo_client.py
```diff
@@ -52,6 +52,35 @@ def get_workflow_template(self, name):
...
+                namespace=self._namespace,
+                plural="workflowtemplates",
+            )["items"]
+            return [
+                result["metadata"]["name"]
+                for result in results
+                if flow_name is None
+                or flow_name
+                == result["metadata"]
+                .get("annotations", {})
```
savingoyal, line 74: is it possible to push this filter server-side? if we are only doing client-side filtering, then we can simply have this filtering logic in `argo_workflows.py`

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -833,6 +833,23 @@ def _token_instructions(flow_name, prev_user):
         obj.echo("\nRun terminated.")
 
 
+@argo_workflows.command(help="Terminate flow execution on Argo Workflows.")
```
savingoyal, line 836: The help string needs fixing

metaflow/plugins/argo/argo_client.py
saikonen: unfortunately not possible to push this server-side, as annotations are not supported for `field_selectors` for argo.

I also wanted to do the trimming of the return in the client as early as possible, as the API Response contains the full workflow templates, which can be quite hefty. Can move the filtering to `argo_workflows.py` too if it makes more sense.

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -833,6 +833,23 @@ def _token_instructions(flow_name, prev_user):
...
+    default=False,
+    is_flag=True,
+    type=bool,
+    help="list all Workflow templates (not limited to this Flow)",
+)
+@click.pass_obj
+def list_templates(obj, all=None):
+    templates = ArgoWorkflows.list_templates(obj.flow.name, all)
+    if templates:
+        obj.echo("Argo Workflow Templates:")
```
savingoyal, line 848: We can skip this echo so that the output is easily parseable by awk, grep etc.

savingoyal (approved):

PR #1578, opened 2023-10-09
chore: cleanup filter logic for argo

saikonen:
cleanup for the list-templates filter logic


metaflow/plugins/argo/argo_workflows.py
```diff
@@ -187,13 +187,19 @@ def _sanitize(name):
     def list_templates(flow_name, all=False):
```
savingoyal, line 187: rather than use `flow_name`, we can rely on `self.flow.name` - tad bit more safe

savingoyal (dismissed):
lgtm! a minor comment

savingoyal (approved):

PR #1585, opened 2023-10-12
Remove use of distutils from code (not setup.py yet)

romain-intel:
Also improve performance by removing pkg_resources and replacing with another way of getting the package version.

Finally, remove spurious includes that have no business being there :)

### DIFF
--- a/metaflow/version.py
+++ b/metaflow/version.py
@@ -1,0 +1,1 @@
+../version.py
--- a/metaflow/version.py
+++ b/metaflow/version.py
@@ -1,1 +1,1 @@
-../version.py
+metaflow_version = "2.10.2"
--- a/metaflow/version.py
+++ b/metaflow/version.py
@@ -1,1 +1,1 @@
-metaflow_version = "2.10.2"
+metaflow_version = "2.10.3"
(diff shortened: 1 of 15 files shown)


metaflow/version.py
```diff
@@ -0,0 +1 @@
+metaflow_version = "2.10.2"
```
savingoyal, line 1: good to go - this needs updating

savingoyal (dismissed):
a small comment to correct the metaflow version

savingoyal:
@romain-intel can you introduce tests for 3.12 as part of this PR - https://github.com/Netflix/metaflow/blob/master/.github/workflows/test.yml#L28

setup.py
```diff
@@ -1,6 +1,7 @@
 from setuptools import setup, find_packages
 
-version = "2.10.3"
+with open("metaflow/version.py", mode="r") as f:
```
saikonen, line 3: was there some reason why `version.py` couldn't be imported here instead of resorting to string parsing? This seems to work for me at least, does it have some unwanted side-effects?

```python
from metaflow.version import metaflow_version

version = metaflow_version
```

romain-intel: Am fine with that. I just didn't want to "import" a package being built. I wasn't sure if there would be issues with it but if that works well, we can do that too.

PR #1586, opened 2023-10-12
fix: apply only supported attributes from resources decorator in batch

saikonen:
Changes the way `BatchDecorator` pulls values from `ResourcesDecorator` if any are set. This limits the values to only ones supported through `resource_defaults` in batchdeco.

Should unblock #1500 which currently breaks for batch as it introduces a `disk=` attribute which is not supported for the batch deco, yet gets pulled in by the current implementation.

### DIFF
--- a/metaflow/plugins/aws/aws_utils.py
+++ b/metaflow/plugins/aws/aws_utils.py
@@ -109,6 +109,11 @@ def compute_resource_attributes(decos, compute_deco, resource_defaults):
                 # We use the non None value if there is only one or the larger value
                 # if they are both non None. Note this considers "" to be equivalent to
                 # the value zero.
+                #
+                # Skip attributes that are not supported by the decorator.
+                if k not in resource_defaults:
+                    continue
+
                 if my_val is None and v is None:
                     continue
                 if my_val is not None and v is not None:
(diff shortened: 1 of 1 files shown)


saikonen:
also relates to #1609

metaflow/plugins/aws/aws_utils.py
```diff
@@ -109,6 +109,11 @@ def compute_resource_attributes(decos, compute_deco, resource_defaults):
                 # We use the non None value if there is only one or the larger value
                 # if they are both non None. Note this considers "" to be equivalent to
                 # the value zero.
+                #
+                # Skip attributes that are not supported by the decorator.
+                if k not in [*resource_defaults.keys(), *deco.attributes.keys()]:
```
romain-intel, line 114: nit: this will be recomputed everytime. May be better to make it a set initially and then compare for inclusion. Here for every k, we will recompute the full list.

This also doesn't seem to do what the comment says. My understanding was that it would only let through attributes that are known to, for example, the batch decorator (that's the resource_defaults part) whereas here it lets through anything that is known to the batch_decorator OR the resources decorator which seems to not be exactly what we want (ie: no real different from what it currently is.

Issue #1521, opened 2023-09-04
Flows with extremely long names can not be launched through Argo Workflows UI


saikonen:
The current Metaflow client truncates flow names to the maximum length of 253 characters for `metadata.name` for Kubernetes objects. This leads to an issue when deploying on Argo Workflows, where flows cannot be launched through the UI, as Argo tries to add the workflow template name as a Kubernetes label (`workflows.argoproj.io/workflow-template`) for the workflow object. Max length for labels is only 63.

**Solution:**
We should truncate flow names to 63 characters when deploying on Argo Workflows.

**Issues:**

- Not backward compatible. f.ex. redeploying a flow of a long name with the new limit would leave a dangling flow of the old length behind, possibly executing on a schedule.
- 63 characters can be quite short when using the project decorator, as appending project/user/branch name to the template rapidly increases the overall character count. This could lead to collisions in the names
- one way to reduce the chance of collisions is to append a hash at the end of the workflow name. This in turn reduces the readability of workflow names in the Argo UI

saikonen:
fixed since introducing name-length truncation

PR #1588, opened 2023-10-12
Add a uid for the kubernetes job creation

### LINKED ISSUE #1554: Name collisions on kubernetes jobs cause flow to fail
If there is a random UID created for a kubernetes job that is the same as an existing job, the job will fail:

```
2023-09-26 19:59:06.554 [26708/write_refl_store/232688 (pid 332902)] metaflow.plugins.kubernetes.kubernetes_job.KubernetesJobException: Unable to launch Kubernetes job.
2023-09-26 19:59:06.554 [26708/write_refl_store/232688 (pid 332902)] jobs.batch "t-hcwfq" already exists, the server was not able to generate a unique name for the object
```

tylerpotts:
Fixes #1554 

When running large for-each branches (3000+) Metaflow will sometimes cause job name collisions on kubernetes. This adds a UUID to the prefix which should add sufficient entropy to prevent this from happening.

### DIFF
--- a/metaflow/plugins/kubernetes/kubernetes.py
+++ b/metaflow/plugins/kubernetes/kubernetes.py
@@ -2,11 +2,11 @@ import json
 import math
 import os
 import re
-import secrets
 import shlex
 import time
 from typing import Dict, List, Optional
-
+import uuid
+from uuid import uuid4
 
 from metaflow import current, util
 from metaflow.exception import MetaflowException
@@ -180,7 +180,7 @@ class Kubernetes(object):
         job = (
             KubernetesClient()
             .job(
-                generate_name="t-{entropy}".format(entropy=secrets.token_hex(2)),
+                generate_name="t-{uid}".format(uid=str(uuid4())[:3]),
                 namespace=namespace,
                 service_account=service_account,
                 secrets=secrets,
--- a/metaflow/plugins/kubernetes/kubernetes.py
+++ b/metaflow/plugins/kubernetes/kubernetes.py
@@ -180,7 +180,7 @@ class Kubernetes(object):
         job = (
             KubernetesClient()
             .job(
-                generate_name="t-{uid}".format(uid=str(uuid4())[:3]),
+                generate_name="t-{uid}-".format(uid=str(uuid4())[:8]),
                 namespace=namespace,
                 service_account=service_account,
                 secrets=secrets,
--- a/metaflow/plugins/kubernetes/kubernetes.py
+++ b/metaflow/plugins/kubernetes/kubernetes.py
@@ -2,10 +2,11 @@ import json
 import math
 import os
 import re
+import secrets
 import shlex
 import time
 from typing import Dict, List, Optional
-import uuid
+
 
 from metaflow import current, util
 from metaflow.exception import MetaflowException
@@ -179,7 +180,7 @@ class Kubernetes(object):
         job = (
             KubernetesClient()
             .job(
-                generate_name="t-{uid}".format(uid=str(uuid.uuid4())),
+                generate_name="t-{entropy}".format(entropy=secrets.token_hex(2)),
                 namespace=namespace,
                 service_account=service_account,
                 secrets=secrets,
--- a/metaflow/plugins/kubernetes/kubernetes.py
+++ b/metaflow/plugins/kubernetes/kubernetes.py
@@ -5,6 +5,7 @@ import re
 import shlex
 import time
 from typing import Dict, List, Optional
+import uuid
 
 from metaflow import current, util
 from metaflow.exception import MetaflowException
@@ -178,7 +179,7 @@ class Kubernetes(object):
         job = (
             KubernetesClient()
             .job(
-                generate_name="t-",
+                generate_name="t-{uid}".format(uid=str(uuid.uuid4())),
                 namespace=namespace,
                 service_account=service_account,
                 secrets=secrets,


elgehelge:
This is quite serious. Running _only 25000 jobs within a week_ results in a 95% chance of error due to job name collisions.

We have run into this problem many times already, and it is super hard to debug because the join-step that fails does not tell you why.

Foot notes:

* TTL (time to live) is set to 7 days and this cannot be changed without making an extension.
* Computation of collision chances was computed using the formula from [Wikipedia (equation marked with "(1)")](https://en.wikipedia.org/wiki/Birthday_problem) like so:
```python
n_possible_job_names = 40**5
n_jobs = 25000
1 - np.prod([(n_possible_job_names - j) / n_possible_job_names for j in range(n_jobs)])

Out[]: 0.9527302210394215
```

metaflow/plugins/kubernetes/kubernetes.py
```diff
@@ -178,7 +179,7 @@ def create_job(
         job = (
             KubernetesClient()
             .job(
-                generate_name="t-",
+                generate_name="t-{uid}".format(uid=str(uuid.uuid4())),
```
saikonen, line 182: another approach here would be to rely on the existing id's for namespacing purposes (added benefit of making the pod name more descriptive). e.g.
```python
generate_name="t-{run_id}-{step_name}-{task_id}-".format(run_id=run_id, step_name=step_name, task_id=task_id),
```
This is similar to what is happening with Argo Workflows pod naming.

there's a max length of 63 characters for the name, and with a suffix of 5 characters there might be truncation in some cases. From what I've understood, the truncation only happens for the generatedName part, so we would still get a random suffix.

saikonen: alternatively we can just chop up the first one or two parts of the uuid, as we're still receiving the suffix from the apiserver on top of it.

elgehelge: I like adding this to the name 👍

Potential issue:
1. If the step name is super long, then `truncate("t-{run_id}-{step_name}-{task_id}")` could end up not containing the task id, and thus we would have massive collisions on branching steps. So we would need something like `"t-{truncate(run_id)}-{truncate(step_name)}-{task_id}")` instead.
2. What if you have two or more metadata services running (I would even try to imagine why, but I think we have to assume people will do it), then the run_id, step_name, and task_id might be equal. I know it is probably unlikely, but I still kind of like having the 4 random characters to be on the safe side. 4 should be enough at this point :D

Unimportant side questions: What does the `t-` mean? And why not use f-strings?

savingoyal: to introduce enough entropy - how about simply doing `generate_name="t-{entropy}".format(entropy=secrets.token_hex(2))`. This should be good enough for a vast majority of use cases till kubernetes address the [collision problem](https://github.com/kubernetes/kubernetes/issues/115489) on their end.

saikonen: The `token_hex` seems like a good way forward. Doesn't stretch the id's that much while providing enough entropy.

My earlier note about embedding more info to the pod id is a bit moot; as noted it is subject to truncation, and the relevant bits are recorded in pod annotations in full detail.

romain-intel: The context was slightly different but we did face a similar issue here at Netflix and we solved it by doing something like this: 
- take 18 characters of a sha1 hash (or something like it)
- convert it to a smaller 12 characters using base64 encoding
- this gives you 72 bits of entropy and if you look here: https://kevingal.com/apps/collision.html, that is good enough for a billion tasks (well you have a 1e-6 chance of collision which is good enough for me :) ).

I don't fully understand how long lived the job name is but 16 bits of entropy (2 bytes for token_hex) seems low and you are likely to get a collision for 12000 jobs (one of the use cases brought on slack). The "trick" of using base64 encoding gives you higher entropy for fairly low cost (you just have to pick your other two characters carefully in case of naming conventions but that's not impossible). Not saying you need the full 72 bits but 16 feels low.

saikonen: the configurable part here is just a prefix supplied to the Kubernetes apiserver, it will still generate a 5-character suffix which it appends to it so overall it should be sufficient (`(16**4)*(40**5)` items if my midnight math is correct)

romain-intel: Ah, I had missed the addition of 5 characters. If those characters have a 6 bit entropy (which covers alphanumerical strings using the full 52 characters and 10 digits) that would be 46 bits of entropy that would definitely be better :) (chances of collision under 1% for 1M jobs).

elgehelge: The 5 characters provided by Kubernetes are all lowercase, thus 26 + 10 digits, I believe. So there are 36^5 total possible outcomes? With an additional 16 bits of entropy, I could run 100.000 jobs per day for 7 days and that would give me a 0.06% chance of collision - small enough. However, is 100.000 jobs per day an acceptable scaling limit? If I run 10x that I would get a collision probability of around 1%, which is too high in my point of view.

Personally, at my current company, I would be happy with 16 bits of entropy. But if we want to support more branching-heavy tasks I would vote for a few more bits of entropy.

```python
n_possible_job_names = (36**5) * (2**16) 
n_jobs = 7 * 100000
round(1 - np.prod([(n_possible_job_names - j) / n_possible_job_names for j in range(n_jobs)]), 2)
# = 0.06
```

roofurmston: Perhaps a unix timestamp would be an option?

elgehelge: Timestamps would be good if we could guarantee at most one job per second. But that is not really the case here.

As I see it, either we go with:

**1. Complete randomness**. In which case we should try to be on the safe side regarding entropy, eg.
```python
generate_name="t-{entropy}".format(entropy=secrets.token_hex(4))
```
which will give us around 0.01% chance of collisions with 7M jobs within 7 days TTL. 

**2. Some kind of increment like the task id**. Plus a little entropy to allow for eg. having two metadata services. So could be:
```python
generate_name="t-{task_id}-{entropy}".format(entropy=secrets.token_hex(2), task_id=task_id)
```

roofurmston: Wouldn't the the random 5 character suffix (generated via`generateName`) mean you wouldn't need to to be restricted to one job per second?

elgehelge: Sorry. You are right @roofurmston. Keep forgetting that. Then a UNIX timestamp should not be a problem.

tylerpotts: Since this is a temporary solution until the Kubernetes folks fix this, @savingoyal's simple solution seem sufficient

metaflow/plugins/kubernetes/kubernetes.py
```diff
@@ -178,7 +180,7 @@ def create_job(
         job = (
             KubernetesClient()
             .job(
-                generate_name="t-",
+                generate_name="t-{entropy}".format(entropy=secrets.token_hex(2)),
```
saikonen, line 183: `create_job` receives `secrets` as a parameter, so this won't work. Small change to directly import and use `token_hex` should fix it.

```python
from secrets import token_hex
```

saikonen (changes requested):

metaflow/plugins/kubernetes/kubernetes.py
tylerpotts: This won't work either if Metaflow still plans on supporting python 3.5.x

Running

```
mamba create -p /tmp/test 'python==3.5.5'
mamba activate /tmp/test
python -c "import secrets"
```

Results in module not found

tylerpotts:
Went back to UUID to maintain backwards compatibility with python 3.5. Using 3 characters since UUID doesn't have as large a character set as secrets

elgehelge:
> Went back to UUID to maintain backwards compatibility with python 3.5. Using 3 characters since UUID doesn't have as large a character set as secrets

If I am not mistaken, the character set is the same. Both things are hexadecimal. `secrets.token_hex(2)` returns 4 characters. And the current code using `str(uuid4())[:3]` only return 3 characters. Is that on purpose?

elgehelge:
I really really appreciate that all of you are pushing this forward. High five 🙌
But I don't understand why you don't want to make it scalable? With this change, we get `36**5 * 16**3` possible job names, which means that running only 25.000 jobs per day will give you a chance of collisions of 6% over the course of a week. With 4 characters as originally proposed, it is 0.39% which is somewhat acceptable if we think 25000 per day is our upper limit. But isn't that quite low?

Apologies, if you find me annoying, I am really just trying to help and avoid future Github issues. As I say, I appreciate what you are doing.

roofurmston:
> I really really appreciate that all of you are pushing this forward. High five 🙌 But I don't understand why you don't want to make it scalable? With this change, we get `36**5 * 16**3` possible job names, which means that running only 25.000 jobs per day will give you a chance of collisions of 6% over the course of a week. With 4 characters as originally proposed, it is 0.39% which is somewhat acceptable if we think 25000 per day is our upper limit. But isn't that quite low?
> 
> Apologies, if you find me annoying, I am really just trying to help and avoid future Github issues. As I say, I appreciate what you are doing.

I was wondering the same., to be fair. 

It is still not clear to me why a unix timestamp wouldn't just solve the issue, to be honest,

If you don't want to take that approach though, surely using more characters from the UID would be preferable?

metaflow/plugins/kubernetes/kubernetes.py
```diff
@@ -178,7 +180,7 @@ def create_job(
         job = (
             KubernetesClient()
             .job(
-                generate_name="t-",
+                generate_name="t-{uid}".format(uid=str(uuid4())[:3]),
```
saikonen, line 183: good catch with the python3.5 support. To echo the other comments, three characters won't be sufficient entropy for some cases. Proposing to pick the whole first section of the uuid instead so we're on the safe side:

```python
generate_name="t-{uid}-".format(uid=str(uuid4())[:8]),
```
(note the trailing dash as it's not appended automatically). First 8 chars should be sufficiently random for the purpose based on the V4 implementation, and this will still keep the generated pod name length reasonable.

saikonen (approved):

metaflow/plugins/kubernetes/kubernetes.py
saikonen: Was able to push the change to the branch, assuming no objections to the 8char length as the original proposal was even longer.

I'll go forward with this once tests pass so we don't keep this dangling any longer.

tylerpotts:
@elgehelge Not finding you annoying at all!

Looks like I made 2 missteps when pushing my latest changes.

1. I misread the secrets code and assumed the `2` in `secrets.token_hex(2)` meant it was returning 2 characters
2. Didn't notice that secrets also uses hexadecimal

It was actually my attempt to increase rather than decrease when using 3 characters

@roofurmston 
> It is still not clear to me why a unix timestamp wouldn't just solve the issue, to be honest,

I spoke with @savingoyal yesterday and now understand that changes to the codebase need to be carefully considered in order to maintain the stability of the project, particularly because Metaflow interacts with so many external systems (AWS Batch, Kubernetes, local machines, etc.) An example of a thorny problem when it comes to UID length is an active issue with Argo Workflows: https://github.com/Netflix/metaflow/issues/1521. A timestamp would bump us up to 10 characters, and it would limit our character set to numbers (10 possibilities vs 16 per character.)

Looks like @saikonen bumped the UID up to 8 characters yesterday, so we should be good to go with entropy

PR #1596, opened 2023-10-18
student attempt at implementing termination of step function execution

mikesteroonie:
It seemed like there was an open issue with regards to the lack of implementation of the termination of a step function and I made an attempt to implement it. Would love feedback / help in getting this implemented!! Im still learning so I appreciate any and all feedback :)

### DIFF
--- a/metaflow/plugins/aws/step_functions/step_functions_client.py
+++ b/metaflow/plugins/aws/step_functions/step_functions_client.py
@@ -82,8 +82,18 @@ class StepFunctionsClient(object):
         )
 
     def terminate_execution(self, state_machine_arn, execution_arn):
-        # TODO
-        pass
+        try:
+            # Use the AWS Step Functions client to stop the execution
+            response = self._client.stop_execution(
+                executionArn=execution_arn
+            )
+            return response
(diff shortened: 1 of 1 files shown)


metaflow/plugins/aws/step_functions/step_functions_client.py
```diff
@@ -82,8 +82,18 @@ def list_executions(self, state_machine_arn, states):
         )
 
     def terminate_execution(self, state_machine_arn, execution_arn):
-        # TODO
-        pass
+        try:
```
saikonen, line 85: this still needs a counterpart in the `step_functions_cli.py` so the function can be called through the CLI. You can see how the command has been implemented in `argo_workflows_cli.py` for some example on the plumbing.

Another thing that needs to be implemented for this feature is a `validate_run_id` (again, similar to argo CLI) which checks that the run_id the user has provided belongs to the correct flow/project/user and has a matching token.
This is to guard against being able to issue arbitrary commands such as
```sh
python anyflow.py step-functions terminate not-my-run-id
```

madhur-ob:
Closing since https://github.com/Netflix/metaflow/pull/1695 was merged.

Issue #1591, opened 2023-10-16
Long argo-events sensor names can cause sensor deployments to fail


shrinandj:
Currently, sensor names are created by replacing `.` with `-` in the argo-workflow names.

```
            # Register sensor. Unfortunately, Argo Events Sensor names don't allow for
            # dots (sensors run into an error) which rules out self.name :(
            # Metaflow will overwrite any existing sensor.
            sensor_name = self.name.replace(".", "-")
            if self._sensor:
                argo_client.register_sensor(sensor_name, self._sensor.to_json())
```

Subsequently, argo-events creates a deployment for each sensor. This deployment's name is created by adding a suffix to the sensor name.

If the sensor name is <= 63 characters, things seem to work fine. However, if the sensor name > 63 characters, the deployment object fails to get created:

```
[...]

PR #1597, opened 2023-10-18
Sensor name convention hashed

mikesteroonie:
Made a new implementation for the naming convention of the sensor objects with a hash, to ensure all instances fit within th 63 character limit as issue #1591 outlined that metaflow executed normal operations despite the object not being deployed. 

Label added to sensor object to maintain data integrity of the original workflow name.

### DIFF
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -4,6 +4,7 @@ import os
 import re
 import shlex
 import sys
+import hashlib
 from collections import defaultdict
 from hashlib import sha1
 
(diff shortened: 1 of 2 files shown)


metaflow/plugins/argo/argo_workflows.py
```diff
@@ -320,8 +321,26 @@ def schedule(self):
...
+
+            # -------------------------
+            # ------After the PR-------
+
+            # Hash the name to ensure a short and unique sensor name
+            # The label apparently allows us to add an additional layer of metadata for querying in these kubernetes clusters
+            hashed_name = hashlib.sha256(self.name.encode()).hexdigest()[
+                :10
+            ]  # Get first 10 chars of the hash
+            sensor_name = f"s-{hashed_name}"
```
saikonen, line 337: does argo append the random suffix in all sensor deployments, or only in cases where the length exceeds 63 characters? what we usually aim to do is retain the original name, but if truncation needs to happen, chop it up and append a hash of the value as a suffix.

This way at at least part of the name remains descriptive

saikonen (commented):
All in all great stuff. I would hold off on this for just a bit until #1608 moves forward so we know how to proceed with regards to sensors.

A big question with both the general argo naming convention changes, and this PR is, how to cleanly handle existing deployments so that no resources are left dangling when redeploying with a new naming schema.

PR #1608, opened 2023-10-23
feature: limit argo workflow name length

saikonen:
first draft for fixing argo workflows workflow name maxLength. Previously 253 characters have been allowed, but this causes  launching workflows through Argo UI to fail for names longer than 63 characters, as Argo wants to set the workflow template name as a Kubernetes label for the workflow.

Idea is to introduce a new limit of 63 characters, which checks for existence of a >63char deployment upon name resolving and raises an exception in case one is found. User can circumvent this by applying a flag to use the old name lengths for argo workflow commands:
```
argo-workflows --old-name-length [command]
```

Note: This can in some cases affect GitOps setups and other CI solutions for deploying flows. An easy workaround for those would to apply the `--old-name-length` flag for deployments to retain the original functionality.

attempts to address #1521 and unblock implementing #1416

### DIFF
--- a/metaflow/plugins/argo/argo_workflows_cli.py
+++ b/metaflow/plugins/argo/argo_workflows_cli.py
@@ -66,15 +66,38 @@ def cli():
     help="Argo Workflow name. The flow name is used instead if "
     "this option is not specified.",
 )
+@click.option(
+    "--old-name-length",
+    default=False,
+    is_flag=True,
+    type=bool,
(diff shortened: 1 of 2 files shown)


saikonen:
- [ ] requires some guard against truncating project flows to invalid DNS subdomain names (dot followed by dash)
- [ ] possibly rethink how project flows are truncated (visibility: project part vs. flow name). we could opt for truncating the project part and keeping flow name visible instead

saikonen:
closing this in favour of #2082

PR #1610, opened 2023-10-24
fix: resource decorator step functions issue

### LINKED ISSUE #1609: step-functions fails to deploy with `resources` decorator and integer values
Two examples that break with deploying to step-functions:

```python
@resources
@batch(use_tmpfs=True)
```

```python
@resources(shared_memory=1024)
```

The culprit is a combination of how `compute_resource_attributes` tries casting any numeric value to float, and outputting the max as a string, and step-functions relying on the batch implementation, which expects the correct types from the batch decorator for attributes (namely integers)

No issue exists with `run --with batch` due to the resources decorator attributes going through a json encode/decode before hitting the batch implementation.

saikonen:
fixes the use of `shared_memory` and `use_tmpfs` in combination with a resources decorator for step-functions.

closes #1609

### DIFF
--- a/metaflow/plugins/aws/batch/batch_client.py
+++ b/metaflow/plugins/aws/batch/batch_client.py
@@ -216,7 +216,7 @@ class BatchJob(object):
             if shared_memory is not None:
                 if not (
                     isinstance(shared_memory, (int, unicode, basestring))
-                    and int(shared_memory) > 0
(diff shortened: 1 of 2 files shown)


saikonen:
An alternative fix would be to use `json.loads` on all values returned from `compute_resource_attributes` so that they retain some types.

PR #1622, opened 2023-10-31
[argo] Add k8s labels to tempalte.metadata instead of spec.podMetadata

jhasm:
### Summary
Argo does not propagate the spec.podLabels to the pods. Only the sections under templates from the WorkflowTemplate are propagated to the Workflow and the pods. Their docs are misleading.

Here are related argo issues,
1. https://github.com/argoproj/argo-workflows/issues/7091
2. https://github.com/argoproj/argo-workflows/issues/7013

Particularly, this [comment](https://github.com/argoproj/argo-workflows/issues/7091#issuecomment-978687939) is helpful in demystifying some confusion of the argo terminology

### Testing
I tested against my argo installation. I did not see my custom labels before the change and I saw the labels applied after the change. I used `METAFLOW_KUBERNETES_LABELS` property in the `~/.metaflowconfig/config.json` file.

### DIFF
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -1391,13 +1391,15 @@ class ArgoWorkflows(object):
                     minutes_between_retries=minutes_between_retries,
                 )
                 .metadata(
-                    ObjectMeta().annotation("metaflow/step_name", node.name)
+                    ObjectMeta()
+                    .annotation("metaflow/step_name", node.name)
                     # Unfortunately, we can't set the task_id since it is generated
                     # inside the pod. However, it can be inferred from the annotation
                     # set by argo-workflows - `workflows.argoproj.io/outputs` - refer
                     # the field 'task-id' in 'parameters'
                     # .annotation("metaflow/task_id", ...)
                     .annotation("metaflow/attempt", retry_count)
+                    .labels(self.kubernetes_labels)
                 )
                 # Set emptyDir volume for state management
                 .empty_dir_volume("out")


saikonen:
Took this for a quick spin but I have a few questions just so I understand the situation correctly.
Tried with something along the lines of
```
METAFLOW_KUBERNETES_LABELS="customlabelA=testA,customlabelB=123" python test.py argo-workflows create
python test.py argo-workflows trigger
```

then grabbing the Kubernetes pod name for f.ex. the start task and running `kubectl describe pod -n metaflow-ns start-task-pod-name`.  For my case the pod has both of the custom labels. Is this the case that was not working for you?

Which version of Argo Workflows are you running on? I noticed that there was mention of this behavior being a bug in `v3.1.3`

jhasm:
> Took this for a quick spin but I have a few questions just so I understand the situation correctly. Tried with something along the lines of
> 
> ```
> METAFLOW_KUBERNETES_LABELS="customlabelA=testA,customlabelB=123" python test.py argo-workflows create
> python test.py argo-workflows trigger
> ```
> 
> then grabbing the Kubernetes pod name for f.ex. the start task and running `kubectl describe pod -n metaflow-ns start-task-pod-name`. For my case the pod has both of the custom labels. Is this the case that was not working for you?
> 
> Which version of Argo Workflows are you running on? I noticed that there was mention of this behavior being a bug in `v3.1.3`

Yeah, I did not see the custom labels in my case. Our argo version is `3.0.7`. I know this is kind of old, but adding the labels under metadata makes it work for all argo version.

jhasm:
@saikonen any thoughts?

saikonen:
Going through some old PR's and noticed this was left pending my comment. Extremely sorry about that 😓 

This should have been inadvertently fixed as part of the annotations/labels reworking in version `2.13.1`.

As the original issue was due to a bug with Argo there hasn't been much appetite for introducing workarounds with possible maintenance overhead, where upgrading to a more recent Argo workflows version was also a solution.

PR #1626, opened 2023-11-07
fix: pip config issues with pypi decorator

saikonen:
Adds support for the pip config environment variable `PIP_INDEX_URL`

Reworks the way pip config is treated in order to get around issue https://github.com/pypa/pip/issues/11700 where using `PIP_CONFIG_FILE` does not play well with `pip config get`

### DIFF
--- a/metaflow/plugins/pypi/pip.py
+++ b/metaflow/plugins/pypi/pip.py
@@ -50,8 +50,7 @@ class Pip(object):
                     for tag in pip_tags(python, platform)
                 ]
             )
-            custom_index_url = self.index_url(prefix)
-            extra_index_urls = self.extra_index_urls(prefix)
+            custom_index_url, extra_index_urls = self.indices(prefix)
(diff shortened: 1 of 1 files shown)


metaflow/plugins/pypi/pip.py
```diff
@@ -156,29 +154,30 @@ def metadata(self, id_, packages, python, platform):
...
-            )
+            config = self._call(prefix, args=["config", "list"], isolated=False)
+            for line in config.splitlines():
+                key, value = line.split("=", 1)
+                _, key = key.split(".")
+                value = value.strip("'\"")
+                if key == "index-url":
+                    indices.append(value)
+                elif key == "extra-index-url":
+                    extra_indices.append(value)
```
romain-intel, line 169: Am fixing this a bit so may as well take the fix: https://github.com/Netflix/metaflow-nflx-extensions/pull/29

Issue #1627, opened 2023-11-08
Allow arbitrary AWS tags based on metaflow config or batch decorator


Limess:
We have some cost allocation tags we'd like to add which are not tied to a given user.
As such, we'd like to be able to tag and then allocate AWS batch usage from metaflow correctly based on these tags.

Currently Metaflow supports `METAFLOW_BATCH_EMIT_TAGS` but this only supports `['metaflow.flow_name', 'metaflow.run_id',
 'metaflow.step_name', 'metaflow.version', 'metaflow.run_id.$', 'metaflow.user', 'metaflow.owner', 'metaflow.production_token']` (https://github.com/Netflix/metaflow/pull/631/files).

Ideally it'd be great to be able to do this both at the metaflow config level (so a team can set/forget it) and in the `@batch` decorator, but the latter would be the most useful.

tonyszhang:
This is needed for Autodesk usage as well, as we have many new teams use Metaflow and need a way to manage EC2 tags per user/team via the `.metaflowconfig/config.json` file (or within batch decorator)

mvarma-caris:
This looks like a great feature to have!

anentropic:
+1 my org uses standardised attribution tags across all the cloud infra, would want to be able to automatically tag metaflow resources with the standard org tags for the parent project, not just metaflow's own metadata tags

ryewilson:
+1 from me - same as above, my org tracks cost usage per area based on standardized tags.

danw-mpl:
+1 - I see there's some WIP on this! 🚀

saikonen:
arbitrary tags for batch are now supported with #2578 which should be out with the next release.

PR #1628, opened 2023-11-08
Add capability to add custom tags to batch

Limess:
Fixes https://github.com/Netflix/metaflow/issues/1627.

Adds:
1. `BATCH_DEFAULT_TAGS` to metaflow config - a map of default AWS tags to apply to batch jobs
2. `tags` kwarg to the AWS batch decorator - a map of AWS tags to apply to batch jobs (applied last)
3. `--aws-tags` to the CLI (to avoid clashing with metaflow tags.

All of these only apply when `BATCH_EMIT_TAGS` is set to avoid IAM problems.

I haven't done any manual testing here as I'm not sure the best way to go around that?


saikonen:
closing this as support should now be in with #2578 merged

PR #1632, opened 2023-11-13
[WIP] Add params to connect to zipkin tracing platform

darinyu:
Tested running workflow using the following command

`METAFLOW_ZIPKIN_ENDPOINT="{ENDPOINT_HERE}" METAFLOW_TRACING_URL_TEMPLATE="{END_POINT_HERE}/{trace_id}" python workflow.py run`

Console result looks like the folloinwg:

```
2023-11-13 20:36:10.220 Done! See the run in the UI at http://url/HelloFlow/10180
2023-11-13 20:36:10.220 Tracing URL is https://url/cab34611f479c138c9af2f97575dd56f
```

### DIFF
--- a/metaflow/metaflow_config.py
+++ b/metaflow/metaflow_config.py
@@ -398,9 +398,13 @@ KUBERNETES_SANDBOX_INIT_SCRIPT = from_conf("KUBERNETES_SANDBOX_INIT_SCRIPT")
 
 OTEL_ENDPOINT = from_conf("OTEL_ENDPOINT")
 ZIPKIN_ENDPOINT = from_conf("ZIPKIN_ENDPOINT")
+# URL template for tracing links in the UI. 
+# An example value is "https://zipkin.example.com/trace/{trace_id}".
+TRACING_URL_TEMPLATE = from_conf("TRACING_URL_TEMPLATE")
 CONSOLE_TRACE_ENABLED = from_conf("CONSOLE_TRACE_ENABLED", False)
 # internal env used for preventing the tracing module from loading during Conda bootstrapping.
 DISABLE_TRACING = bool(os.environ.get("DISABLE_TRACING", False))
+DEBUG_TRACING = bool(os.environ.get("DEBUG_TRACING", False))
 
 # MAX_ATTEMPTS is the maximum number of attempts, including the first
 # task, retries, and the final fallback task and its retries.
--- a/metaflow/metaflow_config.py
+++ b/metaflow/metaflow_config.py
@@ -404,7 +404,6 @@ TRACING_URL_TEMPLATE = from_conf("TRACING_URL_TEMPLATE")
 CONSOLE_TRACE_ENABLED = from_conf("CONSOLE_TRACE_ENABLED", False)
 # internal env used for preventing the tracing module from loading during Conda bootstrapping.
 DISABLE_TRACING = bool(os.environ.get("DISABLE_TRACING", False))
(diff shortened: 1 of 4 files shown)


wangchy27:
QQ about the "METAFLOW_TRACING_URL_TEMPLATE": Do we have a universal template for each different endpoint? If so, would it be more reasonable to set this as a part of `tracing/__init__` file?

darinyu:
> QQ about the "METAFLOW_TRACING_URL_TEMPLATE": Do we have a universal template for each different endpoint? If so, would it be more reasonable to set this as a part of `tracing/__init__` file?

We dont have one universal template for different endpoint. I saw that in [Jaegar](https://www.jaegertracing.io/docs/1.50/frontend-ui/) they use `http://localhost:16686/trace/{trace-id}`,  and zipkin use `url/request/{trace_id}` so I decided to wirte a template with `trace_id` to be the variance and let user set up the rest of the template thru config.

metaflow/metaflow_config.py
```diff
@@ -398,9 +398,13 @@
 
 OTEL_ENDPOINT = from_conf("OTEL_ENDPOINT")
 ZIPKIN_ENDPOINT = from_conf("ZIPKIN_ENDPOINT")
+# URL template for tracing links in the UI.
+# An example value is "https://zipkin.example.com/trace/{trace_id}".
```
romain-intel, line 402: nit: I would add that it needs to contain {trace_id} where the trace ID should go.

metaflow/metaflow_config.py
```diff
@@ -398,9 +398,13 @@
...
+TRACING_URL_TEMPLATE = from_conf("TRACING_URL_TEMPLATE")
 CONSOLE_TRACE_ENABLED = from_conf("CONSOLE_TRACE_ENABLED", False)
 # internal env used for preventing the tracing module from loading during Conda bootstrapping.
 DISABLE_TRACING = bool(os.environ.get("DISABLE_TRACING", False))
+DEBUG_TRACING = bool(os.environ.get("DEBUG_TRACING", False))
```
romain-intel, line 407: What is this used for? Is it meant to be used to determine if the URL should print?

metaflow/runtime.py
```diff
@@ -18,7 +18,7 @@
 
 from . import get_namespace
 from .metadata import MetaDatum
-from .metaflow_config import MAX_ATTEMPTS, UI_URL
+from .metaflow_config import MAX_ATTEMPTS, UI_URL, TRACING_URL_TEMPLATE
```
romain-intel, line 21: Not used in this file.

metaflow/runtime.py
```diff
@@ -219,6 +217,13 @@ def execute(self):
                 "Workflow starting (run-id %s):" % self._run_id, system_msg=True
             )
 
+        tracing_url = tracing.get_tracing_url()
+        if tracing_url:
+            self._logger(
```
romain-intel, line 222: we could print only if we are debugging tracing.

metaflow/metaflow_config.py
wangchy27: @savingoyal How does the tracing URL look like in OB's Otel setup? Trying to see if it makes more sense to keep it in OSS config or Netflix internal config

metaflow/runtime.py
darinyu: removed

metaflow/runtime.py
darinyu: decided to move this out of per-task level (instead put this at the same lvl like maestro links).

PR #1643, opened 2023-11-23
fix: pypi decorator handling packages with special characters

### LINKED ISSUE #1642: Issue with pypi decorator and special characters in package name
The current PyPI implementation breaks if a package name or version has special characters in it that require URL encoding due to this leading to a mismatch in the cache path lookup.

Easy reproduction:
publish your favourite Python package to a private repository with a custom version, e.g. `1.0.0.1+devtest` and try to install this with the decorator.

saikonen:
Fixes issues with packages/versions having characteres requiring url encoding.

Closes #1642


Moazzem-Hossain-pixel (commented):
Thanks for contributions.

Moazzem-Hossain-pixel (commented):
Great!

Issue #1648, opened 2023-12-01
Strange inheritance behaviour of `FlowSpec`


sangwoo-joh:
I have a base pipeline class, and two children that inherit this.
And the main calls one of them.

```python
from metaflow import FlowSpec, step, Parameter

class Base(FlowSpec):
    base_param = Parameter("base-param")

class First(Base):
    first_param = Parameter("first")

class Second(Base):
    second_param = Parameter("second")


if __name__ == "__main__":
    Second()
```

When I ran `second run --help`, it strangely shows the `first_param` too!!!

```
python test.py run --help
Metaflow 2.9.11 executing Second for user:sangwoo-joh
Usage: test.py run [OPTIONS]

  Run the workflow locally.

Options:
  --base-param TEXT
  --first TEXT
  --second TEXT
...
```

How can this happen? What should I do?

sangwoo-joh:
Oh.. After I split `First` and `Second` into different file and create flow by importing this, it does show only the necessary parameters.
But I'm curious why this happens...

romain-intel:
I believe it is due to this: https://github.com/Netflix/metaflow/blob/master/metaflow/parameters.py#L38. There is a comment saying we could fix it :).

maxzheng:
@romain-intel Here is a PR to fix this: https://github.com/Netflix/metaflow/pull/1653

saikonen:
should be fixed now.

PR #1653, opened 2023-12-09
Set parameters for running flow instead of any imported

maxzheng:
This fixes the [issue where parameters for multiple imported flows are displayed](https://github.com/Netflix/metaflow/issues/1648) instead of the running one when multple flows are in the same file with parameters.

Also:
1. Change to use Exception as bare except is bad practice / fails flake8.
2. Switch to editable/develop install so any code changes would be effective without reinstalling. Much faster to iterate changes.

Note: I didn't add a test for this as this feature is already tested by many existing tests and they worked fine with this change, but happy to add one specifically for the mentioned issue if desired.

### DIFF
--- a/metaflow/flowspec.py
+++ b/metaflow/flowspec.py
@@ -7,7 +7,7 @@ from itertools import islice
 from types import FunctionType, MethodType
 from typing import Any, Callable, List, Optional, Tuple
 
-from . import cmd_with_io
+from . import cmd_with_io, parameters
 from .parameters import DelayedEvaluationParameter, Parameter
 from .exception import (
     MetaflowException,
@@ -99,6 +99,13 @@ class FlowSpec(object):
 
         self._graph = FlowGraph(self.__class__)
         self._steps = [getattr(self, node.name) for node in self._graph]
+
+        # This must be set before calling cli.main() below (or specifically, add_custom_parameters)
+        parameters.parameters = [
+            getattr(self, a)
+            for a in dir(self)
+            if isinstance(getattr(self, a, None), Parameter)
+        ]
 
         if use_cli:
             # we import cli here to make sure custom parameters in
@@ -197,7 +204,7 @@ class FlowSpec(object):
                 continue
             try:
                 val = getattr(self, var)
-            except:
+            except Exception:
                 continue
             if isinstance(val, Parameter):
                 yield var, val
@@ -559,7 +566,7 @@ class FlowSpec(object):
         for i, dst in enumerate(dsts):
             try:
                 name = dst.__func__.__name__
-            except:
+            except Exception:
                 msg = (
                     "In step *{step}* the {arg}. argument in self.next() is "
                     "not a function. Make sure all arguments in self.next() "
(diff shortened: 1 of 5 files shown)


metaflow/flowspec.py
```diff
@@ -7,7 +7,7 @@
 from types import FunctionType, MethodType
 from typing import Any, Callable, List, Optional, Tuple
 
-from . import cmd_with_io
+from . import cmd_with_io, parameters
 from .parameters import DelayedEvaluationParameter, Parameter
```
madhur-ob, line 11: `from . import cmd_with_io, parameters` --> can we do this instead:
`from .parameters import DelayedEvaluationParameter, Parameter, parameters`

since we already have the snippet for `from .parameters`

metaflow/flowspec.py
```diff
@@ -100,6 +100,13 @@ def __init__(self, use_cli=True):
         self._graph = FlowGraph(self.__class__)
         self._steps = [getattr(self, node.name) for node in self._graph]
 
+        # This must be set before calling cli.main() below (or specifically, add_custom_parameters)
+        parameters.parameters = [
```
madhur-ob, line 104: changing the import above will just require `parameters` instead of `parameters.parameters` now...

metaflow/flowspec.py
```diff
@@ -100,6 +100,13 @@ def __init__(self, use_cli=True):
...
+            getattr(self, a)
+            for a in dir(self)
```
madhur-ob, line 106: There's already a function known as `_get_parameters`, we can probably re-use that instead?

metaflow/flowspec.py
madhur-ob: Actually, the current approach is fine, otherwise we will need to access it as a global variable inside the class..

metaflow/flowspec.py
madhur-ob: thus, this is fine too..

metaflow/flowspec.py
madhur-ob: something like: `parameters.parameters = [p for _, p in self._get_parameters()]`

madhur-ob (approved):
Maybe @savingoyal can comment on changes to `test_runner` and `tox.ini`, LGTM otherwise...

Will prefer re-using `self._get_parameters` too...

setup.cfg
```diff
@@ -1,2 +1,5 @@
 [metadata]
 license_files = LICENSE
+
+[flake8]
+max-line-length = 120
```
saikonen, line 5: seems like an unrelated change to the PR. We're also not using flake8 currently

test_runner
```diff
@@ -5,7 +5,7 @@ set -o errexit -o nounset -o pipefail
 install_deps() {
   for version in 2 3;
   do
-    python$version -m pip install . 
+    python$version -m pip install -e .
```
saikonen, line 8: is it necessary to change to an editable install for this PR?

tox.ini
```diff
@@ -3,4 +3,4 @@
 [testenv]
 allowlist_externals = ./test_runner
 commands = ./test_runner
-
+usedevelop = True
```
saikonen, line 6: required change?

saikonen (changes requested):
some of the changes seem unrelated to the proposed fix. If these are not required, could you please clean them from the PR before we go forward with it?

PR #1654, opened 2023-12-13
chore: use metadata url mapping to install from local cache for PyPI environments

saikonen:
Use the metadata url -> local path mapping for PyPI environments during the environment creation instead of relying on pip http cache to have packages available.

Extracted from #1559 

benchmarked changes with the following change to environment resolving:
```python
# First resolve environments through Conda, before PyPI.
start_ts = time.time()
echo("Bootstrapping virtual environment(s) ...")
...
end_ts = time.time()
echo(f"Bootstraping took: {end_ts-start_ts}")
```
timings for environment resolving (seconds):
|http cache|local paths|
|---|---|
|111.19|115.22|
|119.00|113.97|
|123.63|117.73|

which are within margin to consider that the change should not result in adverse performance.


saikonen:
included in #1681

PR #1655, opened 2023-12-14
fix: Argo workflows joins failing with wide foreach flows 

### LINKED ISSUE #1538: Argo Workflows template exceeds max size with preceding large foreach
The task_id's of the preceding foreach steps for a join appear multiple times in the Argo Workflow init containers `ARGO_TEMPLATE`, which bloats up the size significantly. With wide foreaches the template exceeds the max size, leading to a broken flow.

Possibly a regression bug. Some discussion on the initial report here: https://outerbounds-community.slack.com/archives/C02116BBNTU/p1694529680541219

saikonen:
Partial remedy to the issue with large foreach flows input-paths being repeated in the `ARGO_TEMPLATE` multiple times, exceeding the max size for it.

closes #1538

### DIFF
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -1034,9 +1034,14 @@ class ArgoWorkflows(object):
             # Ideally, we would like these task ids to be the same as node name
             # (modulo retry suffix) on Argo Workflows but that doesn't seem feasible
             # right now.
+
             task_str = node.name + "-{{workflow.creationTimestamp}}"
+            # Note: The input paths might be quite large for foreach joins,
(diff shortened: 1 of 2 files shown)


shrinandj:
This looks good to me!

rohanrebello:
I've tested that this PR works for up to 3000-way foreach's on AWS

tslott:
Can this be merged?

saikonen:
have to close this PR as a dead end due to the direction that Argo Workflows took with the new handling of oversized `ARGO_TEMPLATE` where they offload it to disk as a ConfigMap, making it not parseable in the way I had in mind in this.

This PR being only a partial fix at best lead me to a different approach, one which should be in review now and hopefully a proper solution to the issue at hand: #1704

PR #1656, opened 2023-12-14
fix overwrite of neuron devices when efa devices are also specified

emattia:
When `@batch` contains both `inferentia` and `efa` arguments, the list of `/dev/neuron{N}` (for inferentia and trainium) specifications in `nodeProperties.nodeRangeProperties.container.linuxParameters.devices` are currently being overwritten by the list of  `/dev/infiniband/uverbs{N}` (for efa eni's) devices.


emattia:
NOTE: There is still a bug here. When submitting multinode job with EFA and inferentia it yields error message about node
`botocore.errorfactory.ClientException: An error occurred (ClientException) when calling the SubmitJob operation: NodeOverride targets should match job definition. Non matching targets - 0:0,1:1`

PR #1659, opened 2023-12-19
feature: max cpu and max memory setting in config

saikonen:
Add hard limits for cpu and memory in resource decorator.

### DIFF
--- a/metaflow/metaflow_config.py
+++ b/metaflow/metaflow_config.py
@@ -403,6 +403,8 @@ ZIPKIN_ENDPOINT = from_conf("ZIPKIN_ENDPOINT")
 CONSOLE_TRACE_ENABLED = from_conf("CONSOLE_TRACE_ENABLED", False)
 # internal env used for preventing the tracing module from loading during Conda bootstrapping.
 DISABLE_TRACING = bool(os.environ.get("DISABLE_TRACING", False))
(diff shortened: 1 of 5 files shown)


metaflow/metaflow_config.py
```diff
@@ -403,6 +403,8 @@
 CONSOLE_TRACE_ENABLED = from_conf("CONSOLE_TRACE_ENABLED", False)
 # internal env used for preventing the tracing module from loading during Conda bootstrapping.
 DISABLE_TRACING = bool(os.environ.get("DISABLE_TRACING", False))
+MAX_MEMORY_PER_TASK = from_conf("MAX_MEMORY_PER_TASK")
```
romain-intel, line 406: It would be better to have a default value because that is used to then convert env var values so we could set it to 0.0

metaflow/plugins/aws/aws_utils.py
```diff
@@ -88,7 +89,7 @@ def get_docker_registry(image_uri):
     return registry
 
 
-def compute_resource_attributes(decos, compute_deco, resource_defaults):
+def compute_resource_attributes(decos, compute_deco, step_name, resource_defaults):
```
romain-intel, line 92: I may have to make changes internally to be compatible with this. Not hard though.

romain-intel (commented):
Mostly ok. I'll have to change our code to be compatible but straightforward change.

saikonen:
not required for OSS

PR #1660, opened 2023-12-19
feature: higher resource defaults for k8s

saikonen:
increase resource defaults for Kubernetes


romain-intel (approved):

saikonen:
not required for OSS

PR #1664, opened 2024-01-03
[Ready for Review] Support more convenient foreach variable/value search-ability.

darinyu:
In this change, we

1. add values of foreach items to flowspec for convenient access to values in foreach steps. 
2. Update foreach stack to include value.
3. Add formatted foreach stack (path from root to current level) to metadata, guarded by a flag.

Currently,  we try to get "value" of the for each in the following order:
1. If the item is list of primitive types like `["a", "b", "c"]`, we will use the literal value directly
2. Otherwise, we use replib to get a serialized value of the iterable, limited to 30 charactors.

The end result will be of the form `[value]`.  We will then

1. Use the list of value (generated in a split/parent node) in the `task_finish` function inside a custom decorator.  
2. Use the list of value when constructing foreach stack, which will eventually be used to construct the for-loop path in metadata.

An example for the updated search UI is like the following.
Metaflow-UI (2)

### DIFF
--- a/metaflow/flowspec.py
+++ b/metaflow/flowspec.py
@@ -504,7 +504,7 @@ class FlowSpec(object):
             )
             raise InvalidNextException(msg)
 
-    def _get_foreach_item_value(self, item):
+    def _get_foreach_item_value(self, item: Any):
         """
         Get the unique value for the item in the foreach iterator.  If no suitable value
         is found, return the value formatted by reprlib, which is at most 30 characters long.
--- a/metaflow/flowspec.py
+++ b/metaflow/flowspec.py
@@ -17,6 +17,7 @@ from .exception import (
 )
 from .graph import FlowGraph
 from .unbounded_foreach import UnboundedForeachInput
+from .metaflow_config import INCLUDE_FOREACH_STACK
 
 # For Python 3 compatibility
 try:
@@ -503,27 +504,15 @@ class FlowSpec(object):
             )
             raise InvalidNextException(msg)
 
-    def _get_foreach_item_value(self, item, foreach_var):
+    def _get_foreach_item_value(self, item):
         """
-        Get the unique value for the item in the foreach iterator. We will use
-        the following order of precedence:
-
-        1) If the item is a primitive type, use the item itself.
-        2) If the item is a dictionary and has a key named `foreach_var`, use the value of that key.
-        3) If the item is a dictionary and has a key named `key`, use the value of that key.
-        4) If the item has an attribute named `key`, use the value of that attribute.
-
-        If no suitable value is found, return the value formatted by reprlib, which is at
-        most 30 characters long.
+        Get the unique value for the item in the foreach iterator.  If no suitable value
+        is found, return the value formatted by reprlib, which is at most 30 characters long.
 
         Parameters
         ----------
         item : Any
             The item to get the value from.
-        foreach_var : str
-            The name of the variable that is being iterated over.
-        default_value : str
-            The default value to use if no suitable value is found.
 
         Returns
         -------
@@ -539,23 +528,7 @@ class FlowSpec(object):
                 or isinstance(item, bool)
             )
 
-        value = reprlib.Repr().repr(item)
-        if _is_primitive_type(item):
-            value = item
-        elif (
-            isinstance(item, dict)
-            and foreach_var in item
-            and _is_primitive_type(item[foreach_var])
-        ):
-            value = item[foreach_var]
-        elif (
-            isinstance(item, dict)
-            and "value" in item
-            and _is_primitive_type(item["value"])
-        ):
-            value = item["value"]
-        elif hasattr(item, "value") and _is_primitive_type(item.value):
-            value = item.value
+        value = item if _is_primitive_type(item) else reprlib.Repr().repr(item)
         return basestring(value)
 
     def next(self, *dsts: Callable[..., None], **kwargs) -> None:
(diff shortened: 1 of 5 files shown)


metaflow/flowspec.py
```diff
@@ -499,6 +501,33 @@ def _validate_ubf_step(self, step_name):
             )
             raise InvalidNextException(msg)
 
+    def _get_foreach_item_value(self, item):
```
romain-intel, line 504: nit: Add typing information here as well since you add it in the docs (I know we are not consistent but little by little :) )

metaflow/task.py
```diff
@@ -188,10 +197,10 @@ def lineage():
                     if join_type == "foreach":
                         top = i["_foreach_stack"][-1]
                         bottom = i["_foreach_stack"][:-1]
-                        # the topmost indices in the stack are all
-                        # different naturally, so ignore them in the
+                        # the topmost indices and values in the stack are
+                        # all different naturally, so ignore them in the
                         # assertion
-                        yield bottom + [top._replace(index=0)]
+                        yield bottom + [top._replace(index=0, value=0)]
```
romain-intel, line 203: shouldn't it be value=None?

metaflow/task.py
```diff
@@ -463,6 +471,37 @@ def run_step(
             # 3. initialize foreach state
             self._init_foreach(step_name, join_type, inputs, split_index)
 
+            # Add foreach stack to metadata of the task
+
+            foreach_stack = (
```
romain-intel, line 476: I've been thinking about this and trying to find a way to do this without actually storing a potentially large (and worse, unbounded) string in the metadata. I think we can probably do the following. In the metadata, we can store the split index instead. The `_foreach_stack` is actually available above as well as the `_foreach_values` so we would then be able, in the UI, to reconstruct the string that is being constructed below. This would be a tad be more complicated in the UI (but not that much and it would have the added benefit of being able to use the cache so fewer requests (ie: for each foreach nest, we would only need to load 2 artifacts to have all the paths for all the tasks in the foreach).

Thoughts?

metaflow/task.py
```diff
@@ -188,10 +197,10 @@ def lineage():
                     if join_type == "foreach":
                         top = i["_foreach_stack"][-1]
                         bottom = i["_foreach_stack"][:-1]
-                        # the topmost indices in the stack are all
-                        # different naturally, so ignore them in the
+                        # the topmost indices and values in the stack are
```
romain-intel, line 200: nit: I know this is not in your modification but what assert?

metaflow/runtime.py
```diff
@@ -1027,7 +1027,13 @@ def results(self):
     @property
     def finished_id(self):
         # note: id is not available before the task has finished
-        return (self.step, tuple(self.results["_foreach_stack"]))
+        # index already identifies the task within the foreach,
```
romain-intel, line 1030: why is this needed? Isn't the presence of the value backward compatible with the old code?

metaflow/flowspec.py
darinyu: added

metaflow/runtime.py
darinyu: for a node, we may want to quickly get the finish_id of the sibling node and check if they all finished (can check usage of `finished_id` within `_queue_task_join`). It is non-trivial to get value of the sibling, so we may as well not compare value at all (afterall, the index of the foreach already identifies the node).

metaflow/task.py
darinyu: I think its referring to the `not all_equal(lineage())` below.

metaflow/task.py
darinyu: we dont use index/value in this `lineage()`, so it should be the same giving them whatever value. (keeping it 0 to be consistent with index).

metaflow/task.py
darinyu: Chatted during meeting, This could be achieved but may be overkill. We also want people to be able to more easily access the foreach path, so putting them in metadata also makes sense.

I added a limit to how much data we store in metadata. If we encounter the arbitrary large string problem we can revisit this in the future.

metaflow/task.py
```diff
@@ -24,7 +24,20 @@
...
+    foreach_frame_field_list,
+    defaults=(None,) * len(foreach_frame_field_list),
+)
+ForeachFrame.__reduce__ = lambda frame: (construct_frame, tuple(frame))
+# To be backwards compatible.
+def construct_frame(*args):
+    return ForeachFrame(*args[: len(ForeachFrame._fields)])
+
+
+# Maximum number of charactors of the foreach path that we store in the metadata.
```
romain-intel, line 39: nit: characters.

metaflow/task.py
romain-intel: Yep -- and now that w have a limit of characters this should be fine.

metaflow/task.py
```diff
@@ -463,6 +475,46 @@ def run_step(
...
+            foreach_stack = (
+                self.flow._foreach_stack
+                if hasattr(self.flow, "_foreach_stack") and self.flow._foreach_stack
+                else []
+            )
+
+            foreach_stack_formatted = []
+            current_foreach_path_length = 0
+            for frame in foreach_stack:
+                if not (frame.var and frame.var):
```
romain-intel, line 489: frame.var and frame.value

romain-intel (approved):
LGTM with a small typo. Let's wait for someone from OB but we can merge afterwards.

metaflow/task.py
darinyu: done

metaflow/task.py
darinyu: done

Issue #1058, opened 2022-06-02
run id is assigned even if `boto3` doesn't exist for flow execution


savingoyal:
If `boto3` isn't found (it's a soft dependency), Metaflow throws an error `You need to install 'boto3' in order to use S3.` after assigning a run id. We can move the check before a run id is assigned.

madhur-ob:
This issue is already solved, but the sandbox is not upgraded yet. Below, I am attaching two screenshots.

- **With older metaflow version** (2.9.7.2+ob):

Run IDs 5 and 6 are present...

<img width="1206" alt="Screenshot 2024-01-07 at 4 30 28 AM" src="https://github.com/Netflix/metaflow/assets/155637867/fc678930-7586-4a82-89ac-e8a371770a54">

<img width="1206" alt="Screenshot 2024-01-07 at 4 30 50 AM" src="https://github.com/Netflix/metaflow/assets/155637867/499f0bab-dc67-42b2-9f8a-7ef40d13248a">


- **With newer metaflow version** (2.10.7.4+ob):

Run IDs 5 and 6 are not present...

<img width="589" alt="Screenshot 2024-01-07 at 4 14 56 AM" src="https://github.com/Netflix/metaflow/assets/155637867/2a2b1907-2f12-4b5f-a48c-628e7741b81e">

<img width="1206" alt="Screenshot 2024-01-07 at 4 29 24 AM" src="https://github.com/Netflix/metaflow/assets/155637867/b76b58cd-89d2-405d-a610-0b180b9047a9">

madhur-ob:
for a normal flow that doesn't use `conda`, the issue still persists...

PR #1666, opened 2024-01-07
don't create run ids if dependencies for data store are missing i.e. fail fast

madhur-ob:
Fixes https://github.com/Netflix/metaflow/issues/1058


madhur-ob:
TLDR: the following approach is wrong, which was already present in various places of the codebase:
```
# cannot decorate __init__... invoke it with dummy decoratee
check_gs_deps(lambda: 0)
```

and it turns out that we CAN actually decorate the `__init__` method of a class...
(also confirmed locally via a small snippet and verified via ChatGPT too!)

saikonen (approved):

romain-intel (approved):

PR #1670, opened 2024-01-08
[@card] [card-core] always call `refresh` after first `render_runtime` (5/N)

valayDave:
- Added a `sync` argument to the card creator

## Why make this change 

Lets assume the case that the user is only doing `current.card.append` followed by `refresh`. In this case, there will be no process executed in `refresh` mode since only the `layout_has_changed` will always be true and we will keep running `render_runtime` instead of any `refresh` call. As a result, there will be no data update that informs the UI of the `METAFLOW_RELOAD_TOKEN` change. In the meantime the UI will be trying to seek for the data update object but it will constantly find None. Due to this the card Iframe will not reload and keep being in it's initial state.

This change ensures that after the first `render_runtime`, any time a `render_runtime` is called, we follow it up with a subsequent SYNCHRONOUS `refresh` call. This will ensure that the UI is always updated with the latest card.

### DIFF
--- a/metaflow/plugins/cards/card_cli.py
+++ b/metaflow/plugins/cards/card_cli.py
@@ -965,7 +965,7 @@ def list(
     default=8324,
     show_default=True,
     type=int,
-    help="Port on which Metaflow card server will run",
+    help="Port on which Metaflow card viewer server will run",
 )
 @click.option(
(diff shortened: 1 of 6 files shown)


romain-intel (commented):
LGTM

PR #1671, opened 2024-01-08
[WIP]: validation of stubs with pyright

madhur-ob:


romain-intel:
I think what you are doing is a good idea (to get to a state where we have good validation checks) but I wonder what the balance is between ignoring errors and fixing them. Also -- I believe I figured out how to get pyright to read the stubs (they also need to have a py.typed it seems -- not spec friendly but whatever). We should make sure it reads those.

Quick suggestion on where the line should/could be: things that are user facing (like core.py), we probably want to document right and be more surgical about what we don't document. Things that are internal like runtime.py, we can probably ignore everything. There are not that many user facing things. It's mostly the __init__.pyi in the stubs (top level in metaflow-stubs), things in the client, things like current, things in the flowspec, and maybe some things in cards (like the type of cards). I may be missing a few but we should have good coverage with that.

madhur-ob:
> I wonder what the balance is between ignoring errors and fixing them.

For the purpose of getting the PR merged, I think we can start with ignoring all (after taking into consideration the generated stubs), and then coming back later to fix them in future PRs.

> need to have a py.typed

Can you expand on this a bit?

PS -- I have fixed some errors in this PR itself where names of arguments didn't comply with the super / parent class...

madhur-ob:
Closing in favour of https://github.com/Netflix/metaflow/pull/1676

PR #1681, opened 2024-01-12
support git repos for @pypi

savingoyal:


metaflow/plugins/pypi/pip.py
```diff
@@ -89,21 +109,76 @@ def download(self, id_, packages, python, platform):
...
+                    lambda x: _build(*x),
+                    enumerate(
+                        package for package in packages if package["require_build"]
+                    ),
+                )
+            )
+            os.makedirs("%s/.pip/wheels" % prefix, exist_ok=True)
+
+            for package, path in results:
+                (wheel,) = [
```
saikonen, line 149: this will outright fail if the folder doesn't contain exactly one .whl file. Unsure if that is a relevant case to cover at this point though

metaflow/plugins/pypi/pip.py
```diff
@@ -89,21 +109,76 @@ def download(self, id_, packages, python, platform):
...
+                ):
+                    raise PipException(
+                        "The built wheel %s is not supported for %s with Python %s"
+                        % (wheel, platform, python)
+                    )
+                target = "{prefix}/.pip/wheels/{wheel}".format(
+                    prefix=prefix, wheel=wheel
+                )
+                shutil.move(os.path.join(path, wheel), target)
+                metadata["{url}".format(**package)] = target
```
saikonen, line 166: slight note on comparing this to #1559 , keeping a separate `BUILD_METADATA` file has the benefit of being able to skip only sources requiring building if they've been built already for the environment. Consider the case:
- user has both pypi and git sources
- git sources get built first, successfully
- pypi sources fail to download due to network issues
- environment creation fails

retrying this would require rebuilding the git sources as well unnecessarily?

savingoyal: yes, that's a very valid edge case but comes at the expense of managing state across multiple files. hopefully re-building a pure-python package should be rather quick but if that isn't the case, we can always split the file into two.

PR #1687, opened 2024-01-16
bump Metaflow version

valayDave:
Doing a minor jump in semver because this will mark the official version from which dynamic cards will be supported


romain-intel:
WAIT -- I thought we were first doing a silent release so we could test it. Can we do that this week and THEN do a minor bump.

PR #1691, opened 2024-01-18
Set parameters for running flow instead of any imported

madhur-ob:
Suggested changes on top of the work of @maxzheng on https://github.com/Netflix/metaflow/pull/1653

### DIFF
--- a/metaflow/flowspec.py
+++ b/metaflow/flowspec.py
@@ -200,7 +200,7 @@ class FlowSpec(object):
                 continue
             try:
                 val = getattr(self, var)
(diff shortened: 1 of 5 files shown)


saikonen (approved):

saikonen (commented):
diff for `test_runner` and `tox.ini` should be empty, but the current changes are only formatting so its fine.

madhur-ob:
`except` blocks have been restored, CC: @savingoyal

savingoyal (approved):

PR #1695, opened 2024-01-24
add ability to terminate execution of a step-fn state machine

madhur-ob:
This PR should over-ride https://github.com/Netflix/metaflow/pull/1596

Usage:
`python flow.py step-functions terminate sfn-d41004cd-1bbf-4c25-b284-94bfc921d78b`

### DIFF
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -227,8 +227,8 @@ class ArgoWorkflows(object):
 
         return schedule_deleted, sensor_deleted, workflow_deleted
 
-    @staticmethod
-    def terminate(flow_name, name):
+    @classmethod
+    def terminate(cls, flow_name, name):
         client = ArgoClient(namespace=KUBERNETES_NAMESPACE)
 
         response = client.terminate_workflow(name)
--- a/metaflow/plugins/aws/step_functions/step_functions.py
+++ b/metaflow/plugins/aws/step_functions/step_functions.py
@@ -167,6 +167,13 @@ class StepFunctions(object):
         return schedule_deleted, sfn_deleted
 
     @classmethod
+    def terminate(cls, flow_name, name):
+        client = StepFunctionsClient()
+        execution_arn, _, _, _, _, _ = cls.get_execution(flow_name, name)
+        response = client.terminate_execution(execution_arn)
+        return response
+
+    @classmethod
     def trigger(cls, name, parameters):
         try:
             state_machine = StepFunctionsClient().get(name)
@@ -233,6 +240,48 @@ class StepFunctions(object):
                     "Functions." % name
                 )
         return None
+
+    @classmethod
+    def get_execution(cls, state_machine_name, name):
+        client = StepFunctionsClient()
+        try:
+            state_machine = client.get(state_machine_name)
+        except Exception as e:
+            raise StepFunctionsException(repr(e))
+        if state_machine is None:
+            raise StepFunctionsException(
+                "The workflow *%s* doesn't exist " "on AWS Step Functions." % name
+            )
+        try:
+            state_machine_arn = state_machine.get("stateMachineArn")
+            parameters = (
+                json.loads(state_machine.get("definition"))
+                .get("States")
+                .get("start")
+                .get("Parameters")
+                .get("Parameters")
+            )
+
+            executions = client.list_executions(state_machine_arn, states=["RUNNING"])
+            for execution in executions:
+                if execution.get("name") == name:
+                    try:
+                        return (
+                            execution.get("executionArn"),
+                            parameters.get("metaflow.owner"),
+                            parameters.get("metaflow.production_token"),
+                            parameters.get("metaflow.flow_name"),
+                            parameters.get("metaflow.branch_name", None),
+                            parameters.get("metaflow.project_name", None),
+                        )
+                    except KeyError:
+                        raise StepFunctionsException(
+                            "A non-metaflow workflow *%s* already exists in AWS Step Functions."
+                            % name
+                        )
+            return None
+        except Exception as e:
+            raise StepFunctionsException(repr(e))
 
     def _compile(self):
         if self.flow._flow_decorators.get("trigger") or self.flow._flow_decorators.get(
--- a/metaflow/plugins/aws/step_functions/step_functions.py
+++ b/metaflow/plugins/aws/step_functions/step_functions.py
@@ -250,7 +250,8 @@ class StepFunctions(object):
             raise StepFunctionsException(repr(e))
         if state_machine is None:
             raise StepFunctionsException(
-                "The workflow *%s* doesn't exist on AWS Step Functions." % name
+                "The state machine *%s* doesn't exist on AWS Step Functions."
+                % state_machine_name
             )
         try:
             state_machine_arn = state_machine.get("stateMachineArn")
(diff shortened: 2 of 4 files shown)


metaflow/plugins/argo/argo_workflows.py
```diff
@@ -227,8 +227,8 @@ def delete(name):
 
         return schedule_deleted, sensor_deleted, workflow_deleted
 
-    @staticmethod
-    def terminate(flow_name, name):
+    @classmethod
```
saikonen, line 230: does this need to be a class method if the `cls` is not accessed?

metaflow/plugins/aws/step_functions/step_functions.py
```diff
@@ -166,6 +166,13 @@ def delete(cls, name):
 
         return schedule_deleted, sfn_deleted
 
+    @classmethod
```
saikonen, line 169: same here, rather static method if `cls` is not needed?

saikonen (commented):
had a quick look so far, a few comments but overall looking good. I'll take this for a spin tomorrow when I have more time.

metaflow/plugins/argo/argo_workflows.py
madhur-ob: I was making it compatible with what I had for the `terminate` method in `step_functions.py`

metaflow/plugins/aws/step_functions/step_functions.py
madhur-ob: Here, I do use `cls.get_execution(flow_name, name)`

metaflow/plugins/aws/step_functions/step_functions_cli.py
```diff
@@ -614,6 +618,104 @@ def _token_instructions(flow_name, prev_user):
...
+    obj.echo(
+        "Terminating run *{run_id}* for {flow_name} ...".format(
+            run_id=run_id, flow_name=obj.flow.name
+        ),
+        bold=True,
+    )
+
+    terminated = StepFunctions.terminate(obj.state_machine_name, name)
+    if terminated:
+        obj.echo(f"\nRun terminated at {terminated.get('stopDate')}.")
```
saikonen, line 664: python 3.5 borks with the f-strings so switch this to an older style.

metaflow/plugins/aws/step_functions/step_functions_client.py
```diff
@@ -81,9 +81,14 @@ def list_executions(self, state_machine_arn, states):
...
 
-    def terminate_execution(self, state_machine_arn, execution_arn):
-        # TODO
-        pass
+    def terminate_execution(self, execution_arn):
+        try:
+            response = self._client.stop_execution(executionArn=execution_arn)
+            return response
+        except self._client.exceptions.ExecutionDoesNotExist:
+            raise ValueError(f"The execution ARN {execution_arn} does not exist.")
```
saikonen, line 89: same here with the f-string

metaflow/plugins/aws/step_functions/step_functions.py
saikonen: ahh, I see it now! my bad 😓

metaflow/plugins/aws/step_functions/step_functions_cli.py
madhur-ob: done

metaflow/plugins/aws/step_functions/step_functions_client.py
madhur-ob: done

metaflow/plugins/aws/step_functions/step_functions_cli.py
```diff
@@ -614,6 +618,104 @@ def _token_instructions(flow_name, prev_user):
...
+):
+    if not run_id.startswith("sfn-"):
+        raise RunIdMismatch(
+            "Run IDs for flows executed through AWS Step Functions begin with 'sfn-'"
+        )
+
+    name = run_id[4:]
+    execution = StepFunctions.get_execution(state_machine_name, name)
+    if execution is None:
+        raise MetaflowException(
```
saikonen, line 678: we actually end up hitting this exception mmore often than any of the more specific ones. The reason being that the state_machine_name is used as part of the execution lookup.

the state machine name contains the project branch, so if current project is a mismatch, we wont find the execution to begin with.

Is it possible to lookup the execution directly with the execution ID instead?

saikonen (changes requested):

metaflow/plugins/aws/step_functions/step_functions_cli.py
madhur-ob: Nope, unfortunately it is not

metaflow/plugins/aws/step_functions/step_functions_cli.py
```diff
@@ -614,6 +618,91 @@ def _token_instructions(flow_name, prev_user):
...
+            "Could not find the execution *%s* (in RUNNING state) on AWS Step Functions"
```
saikonen, line 679:
```
"Could not find the execution *%s* (in RUNNING state) for the state machine *%s* on AWS Step Functions" % (name, state_machine_name)
```
for clarity?

metaflow/plugins/aws/step_functions/step_functions_cli.py
```diff
@@ -614,6 +618,91 @@ def _token_instructions(flow_name, prev_user):
...
+
+    _, owner, token, sfn_state_machine = execution
+
+    # this snippet is probably never triggered since we fail early with not
+    # being able to find the state_machine itself
+    if state_machine_name != sfn_state_machine:
+        raise RunIdMismatch(
+            "The workflow with the run_id *%s* belongs to the state machine *%s*, not for the state machine *%s*."
+            % (run_id, sfn_state_machine, state_machine_name)
+        )
```
saikonen, line 691: Because we search for the execution with the `state_machine_name`, then `state_machine_name==sfn_state_machine` always if an execution was found. It should be okay to remove this check.

saikonen (changes requested):

saikonen:
> @saikonen can you confirm if you don't need the `--authorize` flag if you are trying to terminate as another user. If you confirm, then we need to discuss -- since this is probably a bug or something that I am not aware of...

tested this and the token authorization and user/owner checks are working correctly. The issue with testing this is that the production token is cached locally, so in order to test a case where both user and token do not match, you need to invalidate the locally cached token

metaflow/plugins/aws/step_functions/step_functions_cli.py
madhur-ob: removed...

metaflow/plugins/aws/step_functions/step_functions_cli.py
madhur-ob: thanks

saikonen (approved):

metaflow/plugins/aws/step_functions/step_functions_cli.py
```diff
@@ -614,6 +618,83 @@ def _token_instructions(flow_name, prev_user):
         )
 
 
+@step_functions.command(help="Terminate flow execution on Step Functions.")
```
savingoyal, line 621: minor nit - but for consistency `AWS Step Functions`. not blocking for merging this PR in, but would be good if you can sweep this and various CLIs to polish up the help strings and make them consistent.

PR #1704, opened 2024-01-27
fix: Deterministic foreach task id's for Argo Workflows

### LINKED ISSUE #1538: Argo Workflows template exceeds max size with preceding large foreach

saikonen:
changes the Task ID generation  for tasks in foreach branches to be deterministic. This is so that the following join steps do not require an enormous list of ID's to be passed in through the `input-paths` input Parameter, which has limited the number  of foreach tasks we can run on Argo Workflows until now.

Note: Still requires rigorous testing for different split cases to confirm this is ready for prime time.

Closes #1538 properly. Supersedes #1655


romain-intel:
I'll check but failure is probably that card test that I need to increase timeout for. There is no change to common code.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -883,6 +885,34 @@ def _visit(node, exit_node=None, templates=None, dag_tasks=None):
                         )
                     )
                 ]
+                # We need to add the split-index and root-input-path for the last step in any foreach to be able to generate
```
savingoyal, line 888: Can you add more comments in the code about the need for `root-input-path`. Given the complexity of this work, it will be non-trivial to ship further changes without full context on this change.

savingoyal (changes requested):

metaflow/plugins/argo/generate_input_paths.py
```diff
@@ -0,0 +1,23 @@
+import sys
+from hashlib import md5
+
+
+def generate_input_paths(step_name, timestamp, input_paths, max_splits):
```
savingoyal, line 5: `input_paths` have been `flow/run/step/task` previously but it seems that now `input_paths` are `run/step/task`?

metaflow/plugins/argo/generate_input_paths.py
```diff
@@ -0,0 +1,23 @@
...
+
+    ids = [_generate_task_id(foreach_base_id, i) for i in range(int(max_splits))]
+    return "{}/{}/:{}".format(run_id, step_name, ",".join(ids))
+
+
+def _generate_task_id(base, idx):
+    # For foreach splits generate the expected input-paths based on max_split and base_id via.
+    # newline required at the end due to 'echo' appending one in the shell side task_id creation.
+    task_str = "%s-%s\n" % (base, idx)
+    hash = md5(task_str.encode("utf-8")).hexdigest()[-8:]
```
savingoyal, line 18: it might be desirable to get task ids of the nature - `t-<hash>-<idx>` - in that scenario - we can probably get a much tighter compression of input_paths (assuming we change the input_paths compression/decompression logic) in addition to the UI looking a lot more nicer.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -1339,20 +1455,51 @@ def _container_templates(self):
...
                 outputs = [Parameter("task-id").valueFrom({"path": "/mnt/out/task_id"})]
             if node.type == "foreach":
                 # Emit split cardinality from foreach task
+                # list of split indices
                 outputs.append(
                     Parameter("num-splits").valueFrom({"path": "/mnt/out/splits"})
                 )
+                # maximum of the splits
+                outputs.append(
+                    Parameter("max-split").valueFrom({"path": "/mnt/out/max_split"})
```
savingoyal, line 1501: maybe rename it to `split-cardinality`?

savingoyal (commented):
still reviewing

metaflow/plugins/argo/argo_workflows.py
saikonen: renamed.

savingoyal (approved):

PR #1707, opened 2024-01-30
fix github action for publishing stubs

madhur-ob:


.github/workflows/publish.yml
```diff
@@ -29,13 +29,13 @@ jobs:
       run: |
         cd ./stubs && python3 setup.py sdist bdist_wheel --universal && cd -
     - name: Publish metaflow-stubs package
-      uses: pypa/gh-action-pypi-publish@9b8e7336db3f96a2939a3e9fa827c62f466ca60d # master
```
romain-intel, line 32: - I think we just need to do release/v1 according to the doc

.github/workflows/publish.yml
```diff
@@ -29,13 +29,13 @@ jobs:
...
+      uses: pypa/gh-action-pypi-publish@release/v1.8
       with:
         user: __token__
         password: ${{ secrets.pypi_password }}
-        packages-dir: ./stubs/dist
```
romain-intel, line 36: the doc does say `packages-dir`. I think it was "fixed": https://github.com/pypa/gh-action-pypi-publish/blob/c12cc61414480c03e10ea76e2a0a1a17d6c764e2/action.yml#L28

.github/workflows/publish.yml
madhur-ob: using hash for `v1.8.11`

.github/workflows/publish.yml
madhur-ob: reverted...

PR #1708, opened 2024-01-30
[WIP] Support user specified python executables for remote bootstrapping

talsperre:
This PR allows users to pass custom python executable paths for saving logs and remote bootstrapping.

### DIFF
--- a/metaflow/mflog/__init__.py
+++ b/metaflow/mflog/__init__.py
@@ -42,22 +42,36 @@ BASH_MFLOG = (
 )
 
 
-def _get_bash_capture_log(default_python_executable="python"):
-    bash_save_log_args = [default_python_executable, "-m", "metaflow.mflog.save_logs"]
-    return bash_save_log_args, " ".join(bash_save_log_args)
+def _get_bash_capture_log(default_python_executable=None):
+    python_executable = default_python_executable or "python"
+    bash_save_log_args = [python_executable, "-m", "metaflow.mflog.save_logs"]
+    return bash_save_log_args
 
 
 # this function returns a bash expression that redirects stdout
 # and stderr of the given bash expression to mflog.tee
-def bash_capture_logs(bash_expr, var_transform=None, default_python_executable_path="python"):
+def bash_capture_logs(
+    bash_expr, var_transform=None, default_python_executable_path=None
+):
+    python_executable = default_python_executable_path or "python"
     if var_transform is None:
         var_transform = lambda s: "$%s" % s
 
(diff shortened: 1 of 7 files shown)


metaflow/plugins/kubernetes/kubernetes.py
```diff
@@ -130,7 +130,7 @@ def _command(
         # We lose the last logs in this scenario.
         #
         # TODO: Capture hard exit logs in Kubernetes.
-        cmd_str += "c=$?; %s; exit $c" % BASH_SAVE_LOGS
+        cmd_str += "c=$?; %s; exit $c" % _get_bash_capture_log()
```
romain-intel, line 133: should be [1]

metaflow/mflog/__init__.py
```diff
@@ -41,21 +41,23 @@
...
+
+def _get_bash_capture_log(default_python_executable="python"):
+    bash_save_log_args = [default_python_executable, "-m", "metaflow.mflog.save_logs"]
+    return bash_save_log_args, " ".join(bash_save_log_args)
 
 
 # this function returns a bash expression that redirects stdout
 # and stderr of the given bash expression to mflog.tee
-def bash_capture_logs(bash_expr, var_transform=None):
+def bash_capture_logs(bash_expr, var_transform=None, default_python_executable_path="python"):
```
romain-intel, line 52: nit: consistency: for bootstrap, it defaults to none but here to python. We can make it python for both or none for both.

metaflow/mflog/__init__.py
```diff
@@ -41,21 +41,23 @@
     " }" % TASK_LOG_SOURCE
 )
 
-BASH_SAVE_LOGS_ARGS = ["python", "-m", "metaflow.mflog.save_logs"]
-BASH_SAVE_LOGS = " ".join(BASH_SAVE_LOGS_ARGS)
+
+def _get_bash_capture_log(default_python_executable="python"):
+    bash_save_log_args = [default_python_executable, "-m", "metaflow.mflog.save_logs"]
+    return bash_save_log_args, " ".join(bash_save_log_args)
```
romain-intel, line 47: we could just return the array part and have the user do the join. I don't have a strong preference but the [0], and [1] on use are a bit weird.

romain-intel (commented):
Minor comments but otherwsie lgtm.

romain-intel:
@talsperre -- as discussed, I think we don't need this anymore and we, instead, need to start the command line with an aliasing (or something equivalent since alias didn't work).

PR #1709, opened 2024-01-31
Add support for shared_memory for Kubernetes workloads.

shrinandj:
This includes support for Kubernetes jobs as well as Argo workflows.

Testing Done:

- Verified that setting shared_memory=N in @kubernetes works as expected
- Verified that setting shared_memory=N in @kubernetes works with argo-workflows
- Verified that setting shared_memory=N on the cli works as expected
- Verified that not setting shared_memory works fine


roofurmston:
Perhaps I am misunderstanding the PR, apologies if that is the case, but isn't shared memory already supported in kubernetes through the `use_tmpfs`, `tmpfs_path` and `tmpfs_size` parameters?

saikonen:
> Perhaps I am misunderstanding the PR, apologies if that is the case, but isn't shared memory already supported in kubernetes through the `use_tmpfs`, `tmpfs_path` and `tmpfs_size` parameters?

functionally there is no difference, and it can be supported with the tmpfs parameters as you noted. This is most likely a convenience/convergence feature, as the `shared_memory` is a separate parameter on `batch` for example

saikonen (commented):
this doesn't seem to work with the `resources` decorator yet? Is it in scope for the PR?

```python
@resources(shared_memory=120)
@kubernetes(shared_memory=100)
@step
```
the expectation is that the larger of the two being picked.

roofurmston:
> > Perhaps I am misunderstanding the PR, apologies if that is the case, but isn't shared memory already supported in kubernetes through the `use_tmpfs`, `tmpfs_path` and `tmpfs_size` parameters?
> 
> functionally there is no difference, and it can be supported with the tmpfs parameters as you noted. This is most likely a convenience/convergence feature, as the `shared_memory` is a separate parameter on `batch` for example

OK, fair enough. Just want to raise it really.

shrinandj:
Thanks for the questions/review.

One reason to have an explicit `shared_memory` option would also be if someone wants the use both: tmpfs and shared_memory. tmpfs is be more generic. `shared_memory` specifically gets used by some of the existing tools like deepspeed and MPI.

saikonen (changes requested):
the stubs test cases need a fix before we go ahead with this, as its going to block the release otherwise due to failures. Not sure why they're failing with a simple change to the decorator func signature. @madhur-ob do you have a chance to take a look at the failures?

metaflow/plugins/kubernetes/kubernetes_cli.py
```diff
@@ -107,6 +107,7 @@ def kubernetes():
     type=JSONTypeClass(),
     multiple=False,
 )
+@click.option("--shared-memory", default=None, help="Size of shared memory in MB")
```
savingoyal, line 110: I believe it's not MB but MiB. We should look into other values like disk, tmpfs, memory and ensure that all fields have the same dimension

savingoyal (dismissed):

metaflow/plugins/kubernetes/kubernetes_cli.py
shrinandj: Fixed.

madhur-ob:
@shrinandj I guess you reverted my changes by accident?

saikonen (approved):

PR #1716, opened 2024-02-05
[kubernetes] Use KUBECONFIG if set

### LINKED ISSUE #1715: "Service token file does not exist" error when deploying flow to Argo from CI
When running `argo-workflows create` from a Bitbucket Cloud pipeline, the following error is raised:

```
```

leogargu:
Closes #1715

Not sure where to document this (it'd be useful for folks setting up their CI pipelines) -  any ideas?

### DIFF
--- a/metaflow/plugins/kubernetes/kubernetes_client.py
+++ b/metaflow/plugins/kubernetes/kubernetes_client.py
@@ -34,6 +34,11 @@ class KubernetesClient(object):
         from kubernetes import client, config
 
         if os.getenv("KUBECONFIG"):
+            # There are cases where we're running inside a pod, but can't use
+            # the kubernetes client for that pod's cluster: for example when
+            # running in Bitbucket Cloud or other CI system.
+            # In this scenario, the user can set a KUBECONFIG environment variable
+            # to load the kubeconfig, regardless of whether we're in a pod or not.
             config.load_kube_config()
         elif os.getenv("KUBERNETES_SERVICE_HOST"):
             # We are inside a pod, authenticate via ServiceAccount assigned to us


metaflow/plugins/kubernetes/kubernetes_client.py
```diff
@@ -32,15 +34,19 @@ def __init__(self):
...
+        if os.getenv("KUBECONFIG"):
             # Use kubeconfig, likely $HOME/.kube/config
             # TODO (savin):
             #  1. Support generating kubeconfig on the fly using boto3
             #  2. Support auth via OIDC - https://docs.aws.amazon.com/eks/latest/userguide/authenticate-oidc-identity-provider.html
             config.load_kube_config()
+        elif os.getenv("KUBERNETES_SERVICE_HOST"):
+            # We are inside a pod, authenticate via ServiceAccount assigned to us
+            config.load_incluster_config()
+        else:
```
savingoyal, line 46: this might be a backwards incompatible change and will break many currently well functioning flows. do we expect `KUBECONFIG` env var to be always available?

savingoyal (changes requested):

metaflow/plugins/kubernetes/kubernetes_client.py
shrinandj: This could be made backward compatible if the original call to `config.load_kube_config` in the `else` case is retained instead of throwing the KubernetesException.

savingoyal (commented):
can you add a comment in the code base reflecting why this change is needed - it will help in preserving the semantics going forward

PR #1719, opened 2024-02-06
[Ready for review] Make Clone-only resume more efficient.

darinyu:
Current clone-only resume has a few flaws we need to solve:

1. Each task to be resume will create a resume runtime and they all want to resume the previously completed execution tree in topological order. This resulted in unnecessary competition between resume tasks.
2. If one resume wins the race (registered new task id) but never completed the cloning, the future attempts will never run and the resume will never finish.

To tackle this, we decided to elect a resume leader (whichever tasks register _parameters task first) and also remembers the resume leader path. All non-leader will need to wait for the leader to finish resuming (which essentially is copying the datastore pointer to new place).

In the case of a failed clone, we will check whether task_id is registered and task is complete to ensure that we will still clone the old run when task_id is already registered but never complete.

To test this, one can try this [flow](https://gist.github.com/darinyu/315ad0b85cc0979a6cad1a4fbc91d76c).
```
python flow.py run
python flow.py resume
```

### DIFF
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -791,6 +791,7 @@ def resume(
         max_log_size=max_log_size * 1024 * 1024,
     )
     write_run_id(run_id_file, runtime.run_id)
+    runtime.print_workflow_info()
     runtime.persist_constants()
     runtime.execute()
 
@@ -845,6 +846,7 @@ def run(
     write_run_id(run_id_file, runtime.run_id)
 
     obj.flow._set_constants(obj.graph, kwargs)
+    runtime.print_workflow_info()
     runtime.persist_constants()
     runtime.execute()
 
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -474,13 +474,6 @@ def logs(obj, input_path, stdout=None, stderr=None, both=None, timestamps=False)
     "not execute anything.",
 )
 @click.option(
-    "--clone-wait-only/--no-clone-wait-only",
-    default=False,
-    show_default=True,
-    help="If specified, waits for an external process to clone the task",
-    hidden=True,
-)
-@click.option(
     "--clone-run-id",
     default=None,
     help="Run id of the origin flow, if this task is part of a flow being resumed.",
@@ -519,7 +512,6 @@ def step(
     retry_count=None,
     max_user_code_retries=None,
     clone_only=None,
-    clone_wait_only=False,
     clone_run_id=None,
     decospecs=None,
     ubf_context="none",
@@ -568,6 +560,8 @@ def step(
         ctx.obj.monitor,
         ubf_context,
     )
+    print("step!!!")
+    print("step_name", step_name)
     if clone_only:
         task.clone_only(
             step_name,
@@ -575,7 +569,6 @@ def step(
             task_id,
             clone_only,
             retry_count,
-            wait_only=clone_wait_only,
         )
     else:
         task.run_step(
@@ -719,6 +712,13 @@ def common_run_options(func):
     hidden=True,
     help="If specified, allows this call to be called in parallel",
 )
+@click.option(
+    "--resume-identifier",
+    default=None,
+    show_default=True,
+    hidden=True,
+    help="If specified, it identifies the task that started this resume call",
+)
 @click.argument("step-to-rerun", required=False)
 @cli.command(help="Resume execution of a previous run of this flow.")
 @common_run_options
@@ -736,6 +736,7 @@ def resume(
     max_log_size=None,
     decospecs=None,
     run_id_file=None,
+    resume_identifier=None,
 ):
     before_run(obj, tags, decospecs + obj.environment.decospecs())
 
@@ -758,7 +759,8 @@ def resume(
                 )
             )
         clone_steps = {step_to_rerun}
-
+    if resume_identifier:
+        print("resume_identifier: ", resume_identifier)
     if run_id:
         # Run-ids that are provided by the metadata service are always integers.
         # External providers or run-ids (like external schedulers) always need to
@@ -789,6 +791,7 @@ def resume(
         max_workers=max_workers,
         max_num_splits=max_num_splits,
         max_log_size=max_log_size * 1024 * 1024,
+        resume_identifier=resume_identifier,
     )
     write_run_id(run_id_file, runtime.run_id)
     runtime.print_workflow_info()
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -724,7 +724,7 @@ def common_run_options(func):
     default=None,
     show_default=True,
     hidden=True,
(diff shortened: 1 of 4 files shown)


metaflow/runtime.py
```diff
@@ -916,6 +1008,31 @@ def log(self, msg, system_msg=False, pid=None, timestamp=True):
...
+
+    def mark_resume_done(self):
+        assert (
+            self.step == "_parameters"
+        ), "Only _parameters step can mark resume as done."
+        assert self.is_resume_leader(), "Only resume leader can mark resume as done."
+
+        # Mark the resume as done. This is called at the end of the resume flow and after
+        # the _parameters step was successfully cloned, so we need to 'dangerously' save
+        # this done file, but the risk should be minimum.
```
romain-intel, line 1031: nit: minimal.

metaflow/cli.py
```diff
@@ -719,6 +719,13 @@ def wrapper(*args, **kwargs):
     hidden=True,
     help="If specified, allows this call to be called in parallel",
 )
+@click.option(
+    "--resume-identifier",
+    default=None,
+    show_default=True,
+    hidden=True,
+    help="If specified, it identifies the task that started this resume call",
```
romain-intel, line 727: nit: should be something about pathspec -- I suspect we need to actually list step_name/task_id here right?

metaflow/runtime.py
```diff
@@ -798,56 +836,110 @@ def __init__(
...
+                if self._is_resume_leader:
+                    self.log(
+                        "Selected as the reentrant clone leader.",
+                        system_msg=True,
+                    )
                     # Clone in place without relying on run_queue.
                     self.new_attempt()
                     self._ds.clone(origin)
+                    # Set the resume leader be the task that calls the resume (first task to clone _parameters task).
+                    if resume_identifier:
```
romain-intel, line 895: What is this check for? I thought the check should be more like: "if we actually created the task ID, we know we are the "first elected" resume leader so we write our identifier.

Also, shouldn't we actually error out if resume_identifier is not set in this case?

metaflow/runtime.py
```diff
@@ -798,56 +836,110 @@ def __init__(
...
+                    ds = self._flow_datastore.get_task_datastore(
+                        self.run_id, self.step, self.task_id
+                    )
+                    if not ds["_task_ok"]:
+                        raise MetaflowInternalError(
+                            "Externally cloned _parameters task did not succeed"
+                        )
+
+                    # Check if we should be the resume leader (maybe from previous attempt)
+                    if ds.has_metadata("_resume_leader", add_attempt=False):
```
romain-intel, line 871: This is a bit tricky here -- there is a possible race where the resume leader is elected but never writes out the metadata which would mean that then no one is the resume leader. We may want to comment on this here but I think we can ignore it for now as it really shouldn't happen.

metaflow/runtime.py
```diff
@@ -798,56 +836,110 @@ def __init__(
...
-                    )
+                    # Wait for the resume leader to complete
                     while True:
-                        try:
-                            ds = self._flow_datastore.get_task_datastore(
-                                self.run_id, self.step, self.task_id
+                        ds = self._flow_datastore.get_task_datastore(
+                            self.run_id, self.step, self.task_id
+                        )
+                        if not ds["_task_ok"]:
```
romain-intel, line 907: why error out so fast here? Shouldn't we just retry. Isn't this scenario possible:
- resume leader elected (task A)
- task B gets here before task A executes line 900

This would error out here but is not an error per say. We just need to try again.

metaflow/runtime.py
```diff
@@ -749,32 +775,44 @@ def __init__(
...
+        self._resume_identifier = resume_identifier
+
         origin = None
         if clone_run_id and may_clone:
             origin = self._find_origin_task(clone_run_id, join_type)
         if origin and origin["_task_ok"]:
             # At this point, we know we are going to clone
             self._is_cloned = True
+
+            if resume_identifier:
```
romain-intel, line 790: we can probably move this into reentrant case. It should be set if we are reentrant. This will also make it slightly cleaner insofar as reentrant is the main flag controlling this distributed resume thingy.

metaflow/task.py
```diff
@@ -291,22 +291,9 @@ def clone_only(
...
-                    )
-                    if not ds["_task_ok"]:
-                        raise MetaflowInternalError(
-                            "Externally cloned task did not succeed"
-                        )
-                    break
-                except DataException:
-                    # No need to get fancy with the sleep here.
-                    time.sleep(5)
+            print("Not cloning anything in wait_only mode.")
```
romain-intel, line 294: I guess we could jsut totally skip the subprocess but I suppose the next PR does that so no worries here.

romain-intel (commented):
Some smallish nits but otherwise looks good.

metaflow/task.py
darinyu: yeah

metaflow/runtime.py
darinyu: moved under reentrant, and raise error if not set.

metaflow/cli.py
darinyu: done

metaflow/runtime.py
darinyu: done

metaflow/runtime.py
darinyu: this is not within the reentrant section, so it could still be empty, I feel that overwriting the same value twice should be fine so I did not check if we are "first elected" or first attempt (feel that checking this scenario makes the logic more complex and this operation is not too expensive).

metaflow/runtime.py
darinyu: I think it is fine here because it is under "task_id_exists_already=True", which means
1. We are in first attempt, since task_id_exists_already is true, we can not be leader anyways.
2. We are in second attempt, and we could be leader from writes in previous attempt (no race here)

metaflow/runtime.py
darinyu: Thats a good point. I actually want to check if "_task_ok" is set and the value is False (meaning task failed) (maybe this doesn't do that?) let me check what is the default behavior of `ds["_task_ok"]`.

Also, thinking about it twice, I may not need to check this at all because if the leader fails, the other task will be stopped anyways. WDYT?

metaflow/runtime.py
darinyu: talked offline, added code to retry (to avoid race condition)

metaflow/runtime.py
darinyu: added more comments

metaflow/runtime.py
darinyu: removed

saikonen (approved):
seems to be working as expected based on my limited testing, should be good to go.

PR #1728, opened 2024-02-08
[Ready for review] Retrieve original run tasks and clone successful ones directly

darinyu:


metaflow/cli.py
```diff
@@ -802,7 +802,10 @@ def resume(
     write_run_id(run_id_file, runtime.run_id)
     runtime.print_workflow_info()
     runtime.persist_constants()
-    runtime.execute()
+    if clone_only:
```
romain-intel, line 805: This seems to change the behavior only for our internal use case of clone first and then execute but it feels like this mechanism would be able to work even for a regular resume by first cloning and then continuing execution?

metaflow/runtime.py
```diff
@@ -235,6 +237,52 @@ def _should_skip_clone_only_execution(self):
...
+
+    def clone_original_run(self):
+        (
+            should_skip_clone_only_execution,
+            skip_reason,
+        ) = self._should_skip_clone_only_execution()
+        if should_skip_clone_only_execution:
+            self._logger(skip_reason, system_msg=True)
+            return
+        self._metadata.start_run_heartbeat(self._flow.name, self._run_id)
```
romain-intel, line 264: shoudl we stop it?

metaflow/task.py
```diff
@@ -293,48 +294,19 @@ def clone_only(
         if wait_only:
```
romain-intel, line 294: Would this ever happen now?

romain-intel (commented):
I think this looks fine for the most part. My main comment/concern is that I think it could be expanded to cover the regular resume as well to speed that one up too. Right now, this only speeds up the reentrant resume if I am reading this right. WIth OB's approval of course but I feel that we would also benefit even for the regular resume (which is also a legit use case if not the most pressing one).

metaflow/runtime.py
darinyu: added stop heartbeat in clone only run.

metaflow/cli.py
darinyu: talked offline, lets make this PR focusing on clone-only behavior first.

metaflow/task.py
darinyu: deleted

metaflow/clone_util.py
```diff
@@ -0,0 +1,74 @@
...
+    step_name,
+    clone_task_id,
+    task_id,
+    flow_datastore,
+    metadata_service,
+    origin_ds_set=None,
+    attempt_id=0,
+):
+    print(
+        f"Cloning task from {flow_name}/{clone_run_id}/{step_name}/{task_id} to {flow_name}/{run_id}/{step_name}/{task_id}"
```
saikonen, line 18: Is this an intentional print or for debugging purposes? The stdout for this seems to differ from regular MFLOG lines.

also as this is called as part of core, the f-string will break for old python so at least change that to older style if the output is intentional.

saikonen: a few other occurences of f-strings need changes as well.

saikonen (changes requested):
missing context on `--clone-wait-only` use case, but apart from the minor changes this should be good to go.

romain-intel:
For reference and posterity: the way we use resume is in one of two ways:
- the usual: "clone all tasks and continue executing in runtime.py"
- a distributed manner where a scheduler (argo/airflow) "resumes" the run (the execution frontier of tasks is relaunched by the scheduler and then all those tasks try to "resume") and then each task will execute its own thing and the scheduler will then continue. This is effectively a resume but without a runtime.py to drive it.

This PR is the second in the line to improve that second behavior:
- the first elected a leader to do the cloning as opposed to doing it in a distributed fashion as we used to do
- this one makes it also a lot faster to clone the tasks (ie: the job of the leader is simplified).

metaflow/clone_util.py
darinyu: Updated print and f-string

saikonen (approved):
needs a rebase due to target branch getting merged, but otherwise looks good to go.

PR #1729, opened 2024-02-12
[Ready for review] Ignore _foreach_values in merge artifacts

darinyu:
`_foreach_values` is now stored as artifacts for each step but this is not intended for users to consume. In merge artifacts, however, we will try to merge them by default, which could fail if the foreach are dynamic. This PR will make merge artifacts skip the foreach value (at join step). This foreach value will have no meaning after joining anyways, so there should be no impact to the entire flow.

### DIFF
--- a/metaflow/flowspec.py
+++ b/metaflow/flowspec.py
@@ -28,6 +28,8 @@ except NameError:
 
 from .datastore.inputs import Inputs
 
+INTERNAL_ARTIFACTS_SET = set(["_foreach_values"])
+
 
 class InvalidNextException(MetaflowException):
     headline = "Invalid self.next() transition detected"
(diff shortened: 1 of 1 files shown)


metaflow/flowspec.py
```diff
@@ -419,6 +419,8 @@ def merge_artifacts(
             This exception is thrown in case an artifact specified in `include` cannot
             be found.
         """
+        INTERNAL_ARTIFACTS_SET = set(["_foreach_values"])
```
romain-intel, line 422: nit: move this out of the function so it doesn't get recreated every time.

romain-intel (approved):
Minor nit but LGTM.

metaflow/flowspec.py
darinyu: updated

romain-intel (approved):
LGTM

PR #1732, opened 2024-02-15
initial runner api

madhur-ob:
The Usage looks something like this:

```
from metaflow import metaflow_runner

params = [{'alpha': 40}, {'alpha': 20}, {'alpha': 10}, {'alpha': 30}]
with metaflow_runner.Pool('../try.py', num_processes=2) as pool:
    results = pool.map(params)
    for i in results:
        print(i)
```

where the contents of `try.py` for the example are:
```
import time
from metaflow import FlowSpec, Parameter, step

class ParameterFlow(FlowSpec):
    alpha = Parameter('alpha',
                      help='Time to Sleep',
                      default=1)

    @step
    def start(self):
        print('alpha is %f' % self.alpha)
        print(f"Let me sleep for {self.alpha} seconds")
        time.sleep(self.alpha)
        self.next(self.end)

    @step
    def end(self):
        print('alpha is still %f' % self.alpha)

if __name__ == '__main__':
    ParameterFlow()
```

In the above example, the runs are executed 2 at a time. The order of them being completed is:
- run that sleeps for 20 seconds
- run that sleeps for 10 seconds
- run that sleeps for 40 seconds
- run that sleeps for 30 seconds

The total running time is 60 seconds. This is as follows:
- 40 seconds that includes
  - run for 20 seconds
  - run for 10 seconds
  - 10 seconds of the run for 30 seconds
- last 20 seconds of the run for 30 seconds

## Assumptions:
- Considering the use-case of `Controlling` and not `Authoring` metaflow runs... at least for now...
- Only accept a file which contains the flow, and not an instance of the class which is derived from `FlowSpec`
- Multiple concurrent runs can be spin off with the API

## Things that need to be discussed
- What other stuff should the API accept? (Eg: more environment variables, other options such as `--with`, etc.)
- This is only the API for `Pool`, what about just `Run`?
- What should a successful run return? Currently, we return the status code...
    - If the run is supposed to produce some data / artefact, how can we pass that further?
    Eg: this API is used as a step in some flow to produce an artefact which is consumed and used by another step
- Error handling / Crashing aspects...


romain-intel:
I'm not going to comment on the implementation (which is fairly straightforward at this time anyways) because I want to align on some UX things first. 

These are my initial thoughts on this but feel free to disagree with anything:
- I think for the initial scope, we can restrict to basically the `run` command (so `python myflow.py ... run ...`). We should, however, try to design a UX that is then extensible to other command line options (resume, dump, step-functions, etc.) that are "flow based" and we can then later also take care of the "metaflow xyz" command line (although that one can probably be addressed in a different way so I am excluding it from too much consideration for now.
- It doesn't feel like the current UX would extend to these other options. It is not so much a UX wrapping the "run" command but more an abstraction even on top of that that takes care of parallel executions of flows. In other words, I think the concept of having something like with Pool(...) is great, it should probably be built on top of a more direct overlay on the "run" CLI.
- One aspect that I feel may be important to capture is the hierarchical nature of the command line (or at least it would be closer to what click does which means it may be easier to map functionality. For example, one option could be the following: 

```
mf_cli = MetaflowCli(flow_file="myflow.py", environment="conda")
mf_cli.run(max_workers=5, alpha=123)
```

Not saying something exactly like this but this more cleanly maps to what we have for click and doesn't feel too un-natural (you have a Cli object -- or call it something else -- and from there you can run with various options exactly like you would on the command line). We can separately argue if we want to infer "environment" for example but that's a separate conversation :). Anyways, hopefully this shows the idea of having a multi level approach that maps more closely to click and can support any arbitrary options. It could be a very thin layer and should be reusable for various commands.
- I also think that `mf_cli.run` should return immediately *something* akin to a future but also akin to the client's Run object. We could (with some modifications) actually return a Run object and you could do something like this:

```
my_run = mf_cli.run(max_workers=5, alpha=123)
while not my_run.finished
  ...
```

(note this is probably not a good idea since it can loop forever but you get the idea). We should also seriously revisit some of these status things (we had a *long conversation* in the past about this but it may become more useful as part of this as well -- for example, we can actually return a better "finished" value here because we know if the subprocess dies/goes away and so it may not run forever :)).
You could also get the log, etc. We could consider adding other functions to list things like "in process steps" for example or what not. Basically it's an object that has value *during* execution as well as *after*. If we want to use a recent analogy, it's like a "dynamic card" (it updates through the execution and is then frozen at the end of execution.
- With something like that, it then becomes quite easy to write a Pool abstraction on top of it by having the async workers do something akin to what is above and just loop around until it is finished. As discussed, Pool should also return Run objects.

madhur-ob:
The usage looks something like this, we can of-course pass `blocking` as True if we wanna wait...

```
with metaflow_runner.Runner('../try.py') as runner:
    result = runner.run(params={'alpha': 1}, blocking=False)
    print(result) # this is the Run object
    while not result.finished:
        print("waiting for run to finish...")
        time.sleep(1)
```

or perhaps...

```
mf = metaflow_runner.Runner('../try.py')
result = mf.run(params={'alpha': 1}, blocking=False)
print(result) # this is the Run object
while not result.finished:
    print("waiting for run to finish...")
    time.sleep(1)
```

We also have a Pool API such as:

```
params = [{'alpha': 40}, {'alpha': 20}, {'alpha': 10}, {'alpha': 30}]
with metaflow_runner.Pool('../try.py', num_processes=2) as pool:
    results = pool.map(params, blocking=False) # these are the Run objects
    for each_run_obj in results:
        print(each_run_obj)
```

madhur-ob:
Pushed out a subprocess manager that allows us to tail logs via streaming them into a queue first...
The higher level runner API uses this subprocess manager...

An example of how the higher level UX is as follows:
```
import asyncio
from metaflow import metaflow_runner

async def main():
    with metaflow_runner.Runner('../try.py',  metadata="local") as runner:
        result = await runner.run(blocking=False, alpha=5, tags=["abc", "def"], max_workers=5)
        print(result) # this is the Run object
        print(result.finished) # this should be False
        await runner.tail_logs(stream="stdout")
        print(result.finished) # this should be True

        # the below can also be done...

        # while not result.finished:
        #     print("waiting for run to finish...")
        #     `asyncio.sleep(0.1)` or perhaps `time.sleep(0.1)`

if __name__ == "__main__":
    asyncio.run(main())
```

the output looks like:
```
Run('ParameterFlow/1710272507865687')
False
2024-03-13 01:11:47.872
Workflow starting (run-id 1710272507865687):
2024-03-13 01:11:47.878 [1710272507865687/start/1 (pid 85339)] Task is starting.
2024-03-13 01:11:48.081 [1710272507865687/start/1 (pid 85339)] alpha is 5.000000
2024-03-13 01:11:53.112 [1710272507865687/start/1 (pid 85339)] Let me sleep for 5 seconds
2024-03-13 01:11:53.114
[1710272507865687/start/1 (pid 85339)] Task finished successfully.
2024-03-13 01:11:53.121 [1710272507865687/end/2 (pid 85381)] Task is starting.
2024-03-13 01:11:53.359 [1710272507865687/end/2 (pid 85381)] alpha is still 5.000000
2024-03-13 01:11:53.393 [1710272507865687/end/2 (pid 85381)] Task finished successfully.
2024-03-13 01:11:53.394 Done!
True
```

metaflow/cli.py
```diff
@@ -842,17 +853,19 @@ def run(
...
 
-def write_run_id(run_id_file, run_id):
-    if run_id_file is not None:
-        with open(run_id_file, "w") as f:
-            f.write(str(run_id))
+def write_file(file_path, content):
+    if file_path is not None:
+        with open(file_path, "w") as f:
+            f.write(str(content))
+            f.close()
```
romain-intel, line 868: Not needed. File is closed when exiting context.

metaflow/subprocess_manager.py
```diff
@@ -0,0 +1,125 @@
...
+                    raise Exception("Ran into an issue...")
+
+    async def consume_queue(self, queue: Queue):
+        while True:
+            item = await queue.get()
+            # break out of loop when we get the `indicator`
+            if item is None:
+                break
+            print(item)
+            queue.task_done()
```
romain-intel, line 65: probably also need before the break.

metaflow/subprocess_manager.py
```diff
@@ -0,0 +1,125 @@
...
+        self.stderr_queue = Queue()
+
+        try:
+            # returns when subprocess has been started, not
+            # when it is finished...
+            self.process = await asyncio.create_subprocess_exec(
+                *command,
+                cwd=self.cwd,
+                env=self.env,
+                stdout=await aiofiles.open(stdout_logfile, "w"),
```
romain-intel, line 82: We probably don't need to be that async...

metaflow/subprocess_manager.py
```diff
@@ -0,0 +1,125 @@
+import os
```
romain-intel, line 1: I am not quite sure the log management is exactly what we want. You seem to be:
- streaming logs to a file
- reading from those files to a queue
- and then streaming all the queue out at once

Granted, it only streams up to that point but it feels a little complex. A better approach may be to read the file nad remember where you are when reading so you can then read it again starting at the right location and continuing. That would enable use to have a cheaper streaming with a lot less overhead in terms of memory.

romain-intel (commented):
Just looked at the process manager for now.

metaflow/metaflow_runner.py
```diff
@@ -0,0 +1,67 @@
...
+            process = await self.spm.run_command([sys.executable, *command.split()])
+
+            if blocking:
+                await process.wait()
+
+            flow_name = await read_from_file_when_ready(tfp_flow.name)
+            run_id = await read_from_file_when_ready(tfp_run_id.name)
+
+            pathspec_components = (flow_name, run_id)
+            run_object = Run("/".join(pathspec_components), _namespace_check=False)
```
romain-intel, line 60: Should be some object like ExecutingRun or something like that and should be independently accessible outside of the runner.

metaflow/click_api.py
```diff
@@ -0,0 +1,362 @@
...
+                if isinstance(obj, Parameter):
+                    parameters.append(obj)
+
+    return parameters
+
+
+class MetaflowAPI(object):
+    def __init__(self, parent=None, **kwargs):
+        self._parent = parent
+        self._chain = [{self._API_NAME: kwargs}]
```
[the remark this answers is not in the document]
madhur-ob: Yes, we do it here: `to_return(parent=None, **method_params)` where `to_return` is 

```
class_dict = {"__module__": "metaflow", "_API_NAME": flow_file}
to_return = type(flow_file, (MetaflowAPI,), class_dict)
```

metaflow/click_api.py
```diff
@@ -0,0 +1,362 @@
+import inspect
+import importlib
+import itertools
```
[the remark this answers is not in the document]
madhur-ob: showcased the demo, resolving...

metaflow/subprocess_manager.py
```diff
@@ -0,0 +1,150 @@
...
+            if stream is None:
+                tasks.append(asyncio.create_task(self.process.wait()))
+            else:
+                tasks.append(asyncio.create_task(self.get_logs(stream)))
+
+            await asyncio.wait(tasks, return_when="FIRST_COMPLETED")
+
+        await self.cleanup()
+
+    async def run_command(self, command: List[str]):
```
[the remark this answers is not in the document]
madhur-ob: This is not an issue because we always do `process = await spm.run_command(cmd)`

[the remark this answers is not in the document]
madhur-ob: and even if `asyncio.create_subprocess_exec` has an issue, we would never return `self.process`

thus stalling on `process = await spm.run_command(cmd)`

metaflow/subprocess_manager.py
```diff
@@ -0,0 +1,150 @@
...
+            self.log_files["stdout"] = stdout_logfile
+            self.log_files["stderr"] = stderr_logfile
+
+            self.run_command_called = True
+            return self.process
+        except Exception as e:
+            print(f"Error starting subprocess: {e}")
+            await self.cleanup()
+
+    async def stream_logs(self, stream):
```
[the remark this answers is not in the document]
madhur-ob: We `yield` the lines and positions now..

metaflow/subprocess_manager.py
```diff
@@ -0,0 +1,150 @@
...
+        if self.run_command_called is False:
+            raise ValueError("No command run yet to get the logs for...")
+
+        if stream not in self.log_files:
+            raise ValueError(f"No log file found for {stream}")
+
+        log_file = self.log_files[stream]
+
+        with open(log_file, mode="r") as f:
```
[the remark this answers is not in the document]
madhur-ob: We yield the `lines` and `positions` now...

metaflow/subprocess_manager.py
```diff
@@ -0,0 +1,150 @@
...
+        else:
+            tasks = [asyncio.create_task(asyncio.sleep(timeout))]
+            if stream is None:
+                tasks.append(asyncio.create_task(self.process.wait()))
+            else:
+                tasks.append(asyncio.create_task(self.get_logs(stream)))
+
+            await asyncio.wait(tasks, return_when="FIRST_COMPLETED")
+
+        await self.cleanup()
```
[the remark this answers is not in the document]
madhur-ob: this is fixed, we clean up if context manager is being used. Else, user has to do it explicitly.

metaflow/click_api.py
```diff
@@ -0,0 +1,362 @@
...
+                for _, v in args.items():
+                    if isinstance(v, list):
+                        for i in v:
+                            components.append(i)
+                    else:
+                        components.append(v)
+                for k, v in options.items():
+                    if isinstance(v, list):
+                        for i in v:
+                            components.append(f"--{k} {i}")
```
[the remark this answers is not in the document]
madhur-ob: Done..

metaflow/click_api.py
```diff
@@ -0,0 +1,362 @@
...
+                    else:
+                        components.append(f"--{k} {v}")
+
+        return " ".join(components)
```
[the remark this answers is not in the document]
madhur-ob: Done...

metaflow/click_api.py
```diff
@@ -0,0 +1,362 @@
...
+        final_chain = list(itertools.chain.from_iterable([p.chain for p in parents]))
+        final_chain.extend(self.chain)
+
+        components = []
+        for each_cmd in final_chain:
+            for cmd, params in each_cmd.items():
+                components.append(cmd)
+                args = params.pop("args", {})
+                options = params.pop("options", {})
+                assert len(params) == 0
```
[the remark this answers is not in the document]
madhur-ob: this is done..

metaflow/click_api.py
```diff
@@ -0,0 +1,362 @@
+import inspect
```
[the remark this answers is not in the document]
madhur-ob: added in most places.

metaflow/metaflow_runner.py
madhur-ob: this is done

madhur-ob:
The latest usage looks like this:

```
import time
import asyncio
from metaflow import metaflow_runner

async def main():
    async with metaflow_runner.Runner('../try.py',  metadata="local") as runner:
        # returns immediately
        result = await runner.run(alpha=5, tags=["abc", "def"], max_workers=5)

        # can access properties of metaflow's Run object..
        print(result.finished)
        while not result.finished:
            print("waiting for run to finish...")
            time.sleep(1)
        print(result.successful)

        # can access properties of the CommandManager as well..
        print(result.log_files)

        await result.wait()
        await result.wait(timeout=2)
        await result.wait(stream="stderr")
        await result.wait(stream="stderr", timeout=3)

        async for _, line in result.stream_logs(stream="stdout"):
            print(line)

if __name__ == "__main__":
    asyncio.run(main())
```

metaflow/subprocess_manager.py
```diff
@@ -0,0 +1,278 @@
...
+        command_id = hash_command_invocation(command)
+        self.commands[command_id] = CommandManager(command, env, cwd)
+        await self.commands[command_id].run()
+        return command_id
+
+    def get(self, command_id: str) -> "CommandManager":
+        return self.commands.get(command_id, None)
+
+    async def cleanup(self):
+        for _, v in self.commands.items():
```
romain-intel, line 44: nit: for v in self.commands.values()

metaflow/subprocess_manager.py
```diff
@@ -0,0 +1,278 @@
...
+
+        if cwd is None:
+            cwd = os.getcwd()
+        self.cwd = cwd
+
+        self.process = None
+        self.run_called = False
+        self.log_files = {}
+
+        signal.signal(signal.SIGINT, self.handle_sigint)
```
romain-intel, line 64: This would technically register this multiple times. I don't think this is supported or what you want. I am not sure how best to handle multiple commands though.

metaflow/subprocess_manager.py
```diff
@@ -0,0 +1,278 @@
...
+        asyncio.create_task(self.kill())
+
+    async def wait(self, timeout=None, stream=None):
+        if timeout is None:
+            if stream is None:
+                await self.process.wait()
+            else:
+                await self.emit_logs(stream)
+        else:
+            tasks = [asyncio.create_task(asyncio.sleep(timeout))]
```
romain-intel, line 83: way of doing the same thing at line 143 seems cleaner :)

metaflow/subprocess_manager.py
```diff
@@ -0,0 +1,278 @@
...
+                # attempting to read it. This is not a problem, but improves readability.
+                await asyncio.sleep(log_write_delay)
+
+                try:
+                    if timeout_per_line is None:
+                        line = f.readline()
+                    else:
+                        line = await asyncio.wait_for(f.readline(), timeout_per_line)
+                except asyncio.TimeoutError as e:
+                    raise LogReadTimeoutError(
```
romain-intel, line 145: we could also return the position and None?

romain-intel: So we would return:
- the actual line
- an empty line if EOF or just an actual empty line
- None if a timeout.

I don't know if we want to distinguish the cases in 2.

metaflow/subprocess_manager.py
```diff
@@ -0,0 +1,278 @@
...
+    def get(self, command_id: str) -> "CommandManager":
+        return self.commands.get(command_id, None)
+
+    async def cleanup(self):
+        for _, v in self.commands.items():
+            await v.cleanup()
+
+
+class CommandManager(object):
+    def __init__(self, command: List[str], env=None, cwd=None):
```
romain-intel, line 49: Type hints and doc strings for all user-exposed code.

metaflow/metaflow_runner.py
```diff
@@ -0,0 +1,83 @@
...
+        return content
+
+
+class ExecutingRun(object):
+    def __init__(self, command_obj: CommandManager, run_obj: Run) -> None:
+        self.command_obj = command_obj
+        self.run_obj = run_obj
+
+    def __getattr__(self, name: str):
+        if hasattr(self.run_obj, name):
```
romain-intel, line 28: need to think a bit more about this. We may want a more explicit control to make sure we expose what we want. It's also a bit harder to document this way.

romain-intel (commented):
Initial comments. WIll need a bit more time but looks promising :)

metaflow/subprocess_manager.py
madhur-ob: done

metaflow/subprocess_manager.py
madhur-ob: let's say if you press CTRL+C on the main terminal while using the subprocess manager to run 2 commands -- both of them will terminate.

However, if you specifically do `kill -SIGINT PID` where PID is for any of the process we spawned out of the above two -- only that particular process will be killed.

This is the current behaviour. As we can see, only the process pertaining to stream B with alpha as 20 gets killed.

Screenshot 2024-04-17 at 2 26 11 AM

metaflow/subprocess_manager.py
madhur-ob: thanks, done

metaflow/subprocess_manager.py
madhur-ob: done

metaflow/subprocess_manager.py
madhur-ob: can discuss for more clarity...

metaflow/metaflow_runner.py
madhur-ob: can discuss more in the meeting...

metaflow/subprocess_manager.py
madhur-ob: resolving this, but can demo in the meeting...

madhur-ob:
New Higher Level UX:

```
import time
import asyncio
from metaflow import metaflow_runner

if __name__ == "__main__":

    # sync version that calls run() directly
    with metaflow_runner.Runner('../try.py', metadata="local").run(alpha=5, tags=["abc", "def"], max_workers=5) as result:
        print(result.run.finished)
        print(result.stdout)

    # sync version using runner
    with metaflow_runner.Runner('../try.py', metadata="local") as runner:
        result = runner.run(alpha=5, tags=["abc", "def"], max_workers=5)
        print(result.run.finished)
        print(result.stdout)

    # async version that calls async_run() directly
    async def async_run_directly():
        with await metaflow_runner.Runner('../try.py', metadata="local").async_run(alpha=5, tags=["abc", "def"], max_workers=5) as result:
            print(result.run.data)
            print(result.run.finished)

            print(result.stdout)
            time.sleep(3)
            print(result.stdout)

            print((await result.wait()).stdout)
            # OR we can also do:
            #    await result.wait()
            #    print(result.stdout)

            while not result.run.finished:
                print("waiting for run to finish...")
                time.sleep(1)
            print(result.run.successful)

            print((await result.wait(timeout=3, stream="stdout")).run.data)
            print(result.run.finished)

            await result.wait(timeout=3, stream="stdout")

            async for _, line in result.stream_logs(stream="stdout"):
                print(line)

    asyncio.run(async_run_directly())

    # async version using runner
    async def async_runner():
        async with metaflow_runner.Runner('../try.py', metadata="local") as runner:
            result = await runner.async_run(alpha=5, tags=["abc", "def"], max_workers=5)

            print(result.run.data)
            print(result.run.finished)

            print(result.stdout)
            time.sleep(3)
            print(result.stdout)

            print((await result.wait()).stdout)
            # OR we can also do:
            #    await result.wait()
            #    print(result.stdout)

            while not result.run.finished:
                print("waiting for run to finish...")
                time.sleep(1)
            print(result.run.successful)

            print((await result.wait(timeout=3, stream="stdout")).run.data)
            print(result.run.finished)

            await result.wait(timeout=3, stream="stdout")

            async for _, line in result.stream_logs(stream="stdout"):
                print(line)

    asyncio.run(async_runner())
```

metaflow/subprocess_manager.py
```diff
@@ -0,0 +1,330 @@
...
+        _, alive = psutil.wait_procs(
+            children, timeout=termination_timeout, callback=on_terminate
+        )
+        for p in alive:
+            p.kill()
+    except psutil.NoSuchProcess:
+        pass
+
+
+class LogReadTimeoutError(Exception):
```
romain-intel, line 29: If we expect the user to catch this, maybe move to exceptions.py?

metaflow/subprocess_manager.py
```diff
@@ -0,0 +1,330 @@
...
+    async def __aenter__(self) -> "CommandManager":
+        return self
+
+    async def __aexit__(self, exc_type, exc_value, traceback):
+        self.cleanup()
+
+    def handle_sigint(self, signum, frame):
+        """Handle the SIGINT signal."""
+
+        print("SIGINT received.")
```
romain-intel, line 101: Do we still want the print?

metaflow/subprocess_manager.py
```diff
@@ -0,0 +1,330 @@
...
+        pass
+
+
+class LogReadTimeoutError(Exception):
+    """Exception raised when reading logs times out."""
+
+    pass
+
+
+class SubprocessManager(object):
```
romain-intel, line 35: If we are exposing this to user, please add docstrings. I am not sure we are though in which case we can add to make clean but not as mandatory.

metaflow/subprocess_manager.py
```diff
@@ -0,0 +1,330 @@
...
+            try:
+                if stream is None:
+                    await asyncio.wait_for(self.process.wait(), timeout)
+                else:
+                    await asyncio.wait_for(self.emit_logs(stream), timeout)
+            except asyncio.TimeoutError:
+                command_string = " ".join(self.command)
+                await self.kill()
+                print(
+                    "Timeout: The process: %s with command: '%s' didn't complete within %s seconds."
```
romain-intel, line 124: nit: "the process (PID %d; command: '%s') did not complete ..."

metaflow/subprocess_manager.py
```diff
@@ -0,0 +1,330 @@
...
+    async def emit_logs(self, stream: str = "stdout", custom_logger: Callable = print):
+        """Helper function to iterate over stream_logs"""
+
+        async for _, line in self.stream_logs(stream):
+            custom_logger(line)
+
+    def cleanup(self):
+        """Clean up log files for a running subprocesses."""
+
+        if hasattr(self, "temp_dir"):
```
romain-intel, line 220: nit: I would set self.temp_dir to None in __init__ and make this a slightly clearer: `if self.temp_dir:` You can also probably use self.run_called.

metaflow/subprocess_manager.py
```diff
@@ -0,0 +1,330 @@
...
+                await self.kill()
+                print(
+                    "Timeout: The process: %s with command: '%s' didn't complete within %s seconds."
+                    % (self.process.pid, command_string, timeout)
+                )
+
+    async def run(self):
+        """Run the subprocess, streaming the logs to temporary files"""
+
+        self.temp_dir = tempfile.mkdtemp()
```
romain-intel, line 131: we may want to prevent multiple calls of `run`. Maybe check on `self.run_called`

metaflow/subprocess_manager.py
```diff
@@ -0,0 +1,330 @@
...
+    async def stream_logs(
+        self,
+        stream: str,
+        position: Optional[int] = None,
+        timeout_per_line: Optional[float] = None,
+        log_write_delay: float = 0.01,
+    ):
+        """Stream logs from the subprocess using the log files"""
+
+        if self.run_called is False:
```
romain-intel, line 164: nit: `not self.run_called`

metaflow/subprocess_manager.py
```diff
@@ -0,0 +1,330 @@
...
+            raise ValueError("No command run yet to get the logs for...")
```
romain-intel, line 165: maybe `RuntimeError`. Not hugely important.

metaflow/subprocess_manager.py
```diff
@@ -0,0 +1,330 @@
...
+
+        if stream not in self.log_files:
+            raise ValueError(
+                "No log file found for '%s', valid values are: %s"
+                % (stream, list(self.log_files.keys()))
```
romain-intel, line 170: nit: `join(", ", self.log_files.keys())`

metaflow/subprocess_manager.py
```diff
@@ -0,0 +1,330 @@
...
+                env=self.env,
+                stdout=open(stdout_logfile, "w"),
+                stderr=open(stderr_logfile, "w"),
+            )
+
+            self.log_files["stdout"] = stdout_logfile
+            self.log_files["stderr"] = stderr_logfile
+
+            self.run_called = True
+            return self.process
```
romain-intel, line 150: we seem to only use the PID where we call this. Can we just return that?

metaflow/subprocess_manager.py
```diff
@@ -0,0 +1,330 @@
...
+                    # will be written to it
+                    if self.process.returncode is not None:
+                        break
+                    # or the process is still running and more logs could be written to
+                    # the file, in which case we continue reading the log file
+                    else:
+                        continue
+
+                position = f.tell()
+                yield position, line.strip()
```
romain-intel, line 209: I would not do a strip. Maybe rstrip but indentation may be relevant to the user.

metaflow/click_api.py
```diff
@@ -0,0 +1,416 @@
+import sys
+
+if sys.version_info < (3, 8):
+    raise RuntimeError(
+        """
+        The Metaflow Programmatic API requires 'typeguard', which is not supported for python 3.5, 3.6 and 3.7.
```
romain-intel, line 6: We can definitely still support 3.7 unless we use some very weird stuff.

metaflow/click_api.py
```diff
@@ -0,0 +1,416 @@
...
+    possible_params = OrderedDict()
+    possible_params.update(possible_arg_params)
+    possible_params.update(possible_opt_params)
+
+    # supplied kwargs
+    for supplied_k, supplied_v in kwargs.items():
+        if supplied_k not in possible_params:
+            raise ValueError(
+                "Unknown argument: '%s', possible args are: %s"
+                % (supplied_k, list(possible_params.keys()))
```
romain-intel, line 83: same, comma separated.

metaflow/_vendor/vendor_any.txt
```diff
@@ -1,2 +1,3 @@
 click==7.1.2
-packaging==23.0
\ No newline at end of file
+packaging==23.0
+typeguard==4.2.1
```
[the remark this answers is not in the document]
madhur-ob: Thanks, using 4.0.1 for typeguard.

metaflow/_vendor/vendor_v3_8.txt
```diff
@@ -0,0 +1,2 @@
+importlib_metadata>=3.6
```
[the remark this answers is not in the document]
madhur-ob: removed

metaflow/metaflow_runner.py
```diff
@@ -0,0 +1,137 @@
...
+        with open(self.command_obj.log_files.get("stderr"), "r") as fp:
+            return fp.read()
+
+    async def stream_logs(self, stream: str, position: Optional[int] = None):
+        async for position, line in self.command_obj.stream_logs(stream, position):
+            yield position, line
+
+
+class Runner(object):
+    def __init__(
```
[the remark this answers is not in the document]
madhur-ob: thanks

metaflow/metaflow_runner.py
```diff
@@ -0,0 +1,137 @@
...
+            try:
+                await asyncio.wait_for(command_obj.process.wait(), timeout=0.5)
+            except asyncio.TimeoutError:
+                pass
+
+            # if the returncode is None, the process has encountered no error within the
+            # initial 0.5 seconds and we proceed to run it in the background
+            # during which it would have written to the run_id and flow_name files
+            if command_obj.process.returncode is not None:
+                stderr_log = open(command_obj.log_files["stderr"]).read()
```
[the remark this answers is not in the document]
madhur-ob: thanks

metaflow/metaflow_runner.py
```diff
@@ -0,0 +1,137 @@
...
+                command = " ".join(command_obj.command)
+                raise RuntimeError(f"Error executing: '{command}':\n\n{stderr_log}")
```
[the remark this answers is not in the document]
madhur-ob: done

metaflow/cli.py
```diff
@@ -677,6 +677,13 @@ def common_run_options(func):
         type=str,
         help="Write the ID of this run to the file specified.",
     )
+    @click.option(
```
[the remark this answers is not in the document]
madhur-ob: done

metaflow/metaflow_runner.py
```diff
@@ -0,0 +1,137 @@
...
+    def stdout(self):
+        with open(self.command_obj.log_files.get("stdout"), "r") as fp:
+            return fp.read()
+
+    @property
+    def stderr(self):
+        with open(self.command_obj.log_files.get("stderr"), "r") as fp:
+            return fp.read()
+
+    async def stream_logs(self, stream: str, position: Optional[int] = None):
```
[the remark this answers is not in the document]
madhur-ob: done

metaflow/metaflow_runner.py
```diff
@@ -0,0 +1,137 @@
...
+def read_from_file_when_ready(file_path):
+    with open(file_path, "r") as file_pointer:
+        content = file_pointer.read()
+        while not content:
+            time.sleep(0.1)
+            content = file_pointer.read()
+        return content
+
+
+class ExecutingRun(object):
```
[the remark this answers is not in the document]
madhur-ob: done

metaflow/subprocess_manager.py
madhur-ob: No, we don't. There is a very rare chance that this occurs. And in the event that it does, user should know what happened.

metaflow/subprocess_manager.py
madhur-ob: removed

metaflow/subprocess_manager.py
madhur-ob: done

metaflow/subprocess_manager.py
madhur-ob: done

metaflow/subprocess_manager.py
madhur-ob: done

metaflow/click_api.py
madhur-ob: using typeguard 4.0.1

metaflow/subprocess_manager.py
madhur-ob: done

metaflow/click_api.py
madhur-ob: done

metaflow/subprocess_manager.py
madhur-ob: done

metaflow/subprocess_manager.py
madhur-ob: done

metaflow/subprocess_manager.py
madhur-ob: using `self.run_called`

metaflow/subprocess_manager.py
madhur-ob: done

metaflow/subprocess_manager.py
madhur-ob: they are already present, but we can improve them in a different PR

metaflow/metaflow_runner.py
```diff
@@ -0,0 +1,137 @@
...
+import time
+import asyncio
+import tempfile
+from typing import Dict, Optional
+from metaflow import Run
+from metaflow.subprocess_manager import SubprocessManager, CommandManager
+
+
+def read_from_file_when_ready(file_path):
+    with open(file_path, "r") as file_pointer:
```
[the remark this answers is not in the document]
madhur-ob: done, added a timeout

savingoyal (approved):

metaflow/click_api.py
```diff
@@ -0,0 +1,410 @@
+import sys
+
+if sys.version_info < (3, 7):
+    raise RuntimeError(
+        """
+        The Metaflow Programmatic API requires 'typeguard', which is not supported for python 3.5 and 3.6.
```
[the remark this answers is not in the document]
madhur-ob: Changing...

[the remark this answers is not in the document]
madhur-ob: Done

PR #1733, opened 2024-02-15
[Ready for review] Add explicit max length control for foreach value.

darinyu:
Add a length control for final result as string. This will prevent us from generating a list of foreach value with crazy long value.

### DIFF
--- a/metaflow/flowspec.py
+++ b/metaflow/flowspec.py
@@ -519,6 +519,7 @@ class FlowSpec(object):
         str
             The value to use for the item.
         """
+        MAXIMUM_VALUE_CHARACTERS = 30
 
         def _is_primitive_type(item):
             return (
@@ -529,7 +530,7 @@ class FlowSpec(object):
             )
 
         value = item if _is_primitive_type(item) else reprlib.Repr().repr(item)
-        return basestring(value)
+        return basestring(value)[:MAXIMUM_VALUE_CHARACTERS]
 
     def next(self, *dsts: Callable[..., None], **kwargs) -> None:
         """
--- a/metaflow/flowspec.py
+++ b/metaflow/flowspec.py
@@ -17,7 +17,7 @@ from .exception import (
 )
 from .graph import FlowGraph
 from .unbounded_foreach import UnboundedForeachInput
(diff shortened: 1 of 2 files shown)


metaflow/flowspec.py
```diff
@@ -529,7 +530,7 @@ def _is_primitive_type(item):
             )
 
         value = item if _is_primitive_type(item) else reprlib.Repr().repr(item)
-        return basestring(value)
+        return basestring(value)[:MAXIMUM_VALUE_CHARACTERS]
```
savingoyal, line 533: Repr already has a default limit of 30 already - I imagine this change only affects primitives right now. I would recommend setting the value of `maxstring` to `MAX_CHARS` so that if the value of `MAX_CHARS` is modified in the future - you can expect a consistent response if `repr` is in action.

savingoyal (changes requested):

metaflow/flowspec.py
darinyu: Repr doc is ambiguous, the `maxstring=30` for Repr means that the string inside the represented result is <=30 (it doesn't truncate the entire representation to maxstring.  for example, 

```
reprlib.Repr().repr({"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa": "bbbbbbbbbbbbbbbbbbbbbbb"})
"{'aaaaaaaaaaaa...aaaaaaaaaaaaa': 'bbbbbbbbbbbbbbbbbbbbbbb'}"
```

metaflow/flowspec.py
```diff
@@ -519,6 +519,7 @@ def _get_foreach_item_value(self, item: Any):
         str
             The value to use for the item.
         """
+        MAXIMUM_VALUE_CHARACTERS = 30
```
romain-intel, line 522: As discussed, I would move this to config and keep the default to 30 but internally we can set to 25. Apart from that, LGTM.

romain-intel (dismissed):

metaflow/flowspec.py
darinyu: Added the config.

romain-intel (approved):
LGTM

savingoyal (approved):

PR #1736, opened 2024-02-15
fix: update list of reserved parameter names

### LINKED ISSUE #1735: Using `tags` as a Parameter name breaks flow.
original discussion here: https://outerbounds-community.slack.com/archives/C02116BBNTU/p1707942631741849

cause is due to the calls to `obj.flow._set_constants(obj.graph, kwargs)` in `cli.py`, as `tags` is an argument for the function, so it is not a part of the trailing kwargs that is passed to `_set_constants`.

List of affected (restricted) parameter names that are not covered, and are affected by the same bug:

```python
    # Not Covered
    tags = Parameter("tags")
    decospecs = Parameter("decospecs")
    run_id_file = Parameter("run_id_file")
    maxworkers = Parameter("max_workers")
    maxnumsplits = Parameter("max_num_splits")
    maxlogsize = Parameter("max_log_size")
    user_namespace = Parameter("user_namespace")
    obj = Parameter("obj")
```

saikonen:
updated list of reserved parameter names.

closes #1735

### DIFF
--- a/metaflow/parameters.py
+++ b/metaflow/parameters.py
@@ -289,26 +289,8 @@ class Parameter(object):
 
         # TODO: check that the type is one of the supported types
         param_type = self.kwargs["type"] = self._get_type(kwargs)
-        reserved_params = [
-            "params",
-            "with",
-            "max-num-splits",
-            "max_num_splits",
-            "max-workers",
-            "max_workers",
-            "tag",
-            "run-id-file",
-            "run_id_file",
-            "namespace",
-            "obj",
-            "decospecs",
-            "max-log-size",
-            "max_log_size",
-            "tags",
-            "user_namespace",
-        ]
 
-        if self.name in reserved_params:
+        if self.name in _reserved_params():
             raise MetaflowException(
                 "Parameter name '%s' is a reserved "
                 "word. Please use a different "
@@ -381,6 +363,28 @@ class Parameter(object):
     # which may do self.param['foobar']
     def __getitem__(self, x):
         pass
+
+
+def _reserved_params():
+    reserved_params = [
+        "params",
+        "with",
+        "tag",
+        "namespace",
+        "obj",
+        "tags",
+        "decospecs",
+        "run-id-file",
+        "max-num-splits",
+        "max-workers",
+        "max-log-size",
+        "user-namespace",
+    ]
+    reserved = set(reserved_params)
+    # due to the way Click maps cli args to function args we also want to add underscored params to the set
+    for param in reserved_params:
+        reserved.add(param.replace("-", "_"))
+    return reserved
 
 
 def add_custom_parameters(deploy_mode=False):
--- a/metaflow/parameters.py
+++ b/metaflow/parameters.py
@@ -302,6 +302,7 @@ class Parameter(object):
             "namespace",
             "obj",
             "decospecs",
+            "max-log-size",
             "max_log_size",
             "tags",
             "user_namespace",
--- a/metaflow/parameters.py
+++ b/metaflow/parameters.py
@@ -293,10 +293,18 @@ class Parameter(object):
             "params",
             "with",
             "max-num-splits",
+            "max_num_splits",
             "max-workers",
+            "max_workers",
             "tag",
             "run-id-file",
+            "run_id_file",
             "namespace",
+            "obj",
+            "decospecs",
+            "max_log_size",
+            "tags",
+            "user_namespace",
         ]
 
         if self.name in reserved_params:
--- a/metaflow/parameters.py
+++ b/metaflow/parameters.py
@@ -290,7 +290,26 @@ class Parameter(object):
         # TODO: check that the type is one of the supported types
         param_type = self.kwargs["type"] = self._get_type(kwargs)
 
-        if self.name in _reserved_params():
+        reserved_params = [
+            "params",
+            "with",
+            "tag",
+            "namespace",
+            "obj",
+            "tags",
+            "decospecs",
+            "run-id-file",
+            "max-num-splits",
+            "max-workers",
+            "max-log-size",
+            "user-namespace",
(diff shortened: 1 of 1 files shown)


metaflow/parameters.py
```diff
@@ -293,10 +293,19 @@ def __init__(
...
             "tag",
             "run-id-file",
+            "run_id_file",
             "namespace",
+            "obj",
+            "decospecs",
+            "max-log-size",
+            "max_log_size",
+            "tags",
+            "user_namespace",
```
savingoyal, line 308: should we also have `user-namespace`? it might be easier if `reserved_params` is a set that adds the dashed versions of all its member by itself so that it is easy to maintain longer term.

metaflow/parameters.py
```diff
@@ -293,10 +293,19 @@ def __init__(
...
+            "max_workers",
             "tag",
             "run-id-file",
+            "run_id_file",
             "namespace",
+            "obj",
+            "decospecs",
+            "max-log-size",
+            "max_log_size",
+            "tags",
```
romain-intel, line 307: hum -- we should maybe discuss this. The others were parameters to run but now these are parameters to other parts of the CLI (argo in this case). I haven't checked if this is used but it is not broken right now internally and so I am a bit leary of removing a perfectly functional name because argo has this restriction.

metaflow/parameters.py
romain-intel: Agreed we should standardize _ and -

romain-intel (changes requested):
Needs a bit more discussion since the issue comes from non-core plugins and this is modifying a core component based on that.

metaflow/parameters.py
savingoyal: which restriction?

romain-intel: I mean that "--tags" is not something that is part of https://github.com/Netflix/metaflow/blob/master/metaflow/cli.py. My understanding of this issue is that it only happens because argo has a `--tags` option. The others are typically here because they appear in the aforementioned cli.py.

savingoyal: argo also doesn't have a `--tags` option. actually, I don't think we need to list `tags` and `decospecs` here in this list.

saikonen: here is a test flow to try out, simply comment out any of the parameters and run it locally:
`python TagsParamFlow.py --metadata local --datastore local run`

```python
from metaflow import step, FlowSpec, Parameter

class TagsParamFlow(FlowSpec):
    # Covered
    # run_id_file = Parameter("run-id-file")
    # maxworkers = Parameter("max-workers")
    # maxnumsplits = Parameter("max-num-splits")
    # user_namespace = Parameter("user-namespace")
    
    # Not Covered
    # tags = Parameter("tags")
    # decospecs = Parameter("decospecs")
    # run_id_file = Parameter("run_id_file")
    # maxworkers = Parameter("max_workers")
    # maxnumsplits = Parameter("max_num_splits")
    # maxlogsize = Parameter("max_log_size")
    # maxlogsize = Parameter("max-log-size")
    # user_namespace = Parameter("user_namespace")
    # obj = Parameter("obj")

    @step
    def start(self):
        print("Starting 👋")
        self.next(self.end)

    @step
    def end(self):
        print("Done! 🏁")


if __name__ == "__main__":
    TagsParamFlow()
```

romain-intel: OK -- I take it back and I shouldn't do fly-by reviews of error messages. Anyways, @saikonen is right. "tags" is the problem but not because it is passed to anything but because it is a named argument in `def run` in cli.py. That said, I would be more in favor of changing those internal names than restricting here (clearly --tags proved useful). So maybe we can make the names in `cli.py` (the ones not passed to kwargs) a bit more exotic (maybe even start them with a _) and that way we can reduce the number of names that are excluded.

Thoughts?

saikonen: the issue is coming from `cli.py#common_run_options` that maps click options to specific args. Note that we are not having issues with a cli arg `--tags` (which does not exist) but that multiple `--tag` are mapped to a `tags` argument for functions using the `@common_run_options`

these functions then try to call `obj.flow._set_constants(obj.graph, kwargs)` with their kwargs, which do not include the explicit args anymore

saikonen: ah @romain-intel beat me to it ;)

I'm all for prefixing the internal args with `_` to avoid collisions, though it seems like a bit of a heavy change across the board given that we've only had collisions with `--tags`, and the others seem less likely to be used.

Prefixing would probably require going through a bunch of the cli functions to check for other collisions besides the base `cli.py` entrypoints though?

romain-intel: changes should be limited to cli.py so may not be too heavy a lift. Am fine either ways since it seems it was already "broken" before with a slight preference for prefixing to make them less colliding.

saikonen: been scratching my head over the UX of this a bit. We can prefix the internal args to avoid collisions, but this only applies to the args of the functions, not cli options. This can lead to some funky UX:

User defines a custom parameter for their flow called `tags`. Now they can deploy a flow supplying the options 
```
--tag some_internal_flow_tag --tags values,for,parameter
```

which is why I'm starting to lean towards simply updating the reserved params list for now, and relaxing the restrictions in a future change if this kind of use case starts popping up more.

metaflow/parameters.py
saikonen: done. Lifted the list construction to a separate function

metaflow/parameters.py
```diff
@@ -374,6 +365,28 @@ def __getitem__(self, x):
         pass
 
 
+def _reserved_params():
```
savingoyal, line 368: does this need to be a function?

savingoyal (approved):

PR #1737, opened 2024-02-16
fix: ensure linuxParameters dict entry exists when creating tmpfs

stev-0:
Had some discussion on [metaflow chat](https://outerbounds-community.slack.com/archives/C02116BBNTU/p1708015812312309?thread_ts=1707995984.923579&cid=C02116BBNTU) about an error I was getting on 2.9.7:

File ".../metaflow/plugins/aws/batch/batch_client.py", line 302, in _register_job_definition
job_definition["containerProperties"]["linuxParameters"]["tmpfs"] = [
KeyError: 'linuxParameters'

I couldn't see anything that would fix it in master, so have put in this PR, but would be good if anyone else can confirm the bug in their environment.


stev-0:
Might repurpose this as a docs update, because this isn't applicable to Fargate, so not necessary here.

Issue #1738, opened 2024-02-16
Add option to batch decorator to increase ephemeralStorage on Fargate


stev-0:
It would be good to have an option to vary the ephemeral storage that a batch job needs on Fargate, since the default (20GB) is relatively low for our needs, and there is no other option on Fargate to increase storage (tmpfs is not an option). Further conversation about this is available on the [ask-metaflow chat](https://outerbounds-community.slack.com/archives/C02116BBNTU/p1707995984923579), but essentially the [API is there](https://docs.aws.amazon.com/batch/latest/APIReference/API_ContainerProperties.html)  (was added about[ a year ago](https://aws.amazon.com/about-aws/whats-new/2023/03/aws-batch-configurable-ephemeral-storage-fargate/)) so it just requires wiring up in metaflow.

I am happy to attempt a PR to do this, I'm not super familiar with the code base but on the other hand I don't think it's a huge job because it's a relatively simple parameter.

PR #1739, opened 2024-02-16
feat: add ephemeral_storage parameter for Batch (Fargate-only)

stev-0:
Fixes issue #1738.

Please note I have not been able to test this for real as I don't have the right permissions in our AWS environment yet, but these are the parameters the code gives to the `register_job_definition` call:

`{'type': 'container', 'containerProperties': {'image': 'public.ecr.aws/docker/library/python:3.11', 'jobRoleArn': 'arn:aws:iam::xxxx', 'command': ['echo', 'hello world'], 'resourceRequirements': [{'value': '1', 'type': 'VCPU'}, {'value': '4096', 'type': 'MEMORY'}], 'executionRoleArn': 'arn:aws:iam::xxxx', 'networkConfiguration': {'assignPublicIp': 'ENABLED'}, 'ephemeralStorage': {'sizeInGiB': 50}}, 'propagateTags': True, 'platformCapabilities': ['FARGATE'], 'jobDefinitionName': 'metaflow_xxxx'}`

Also fixes some documentation for the tempfs parameter, and clarifies where this parameter can be used.

### DIFF
--- a/metaflow/plugins/aws/batch/batch.py
+++ b/metaflow/plugins/aws/batch/batch.py
@@ -191,6 +191,7 @@ class Batch(object):
         tmpfs_size=None,
         tmpfs_path=None,
         num_parallel=0,
+        ephemeral_storage=None,
     ):
         job_name = self._job_name(
             attrs.get("metaflow.user"),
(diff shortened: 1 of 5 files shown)


stev-0:
I have access to the correct permissions now, and this gives a batch job with ephemeral permissions as expected when run

romain-intel:

saikonen (approved):

PR #1742, opened 2024-02-19
fix: allow dot files with `--package-suffixes`

### LINKED ISSUE #1741: `--package-suffixes` omits dotfiles
There is an issue when trying to append dotfiles with `--package-suffixes` as it skips any filename that starts with a `.`

*Expected*
`--package-suffixes=.env` should include a file with the name `.env` to the package

saikonen:
allow files that start with . to be added via `--package-suffixes`

closes #1741

### DIFF
--- a/metaflow/package.py
+++ b/metaflow/package.py
@@ -88,8 +88,6 @@ class MetaflowPackage(object):
             # if path and (path[0] == '.' or './' in path):
             #    continue
             for fname in files:
-                if fname[0] == ".":
-                    continue
                 if any(fname.endswith(suffix) for suffix in suffixes):
                     p = os.path.join(path, fname)
                     yield p, p[prefixlen:]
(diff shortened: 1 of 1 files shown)


madhur-ob (approved):

metaflow/package.py
```diff
@@ -88,8 +88,6 @@ def _walk(self, root, exclude_hidden=True, suffixes=None):
             # if path and (path[0] == '.' or './' in path):
             #    continue
             for fname in files:
```
romain-intel, line 90: maybe a slightly safer and less expensive check would be something like:
```
if (fname[0] == "." and fname in suffixes) or (fname[0] != "." and any(fname.endswith(suffix) for suffix in suffixes):
```
This would restrict to including just files like `.env` and not check all random . files.

But am fine with this change as well.

romain-intel (approved):
One possible alternative but mergeable in both cases.

metaflow/package.py
saikonen: going with your suggestion as the initial implementation ended up including arbitrary hidden files with known extensions, e.g. `.hidden.py`

romain-intel: Good point -- I hadn't even thought of that.

PR #1744, opened 2024-02-19
Support @parallel on Kubernetes with jobsets

shrinandj:
This commit adds support for @parallel when flows are run `--with kubernetes` Support for Argo workflows will follow in a separate commit.

A user can run a flow with the following:

    @step
    def start(self):
        self.next(self.parallel_step, num_parallel=3)

    @kubernetes(cpu=1, memory=512)
    @parallel
    @step
    def parallel_step(self):
    ...

Testing Done:
- Ran a flow with @parallel on Kubernetes. Verified that it works correctly
- Ran a flow without @parallel on Kubernetes. Verified that it works as expected.
- Verified that jobsets based @parallel step gets scaled down if user kills it with a Ctrl-C

### DIFF
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -836,6 +836,11 @@ class ArgoWorkflows(object):
     # Visit every node and yield the uber DAGTemplate(s).
     def _dag_templates(self):
         def _visit(node, exit_node=None, templates=None, dag_tasks=None):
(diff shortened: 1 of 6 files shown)


romain-intel:

shrinandj:
There will be more testing done on this PR once https://github.com/Netflix/metaflow/pull/1745 gets merged. Distributed training workloads use an open port communicating with each other.

PR #1756, opened 2024-03-05
pluggable azure credentials provider

oavdeev:
Refactors azure plugin to support pluggable credential providers like we do for AWS


saikonen:
Does this supersede #1661 ?

romain-intel:
This PR is fine from my POV. I'll monitor when it merges so I can strip out all the plugin internally but ow can merge whenever from my side.

PR #1757, opened 2024-03-06
[Ready for review] add an option to make sidecar send msg func threadsafe

darinyu:
Current send_msg is not threadsafe. This PR add an optional to use lock that guards the shared ressource (stdin of the subprocess).

### DIFF
--- a/metaflow/sidecar/sidecar.py
+++ b/metaflow/sidecar/sidecar.py
@@ -13,7 +13,7 @@ class Sidecar(object):
         if t is not None and t.get_worker() is not None:
             self._has_valid_worker = True
         self.sidecar_process = None
-        # Whether to send msg in a thread-safe fasion.
+        # Whether to send msg in a thread-safe fashion.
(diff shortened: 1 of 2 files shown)


metaflow/sidecar/sidecar.py
```diff
@@ -13,14 +13,24 @@ def __init__(self, sidecar_type):
         if t is not None and t.get_worker() is not None:
             self._has_valid_worker = True
         self.sidecar_process = None
+        # Whether to send msg in a thread-safe fasion.
```
madhur-ob, line 16: nitpick: `fasion` --> `fashion`

madhur-ob (approved):

metaflow/sidecar/sidecar.py
darinyu: done

saikonen (approved):
seems good to go, default behavior for sidecar messages stays unchanged so assuming that some extensions are going to be using the threadsafe messaging?

PR #1759, opened 2024-03-12
[Ready for review] make default s3 worker count configurable

darinyu:
No-op to current behavior, but configuration on S3 parallel process count at top level.


romain-intel (approved):

metaflow/metaflow_config.py
```diff
@@ -81,6 +81,9 @@
 # so setting it to 0 means each operation will be tried once.
 S3_RETRY_COUNT = from_conf("S3_RETRY_COUNT", 7)
 
+# S3 worker configuration. Number of concurrent S3 connections.
```
romain-intel, line 84: nit: "Number of concurrent S3 processes for parallel operations"

romain-intel (dismissed):

metaflow/metaflow_config.py
darinyu: done

madhur-ob (approved):

PR #1764, opened 2024-03-14
fix format error for py3.5

wangchy27:


.pre-commit-config.yaml
```diff
@@ -12,3 +12,7 @@ repos:
         language_version: python3
         exclude: "^metaflow/_vendor/"
         additional_dependencies: ["click<8.1.0"]
+        # python3.12 is not supported in black 22.10.0
+        args: [-t, py34, -t, py35, -t, py36, -t, py37, -t, py38, -t, py39, -t, py310, -t, py311]
```
romain-intel, line 16: I don't think we need 3.4 but we can leave for now. We will probably need to drop support for some older versions moving forward :)

PR #1765, opened 2024-03-14
Generate stubs properly for extensions with lazy loaded modules

romain-intel:


madhur-ob:
Checked that it works locally

Screenshot 2024-03-18 at 10 27 09 PM

But we do get some squiggly lines on import:

Screenshot 2024-03-18 at 10 28 41 PM

with the following issue:

Screenshot 2024-03-18 at 10 44 46 PM

But, I think those can be resolved later and we can still merge this?

madhur-ob:
Also works for: `from metaflow import get_aws_client`

Screenshot 2024-03-19 at 1 15 45 AM

PR #1766, opened 2024-03-14
fix bug where client can not access foreach stack

darinyu:
To reproduce, have a flow with foreach task, and then  In python client, do
```
>>> task = Task("ForeachFlow/123/foreach_step/task-00000000")
>>> tp.index
  File "<stdin>", line 1, in <module>
  File "/root/metaflow/metaflow/client/core.py", line 1166, in index
    return self["_foreach_stack"].data[-1].index
  File "/root/metaflow/metaflow/client/core.py", line 875, in data
    obj = filecache.get_artifact(ds_type, location[6:], meta, *components)
  File "/root/metaflow/metaflow/client/filecache.py", line 207, in get_artifact
    _, obj = next(
  File "/root/metaflow/metaflow/datastore/task_datastore.py", line 370, in load_artifacts
    yield name, pickle.loads(blob)
AttributeError: Can't get attribute 'ForeachFrame' on <module '__main__' (built-in)>
```

This is because this ForeachFrame is loaded to __main__, and defined in `task.py`. Without a flow, this ForeachFrame is not imported.  This PR refactors the namedtuple util to top level (with minimum dependency) and import it at top level.

### DIFF
--- a/metaflow/__init__.py
+++ b/metaflow/__init__.py
@@ -143,6 +143,9 @@ from .client import (
     DataArtifact,
 )
 
+# Data class needed for Client.
+from .tuple_util import ForeachFrame
+
 __version_addl__ = []
 _ext_debug("Loading top-level modules")
 for m in _tl_modules:
--- a/metaflow/plugins/datatools/s3/s3.py
+++ b/metaflow/plugins/datatools/s3/s3.py
@@ -21,7 +21,6 @@ from metaflow.metaflow_config import (
     TEMPDIR,
 )
 from metaflow.util import (
-    namedtuple_with_defaults,
     is_stringish,
     to_bytes,
     to_unicode,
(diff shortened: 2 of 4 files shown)


metaflow/__init__.py
```diff
@@ -143,6 +143,9 @@ class and related decorators.
     DataArtifact,
 )
 
+# Data class needed for Client.
+from .tuple_util import ForeachFrame
```
romain-intel, line 147: nit, you probably don't need to import ForeachFrame. You can probably simply do: `from . import tuple_util`

romain-intel (dismissed):
LGTM with minor comment to not introduce a new symbol.

metaflow/__init__.py
darinyu: done

romain-intel (approved):

savingoyal (approved):

romain-intel:
@darinyu -- is this no longer something you want merged? I thought it was still relevant.

romain-intel:
@darinyu for when you are back :)

darinyu:
Yeah this still needs to go in, it slipped during before my vacation.  I just rebased this commit and I think it is ready to go in now.

romain-intel (approved):
I think this looks good. I think we should be good to merge.

Issue #1101, opened 2022-08-03
Make `log-group` and `log-stream` configurable for AWS Batch


savingoyal:
https://outerbounds-community.slack.com/archives/C02116BBNTU/p1659523083731559 via chat.metaflow.org

Limess:
This is important for us, as it's very difficult to attribute cost when someone runs a metaflow flow and logs too much data - all the logs go to `/aws/batch/job`

We've seen single jobs costing $1000+ per job in logs due to too much logging granularity and had difficulty tracking it down.

madhur-ob:
done in https://github.com/Netflix/metaflow/pull/1767

Issue #471, opened 2021-04-07
Expose logConfiguration options for jobs running on AWS Batch


corleyma:
By default, AWS Batch uses the `awslogs` log driver to ship logs for a job container to AWS CloudWatch.  However, this behavior [is configurable](https://docs.aws.amazon.com/batch/latest/APIReference/API_LogConfiguration.html), and even the `awslogs` driver itself [has configurability](https://docs.aws.amazon.com/batch/latest/userguide/using_awslogs.html) that may be of interest to Metaflow users.  

Example motivating usecases:
- specifying the log group or stream prefix may be useful for organizations that have schemes dictating how logs are collected in CloudWatch
- users of the Python standard logging library may benefit from the ability to configure multi-line pattern matching to ensure tracebacks and other multi-line log messages are interpreted as single events in CloudWatch.

The ask here is to expose `awslogs` logging config options, preferably in two ways:
1. as properties settable per step when using the `@batch` decorator
2. and, as is the case for other Batch configuration, as defaults in the Metaflow config profile.

Because Metaflow currently has an implicit dependency on logs making their way to CloudWatch, allowing users to specify log drivers other than `awslogs` may not be feasible today.  Still, this is something worth thinking about for the future.

savingoyal:
Makes total sense. We can add this functionality in after we have published #449.

madhur-ob:
https://github.com/Netflix/metaflow/pull/1767 offers as properties settable per step when using the @batch decorator.
Should we close this?

saikonen:
Let's leave it open for now as there is still the other use case that is missing
- Being able to configure defaults for the logging through Metaflow profile (or env variables set by the exec platform)

Faakhir30:
> Being able to configure defaults for the logging through Metaflow profile (or env variables set by the exec platform)

Looking into this
Working on this issue.

npow:
Closing as `logConfiguration` support was added to AWS Batch job definitions in `batch_client.py`.

PR #1767, opened 2024-03-19
ability to customise log driver and log options

madhur-ob:
We can now do the following:

- This can expect a list where each string in the list follows `key:value`
```
@batch(log_options=["awslogs-group:try/batch/custom/demo"])
@step
def mid(self):
    print("yayyyy")
    self.next(self.end)
```

or perhaps just a single string of the form `key:value`

```
@batch(log_options="awslogs-group:try/batch/custom/demo")
@step
def mid(self):
    print("yayyyy")
    self.next(self.end)
```

Of course the assumption is that the log group `try/batch/custom/demo` should exist first on Cloudwatch...

### DIFF
--- a/metaflow/plugins/aws/batch/batch.py
+++ b/metaflow/plugins/aws/batch/batch.py
@@ -193,6 +193,8 @@ class Batch(object):
         tmpfs_path=None,
         num_parallel=0,
         ephemeral_storage=None,
+        log_driver=None,
+        log_options=None,
     ):
         job_name = self._job_name(
             attrs.get("metaflow.user"),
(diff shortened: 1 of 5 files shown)


madhur-ob:
We can see the logs going into the `log-group` of our choice on Cloudwatch

Screenshot 2024-03-19 at 12 27 02 PM

madhur-ob:
Closes https://github.com/Netflix/metaflow/issues/1101

madhur-ob:
Related: https://github.com/Netflix/metaflow/issues/471

PR #1770, opened 2024-03-23
Fix an issue when unpickling functions through the escape hatch

romain-intel:
When a function was unpickled, the name did not match what was expected.


wangchy27 (approved):
LGTM

Issue #1776, opened 2024-03-27
Invalid CloudWatch log group prefix


stathva:
https://github.com/Netflix/metaflow/blob/82898293ff4f6e66278c51d3ab4a5065b50aba3d/metaflow/plugins/aws/batch/batch_client.py#L214

The default CloudWatch log group prefix above is invalid. AWS reserves log groups starting with `AWS/`. Perhaps change it to `/aws/batch/job`

madhur-ob:
According to https://docs.aws.amazon.com/batch/latest/userguide/using_awslogs.html#create_awslogs_logdriver_options
the default is `aws/batch/job`

Moreover, check the usage here: https://github.com/Netflix/metaflow/pull/1767#issue-2194180976
We supply `try/batch/custom/demo` i.e. without the `/` in beginning and it worked as shown in https://github.com/Netflix/metaflow/pull/1767#issuecomment-2005997732

madhur-ob:
@stathva Did you have issues with running flows with what we have here i.e. `aws/batch/job` without the `/` in beginning?

stathva:
>@stathva Did you have issues with running flows with what we have here i.e. aws/batch/job without the / in beginning?

@madhur-ob yes, we deployy Metflow Wofklows as StepFunctions, our workflow is using the `@batch` decorator to define the container image and step resource requirements but it doesn't specify a log group. 

The jobs failed with the following error:
```
"Reason": "CannotStartContainerError: Error response from daemon: failed to initialize logging driver: failed to create Cloudwatch log stream: InvalidParameterException: Log groups starting with AWS/ are reserved for AWS.",
```

So the problem is not missing `/` in general like in the `try/batch/custom/demo` example but using `aws/*` since its reserved.

madhur-ob:
Aah, I think the fix is to not include it at all I guess...let me remove it, that way, it will consider `aws/batch/job` by default..

simonhs87:
This happened to me now for unknown reasons. The "fix" is to assign log_options=["awslogs-group:/aws/batch/job"] to every batch step in the step functions i create. Using metaflow 2.11.6.

madhur-ob:
@simonhs87 Can you please try with `2.11.9`? There was one more fix made in https://github.com/Netflix/metaflow/pull/1786

PR #1777, opened 2024-03-27
fix log options

madhur-ob:
Closes https://github.com/Netflix/metaflow/issues/1776

If nothing is supplied... it defaults to the previous behaviour and should thus automatically set the correct `log driver` and `log options`

### DIFF
--- a/metaflow/plugins/aws/batch/batch_client.py
+++ b/metaflow/plugins/aws/batch/batch_client.py
@@ -206,14 +206,14 @@ class BatchJob(object):
                 k, v = each_log_option.split(":", 1)
                 log_options_dict[k] = v
 
-        job_definition["containerProperties"]["logConfiguration"] = {
-            "logDriver": log_driver if log_driver is not None else "awslogs",
-            "options": log_options_dict
-            if log_options is not None
-            else {
-                "awslogs-group": "aws/batch/job",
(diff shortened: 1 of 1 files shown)


saikonen (approved):

metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -206,14 +206,16 @@ def _register_job_definition(
                 k, v = each_log_option.split(":", 1)
                 log_options_dict[k] = v
 
-        job_definition["containerProperties"]["logConfiguration"] = {
-            "logDriver": log_driver if log_driver is not None else "awslogs",
-            "options": log_options_dict
-            if log_options is not None
-            else {
-                "awslogs-group": "aws/batch/job",
```
npow, line 214: was there a reason for removing this as the default?

madhur-ob: yeah, please refer to https://github.com/Netflix/metaflow/issues/1776

TLDR: AWS doesn't like when we supply the default values explicitly and not supplying is equivalent to supplying them implicitly anyway...

so it does result into using `aws/batch/job` even if we don't supply it...
while supplying it is not allowed because it is AWS specific and not custom...

PR #1778, opened 2024-03-27
Replace awscli v1 with boto3 during environment bootstrapping

### LINKED ISSUE #387: Use boto3 instead of awscli python module to copy packages

trhodeos:
Fixes #387

### DIFF
--- a/metaflow/plugins/aws/aws_utils.py
+++ b/metaflow/plugins/aws/aws_utils.py
@@ -4,7 +4,6 @@ import requests
 from metaflow.exception import MetaflowException
 
 
-
 def parse_s3_full_path(s3_uri):
     from urllib.parse import urlparse
 
--- a/metaflow/plugins/aws/aws_utils.py
+++ b/metaflow/plugins/aws/aws_utils.py
@@ -4,6 +4,23 @@ import requests
 from metaflow.exception import MetaflowException
 
 
+
+def parse_s3_full_path(s3_uri):
+    from urllib.parse import urlparse
+
+    #  <scheme>://<netloc>/<path>;<params>?<query>#<fragment>
+    scheme, netloc, path, _, _, _ = urlparse(s3_uri)
+    assert scheme == "s3"
+    assert netloc is not None
+
+    bucket = netloc
+    path = path.lstrip("/").rstrip("/")
+    if path == "":
+        path = None
+
+    return bucket, path
+
+
 def get_ec2_instance_metadata():
     """
     Fetches the EC2 instance metadata through AWS instance metadata service
--- a/metaflow/metaflow_environment.py
+++ b/metaflow/metaflow_environment.py
@@ -89,10 +89,21 @@ class MetaflowEnvironment(object):
         It should work silently if everything goes well.
         """
         if datastore_type == "s3":
+            from .plugins.aws.aws_utils import parse_s3_full_path
+
+            bucket, s3_object = parse_s3_full_path(code_package_url)
+            # Explicitly _don't_ use awscli here, because of version incompatibilities
+            # between awscli v1 (installed via pypi) and awscli v2 (installed _not_
+            # through pypi).
+            # Instead, just use boto3 which we already have installed.
             return (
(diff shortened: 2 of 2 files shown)


madhur-ob (approved):
LGTM

trhodeos:
@madhur-ob sorry I had to fix some formatting / precommit issues. When you get the chance, would you mind re-reviewing?

madhur-ob (approved):

trhodeos:
Thanks @madhur-ob .. It looks like some tests failed.. Perhaps they are flaky? 

Also, would it be possible to run the `metaflow.s3-tests`? Seems like this'd be a useful gauge to make sure this works as intended!

metaflow/plugins/aws/aws_utils.py
```diff
@@ -4,6 +4,22 @@
 from metaflow.exception import MetaflowException
 
 
+def parse_s3_full_path(s3_uri):
```
savingoyal, line 7: @trhodeos is there any specific reason to create a utility function? maybe we can have this snippet live right next to L:91 in `metaflow_environment.py`. it will help in making the code much more readable (avoiding an extra jump to a different file)

metaflow/plugins/aws/aws_utils.py
```diff
@@ -4,6 +4,22 @@
...
+    from urllib.parse import urlparse
+
+    #  <scheme>://<netloc>/<path>;<params>?<query>#<fragment>
+    scheme, netloc, path, _, _, _ = urlparse(s3_uri)
+    assert scheme == "s3"
```
savingoyal, line 12: within metaflow code base, we have elected to not use assert statements. also, when the assert statement fails, the user is left no better in their understanding of how to fix the issue

metaflow/plugins/aws/aws_utils.py
```diff
@@ -4,6 +4,22 @@
...
+    assert netloc is not None
+
+    bucket = netloc
+    path = path.lstrip("/").rstrip("/")
+    if path == "":
+        path = None
```
savingoyal, line 18: we don't expect this to be `None`

metaflow/plugins/aws/aws_utils.py
```diff
@@ -4,6 +4,22 @@
...
+def parse_s3_full_path(s3_uri):
+    from urllib.parse import urlparse
+
+    #  <scheme>://<netloc>/<path>;<params>?<query>#<fragment>
+    scheme, netloc, path, _, _, _ = urlparse(s3_uri)
+    assert scheme == "s3"
+    assert netloc is not None
+
+    bucket = netloc
+    path = path.lstrip("/").rstrip("/")
```
savingoyal, line 16: is the `rstrip` needed?

savingoyal (changes requested):

metaflow/plugins/aws/aws_utils.py
trhodeos: This mirrors what we do for azure and gcs. I was just following convention here, happy to pull this inline if you think it's better.

metaflow/plugins/aws/aws_utils.py
trhodeos: This is copied from `parse_gcs_full_path`.. Happy to remove it though if we don't want it!

metaflow/plugins/aws/aws_utils.py
trhodeos: Yep, removing it makes sense for this use case.. I sound like a broken record at this point, but I just copied this wholesale from `parse_gcs_full_path`.

saikonen:
closing as these changes were part of #1972

PR #1783, opened 2024-03-28
fix: support specifying subdirectory with PyPI vcs sources.

saikonen:
add support for `#subdirectory=` as part of a pypi package version.


madhur-ob (approved):
LGTM

PR #1786, opened 2024-03-29
check for falsy values instead of just None

madhur-ob:
Reason: a click option that has `default` as `None` and `multiple` as `True` gives an empty tuple i.e. `()` when the option is not supplied, which is weird, thus the issue


romain-intel:
Merging as this is considered a hotfix by @madhur-ob

PR #1794, opened 2024-04-08
Expose metaflow logger and monitor via singleton

talsperre:
## Expose logger and monitor via a `system_current` singleton

Currently, users don't have the ability to use their implementation of logger and monitor sidecars in their own code. Additionally, if platform developers want to instrument their Metaflow extensions, they have to pass in the logger/monitor constructs to each one of their plugins, leading to code duplication.

This PR exposes two new singleton objects called `_system_logger` and `_system_monitor` that can be used to access the monitor and the logger anywhere.  

- The monitor/logger can be accessed both within and outside a Metaflow flow. This allows us to instrument Metaflow plugins like `metaflow.S3`, which is often used outside of a flow as well. 

### Usage
The monitor/logger sidecar can be used in the following manner:

```python
with _system_monitor.count("<your_metric_name>"): 
      # your code
      pass

_system_logger.log(payload):
```

### DIFF
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -23,6 +23,7 @@ from .metaflow_config import (
     DEFAULT_PACKAGE_SUFFIXES,
 )
 from .metaflow_current import current
+from metaflow.system import _system_monitor, _system_logger
 from .metaflow_environment import MetaflowEnvironment
 from .mflog import LOG_SOURCES, mflog
 from .package import MetaflowPackage
@@ -952,11 +953,13 @@ def start(
         flow=ctx.obj.flow, env=ctx.obj.environment
     )
     ctx.obj.event_logger.start()
+    _system_logger.init_system_logger(ctx.obj.flow.name, ctx.obj.event_logger)
 
     ctx.obj.monitor = MONITOR_SIDECARS[monitor](
         flow=ctx.obj.flow, env=ctx.obj.environment
     )
     ctx.obj.monitor.start()
+    _system_monitor.init_system_monitor(ctx.obj.flow.name, ctx.obj.monitor)
 
     ctx.obj.metadata = [m for m in METADATA_PROVIDERS if m.TYPE == metadata][0](
         ctx.obj.environment, ctx.obj.flow, ctx.obj.event_logger, ctx.obj.monitor
--- a/metaflow/system/system_monitor.py
+++ b/metaflow/system/system_monitor.py
@@ -26,11 +26,17 @@ class SystemMonitor(object):
 
         """
         from metaflow.sidecar import Message, MessageTypes
+
         self._context.update(context)
-        self.monitor.send(Message(MessageTypes.MUST_SEND, {
-            "is_context_updated": True,
-            **self._context,
-        }))
+        self.monitor.send(
+            Message(
+                MessageTypes.MUST_SEND,
+                {
+                    "is_context_updated": True,
+                    **self._context,
+                },
+            )
+        )
 
     def init_system_monitor(
         self, flow_name: str, monitor: "metaflow.monitor.NullMonitor"
--- a/metaflow/system/system_monitor.py
+++ b/metaflow/system/system_monitor.py
@@ -1,0 +1,122 @@
+import os
+import sys
+from ..debug import debug
+from contextlib import contextmanager
+from typing import Optional, Union, Dict, Any
+
+
+class SystemMonitor(object):
+    def __init__(self):
+        self._monitor = None
+        self._flow_name = None
+        self._context = {}
+
+    def __del__(self):
+        if self._flow_name == "not_a_real_flow":
+            self.monitor.terminate()
+
+    def update_context(self, context: Dict[str, Any]):
+        """
+        Update the global context maintained by the system monitor.
+
+        Parameters
+        ----------
+        context : Dict[str, Any]
+            A dictionary containing the context to update.
+
+        """
+        from metaflow.sidecar import Message, MessageTypes
+        self._context.update(context)
+        self.monitor.send(Message(MessageTypes.MUST_SEND, {
+            "is_context_updated": True,
+            **self._context,
+        }))
+
+    def init_system_monitor(
+        self, flow_name: str, monitor: "metaflow.monitor.NullMonitor"
+    ):
+        self._flow_name = flow_name
+        self._monitor = monitor
+
+    def _init_system_monitor_outside_flow(self):
+        from .system_utils import DummyFlow
+        from .system_utils import init_environment_outside_flow
+        from metaflow.plugins import MONITOR_SIDECARS
+        from metaflow.metaflow_config import DEFAULT_MONITOR
+
+        self._flow_name = "not_a_real_flow"
+        _flow = DummyFlow(self._flow_name)
+        _environment = init_environment_outside_flow(_flow)
+        _monitor = MONITOR_SIDECARS[DEFAULT_MONITOR](_flow, _environment)
+        return _monitor
+
+    @property
+    def monitor(self) -> Optional["metaflow.monitor.NullMonitor"]:
+        if self._monitor is None:
+            # This happens if the monitor is being accessed outside of a flow
+            self._debug("Started monitor outside of a flow")
+            self._monitor = self._init_system_monitor_outside_flow()
+            self._monitor.start()
+        return self._monitor
+
+    @staticmethod
+    def _debug(msg: str):
+        """
+        Log a debug message to stderr.
+
+        Parameters
+        ----------
+        msg : str
+            Message to log.
+
+        """
+        if os.environ.get("METAFLOW_DEBUG_SIDECAR", "0") == "1":
+            print("system monitor: %s" % msg, file=sys.stderr)
+
+    @contextmanager
+    def measure(self, name: str):
+        """
+        Context manager to measure the execution duration and counter of a block of code.
+
+        Parameters
+        ----------
+        name : str
+            The name to associate with the timer and counter.
+
+        Yields
+        ------
+        None
+        """
+        # Delegating the context management to the monitor's measure method
+        with self.monitor.measure(name):
+            yield
+
+    @contextmanager
+    def count(self, name: str):
+        """
+        Context manager to increment a counter.
+
+        Parameters
+        ----------
+        name : str
+            The name of the counter.
+
+        Yields
+        ------
+        None
+        """
+        # Delegating the context management to the monitor's count method
+        with self.monitor.count(name):
+            yield
+
+    def gauge(self, gauge: "metaflow.monitor.Gauge"):
+        """
+        Log a gauge.
+
+        Parameters
+        ----------
+        gauge : metaflow.monitor.Gauge
+            The gauge to log.
+
+        """
+        self.monitor.gauge(gauge)
--- a/metaflow/system/system_monitor.py
+++ b/metaflow/system/system_monitor.py
@@ -76,7 +76,11 @@ class SystemMonitor(object):
             Message to log.
 
         """
-        if os.environ.get("METAFLOW_DEBUG_SIDECAR", "0") == "1":
+        if os.environ.get("METAFLOW_DEBUG_SIDECAR", "0").lower() not in (
+            "0",
+            "false",
+            "",
+        ):
             print("system monitor: %s" % msg, file=sys.stderr)
 
     @contextmanager
--- a/metaflow/task.py
+++ b/metaflow/task.py
@@ -4,9 +4,11 @@ import math
 import sys
 import os
 import time
+import traceback
 
 from types import MethodType, FunctionType
 
+from metaflow.sidecar import Message, MessageTypes
 from metaflow.datastore.exceptions import DataException
 
 from .metaflow_config import MAX_ATTEMPTS
@@ -22,6 +24,7 @@ from .unbounded_foreach import UBF_CONTROL
 from .util import all_equal, get_username, resolve_identity, unicode_type
 from .clone_util import clone_task_helper
 from .metaflow_current import current
+from metaflow.system import _system_logger, _system_monitor
 from metaflow.tracing import get_trace_id
 from metaflow.tuple_util import ForeachFrame
 
@@ -280,25 +283,42 @@ class MetaflowTask(object):
                 "task.clone_only needs a valid clone_origin_task value."
             )
         origin_run_id, _, origin_task_id = clone_origin_task.split("/")
-
-        msg = {
-            "task_id": task_id,
-            "msg": "Cloning task from {}/{}/{}/{} to {}/{}/{}/{}".format(
-                self.flow.name,
-                origin_run_id,
-                step_name,
-                origin_task_id,
-                self.flow.name,
-                run_id,
-                step_name,
-                task_id,
-            ),
-            "step_name": step_name,
+        # Update system logger and monitor context
+        # We also pass this context as part of the task payload to support implementations that
+        # can't access the context directly
+        task_payload = {
             "run_id": run_id,
-            "flow_name": self.flow.name,
-            "ts": round(time.time()),
+            "step_name": step_name,
+            "task_id": task_id,
+            "retry_count": retry_count,
+            "project_name": current.get("project_name"),
+            "branch_name": current.get("branch_name"),
+            "is_user_branch": current.get("is_user_branch"),
+            "is_production": current.get("is_production"),
+            "project_flow_name": current.get("project_flow_name"),
+            "origin_run_id": origin_run_id,
+            "origin_task_id": origin_task_id,
         }
-        self.event_logger.log(msg)
+        _system_logger.update_context(task_payload)
+        _system_monitor.update_context(task_payload)
+
+        msg = "Cloning task from {}/{}/{}/{} to {}/{}/{}/{}".format(
+            self.flow.name,
+            origin_run_id,
+            step_name,
+            origin_task_id,
+            self.flow.name,
+            run_id,
+            step_name,
+            task_id,
+        )
+        with _system_monitor.count("metaflow.task.clone"):
+            _system_logger.log_event(
+                level="info",
+                module="metaflow.task",
+                name="clone",
+                payload={**task_payload,"msg": msg},
+            )
         # If we actually have to do the clone ourselves, proceed...
         clone_task_helper(
             self.flow.name,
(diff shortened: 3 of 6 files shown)


metaflow/task.py
```diff
@@ -513,19 +514,33 @@ def run_step(
...
+            "task_id": task_id,
+            "retry_count": retry_count,
+            "project_name": current.get("project_name"),
+            "branch_name": current.get("branch_name"),
+            "is_user_branch": current.get("is_user_branch"),
+            "is_production": current.get("is_production"),
+            "project_flow_name": current.get("project_flow_name"),
+            "trace_id": trace_id or None,
+        }
+        self.event_logger.send(
```
romain-intel, line 533: nit: could we use `system_current` to be consistent? It feels like we could get rid of the monitor/logger passed in to the decorators/task (it's a good thing) and use the globally present one.

metaflow/task.py
```diff
@@ -631,15 +646,16 @@ def run_step(
...
-                "step_name": step_name,
-                "run_id": run_id,
-                "flow_name": self.flow.name,
-            }
-            logger.log(tsk_msg)
+            with system_current.metrics_manager.count("metaflow.task.exception"):
+                pass
+            system_current.logger.log(
+                {
+                    "log_type": "ERROR",
```
romain-intel, line 653: could we have constants for these?

metaflow/metaflow_metrics_manager.py
```diff
@@ -0,0 +1,106 @@
...
+        logger_msg = self.logger.get_count_payload(
+            monitor_msg.payload if monitor_msg else None, qualifer_name
+        )
+        yield
+        self.monitor.send_metric(monitor_msg)
+        self.logger.log_event(logger_msg)
+
+    @contextmanager
+    def measure(self, metric_name, qualifer_name=None):
+        timer, counter = self.monitor.get_measure_metrics(metric_name)
```
romain-intel, line 88: could we ideally make this look symetric for both?

romain-intel: we can also just return a single "payload". Internally, the send_metric would know to stop the timer for example.

romain-intel (commented):
Some initial comments.

metaflow/__init__.py
```diff
@@ -100,6 +100,12 @@ class and related decorators.
 # current runtime singleton
 from .metaflow_current import current
 
+# system monitor runtime singleton
```
romain-intel, line 103: I would add a comment like (for internal metaflow use only). It's kind of explicit from the name but we can be extra explicit :)

metaflow/cli.py
```diff
@@ -1066,6 +1068,10 @@ def start(
     if decospecs:
         decorators._attach_decorators(ctx.obj.flow, decospecs)
 
+    # We create an instance of SystemMonitor and SystemLogger respectively
```
romain-intel, line 1071: nit, I would move this up to where ctx.obj.monitor and event_logger are created.

metaflow/metaflow_system_logger.py
```diff
@@ -0,0 +1,109 @@
...
+            self.flow_name = "not_a_real_flow"
+        return self._flow
+
+    @flow.setter
+    def flow(self, flow):
+        self._flow = flow
+
+    @property
+    def flow_name(self):
+        return self._flow_name
```
romain-intel, line 49: assumes that flow_name will not be called prior to flow which may not be correct. In general, I would have one init method that inits all 4 (flow, flow_name, environment and logger) and controlled by one single flag.

metaflow/metaflow_system_logger.py
```diff
@@ -0,0 +1,109 @@
...
+        if self._logger is None:
+            self._logger = LOGGING_SIDECARS[DEFAULT_EVENT_LOGGER](
+                flow=self.flow, env=self.environment
+            )
+            self._debug("Started logger outside of a flow")
+            self._logger.start()
+        return self._logger
+
+    @logger.setter
+    def logger(self, logger):
```
romain-intel, line 69: we don't really want this to be settable independently do we? can we just keep set_logger and have things set directly on the _ names? Or is there a use you are thinking of of setting things independently?

metaflow/metaflow_system_logger.py
```diff
@@ -0,0 +1,109 @@
+import sys
+from typing import Dict, Any
+
+
+class SystemLogger(object):
```
romain-intel, line 5: nit: please add type hints where appropriate.

metaflow/metaflow_system_monitor.py
```diff
@@ -0,0 +1,148 @@
+import sys
+from contextlib import contextmanager
+
+
+class SystemMonitor(object):
```
romain-intel, line 5: same comments as for system_logger

metaflow/task.py
```diff
@@ -513,19 +519,39 @@ def run_step(
...
-            msg = {
-                "task_id": task_id,
-                "msg": "task starting",
-                "step_name": step_name,
-                "run_id": run_id,
-                "flow_name": self.flow.name,
-                "ts": round(time.time()),
-            }
-            logger.log(msg)
+            with _system_monitor.count("metaflow.task.start"):
```
romain-intel, line 547: I would keep things the same here -- ie: use self.event_logger and self.logger

metaflow/task.py
```diff
@@ -631,15 +657,19 @@ def run_step(
...
-            tsk_msg = {
-                "task_id": task_id,
-                "exception_msg": str(ex),
-                "msg": "task failed with exception",
-                "step_name": step_name,
-                "run_id": run_id,
-                "flow_name": self.flow.name,
-            }
-            logger.log(tsk_msg)
+            print("I am here in metaflow")
```
romain-intel, line 660: nit: remove.

metaflow/task.py
```diff
@@ -665,18 +695,24 @@ def run_step(
...
-                "task_id": task_id,
-                "msg": "task ending",
-                "step_name": step_name,
-                "run_id": run_id,
-                "flow_name": self.flow.name,
-                "ts": round(time.time()),
-                "runtime": round(end),
-            }
-            logger.log(msg)
+            _system_monitor.gauge("metaflow.task.duration", duration)
```
romain-intel, line 709: A measure would be more appropriate here than a gauge.

romain-intel (commented):
A few initial comments but I think this goes in the direction we agreed to so we shouldn't be too far.

metaflow/task.py
talsperre: Question is should we be measuring the task duration only until the task "ends" or consider the portion dealing with metadata register as part of the task duration as well? 

If it's the former then it would be difficult to implement using the existing context manager construct - hence why we use the gauge metric.

romain-intel: That's a good point so we may need to add non-context specific ones for this. The issue is that a gauge isn't the same as a measure. Gauge are supposed to measure levels of things (like # of machines in a pool, etc) not disparate point measurements like this.

metaflow/metaflow_system_logger.py
romain-intel: This comment is still not addressed I think.

metaflow/metaflow_system_logger.py
```diff
@@ -0,0 +1,101 @@
...
+    @property
+    def logger(self) -> Optional["metaflow.event_logger.NullEventLogger"]:
+        from .plugins import LOGGING_SIDECARS
+        from .metaflow_config import DEFAULT_EVENT_LOGGER
+
+        if self._logger is None:
+            self._logger = LOGGING_SIDECARS[DEFAULT_EVENT_LOGGER](
+                flow=self.flow, env=self.environment
+            )
+            self._debug("Started logger outside of a flow")
```
savingoyal, line 54: Is this always true (or does this distinction even matter)?

metaflow/plugins/project_decorator.py
```diff
@@ -118,6 +118,7 @@ def flow_init(
                 "project_flow_name": project_flow_name,
             }
         )
+
```
savingoyal, line 121: minor nit - we can remove this file from this PR

metaflow/__init__.py
```diff
@@ -100,6 +100,12 @@ class and related decorators.
 # current runtime singleton
 from .metaflow_current import current
 
+# system monitor runtime singleton (meant for internal use)
+from .metaflow_system_monitor import _system_monitor
```
savingoyal, line 104: do we need to expose this as a top level import. we can consider doing - `from metaflow.system import monitor`

metaflow/task.py
```diff
@@ -291,25 +295,27 @@ def clone_only(
...
+            step_name,
+            origin_task_id,
+            self.flow.name,
+            run_id,
+            step_name,
+            task_id,
+        )
+        with _system_monitor.count("metaflow.task.clone"):
+            self.event_logger.log(
+                {
```
savingoyal, line 310: The log message format seems specific to the logging consumer. If that is the case, consider keeping `self.event_logger.log(msg)` as is to maintain flexibility and decorate the payload with `event_value`, `log_type`and `stream_type` inside an `atlas_logger` class.

metaflow/task.py
```diff
@@ -513,204 +519,239 @@ def run_step(
...
+            "task_id": task_id,
+            "retry_count": retry_count,
+            "project_name": current.get("project_name"),
+            "branch_name": current.get("branch_name"),
+            "is_user_branch": current.get("is_user_branch"),
+            "is_production": current.get("is_production"),
+            "project_flow_name": current.get("project_flow_name"),
+            "trace_id": trace_id or None,
+        }
+        self.event_logger.send(
```
savingoyal, line 538: why use `send` instead of `log`? iirc `send` was a workaround for a very specific use case (I need to dig through the slack conversations) and we would want to avoid introducing it too much in the code base.

metaflow/task.py
```diff
@@ -513,204 +519,239 @@ def run_step(
...
+                    input_obj = Inputs(self._clone_flow(inp) for inp in inputs)
+                    self.flow._set_datastore(output)
                     # initialize parameters (if they exist)
                     # We take Parameter values from the first input,
                     # which is always safe since parameters are read-only
                     current._update_env(
                         {
                             "parameter_names": self._init_parameters(
-                                inputs[0], passdown=False
+                                inputs[0], passdown=True
```
savingoyal, line 588: can you highlight the rationale for this change?

metaflow/task.py
```diff
@@ -513,204 +519,239 @@ def run_step(
...
+                    self.event_logger.log(
+                        {
+                            "event_name": "metaflow.task.exception",
+                            "event_value": 1,
+                            "log_type": "ERROR",
+                            "log_value": str(ex),
+                            "traceback": traceback.format_exc(),
+                        }
+                    )
+                    pass
```
savingoyal, line 676: why is this needed?

metaflow/task.py
```diff
@@ -513,204 +519,239 @@ def run_step(
...
+
+                # Emit metrics to logger/monitor sidecar implementations
+                with self.monitor.count("metaflow.task.end"):
+                    self.event_logger.log(
+                        {
+                            "event_name": "metaflow.task.end",
+                            "event_value": 1,
+                        }
+                    )
+                    pass
```
savingoyal, line 710: same here - why is a `pass` needed? is it to signify end of an indent block or something more? if the former, we can avoid introducing pass to maintain consistency of style in the code base

metaflow/task.py
```diff
@@ -513,204 +519,239 @@ def run_step(
...
-                    # This should be captured by static checking but
-                    # let's assert this again
-                    raise MetaflowInternalError(
-                        "Step *%s* is not a join "
-                        "step but it gets multiple "
-                        "inputs." % step_name
+        with self.monitor.measure("metaflow.task.duration"):
+            try:
+                with self.monitor.count("metaflow.task.start"):
+                    self.event_logger.log(
```
savingoyal, line 549: why is this log necessary?

savingoyal (changes requested):

metaflow/task.py
talsperre: Replace it with something like `event_logger.log(msg="", type="")`.

metaflow/task.py
talsperre: Updated to `False` again. Got changed by mistake.

metaflow/task.py
talsperre: Will remove in updated PR

metaflow/task.py
talsperre: We want the event to be available in the log stream in addition to our metrics.

metaflow/task.py
talsperre: `send` is used to update the `event_logger` context so that we have all these values available as additional tags in the event stream. We thus log this event with message type set to `MUST_SEND` instead of `BEST_EFFORT` that is normally used in `log`.

metaflow/task.py
talsperre: Will update the API in the revised PR.

metaflow/metaflow_system_logger.py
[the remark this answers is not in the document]
talsperre: This would be the case when we access `logger` outside of a flow. Otherwise it is started in `cli.py` itself.

talsperre: It is not always true. `self._logger` would be `None` only when the logger is being accessed outside a flow. Otherwise we set `self._logger` to the event logger in `cli.py`.

metaflow/metaflow_system_logger.py
```diff
@@ -0,0 +1,101 @@
+import sys
+from typing import Dict, Any, Optional, Union
+
+
+class SystemLogger(object):
```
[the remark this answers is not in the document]
talsperre: Do we want it as part of the code or in a separate README?

metaflow/event_logger.py
```diff
@@ -24,6 +24,31 @@ def log(self, payload):
             msg = Message(MessageTypes.BEST_EFFORT, payload)
             self._sidecar.send(msg)
 
+    def log_event(self, msg=None, event_name=None, log_stream=None, other_context=None):
+        """
+        Log an event to the event logger.
+
+        Parameters
+        ----------
+        msg : str
```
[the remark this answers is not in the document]
romain-intel: Also add type hints in the signature.

metaflow/system/metaflow_system_logger.py
```diff
@@ -0,0 +1,92 @@
...
+class SystemLogger(object):
+    def __init__(self):
+        self._logger = None
+        self._flow_name = None
+
+    def __del__(self):
+        if self._flow_name == "not_a_real_flow":
+            self.logger.terminate()
+
+    def init_environment_outside_flow(
```
romain-intel, line 14: missing return type?

metaflow/system/metaflow_system_logger.py
```diff
@@ -0,0 +1,92 @@
...
+        Parameters
+        ----------
+        msg : str
+            Message to log.
+
+        Returns
+        -------
+        None
+        """
+        print("system logger: %s" % msg, file=sys.stderr)
```
romain-intel, line 68: nit: maybe squelch unless debug is turned on?

metaflow/system/metaflow_system_logger.py
```diff
@@ -0,0 +1,92 @@
...
+        event_name: Optional[str] = None,
+        log_stream: Optional[str] = None,
+        other_context: Optional[Dict[str, Any]] = None,
+    ):
+        """
+        Log an event to the event logger.
+
+        Parameters
+        ----------
+        msg : str, optional default None
```
romain-intel, line 82: nit: optional, default None

metaflow/system/metaflow_system_monitor.py
```diff
@@ -0,0 +1,121 @@
...
+class SystemMonitor(object):
+    def __init__(self):
+        self._monitor = None
+        self._flow_name = None
+
+    def __del__(self):
+        if self._flow_name == "not_a_real_flow":
+            self.monitor.terminate()
+
+    def init_environment_outside_flow(
```
romain-intel, line 15: this method does not depend on self. I would move it to a util or something since it is common across both logger and monitor.

metaflow/system/metaflow_system_monitor.py
```diff
@@ -0,0 +1,121 @@
...
+        name : str
+            The name of the gauge.
+        value : float
+            The value of the gauge.
+
+        """
+        self.monitor.gauge(gauge)
+
+
+_system_monitor = SystemMonitor()
```
romain-intel, line 121: Remove line -- no longer used since in __init__.py

metaflow/task.py
```diff
@@ -291,25 +294,21 @@ def clone_only(
...
+            origin_run_id,
+            step_name,
+            origin_task_id,
+            self.flow.name,
+            run_id,
+            step_name,
+            task_id,
+        )
+        with _system_monitor.count("metaflow.task.clone"):
+            self.event_logger.log_event(
```
romain-intel, line 308: nit: _system_logger to be consistent.

metaflow/system/metaflow_system_logger.py
```diff
@@ -0,0 +1,92 @@
...
+            Message to log.
```
romain-intel, line 83: What does it mean to log a "none" message, just empty? Maybe add a comment to that effect.

metaflow/task.py
romain-intel: Yes, the intent is to set a "common set of tags" (in the lingo of the other measure thing (datadog)). I think it may make sense to actually have a specific method `self.event_logger.set_common_tags()` or something like that. Backends would be free to do whatever they want with that. Internally, we could then use send to send a MUST_SEND message or just keep it around (the MUST_SEND was to avoid overloading the pipe to the sidecar with all the context all the time).

romain-intel (commented):
Some comments. Let's discuss tomorrow and finalize.

metaflow/event_logger.py
```diff
@@ -24,6 +24,31 @@ def log(self, payload):
             msg = Message(MessageTypes.BEST_EFFORT, payload)
             self._sidecar.send(msg)
 
+    def log_event(self, msg=None, event_name=None, log_stream=None, other_context=None):
```
savingoyal, line 27: do we need this method? can we not reuse `def log(self, payload)` where the payload contains the msg, event_name, log_stream and other_context?

metaflow/system/metaflow_system_logger.py
savingoyal, line 14: can you add comments here for future readers as to why/how the logger needs to be/is constructed in this manner?

savingoyal (changes requested):

metaflow/system/metaflow_system_logger.py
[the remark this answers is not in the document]
talsperre: I think there is reasonable benefit for passing `MetaflowEnvironment` when used outside of a flow as well. Primarily, the environment provides us with additional tags/context like the version of metaflow used, platform, the user name etc.  With that in mind, I just referred to the way the logger/monitor were instantiated in `cli.py` [here](https://github.com/Netflix/metaflow/blob/master/metaflow/cli.py#L951-L959).

talsperre: Will address

metaflow/system/metaflow_system_logger.py
```diff
@@ -0,0 +1,92 @@
...
+            if e.TYPE == DEFAULT_ENVIRONMENT
+        ][0](flow)
+
+    def init_system_logger(
+        self, flow_name: str, logger: "metaflow.event_logger.NullEventLogger"
+    ):
+        self._flow_name = flow_name
+        self._logger = logger
+
+    def init_logger_outside_flow(self):
```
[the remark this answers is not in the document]
talsperre: Agreed - will address

metaflow/task.py
talsperre: Yes, lets discuss this in the call. What you said should be possible - depending on the presence of certain specific key/value pair in the payload we can use `MUST_SEND` or `BEST_EFFORT` appropriately. My only concern would be that if someone reads the code in `cli.py` then it would be confusing for them as to why we use different message types for different payloads.

metaflow/task.py
```diff
@@ -280,25 +283,21 @@ def clone_only(
...
+                event_name="metaflow.task.clone",
+                msg=msg,
```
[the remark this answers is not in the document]
talsperre: The `task_id`, `step_name` etc are not part of the payload of an event usually. They are added once at the beginning as additional tags and then are automatically added to all events. But yes, I should add another call to add these additional tags/context like we did in the `run_step` function.

metaflow/system/metaflow_system_logger.py
romain-intel: still relevant I think?

metaflow/system/metaflow_system_logger.py
```diff
@@ -0,0 +1,103 @@
...
+            Message to log.
+
+        Returns
+        -------
+        None
+        """
+        print("system logger: %s" % msg, file=sys.stderr)
+
+    def log_event(
+        self, level: str, module: str, name: str, payload: Optional[Any] = None
```
romain-intel, line 77: we could make level an enum to validate it.

metaflow/system/metaflow_system_monitor.py
```diff
@@ -0,0 +1,106 @@
...
+        Parameters
+        ----------
+        msg : str
+            Message to log.
+
+        Returns
+        -------
+        None
+        """
+        print("system monitor: %s" % msg, file=sys.stderr)
```
romain-intel, line 56: ditto as above

romain-intel (commented):
Looks very close. minor nits.

metaflow/task.py
```diff
@@ -280,25 +283,42 @@ def clone_only(
                 "task.clone_only needs a valid clone_origin_task value."
             )
         origin_run_id, _, origin_task_id = clone_origin_task.split("/")
+        # Update system logger context
+        # We don't update the context for the system monitor as the cardinality of this context is high
```
savingoyal, line 287: to confirm - your custom monitor implementation ignores whatever context is set? since other monitor implementations may need a context to be set here.

metaflow/task.py
```diff
@@ -280,25 +283,42 @@ def clone_only(
...
+            run_id,
+            step_name,
+            task_id,
+        )
+        with _system_monitor.count("metaflow.task.clone"):
+            _system_logger.log_event(
+                level="info",
+                module="metaflow.task",
+                name="clone",
+                payload={"msg": msg},
```
savingoyal, line 320: to maintain consistency, can you ensure that other elements of the "msg" payload from the tip of the master are also passed in as in - `task_id`, `step_name`, `run_id`, `flow_name`, `ts` etc.

metaflow/task.py
```diff
@@ -502,204 +522,226 @@ def run_step(
...
+                self.flow._task_ok = True
+                self.flow._success = True
+
+            except Exception as ex:
+                with self.monitor.count("metaflow.task.exception"):
+                    _system_logger.log_event(
+                        level="error",
+                        module="metaflow.task",
+                        name="exception",
+                        payload={"msg": traceback.format_exc()},
```
savingoyal, line 667: to maintain consistency, can you ensure that other elements of the "msg" payload from the tip of the master are also passed in as in - task_id, step_name, run_id, flow_name, ts etc.

metaflow/task.py
```diff
@@ -502,204 +522,226 @@ def run_step(
...
+            finally:
+                if self.ubf_context == UBF_CONTROL:
+                    self._finalize_control_task()
+
+                # Emit metrics to logger/monitor sidecar implementations
+                with self.monitor.count("metaflow.task.end"):
+                    _system_logger.log_event(
+                        level="info",
+                        module="metaflow.task",
+                        name="end",
```
savingoyal, line 699: no `msg` here?

metaflow/task.py
```diff
@@ -502,204 +522,226 @@ def run_step(
...
+                self.metadata.stop_heartbeat()
+
+                # Task duration consists of the time taken to run the task as well as the time taken to
+                # persist the task metadata and data to the datastore.
+                duration = time.time() - start
+                _system_logger.log_event(
+                    level="info",
+                    module="metaflow.task",
+                    name="duration",
+                    payload={"msg": str(duration)},
```
savingoyal, line 746: to maintain consistency, can you ensure that other elements of the "msg" payload from the tip of the master are also passed in as in - task_id, step_name, run_id, flow_name, ts etc.

savingoyal: also, can we be consistent in what is encoded in `msg`?

metaflow/task.py
```diff
@@ -280,25 +283,42 @@ def clone_only(
...
-                run_id,
-                step_name,
-                task_id,
-            ),
-            "step_name": step_name,
-            "run_id": run_id,
-            "flow_name": self.flow.name,
-            "ts": round(time.time()),
-        }
-        self.event_logger.log(msg)
```
savingoyal, line 301: curious, what is the impact of still including `self.event_logger.log` if we chose to? would it break anything on your end downstream?

metaflow/task.py
```diff
@@ -502,204 +523,230 @@ def run_step(
...
+                self.metadata.stop_heartbeat()
+
+                # Task duration consists of the time taken to run the task as well as the time taken to
+                # persist the task metadata and data to the datastore.
+                duration = time.time() - start
+                _system_logger.log_event(
+                    level="info",
+                    module="metaflow.task",
+                    name="duration",
+                    payload={**task_payload, "msg": str(duration)},
```
savingoyal, line 751: isn't the `msg` - `task ending`?

savingoyal (approved):
lgtm! one change needed. @saikonen is also taking a quick look

metaflow/system/system_monitor.py
```diff
@@ -49,11 +69,9 @@ def _debug(msg: str):
         msg : str
             Message to log.
 
-        Returns
-        -------
-        None
         """
-        print("system monitor: %s" % msg, file=sys.stderr)
+        if os.environ.get("METAFLOW_DEBUG_SIDECAR", "0") == "1":
```
romain-intel, line 73: minor nit, why not use `debug.sidecar_exec`? The reason I ask is we handle True/0/1 in a slightly different way so it would be more consistent.

talsperre: `debug.sidecar_exec` won't work when used outside of a flow. Open to replacing with the correct check to make it more consistent.

romain-intel:
OK, very close. There seems to be a small conflict as well. 

To clarify btw, there is no assumption/expectation that we pass the context to all messages going forward. For example, if a plugin wants to send a message using the system logger, it would not re-include the task_payload (since it wouldn't have it). Is that the understanding? In other words, the updated context is passed in for the messages in task but that would be it (otherwise it would not be passed in). The idea is that some implementations (like ours) would strip this out since it was just passed in update_context but others would use it but it wouldn't be availabe for other messages.

metaflow/task.py
talsperre: There is a separate message for task ending / ended. This one is meant to just add persist the task duration as part of a log.

metaflow/system/system_monitor.py
```diff
@@ -0,0 +1,124 @@
...
+    def gauge(self, gauge: "metaflow.monitor.Gauge"):
+        """
+        Log a gauge.
+
+        Parameters
+        ----------
+        name : str
+            The name of the gauge.
+        value : float
+            The value of the gauge.
```
saikonen, line 121: nit: this seems to only be taking in a gauge, not name and value

metaflow/task.py
```diff
@@ -280,25 +283,43 @@ def clone_only(
...
+            "task_id": task_id,
+            "retry_count": retry_count,
+            "project_name": current.get("project_name"),
+            "branch_name": current.get("branch_name"),
+            "is_user_branch": current.get("is_user_branch"),
+            "is_production": current.get("is_production"),
+            "project_flow_name": current.get("project_flow_name"),
+            "origin_run_id": origin_run_id,
+            "origin_task_id": origin_task_id,
+            "trace_id": trace_id or None,
```
saikonen, line 301: we don't have a trace_id available here so this fails

savingoyal (approved):

saikonen (approved):

PR #1796, opened 2024-04-10
feature: on demand docker environments

saikonen:
Adds support for on demand Docker images. Open for discussion

open todo's

- [x] add fallback to `conda` environment when executing locally with `--environment docker`
- [x] add support for imagebakery endpoint auth
- [x] cache bakery image tags locally to avoid environment drift with consecutive deployments when using loosely pinned versions


savingoyal:
`python for.py --environment=docker run` without `@kubernetes` immediately returns an error that `Image Bakery` is not configured. Perhaps it shouldn't throw the error?

metaflow/metaflow_config.py
```diff
@@ -281,6 +281,19 @@
     else None,
 )
 ###
+# On Demand Docker image build configuration
+###
+# Image builder service url
+DOCKER_IMAGE_BAKERY_URL = from_conf("DOCKER_IMAGE_BAKERY_URL", None)
```
savingoyal, line 287: nit - change this to `FAST_BAKERY_URL`? and similar changes below?

metaflow/plugins/docker/docker_environment.py
```diff
@@ -0,0 +1,135 @@
...
+        for k in cls.__dict__:
+            obj = getattr(cls, k)
+            if callable(obj):
+                setattr(self.__class__, k, obj)
+        return True
+
+    def init_environment(self, echo):
+        if self._setup_conda_fallback():
+            print(
+                "Some steps would execute locally. Had to fallback to a conda environment"
```
savingoyal, line 71: we can skip the print statement for now? ideally, we can have a debug flag that will print out what kind of environment is being created for which step. we can make it a follow up with a TODO in the code.

metaflow/plugins/docker/docker_environment.py
```diff
@@ -0,0 +1,135 @@
...
+        # replace methods with the delegate class that we determined.
+        cls = CondaEnvironment
+        for k in cls.__dict__:
+            obj = getattr(cls, k)
+            if callable(obj):
+                setattr(self.__class__, k, obj)
+        return True
+
+    def init_environment(self, echo):
+        if self._setup_conda_fallback():
```
savingoyal, line 69: it seems that we fall back to conda environments for all steps if any step is running locally. can we make it such that any step that is marked to run on kubernetes can benefit from fast bakery?

metaflow/plugins/docker/docker_environment.py
```diff
@@ -0,0 +1,135 @@
...
+            print(
+                "Some steps would execute locally. Had to fallback to a conda environment"
+            )
```
savingoyal, line 72: a flow that has a single step marked with @kubernetes, fails right now 

```
Some steps would execute locally. Had to fallback to a conda environment
Bootstrapping virtual environment(s) ...
Virtual environment(s) bootstrapped!
2024-05-02 08:48:39.662 Workflow starting (run-id 217204):
2024-05-02 08:48:42.651 [217204/start/1333327 (pid 68470)] Task is starting.
2024-05-02 08:48:47.021 [217204/start/1333327 (pid 68470)] [pod t-f10eeb12-c9h8b-bv7gc] Task is starting (Pod is pending, Container is waiting - ContainerCreating)...
2024-05-02 08:48:46.955 [217204/start/1333327 (pid 68470)] [pod t-f10eeb12-c9h8b-bv7gc] Setting up task environment.
2024-05-02 08:48:56.265 [217204/start/1333327 (pid 68470)] [pod t-f10eeb12-c9h8b-bv7gc] Downloading code package...
2024-05-02 08:48:57.143 [217204/start/1333327 (pid 68470)] [pod t-f10eeb12-c9h8b-bv7gc] Code package downloaded.
2024-05-02 08:48:57.175 [217204/start/1333327 (pid 68470)] [pod t-f10eeb12-c9h8b-bv7gc] Task is starting.
2024-05-02 08:48:57.658 [217204/start/1333327 (pid 68470)] [pod t-f10eeb12-c9h8b-bv7gc] bash: line 1: /conda-prefix/bin/python: No such file or directory
2024-05-02 08:49:00.594 [217204/start/1333327 (pid 68470)] Kubernetes error:
2024-05-02 08:49:00.683 [217204/start/1333327 (pid 68470)] Error (exit code 127). This could be a transient error. Use @retry to retry.
2024-05-02 08:49:00.683 [217204/start/1333327 (pid 68470)]
2024-05-02 08:49:00.912 [217204/start/1333327 (pid 68470)] Task failed.
```

saikonen: this should now be working after the latest changes

metaflow/plugins/docker/fast_bakery.py
```diff
@@ -0,0 +1,172 @@
+import hashlib
+import json
+import requests
+
+from metaflow.exception import MetaflowException
+from metaflow.metaflow_config import (
+    FAST_BAKERY_TYPE,
+    FAST_BAKERY_URL,
+    FAST_BAKERY_AUTH,
+    get_pinned_conda_libs,
```
savingoyal, line 10: can the environment implementation fetch the pinned dependencies and pass it to the bakery? ideally, we can ensure that this client has no dependencies on metaflow - the auth variables can be populated by docker_environment.py

metaflow/plugins/docker/fast_bakery.py
```diff
@@ -0,0 +1,172 @@
...
+)
+
+BAKERY_METAFILE = ".imagebakery-cache"
+
+
+class FastBakeryException(MetaflowException):
```
savingoyal, line 16: We can throw a generic exception (not MetaflowException) here which can be caught by docker_environment and wrapped into a MetaflowException

metaflow/plugins/docker/fast_bakery.py
```diff
@@ -0,0 +1,172 @@
+import hashlib
+import json
+import requests
+
```
savingoyal, line 4: Would something of this sorts work here -

```suggestion
import json
import requests


class FastBakery(object):
    def __init__(self, url):
        self.url = url

    def bake(
        self,
        python_ver=None, # TODO - Ensure the keys are same in the API spec and client
        pypi_packages=None,
        conda_packages=None,
        base_image=None,
        image_kind="esgz-zstd",
    ):
        deep_clean = (lambda f: lambda d: f(f, d))(
            lambda self, item: {
                k: v
                for k, v in (
                    (k, self(self, v)) if isinstance(v, dict) else (k, v)
                    for k, v in item.items()
                )
                if v is not None and v != {}
            }
        )
        payload = deep_clean(
            {
                "pythonVersion": python_ver,
                "imageKind": image_kind,
                "pipRequirements": pypi_packages,
                "condaMatchspecs": conda_packages,
                "baseImage": {"imageReference": base_image},
            }
        )
        return self.aws_iam_invoker(payload)

    def aws_iam_invoker(self, payload):
        try:
            # AWS_IAM requires a signed request to be made
            headers = {"Content-Type": "application/json"}
            payload = json.dumps(payload)

            import boto3
            from botocore.auth import SigV4Auth
            from botocore.awsrequest import AWSRequest
            from botocore.exceptions import BotoCoreError, NoCredentialsError

            import urllib

            session = boto3.Session()
            credentials = session.get_credentials().get_frozen_credentials()

            # credits to https://github.com/boto/botocore/issues/1784#issuecomment-659132830,
            # We need to jump through some hoops when calling the endpoint with IAM auth
            # as botocore does not offer a direct utility for signing arbitrary requests
            req = AWSRequest("POST", self.url, headers, payload)
            SigV4Auth(
                credentials, service_name="lambda", region_name=session.region_name
            ).add_auth(req)

            response = requests.post(req.url, data=req.body, headers=req.headers)
            
            # Manually check the status code and raise an exception if it's an error
            if response.status_code >= 400:
                raise requests.exceptions.HTTPError(
                    f"HTTP {response.status_code} Error: {response.reason}\nResponse Body: {response.text}",
                    response=response
                )

            return response

        except requests.exceptions.HTTPError as e:
            raise e

        except (requests.exceptions.RequestException, BotoCoreError, NoCredentialsError) as e:
            # Handle specific errors related to requests and AWS services
            raise Exception(f"Failed to invoke AWS service due to network or AWS error: {e}")

        except Exception as e:
            # Handle other possible exceptions
            raise Exception(f"An unexpected error occurred: {e}")
```

metaflow/plugins/docker/docker_environment.py
```diff
@@ -0,0 +1,154 @@
...
+                self.steps_to_delegate.add(step.name)
+        if not self.steps_to_delegate:
+            return
+        # TODO: move to init if possible.
+        self.delegate = CondaEnvironment(self.flow)
+        self.delegate.set_local_root(self.local_root)
+
+    def init_environment(self, echo):
+        self._init_conda_fallback()
+        # First resolve environments through Conda, before PyPI.
```
savingoyal, line 67: this comment might be misplaced?

metaflow/plugins/docker/docker_environment.py
```diff
@@ -0,0 +1,154 @@
...
+        echo("Baking Docker images for environment(s) ...")
+        for step in self.flow:
+            self.bake_image_for_step(step)
+        echo("Environments are ready!")
+        if self.steps_to_delegate:
+            # TODO: add debug echo to output steps that required a conda environment.
```
savingoyal, line 73: these comments too

metaflow/plugins/docker/docker_environment.py
```diff
@@ -0,0 +1,154 @@
...
+            # The delegated conda environment also need to validate and init.
+            # we pass a set of steps we want to init to restrict the conda environments created.
+            self.delegate.validate_environment(echo, self.datastore_type)
+            self.delegate.init_environment(echo, self.steps_to_delegate)
+
+    def bake_image_for_step(self, step):
```
savingoyal, line 79: multiple steps might share the same image - maybe we can check if that is the case and only bake unique images?

metaflow/plugins/docker/fast_bakery.py
```diff
@@ -0,0 +1,172 @@
...
+from metaflow.exception import MetaflowException
+from metaflow.metaflow_config import (
+    FAST_BAKERY_TYPE,
+    FAST_BAKERY_URL,
+    FAST_BAKERY_AUTH,
+    get_pinned_conda_libs,
+)
+
+BAKERY_METAFILE = ".imagebakery-cache"
```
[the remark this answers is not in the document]
savingoyal: also, can you help me understand the contents of this cache?

metaflow/plugins/aws/batch/batch.py
```diff
@@ -23,6 +24,9 @@
     DEFAULT_SECRETS_BACKEND_TYPE,
     AWS_SECRETS_MANAGER_DEFAULT_REGION,
     S3_SERVER_SIDE_ENCRYPTION,
+    FAST_BAKERY_URL,
+    FAST_BAKERY_AUTH,
+    FAST_BAKERY_TYPE,
```
savingoyal, line 29: we can avoid introducing these as env vars here and in @kubernetes with the expectation that using fast bakery may involve using `@secrets` or `@environment` - to configure the URL and auth. The client can try to read from env vars unless explicitly configured with the URL and auth?

metaflow/plugins/docker/docker_environment.py
saikonen: the images are baked in sequence so the consecutive steps with identical packages should be hitting the cache at this point, skipping the baking. previously the caching was happening inside the fast_bakery, but moved to the environment side now so it should be more clear.

metaflow/plugins/docker/fast_bakery.py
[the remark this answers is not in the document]
saikonen: moved caching to the environment side. what is in the cache at the moment is
- a unique hash derived from the requested environment (python, packages, image-type etc.)
- the image tag that was returned from the bakery endpoint for the request
- we also persist the request payload that was sent to the bakery in order to be able to list details of cached images via `docker cache list`

metaflow/plugins/docker/docker_environment.py
savingoyal: rather than baking in sequence, we can perhaps bake them in parallel - that should save precious seconds

metaflow/plugins/docker/docker_environment.py
```diff
@@ -0,0 +1,223 @@
...
+        pypi_deco = next(
+            (deco for deco in step.decorators if isinstance(deco, PyPIStepDecorator)),
+            None,
+        )
+        image = None
+        # pypi packages need to take precedence over conda.
+        # a conda decorator always exists alongside pypi so this needs to be accounted for
+        dependency_deco = pypi_deco if pypi_deco is not None else conda_deco
+        if dependency_deco is not None:
+            python = dependency_deco.attributes["python"]
```
savingoyal, line 110: python comes from conda_deco

metaflow/plugins/pypi/conda_environment.py
```diff
@@ -65,7 +65,7 @@ def validate_environment(self, echo, datastore_type):
         micromamba = Micromamba()
         self.solvers = {"conda": micromamba, "pypi": Pip(micromamba)}
 
-    def init_environment(self, echo):
+    def init_environment(self, echo, only_steps=[]):
```
saikonen, line 68: @romain-intel I'd like your thoughts on this mechanism.

The reason we need something akin to this is to support the environment fallback logic in `docker_environment.py`.

1. docker env gathers a list of steps that it can not act on (conda/pypi but not executing remotely)
2. for the incompatible steps we want to fallback to the conda implementation, and initialize accordingly
3. the only_steps is meant to skip those steps that we were able to bake a docker image for

Would this be a good addition to the environment init flow. Any better alternatives to this?

metaflow/plugins/pypi/pypi_decorator.py
```diff
@@ -64,7 +64,7 @@ def step_init(self, flow, graph, step, decos, environment, flow_datastore, logge
         # one here. We can consider introducing a UX where @pypi is able to consume
         # poetry.lock files in the future.
 
-        _supported_virtual_envs = ["conda"]  # , "venv"]
+        _supported_virtual_envs = ["conda", "docker"]  # , "venv"]
```
saikonen, line 67: @romain-intel this relates to discussions on decorator improvements. Should this be reliably extendable through extensions?

Main motivation is that this whole imagebakery feature could be a standalone extension, except for the few additions it needs to make to the OSS PyPI plugin.

saikonen: `_supported_virtual_envs` is just an implementation detail here, but what I'm pitching is adding a mechanism where a newly introduced environment could add itself to a whitelist for known decorators.

saikonen:
the changes allow an extension to implement
```python
def conda_supported_virtual_envs():
    return ["docker"]
```
in `org_ext/config/mfextinit_org_ext.py`
to support `--environment docker` for `@pypi` and `@conda`

---

saikonen:
also possible to fallback to a Conda environment for steps that can not be handled by another env implementation:

```python
def init_environment(self, echo):
    self.delegate = CondaEnvironment(self.flow)
    ...
    if self.steps_to_delegate:
        self.delegate.validate_environment(echo, self.datastore_type)
        self.delegate.init_environment(echo, self.steps_to_delegate) # skips steps that were already handled.
```

PR #1797, opened 2024-04-11
fix: support default parameter values with argo events

saikonen:
fixes issue where a partial event payload would not correctly use default values for a flow's Parameters.


romain-intel:
No impact on core -- no issues.

Issue #1769, opened 2024-03-20
Make default kubernetes resources configurable


martinhausio:
We recently upgraded our GKE Autopilot cluster to 1.29 and to our surprise, it is no longer calculating the resource allocation the same way, and the default metaflow resources are now "just a little" too large, resulting in our workflows being rejected.

I found the defaults are configured [here](https://github.com/Netflix/metaflow/blob/fbe4aa10280873beeada332390faf3fd45b5d8bb/metaflow/plugins/kubernetes/kubernetes_decorator.py#L96), and we also have many ways to configure the resources, but it would be ideal for us to be able to configure the default resources via an environment variable, much the same way we can set the nodeSelector.

savingoyal:
Great idea! We would be open to a PR if you would like to submit one!

martinhausio:
@savingoyal , I just submitted a quick PR for it! Hope this helps.

PR #1800, opened 2024-04-11
Make Kubernetes default resources configurable

martinhausio:
This fixes https://github.com/Netflix/metaflow/issues/1769.

A quick addition that looks if the METAFLOW_KUBERNETES_CPU,  METAFLOW_KUBERNETES_MEMORY, METAFLOW_KUBERNETES_DISK environment variable is set and if so parses the values. If a value is detected as being an attribute, then assign it, otherwise it will just use the default values as defined.


martinhausio:
I have tested locally by doing `pip install .`, then instantiating the same argo workflow with and without the environment variable and this behaves as intended. 

I tested setting `cpu`, `memory` and `disk` separately, and together successfully.

saikonen (changes requested):

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -174,6 +175,14 @@ def __init__(self, attributes=None, statically_defined=False):
             except (NameError, ImportError):
                 pass
 
+        # parse the CPU, memory, disk, values from the KUBERNETES_RESOURCES environment variable (you would need to export the METAFLOW_KUBERNETES_RESOURCES environment variable with the desired values before running the flow)
+        # find the values in the format "cpu=1,memory=4096,disk=10240", if not set, the default values described above are used
+        if KUBERNETES_RESOURCES:
```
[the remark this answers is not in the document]
martinhausio: thank you @saikonen , I did consider this and merely proceeded this way because of the way it was being passed via the command line arguments. I agree it is cleaner to separate, will fix now.

martinhausio:
@saikonen , changes ready for your review!

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -175,13 +177,14 @@ def __init__(self, attributes=None, statically_defined=False):
...
-                if key in self.attributes:
-                    self.attributes[key] = value
+        # parse the CPU, memory, disk, values from the KUBERNETES_ environment variable (you would need to export the METAFLOW_KUBERNETES_CPU, METAFLOW_KUBERNETES_MEMORY and/or METAFLOW_KUBERNTES_DISK environment variable with the desired values before running the flow)
+        # find the values from the environment variables, if not set, the default values as defined above are used
+        if KUBERNETES_CPU:
+            self.attributes["cpu"] = KUBERNETES_CPU
+        if KUBERNETES_MEMORY:
+            self.attributes["memory"] = KUBERNETES_MEMORY
+        if KUBERNETES_DISK:
+            self.attributes["disk"] = KUBERNETES_DISK
```
saikonen, line 187: One more change I would suggest is having a similar check to existing env var setters, where the default does not override a value that the user has supplied.
as an example:

```
METAFLOW_KUBERNETES_CPU=3

@kubernetes(cpu=2)
@step
```
expectation would be that cpu gets set to 2, not 3.

note though that the check can not be identical to the existing ones, as we have proper default values for some of the attributes, so f.ex. `self.attributes["cpu"]` will always be set.

metaflow/metaflow_config.py
```diff
@@ -315,7 +315,9 @@
 # Default port number to open on the pods
 KUBERNETES_PORT = from_conf("KUBERNETES_PORT", None)
 # Default kubernetes resource requests for CPU, memory and disk
-KUBERNETES_RESOURCES = from_conf("KUBERNETES_RESOURCES", "")
+KUBERNETES_CPU = from_conf("KUBERNETES_CPU", "")
+KUBERNETES_MEMORY = from_conf("KUBERNETES_MEMORY", "")
+KUBERNETES_DISK = from_conf("KUBERNETES_DISK", "")
```
saikonen, line 320: Empty string works as a default too, but just to stick to other defaults I'd change these to `None` instead.

saikonen (changes requested):

metaflow/plugins/kubernetes/kubernetes_decorator.py
martinhausio: found the way to validate that, thanks

saikonen (approved):

PR #1802, opened 2024-04-16
feature: scrub logs

saikonen:
introduces `logs scrub run_id/step/task_id` command for scrubbing logs of tasks.

by default it deletes both stdout/stderr for the latest attempt of a task.

Final feature set:
- `logs [pathspec]` works as previously, with the new default subcommand custom click group
- `logs show [pathspec]` is the default subcommand for logs cmd group
- `logs scrub [pathspec]` will overwrite logs with a redacted message

also adds support for both show/scrub commands to specify `--attempt [number]` to target a specific attempt of a task, instead of the latest.


savingoyal:
@saikonen we could introduce a top-level `delete` command that has a sub command - `logs`? wdyt? also, the default behavior could be for `--latest-attempt` and users can pass in `--all-attempts` to wipe out logs for all attempts?

saikonen:
this now introduces `logs` as a command group, with a default fallback to the new `logs show` command in case no subcommmand is specified.

the following command still works the same
```
python LongLog.py logs 1713532810109387/start/1
python LongLog.py logs 1713532810109387/start/1 --stderr
```
but now we can also support the following

```
python LongLog.py logs delete 1713532810109387/start/1
```

metaflow/plugins/logs_cli.py
```diff
@@ -0,0 +1,292 @@
...
+    help="Scrub stdout/stderr produced by a task or all tasks in a step. "
+    "The format for input-path is either <run_id>/<step_name> or "
+    "<run_id>/<step_name>/<task_id>."
+)
+@click.argument("input-path")
+@click.option(
+    "--stdout/--no-stdout",
+    default=False,
+    show_default=True,
+    help="Show stdout of the task.",
```
savingoyal, line 220: scrub instead of show? (here and below?)

metaflow/plugins/logs_cli.py
```diff
@@ -0,0 +1,292 @@
...
+    # Pathspec can either be run_id/step_name or run_id/step_name/task_id.
+    parts = input_path.split("/")
+    if len(parts) == 2:
+        run_id, step_name = parts
+        task_id = None
+    elif len(parts) == 3:
+        run_id, step_name, task_id = parts
+    else:
+        raise CommandException(
+            "input_path should either be run_id/step_name "
```
savingoyal, line 257: `input-path`?

metaflow/plugins/logs_cli.py
```diff
@@ -0,0 +1,292 @@
...
+        ds_list = list(datastore_set)
+
+    if ds_list:
+        for ds in ds_list:
+            for stream in streams:
+                ds.scrub_logs(LOG_SOURCES, stream)
+        echo("Logs have been scrubbed.")
+
+    else:
+        raise CommandException(
```
savingoyal, line 286: what is the behavior for tasks that haven't finished yet? a few different cases here - 
1. the attempt is still in flight
2. the attempt has finished but we expect further attempts

metaflow/datastore/task_datastore.py
```diff
@@ -750,6 +762,33 @@ def save_logs(self, logsource, stream_data):
...
+
+        # Legacy log paths
+        legacy_log = self._metadata_name_for_attempt(
+            "%s.log" % stream, attempt_override
+        )
+        path_logsources[legacy_log] = stream
+
+        paths = path_logsources.keys()
+
+        deleted_paths = self._delete_file(paths, add_attempt=False)
```
savingoyal, line 783: do we need to delete? can't we just overwrite?

savingoyal (commented):
added a few comments

metaflow/plugins/logs_cli.py
saikonen: same as in the existing 'show' command. should this be reworded to something else? pathspec for example?

metaflow/plugins/logs_cli.py
saikonen: fixed.

metaflow/plugins/logs_cli.py
saikonen: one approach would be that we instead init the datastore with `allow_not_done=False` ?

metaflow/datastore/task_datastore.py
saikonen: turns out yes, we can simply overwrite. I cleaned up all the delete related additions to the datastore now, instead relying on `save_bytes(..., allow_overwrite=True)` to do the heavy lifting

metaflow/plugins/logs_cli.py
saikonen: bunch of fixes regarding this. How things work now:
we rely on `allow_not_done=False` and check when initializing the datastore in `mode="d"` that the latest / specified attempt is done, otherwise we raise an error.

- trying to `logs scrub [pathspec]` on a running task will lead to an error.
- `logs scrub [pathspec] --attempt x` will work even on a running task, as long as the specified attempt has finished.

saikonen: Note: I also tried simply allowing the scrubbing of logs for running tasks, but due to how logs are written in bursts from a buffer, the effects of overwriting a log with the `[REDACTED]` message were being nullified, as every log write cycle would simply rewrite the scrubbed log lines if they were still in buffer.

metaflow/datastore/flow_datastore.py
```diff
@@ -93,6 +93,8 @@ def get_latest_task_datastores(
         allow_not_done : bool, optional
             If True, returns the latest attempt of a task even if that attempt
             wasn't marked as done, by default False
+        mode : str, optional
```
romain-intel, line 96: nit: this is incorrect in terms of type hint, it should be: "mode : str, default "r"" or something like that.

metaflow/datastore/flow_datastore.py
```diff
@@ -93,6 +93,8 @@ def get_latest_task_datastores(
...
+            Mode to initialize the returned TaskDataStores in. Default "r"
```
romain-intel, line 97: can remove default here.

metaflow/datastore/task_datastore.py
```diff
@@ -173,6 +173,26 @@ def __init__(
...
+                for i in range(metaflow_config.MAX_ATTEMPTS):
+                    check_meta = self._metadata_name_for_attempt(
+                        self.METADATA_ATTEMPT_SUFFIX, i
+                    )
+                    if self.has_metadata(check_meta, add_attempt=False):
+                        self._attempt = i
+
+            # Do not allow destructive operations on the datastore if attempt is still in flight
+            # and we explicitly did not allow operating on running tasks.
+            if not allow_not_done and not self.has_metadata(self.METADATA_DONE_SUFFIX):
```
romain-intel, line 190: is there a case where we want to delete logs for running tasks? It seems in delete mode we may want to ingore `allow_not_done`?

metaflow/plugins/logs_cli.py
```diff
@@ -0,0 +1,306 @@
...
+
+from ..mflog import mflog, LOG_SOURCES
+
+# main motivation from https://github.com/pallets/click/issues/430
+# in order to support a default command being called for a Click group.
+#
+# NOTE: We need this in order to not introduce breaking changes to existing CLI, as we wanted to
+# nest both existing `logs` and the new `logs scrub` under a shared group, but `logs` already has
+# a well defined behavior of showing the logs.
+class CustomGroup(click.Group):
```
romain-intel, line 16: we need to check if this is handled by @madhur-ob 's click parsing thing as well (should be but mentioning).

metaflow/plugins/logs_cli.py
```diff
@@ -0,0 +1,306 @@
...
+    help="Scrub stdout/stderr produced by a task or all tasks in a step. "
+    "The format for input-path is either <run_id>/<step_name> or "
+    "<run_id>/<step_name>/<task_id>."
+)
+@click.argument("input-path")
+@click.option(
+    "--stdout/--no-stdout",
+    default=False,
+    show_default=True,
+    help="Scrub stdout of the task.",
```
romain-intel, line 229: nit: step or task

metaflow/plugins/logs_cli.py
```diff
@@ -0,0 +1,306 @@
...
+    help="Show stdout of the task.",
+)
+@click.option(
+    "--stderr/--no-stderr",
+    default=False,
+    show_default=True,
+    help="Show stderr of the task.",
+)
+@click.option(
+    "--both/--no-both",
```
romain-intel, line 109: nit: I would add a comment (and in the next one) that this is overridden by specifying anything wrt stdout/stderr (I know it wasn't there before but unclear on the priority). Not suggesting changing the API but clarifying behavior.

metaflow/plugins/logs_cli.py
romain-intel: I would definiltely not advocate for allowing the redaction of in-flight logs. THere are just too many complications. We would have to prevent writing to a log that has been redacted which we can't do with past versions of metaflow so there would be this big gapping hole.

metaflow/plugins/logs_cli.py
```diff
@@ -0,0 +1,306 @@
...
+    if ds_list:
+        for ds in ds_list:
+            for stream in streams:
+                ds.scrub_logs(LOG_SOURCES, stream)
+        echo("Logs have been scrubbed.")
+
+    else:
+        raise CommandException(
+            "No Tasks found at the given path -- "
+            "either none exist or none have started yet"
```
romain-intel, line 302: or they haven't finished?

metaflow/plugins/logs_cli.py
```diff
@@ -0,0 +1,306 @@
...
+        )
+
+
+# TODO - move step and init under a separate 'internal' subcommand
```
romain-intel, line 306: ?

metaflow/plugins/logs_cli.py
romain-intel: I like the current behavior described @saikonen -- ie: can scrub past completed attempts with `--attempt`; can scrub latest attempt if finished with no arg and can't scrub any other case.

metaflow/plugins/logs_cli.py
```diff
@@ -0,0 +1,306 @@
...
+        ]
+    else:
+        ds_list = obj.flow_datastore.get_latest_task_datastores(
+            run_id=run_id, steps=[step_name], attempt=attempt, mode="d"
+        )
+
+    if ds_list:
+        for ds in ds_list:
+            for stream in streams:
+                ds.scrub_logs(LOG_SOURCES, stream)
```
romain-intel, line 296: This is a bit annoying in the case some logs get scrubbed but not others. I would do this in a try-catch and do as much as possible but report partial failures (maybe with echo_always?)

metaflow/plugins/logs_cli.py
```diff
@@ -0,0 +1,306 @@
...
+    if stdout:
+        types.add("stdout")
+        both = False
+    if stderr:
+        types.add("stderr")
+        both = False
+    if both:
+        types.update(("stdout", "stderr"))
+
+    streams = list(sorted(types, reverse=True))
```
romain-intel, line 262: what is the rationale behind sorting?

romain-intel (commented):
No issues with the larger feature -- a few comments here and there.

metaflow/datastore/task_datastore.py
saikonen: I left this in mainly in order to keep the datastore side more generic (allowing delete for in-flight attempts as well, though no use-case ). It is definitely not required for this feature, so I could remove it

metaflow/plugins/logs_cli.py
saikonen: leftovers from `cli.py`. Cleaned up.

savingoyal (dismissed):

metaflow/plugins/logs_cli.py
```diff
@@ -0,0 +1,356 @@
...
+    show_default=False,
+    help="Attempt number of a task to scrub, defaults to the latest attempt.",
+)
+@click.option(
+    "--latest/--all",
+    default=True,
+    show_default=False,
+    help="Scrub latest/all attempts of a step or task",
+)
+@click.option(
```
savingoyal, line 256: will this also try to scrub in-flight tasks?

metaflow/plugins/logs_cli.py
saikonen: tested, and this does not work as-is with the low level click api due to the fallback logic happening when invoking the command. Calling the explicit subcommand works as expected though, but the group level fallback fails

```
from metaflow.metaflow_runner import MetaflowAPI
from metaflow.cli import start

if __name__ == "__main__":

    api = MetaflowAPI.from_cli("AttemptLog.py", start)

    # works
    cmd = api().logs().show(input_path="217190/start")
    print(cmd)

    # does not
    cmd = api().logs(input_path="217190/start")
    print(cmd)
```

Will look into how easy of a fix there could be to support it

metaflow/datastore/flow_datastore.py
```diff
@@ -93,6 +100,12 @@ def get_latest_task_datastores(
         allow_not_done : bool, optional
             If True, returns the latest attempt of a task even if that attempt
             wasn't marked as done, by default False
+        attempt : int, optional
+            Attempt number of the tasks to return. Overrides returning the latest TaskDataStore.
```
romain-intel, line 104: I would rephrase as, if not provided, returns latest attempt

metaflow/datastore/flow_datastore.py
```diff
@@ -93,6 +100,12 @@ def get_latest_task_datastores(
...
+        include_prior : boolean, default False
+            Flag to determine if previous attempts TaskDataStores should be returned along the latest one
```
romain-intel, line 106: rephrase as: If True, returns all attempts up to and including attempt.

metaflow/datastore/flow_datastore.py
```diff
@@ -68,7 +68,14 @@ def datastore_root(self):
         return self._storage_impl.datastore_root
 
     def get_latest_task_datastores(
```
romain-intel, line 70: rename as `get_task_datastores`

metaflow/datastore/flow_datastore.py
```diff
@@ -127,7 +140,12 @@ def get_latest_task_datastores(
             ]
         urls = []
         for task_url in task_urls:
-            for attempt in range(metaflow_config.MAX_ATTEMPTS):
+            # parse content urls for specific attempt only, or for all attempts in max range
+            attempt_range = range(metaflow_config.MAX_ATTEMPTS)
+            if attempt is not None:
+                attempt_range = range(attempt + 1) if include_prior else [attempt]
+
```
romain-intel, line 147: I would add a check to see if attempt is less than or equal to MAX_ATTEMPTS - 1

romain-intel (commented):
I like this refactor. I think with a couple name changes, it makes a lot more sense and the code looks clear imho.

metaflow/plugins/logs_cli.py
romain-intel: Cool -- we don't need to fix it for this but ya, we need to probbaly have it working at some point. As in, the low level api should be able to handle all our cases.

metaflow/datastore/flow_datastore.py
saikonen: should we also keep `get_latest_task_datastores` which can now just be a call to `get_task_datastores(..., attempt=None, include_prior=False)`

or is `get_latest_task_datastores` not considered public and we should be fine to ditch it? OSS codebase only has two references to this so the internal rename is simple enough. Slightly concerned if something external is relying on this though.

saikonen: renamed based on discussions.

savingoyal (approved):

PR #1816, opened 2024-04-28
tests with higher level chaining api

madhur-ob:
This should only be merged after https://github.com/Netflix/metaflow/pull/1732

Notes: we need `typeguard` which is apparently not available for Python 3.5 and 3.6

https://pypi.org/project/typeguard/

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,7 +49,7 @@ jobs:
         python3 -m pip install --upgrade pip setuptools
         python3 -m pip install tox numpy
 
-    - name: Execute Python tests with different executors
+    - name: Execute Python tests
       run: tox
 
   R:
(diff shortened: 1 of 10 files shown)


.github/workflows/test.yml
```diff
@@ -43,6 +43,9 @@ jobs:
     - name: Execute Python tests
       run: tox
 
+    - name: Execute Python tests with low level chaining API
+      run: tox -e chaining
```
savingoyal, line 47: it might be better not to introduce another test in the yaml but to embed the test execution in the context

savingoyal: you could introduce an `executor` keyword in the context that takes in two options - `cli` and `api` - similar to `checker`

savingoyal: the tests will go against the high level API and not the low level API but hopefully that's a simple enough change

madhur-ob: done, added lower level api as an executor for now, can switch to higher level later once the changes for UX are in.

savingoyal:
#1859 merged with a fix

Issue #1771, opened 2024-03-25
`python hello.py batch step --help` is broken


madhur-ob:
leads to `TypeError: sequence item 0: expected str instance, NoneType found`

PR #1817, opened 2024-04-29
bug fix: unused arg for batch step --help

mae5357:
addresses: https://github.com/Netflix/metaflow/issues/1771

issue:

`python hello.py batch step --help` command was breaking due to wrong type for click.Choice arg.

change:

setting default arg for `ubf-context` from None to "none" in "batch step" make it compatible with [click.Choice()](https://github.com/mae5357/metaflow/blob/4e648627ea6eb6535c0e28e80ce4daf8f1839ee7/metaflow/_vendor/click/types.py#L137-L138) 

however, ubf-context is not needed for the "step" function when using aws-batch, so we can safely remove altogether.

### DIFF
--- a/metaflow/plugins/aws/batch/batch_cli.py
+++ b/metaflow/plugins/aws/batch/batch_cli.py
@@ -211,7 +211,7 @@ def step(
     log_driver=None,
     log_options=None,
     num_parallel=None,
-    **kwargs,
+    **kwargs
 ):
     def echo(msg, stream="stderr", batch_id=None, **kwargs):
         msg = util.to_unicode(msg)
--- a/metaflow/plugins/aws/batch/batch_cli.py
+++ b/metaflow/plugins/aws/batch/batch_cli.py
@@ -151,7 +151,7 @@ def kill(ctx, run_id, user, my_runs):
 @click.option("--tmpfs-size", help="tmpfs requirement for AWS Batch.")
 @click.option("--tmpfs-path", help="tmpfs requirement for AWS Batch.")
 # TODO: Maybe remove it altogether since it's not used here
-@click.option("--ubf-context", default=None, type=click.Choice([None, "ubf_control"]))
+@click.option("--ubf-context", default=None, type=click.Choice(["none", "ubf_control"]))
 @click.option("--host-volumes", multiple=True)
 @click.option("--efs-volumes", multiple=True)
 @click.option(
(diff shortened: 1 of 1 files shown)


madhur-ob (approved):

romain-intel:
If you want to keep the functionality, the proper line should read something like: 
```
type=click.Choice(["none", UBF_CONTROL, UBF_TASK]),
```
I was under the impression that ubf may be used by parallel so may be worth keeping around.

mae5357:
@romain-intel i can revert to if you want https://github.com/Netflix/metaflow/pull/1817/commits/244ff31c9c81ac811ef4b4499c4f819e74a028f3

> I was under the impression that ubf may be used by parallel so may be worth keeping around.


fwiw, the `step()` function this is decorating doesn't take `ubf-context` as an argument. So there's not a possibility this arg is being used ([here](https://github.com/Netflix/metaflow/blob/4e648627ea6eb6535c0e28e80ce4daf8f1839ee7/metaflow/plugins/aws/batch/batch_cli.py#L210)).

saikonen:
Finally had time to test this, and as Romain mentioned, `@parallel` somewhat depends on the option existing, even if it is not explicitly used. Removing the option completely breaks parallel deco on batch. 

Can you instead revert to the commit you mentioned? Also replace the TODO with for example the following:

```
# NOTE: ubf-context is not explicitly used, but @parallel decorator tries to pass this so keep it for now.
```

metaflow/plugins/aws/batch/batch_cli.py
```diff
@@ -150,8 +150,6 @@ def kill(ctx, run_id, user, my_runs):
 @click.option("--tmpfs-tempdir", is_flag=True, help="tmpfs requirement for AWS Batch.")
 @click.option("--tmpfs-size", help="tmpfs requirement for AWS Batch.")
 @click.option("--tmpfs-path", help="tmpfs requirement for AWS Batch.")
-# TODO: Maybe remove it altogether since it's not used here
-@click.option("--ubf-context", default=None, type=click.Choice([None, "ubf_control"]))
```
saikonen, line 154: removing this completely leads to `@parallel` breaking on batch. Keep the original fix instead and change comment to

```
# NOTE: ubf-context is not explicitly used, but @parallel decorator tries to pass this so keep it for now.
```

saikonen (changes requested):

saikonen (approved):

Issue #350, opened 2020-10-14
Add a `@local` decorator


romain-intel:
Currently you can use `@batch` to decorate single steps that you want to force to run on batch even without using `--with batch`. However, you cannot do the opposite and force steps to run locally even when doing `--with batch`.

A possible use-case for this would be to annotate a GPU-heavy step that you want to run locally to save on AWS costs but take advantage of fast S3 access for other steps (from the EC2 machines batch uses).

It is currently possible to have this use case by annotating every other step with `@batch` but this is not very developer friendly (you can't easily run your code locally to test, it's tedious, etc).

Relevant discussion on gitter:
https://gitter.im/metaflow_org/community?at=5f85f41d24a20801a8e5373d

Other questions to address when touching decorators:
https://gitter.im/metaflow_org/community?at=5f7f81e024a20801a8d509a6
https://gitter.im/metaflow_org/community?at=5f05d3baa5ab931e4f7280e3

romain-intel:
This one should be simple and useful.

vardhaman-surana:
hi @romain-intel i want to work on this issue.
I was exploring the code and it seems to me that the intended @local decorator will just prevent @batch and @kubernetes decorators from being added to a step in case it is present. Am i thinking in the right direction, please guide me with the correct solution if my suggestion is wrong.

jakharmonika364:
I am working on it.

PR #1824, opened 2024-05-06
added local decorator

vardhaman-surana:
for issue: https://github.com/Netflix/metaflow/issues/350

1. added a new decorator: `@local`
2. this will force steps to run locally even when doing `--with batch` and  `--with kubernetes`, by preventing attachment of **batch** and **kubernetes** decorators to the steps which have `@local` attached to them already.
to achieve this:
   - a classmethod should_attach is introduced in StepDecorator class which will always return True
   - that method is overridden in batch and kubernetes decorator classes to check for presence of local decorator on the step
   - the logic to check should_attach condition is implemented in _attach_decorators_to_step function inside decorators.py file

### DIFF
--- a/metaflow/decorators.py
+++ b/metaflow/decorators.py
@@ -325,6 +325,14 @@ class StepDecorator(Decorator):
         """
         pass
 
+    @classmethod
+    def should_attach(cls, step):
+        """
+        Can be used to check if the decorator can be attached to the step or not.
+        returns True by default
+        """
+        return True
+
     def task_pre_step(
         self,
         step_name,
@@ -481,6 +489,11 @@ def _attach_decorators_to_step(step, decospecs):
         deconame = splits[0]
         if deconame not in decos:
             raise UnknownStepDecoratorException(deconame)
+
+        # Skip attaching decorator if should_attach returns False.
+        if not decos[deconame].should_attach(step):
+            continue
+
         # Attach the decorator to step if it doesn't have the decorator
         # already. This means that statically defined decorators are always
         # preferred over runtime decorators.
--- a/metaflow/plugins/__init__.py
+++ b/metaflow/plugins/__init__.py
@@ -29,6 +29,7 @@ STEP_DECORATORS_DESC = [
     ("resources", ".resources_decorator.ResourcesDecorator"),
     ("batch", ".aws.batch.batch_decorator.BatchDecorator"),
     ("kubernetes", ".kubernetes.kubernetes_decorator.KubernetesDecorator"),
+    ("local", ".local_decorator.LocalDecorator"),
     (
         "argo_workflows_internal",
         ".argo.argo_workflows_decorator.ArgoWorkflowsInternalDecorator",
--- a/metaflow/plugins/aws/batch/batch_decorator.py
+++ b/metaflow/plugins/aws/batch/batch_decorator.py
@@ -234,6 +234,11 @@ class BatchDecorator(StepDecorator):
             if not R.use_r():
                 cli_args.entrypoint[0] = sys.executable
 
+    @classmethod
+    def should_attach(cls, step):
+        # return False if local decorator is attached to the step already.
+        return not any(deco.name == "local" for deco in step.decorators)
+
     def task_pre_step(
         self,
         step_name,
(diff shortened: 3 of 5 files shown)


metaflow/decorators.py
```diff
@@ -481,6 +483,12 @@ def _attach_decorators_to_step(step, decospecs):
         deconame = splits[0]
         if deconame not in decos:
             raise UnknownStepDecoratorException(deconame)
+
```
romain-intel, line 486: I have a few issues with this and I don't, unfortunately, have an ideal solution for it yet. I don't want to put specific decorators here (in core). There are other decorators that local could impact (for example, at Netflix, we have a `@titus` decorator). Ideally we would have a better way for decorators to communicate between each other (or something).

One thing that I think may be better (but @savingoyal could also comment) is actually putting stuff in the batch and kubernetes decorator and have them "disable" themselves if they see that local is present. I think that type of change may be more palatable and puts less stuff in the core.

vardhaman-surana: I can work on implementing a solution to have batch and Kubernetes operators disable themselves when local decorator  is present. Should I proceed with this approach?

vardhaman-surana: also if in the current solution if i don't put condition directly in the attach_decorators function and add a method (let's say should_attach(step)) in base StepDecorator which always return true.
and override that in BatchDecorator and KubernetesDecorator class  to check for the local decorator presence...
would that be acceptable ?

romain-intel: I'll let @savingoyal chime in. I think the least intrusive one is just:
- add a local decorator
- have the batch/kube/etc decorators check for the presence of the local decorator and disable themselves if present

This doesn't add much to core (apart from the local decorator) and we can then properly evolve it. We are looking at redesigning the decorator interfaces as well so this can probably be a part of that. In the meantime, this seems like a small enough solution. We could even get a tad fancier and check if the decorators are static or not (added with --with batch) and error if there is local and batch statically there.

vardhaman-surana: @savingoyal should i proceed with adding should_attach function in stepDecorator ?
@romain-intel i was trying to disable batch and kubernetes decorators from within themselves but it seems that approach will require changes in each function of those decorators which are called as per the workflow of decorators.

vardhaman-surana:
@romain-intel  @savingoyal I have updated the PR with the should attach class method approach. Please review once.

savingoyal:
@vardhaman-surana, thanks for the PR. I would like to understand the use case better. Currently, we have `@resources` and `@batch`/`@kubernetes` decorators (and more coming). The attributes for `@resources` (CPU, mem, etc.) are a strict subset of `@batch`/`@kubernetes` today and are effectively ignored unless used `--with batch/kubernetes`. If the intention is that `@local` simply ignores the `@batch`/`@kubernetes` decorator - the step may not execute as expected if `@batch`/`@kubernetes` were providing a specific `image` or fiddling with `tempfs` settings, for example.

vardhaman-surana:
> @vardhaman-surana, thanks for the PR. I would like to understand the use case better. Currently, we have `@resources` and `@batch`/`@kubernetes` decorators (and more coming). The attributes for `@resources` (CPU, mem, etc.) are a strict subset of `@batch`/`@kubernetes` today and are effectively ignored unless used `--with batch/kubernetes`. If the intention is that `@local` simply ignores the `@batch`/`@kubernetes` decorator - the step may not execute as expected if `@batch`/`@kubernetes` were providing a specific `image` or fiddling with `tempfs` settings, for example.
> 
> One possible workaround is to maybe have `METAFLOW_DISABLE_DECOSPECS=kubernetes` as an environment variable to disable all references of @kubernetes without providing an explicit @Local decorator.

hi @savingoyal the use case as mentioned in the [issue](https://github.com/Netflix/metaflow/issues/350) is

**A possible use-case for this would be to annotate a GPU-heavy step that you want to run locally to save on AWS costs but take advantage of fast S3 access for other steps (from the EC2 machines batch uses).**

i tried adding the local decorator based on this requirement mentioned in the issue. I am not sure if the env variable approach which you are suggesting will work for this case or not.

PR #1837, opened 2024-05-15
Add the possibility of defining default decorators for steps

romain-intel:
Default decorators can be defined in one of two ways:
  - using the `METAFLOW_DEFAULT_STEP_DECORATORS` configuration (env var or configuration value). If this is set, it will be applied regardless of anything else. This is a comma separated string.
  - having extensions use TOGGLE_DEFAULT_STEP_DECORATORS. This is a list of strings that:
    - start with + or - (or something else which is equivalent to a +) -- the + means add a default decorator and a - means remove one.
    - are a decospec for the decorator to be added or removed.

I will add tests and try it out but putting it here for initial review to see if the method is OK.

### DIFF
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -137,7 +137,7 @@ def config_merge_cb(ctx, param, value):
     splits = DECOSPECS.split()
     if len(splits) == len(value) and all([a == b for (a, b) in zip(splits, value)]):
         return value
-    return tuple(DECOSPECS.split() + list(value))
+    return tuple(list(value) + DECOSPECS.split())
 
 
 @click.group()
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -640,7 +640,7 @@ def resume(
     resume_identifier=None,
     runner_attribute_file=None,
 ):
-    before_run(obj, tags, decospecs + obj.environment.decospecs())
+    before_run(obj, tags, decospecs)
 
     if origin_run_id is None:
         origin_run_id = get_latest_run_id(obj.echo, obj.flow.name)
@@ -738,7 +738,7 @@ def run(
 ):
     if user_namespace is not None:
         namespace(user_namespace or None)
-    before_run(obj, tags, decospecs + obj.environment.decospecs())
+    before_run(obj, tags, decospecs)
 
     runtime = NativeRuntime(
         obj.flow,
@@ -786,9 +786,20 @@ def before_run(obj, tags, decospecs):
     # A downside is that we need to have the following decorators handling
     # in two places in this module and make sure _init_step_decorators
     # doesn't get called twice.
-    if decospecs:
-        decorators._attach_decorators(obj.flow, decospecs)
+
+    # We want the order to be the following:
+    # - run level decospecs
+    # - top level decospecs
+    # - environment decospecs
+    all_decospecs = (
+        list(decospecs or [])
+        + obj.tl_decospecs
+        + list(obj.environment.decospecs() or [])
+    )
+    if all_decospecs:
+        decorators._attach_decorators(obj.flow, all_decospecs)
         obj.graph = FlowGraph(obj.flow.__class__)
+
     obj.check(obj.graph, obj.flow, obj.environment, pylint=obj.pylint)
     # obj.environment.init_environment(obj.logger)
 
@@ -978,8 +989,11 @@ def start(
         deco_options,
     )
 
-    if decospecs:
-        decorators._attach_decorators(ctx.obj.flow, decospecs)
+    # In the case of run/resume, we will want to apply the TL decospecs
+    # *after* the run decospecs so that they don't take precedence. In other
+    # words, for the same decorator, we want `myflow.py run --with foo` to
+    # take precedence over any other `foo` decospec
+    ctx.obj.tl_decospecs = list(decospecs or [])
 
     # initialize current and parameter context for deploy-time parameters
     current._set_env(flow=ctx.obj.flow, is_running=False)
@@ -990,7 +1004,14 @@ def start(
     if ctx.invoked_subcommand not in ("run", "resume"):
         # run/resume are special cases because they can add more decorators with --with,
         # so they have to take care of themselves.
-        decorators._attach_decorators(ctx.obj.flow, ctx.obj.environment.decospecs())
+        all_decospecs = ctx.obj.tl_decospecs + list(
+            ctx.obj.environment.decospecs() or []
+        )
+        if all_decospecs:
+            decorators._attach_decorators(ctx.obj.flow, all_decospecs)
+            # Regenerate graph if we attached more decorators
+            ctx.obj.graph = FlowGraph(ctx.obj.flow.__class__)
+
         decorators._init_step_decorators(
             ctx.obj.flow,
             ctx.obj.graph,
@@ -998,6 +1019,7 @@ def start(
             ctx.obj.flow_datastore,
             ctx.obj.logger,
         )
+
         # TODO (savin): Enable lazy instantiation of package
         ctx.obj.package = None
     if ctx.invoked_subcommand is None:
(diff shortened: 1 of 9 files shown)


metaflow/decorators.py
```diff
@@ -158,13 +158,14 @@ def make_decorator_spec(self):
...
                     attr_list.append("%s=%s" % (k, json.dumps(v).replace('"', '\\"')))
+
             attrstr = ",".join(attr_list)
             return "%s:%s" % (self.name, attrstr)
         else:
             return self.name
 
     def __str__(self):
-        mode = "decorated" if self.statically_defined else "cli"
+        mode = "decorated" if self.statically_defined else "cli or configuration"
```
savingoyal, line 168: maybe `undecorated`?

romain-intel: I find "undecorated" to be less specific. It doesn't really give much information. What do you think of "injected". We could also have "static" and "dynamic" simply. Wdyt?

metaflow/cli.py
```diff
@@ -837,6 +859,7 @@ def version(obj):
     multiple=True,
     help="Add a decorator to all steps. You can specify this option "
     "multiple times to attach multiple decorators in steps.",
+    callback=use_conf_and_merge_cb,
```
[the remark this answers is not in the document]
romain-intel: Don't have a huge preference but my intent was to make this function useful for anything that comes from conf too. Maybe `use_conf_for_multiple_cb`. It also doesn't apply to all decospecs for example.

metaflow/decorators.py
savingoyal: yep - static and dynamic works well.

metaflow/cli.py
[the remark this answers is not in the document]
savingoyal: maybe just `config_callback`?

savingoyal (approved):

saikonen (commented):
not sure on the cause, as no changes touch this codepath directly:
when running a flow remotely with f.ex. `--with kubernetes`, I'm getting to following in stdout
```
'nbrun' requires an interactive python environment (such as Jupyter)
```

metaflow/cli.py
```diff
@@ -118,6 +120,26 @@ def logger(body="", system_msg=False, head="", bad=False, timestamp=True, nl=Tru
...
+    # NOTE: Assumes that ctx.auto_envvar_prefix is set to METAFLOW (same as in
+    # from_conf)
+
+    # Special case where DECOSPECS and value are the same. This happens
+    # when there is no --with option at the TL and DECOSPECS is read from
+    # the env var. In this case, click also passes it as value
+    splits = DECOSPECS.split()
+    if len(splits) == len(value) and all([a == b for (a, b) in zip(splits, value)]):
+        return value
+    return tuple(DECOSPECS.split() + list(value))
```
saikonen, line 140: Maybe this was decided already before, but noting some oddities regarding the hierarchy in the config.

test setup:
```
# metaflow config.json
"METAFLOW_DECOSPECS": "kubernetes:memory=3072"

# cmd
python HelloFlow.py run --with "kubernetes:memory=1024,cpu=2"
```
results in the higher memory value being used, and cpu omitted.

--- 
same setup with additional env var set
```
METAFLOW_DECOSPECS="kubernetes:memory=2048"
```
results in env var overriding value from config.json (as expected?)

---
directly supplying values in a deco seems to override everything else as expected
```python
@kubernetes(memory=512, cpu=3)
```

---

so the succession right now seems to be 
```
explicit decorator in flow file
> METAFLOW_DECOSPEC env
> METAFLOW_DECOSPEC in config.json
> CLI --with
```

saikonen: the only thing I found odd with the succession is treating an explicit CLI `--with` secondary when the config/envvar values are advertised as setting defaults only. My expectation would be that an option passed through CLI would perhaps override defaults

Disregard if this is intended behaviour :)

romain-intel:
@saikonen -- could you try again. I rebased on master and addressed @savingoyal 's comments. I suspect the error you are getting had to do with the previous changes for the runner and I didn't take all the "fixes" and master was, at some point, in a bad state.

saikonen:
> @saikonen -- could you try again. I rebased on master and addressed @savingoyal 's comments. I suspect the error you are getting had to do with the previous changes for the runner and I didn't take all the "fixes" and master was, at some point, in a bad state.

seems to be fixed now 👍

metaflow/cli.py
romain-intel: This is indeed the behavior that is coded. There is no notion of "decorator merge" although there should be. For now though, I think this is good enough but going forward we should see how we can merge decorators. This is the same problem as the @environment decorator where one overrides all others (and they don't merge). There is also no warning of duplicate decorators because they are all non static. In short, we have not changed the behavior at all but it can be a bit weird. You can try for example `--with kubernetes:memory=2048 --with kubernetes:memory=1024,cpu=2`, the second one will be ignored just like it is right now. You can view the default decospecs as stuff that is prepended basically. Am happy to make it post pended though if ppl prefer, as in:
explicit deco in flow file
CLI --with
METAFLOW_DECOSPEC env
METAFLOW_DECOSPEC in config.json.

Given we have no promises on any ordering right now, I don't view that as a huge issue and I think we need to more cleanly address it but that would be for another set of changes.

metaflow/cli.py
```diff
@@ -837,6 +859,7 @@ def version(obj):
     multiple=True,
     help="Add a decorator to all steps. You can specify this option "
     "multiple times to attach multiple decorators in steps.",
+    callback=config_merge_cb,
```
saikonen, line 862: also need the callback for the `run` subcommand I guess? 
right now this works
```
METAFLOW_DECOSPECS="kubernetes:memory=512" python HelloFlow.py --with "kubernetes:memory=4096,cpu=2" run
```
but this does not
```
METAFLOW_DECOSPECS="kubernetes:memory=512" python HelloFlow.py run --with "kubernetes:memory=4096,cpu=2"
```

metaflow/metaflow_config.py
```diff
@@ -519,8 +529,18 @@ def _new_get_pinned_conda_libs(
                     return d1
 
                 globals()[n] = _new_get_pinned_conda_libs
+            elif n == "TOGGLE_DECOSPECS":
+                if any([x.startswith("-") for x in o]):
+                    raise ValueError("Removing decospecs is not currently supported")
+                if any(" " in x for x in o):
+                    raise ValueError("Decospecs cannot contain spaces")
+                _TOGGLE_DECOSPECS.extend(o)
```
saikonen, line 537: what's the latest on this? it seems we can only add decospecs for now through extensions, not remove them.

The usage is a bit different from the PR description as well, not requiring prepending a `+` for adding a decospec. This makes sense for now as `-` is not supported, but should we agree on the final form already, or is there a use case for this yet?

saikonen: for adding decospecs the mechanism is working as expected 👍

metaflow/cli.py
saikonen: fixed with latest commit :)

saikonen (approved):

metaflow/metaflow_config.py
romain-intel: @savingoyal had me remove the - option since he thought it would be confusing. It can easily be added back if we feel the need for it later (I don't need it right now, I had initially done the - to keep in line with the TOGGLE_STEP_DECORATOR for example (but that one I really needed the -)).

saikonen: was more so wondering about the prefixes in general, as currently `+` is not supported either, so do we even need a check against `-`?

not working at the moment. Is this intended?
```
TOGGLE_DECOSPECS = ["+kubernetes:memory=1024"]
```

this works fine
```
TOGGLE_DECOSPECS = ["kubernetes:memory=1024"]
```
---
as its currently append-only, naming the var `toggle` is a bit confusing. We can keep it as is though if the end goal is to support toggling the decospecs.

PR #1845, opened 2024-05-21
synchronous run and resume functionality + nbrun for runner API

madhur-ob:
Should only be merged after https://github.com/Netflix/metaflow/pull/1732

This allows us to do stuff like:

```
    runner = Runner('../try.py', metadata="local")
    result = runner.run(alpha=5, tags=["abc", "def"], max_workers=5)

    while not result.run.finished:
        print('trying again...')
        result = runner.resume()
```

Also allows us to use `run` inside environments which already have a running event loop such as Jupyter notebooks.

### DIFF
--- a/metaflow/runner/nbrun.py
+++ b/metaflow/runner/nbrun.py
@@ -1,6 +1,8 @@
+import os
 import ast
 import shutil
 import tempfile
+from typing import Optional, Dict
 from metaflow import Runner
 
 try:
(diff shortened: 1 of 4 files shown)


metaflow/subprocess_manager.py
```diff
@@ -0,0 +1,399 @@
...
+
+    def run(self):
+        if not self.run_called:
+            self.temp_dir = tempfile.mkdtemp()
+            stdout_logfile = os.path.join(self.temp_dir, "stdout.log")
+            stderr_logfile = os.path.join(self.temp_dir, "stderr.log")
+
+            def stream_to_stdout_and_file(pipe, log_file):
+                with open(log_file, "w") as file:
+                    for line in iter(pipe.readline, ""):
```
tuulos, line 146: @madhur-ob let's add a `show_output=True` flag in `run` so you can suppress outputs optionally. Here it outputs both the `stdout` and `stderr` of the subprocess to `stdout` always, which is not always preferable.

let's do
```
if  show_output:
    sys.stdout.write(line)
```
even if `show_output=False`, you can retrieve the complete logs from `result.stdout` and `result.stderr`

madhur-ob: Thanks, I have updated it...

savingoyal (approved):

metaflow/runner/nbrun.py
```diff
@@ -0,0 +1,131 @@
...
+import tempfile
+from typing import Optional, Dict
+from metaflow import Runner
+
+try:
+    from IPython import get_ipython
+
+    ipython = get_ipython()
+except ModuleNotFoundError:
+    print("'nbrun' requires an interactive python environment (such as Jupyter)")
```
savingoyal, line 13: 'NBRunner'?

PR #1856, opened 2024-05-23
feature: stream logs in chunks

saikonen:
attempts to read logs with a fixed byte offset instead of all at once. Motivation is to decrease memory consumption when handling massive log files.

### DIFF
--- a/metaflow/plugins/datastores/s3_storage.py
+++ b/metaflow/plugins/datastores/s3_storage.py
@@ -2,7 +2,13 @@ import os
 
 from itertools import starmap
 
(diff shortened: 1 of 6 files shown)


metaflow/plugins/datastores/s3_storage.py
```diff
@@ -144,3 +150,34 @@ def iter_results():
                         yield r.key, None, None
 
         return CloseAfterUse(iter_results(), closer=s3)
+
+    def stream_bytes(self, paths, chunk_size=2**1024):
```
saikonen, line 154: default chunk_size choice is arbitrary here, mainly for testing purposes. For the UI Backend case this should probably be made configurable.

savingoyal (approved):

PR #1857, opened 2024-05-24
Remove step level decospecs

romain-intel:
This was not used anywhere and step is an internal command. Also added clarifying comments around attaching decospecs


saikonen:
fixes merged in #2203

PR #1863, opened 2024-05-27
add explicit cleanup() methods in Runners

tuulos:
so user can clean up results without context managers

### DIFF
--- a/metaflow/runner/nbrun.py
+++ b/metaflow/runner/nbrun.py
@@ -97,7 +97,7 @@ class NBRunner(object):
         self.runner.show_output = True
         result = self.runner.run(**kwargs)
         self.runner.show_output = self.old_val_show_output
-        self.runner.spm.cleanup()
+        self.cleanup()
         return result.run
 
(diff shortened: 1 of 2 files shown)


metaflow/runner/nbrun.py
```diff
@@ -97,15 +97,15 @@ def nbrun(self, **kwargs):
...
+        self.cleanup()
         return result.run
 
     def nbresume(self, **kwargs):
         self.old_val_show_output = self.show_output
         self.runner.show_output = True
         result = self.runner.resume(**kwargs)
         self.runner.show_output = self.old_val_show_output
-        self.runner.spm.cleanup()
+        self.cleanup()
```
madhur-ob, line 108: Should we do `os.remove(self.tmp_flow_file.name)` in here as well?

madhur-ob: aah sorry, I confused it for `self.runner.cleanup()` in a hurry, approving..

madhur-ob (approved):

PR #1864, opened 2024-05-27
allow output to be hidden in nbrun()

tuulos:


metaflow/runner/nbrun.py
```diff
@@ -50,7 +50,7 @@ class NBRunner(object):
     def __init__(
         self,
         flow,
-        show_output: bool = False,
+        show_output: bool = True,
```
madhur-ob, line 53: So, the `self.runner` is used in other places too such as `run`, `resume`, `async_run` and `async_resume` -- and the default way here should be to suppress outputs.

It is only for `nbrun` and `nbresume` methods that we want to show outputs by default.

Setting the default value to True will make it show for the `run`, `resume`, `async_run` and `async_resume` methods as well, which we don't want..

tuulos: we want to show output by default for `run` and `resume`. For `async_` calls, `show_output` doesn't have any effect, right?

madhur-ob: yeah, show output should have no effect on `async_`

madhur-ob (approved):

PR #1878, opened 2024-06-05
use ast instead of loading for injecting user params

madhur-ob:


savingoyal (approved):

metaflow/runner/click_api.py
```diff
@@ -136,31 +136,76 @@ def get_inspect_param_obj(p: Union[click.Argument, click.Option], kind: str):
...
+parsed_modules = {}
+
+
+class FlowSpecVisitor(ast.NodeVisitor):
+    def __init__(self):
+        self.parameters = []
+
+    def construct_parameter(self, node):
+        all_values = {}
+        fields = ["name", "default", "type", "help", "required", "show_default"]
```
romain-intel, line 149: we use other parameter values so would need something a bit more generic I believe.

romain-intel (commented):
I think we need to find another way to do this. The main reason being that we override parameters and would at the very least have to override this function as well which feels a bit icky. I am also not 100% comfortable with the AST parsing and feel we may miss cases and spend time chasing things (could be wrong but feels less robust).

The main issue is that FlowSpec.__init__ sets the global parameters. Maybe we can focus on that and not do that if we are loading just to evaluate the class. Thoughts?

PR #1881, opened 2024-06-05
Cosmetic changes for decorator handling in cli.py

savingoyal:
can be safely ignored

### DIFF
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -796,17 +796,17 @@ def before_run(obj, tags, decospecs):
     # in two places in this module and make sure _init_step_decorators
     # doesn't get called twice.
 
-    # We want the order to be the following:
-    # - run level decospecs
-    # - top level decospecs
-    # - environment decospecs
-    all_decospecs = (
+    # decospecs have this precedence order:
+    # - run-level
+    # - top-level
+    # - environment
+    decospecs = (
         list(decospecs or [])
-        + obj.tl_decospecs
+        + obj.decospecs
         + list(obj.environment.decospecs() or [])
     )
-    if all_decospecs:
-        decorators._attach_decorators(obj.flow, all_decospecs)
+    if decospecs:
+        decorators._attach_decorators(obj.flow, decospecs)
         obj.graph = FlowGraph(obj.flow.__class__)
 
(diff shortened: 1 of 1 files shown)


metaflow/cli.py
```diff
@@ -901,6 +899,12 @@ def version(obj):
     type=click.Choice(MONITOR_SIDECARS),
     help="Monitoring backend type",
 )
+@click.option(
```
romain-intel, line 902: I'm not a fan of having options cause other options to be ignored with no warning. I would add a check to see if there are conflicting options. `dry-run` is also a bit misleading (it's not dry run and things may happen, for example conda package caching, anything the code does as well, etc).

metaflow/cli.py
```diff
@@ -1013,11 +1021,9 @@ def start(
     if ctx.invoked_subcommand not in ("run", "resume"):
         # run/resume are special cases because they can add more decorators with --with,
         # so they have to take care of themselves.
-        all_decospecs = ctx.obj.tl_decospecs + list(
-            ctx.obj.environment.decospecs() or []
-        )
-        if all_decospecs:
-            decorators._attach_decorators(ctx.obj.flow, all_decospecs)
+        decospecs = list(decospecs or []) + list(ctx.obj.environment.decospecs() or [])
```
romain-intel, line 1024: nit, use ctx.obj.decospecs (no point redoing the same computation and conversion)

metaflow/cli.py
```diff
@@ -998,11 +1008,9 @@ def start(
         deco_options,
     )
 
-    # In the case of run/resume, we will want to apply the TL decospecs
-    # *after* the run decospecs so that they don't take precedence. In other
-    # words, for the same decorator, we want `myflow.py run --with foo` to
-    # take precedence over any other `foo` decospec
-    ctx.obj.tl_decospecs = list(decospecs or [])
+    # run/resume can add more decorators with --with, which take precedence
```
romain-intel, line 1011: am fine with this (and associated change) but would re-add the comment about how precedence is determined by order right now (first one wins) so we need to save to apply in before_run. (ie: this comment is correct but doesn't say why we do it this way).

PR #1884, opened 2024-06-11
[Ready for Review] Improve native resume 

darinyu:
Extend the resume speedup change to native(local) resume.

Note:
* We will go over all the successful tasks and copy them first. Then we reconstruct runtime queue to continue.
* We will keep UBF resume behavior the same. i.e if some mapper tasks fail, all mapper tasks will reran in the resume.
     
A few test example flows:
* Resume w/ some branch: [link](https://gist.github.com/darinyu/b895a8aca64f4674a147475763a50a85)
* Resume during failure on foreach split :[link](https://gist.github.com/darinyu/5ca42e7bad8e984f6bff942e1f064ff4)
* Resume during failure on UBF split: [link](https://gist.github.com/darinyu/aaf604a9e1e22c51c958135974caa27a)
* Resume during failure on UBF join: [link](https://gist.github.com/darinyu/1393a506b77b874c0cffeadaac2bfb5d)

To test the above flows:
```python
python flow.py run
python flow.py resume
```

Benchmark:
Resume-Speed-Test-Google-Docs

Open question:
* We only consider last "run" but never consider last "resume", so running multiple resume will anchor on the same run_id (instead of previous resume). If running resume consecutively, should we continue on the resume?

### DIFF
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -717,7 +717,7 @@ def resume(
         if clone_only:
             runtime.clone_original_run()
         else:
-            runtime.clone_original_run(generate_task_obj=True)
+            runtime.clone_original_run(generate_task_obj=True, verbose=False)
             runtime.execute()
 
 
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -710,6 +710,7 @@ def resume(
     if clone_only:
         runtime.clone_original_run()
     else:
+        runtime.clone_original_run(generate_task_obj=True)
         runtime.execute()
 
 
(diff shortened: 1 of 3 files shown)


savingoyal:
@darinyu can you walk me through the nature of changes for this PR so that we can have a quick review turnaround?

savingoyal:
not related to this PR but maybe we can mute [this](https://github.com/Netflix/metaflow/blob/86a6991086af3edf76648d25c5caf07c7475df9b/metaflow/runtime.py#L963) log line

savingoyal:
ready to go

PR #1886, opened 2024-06-14
Remove global nature of parameters and flow decorators

romain-intel:
This scopes decorators and parameters to the flow itself which should improve compatibility with the runner

### DIFF
--- a/metaflow/R.py
+++ b/metaflow/R.py
@@ -129,7 +129,7 @@ def run(
     from . import cli
 
     try:
-        with parameters.with_flow(flow) as _:
+        with parameters.flow_context(flow.__class__) as _:
             cli.main(
                 flow,
                 args=metaflow_args,
--- a/metaflow/R.py
+++ b/metaflow/R.py
@@ -3,6 +3,7 @@ import sys
 from importlib import util as imp_util, machinery as imp_machinery
 from tempfile import NamedTemporaryFile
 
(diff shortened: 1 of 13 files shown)


romain-intel:
Duh -- I think I need to import the cli inside the context in the R code. I'll make the change.

metaflow/runner/click_api.py
```diff
@@ -15,6 +15,7 @@
 from collections import OrderedDict
 from typing import Any, Callable, Dict, List, Optional
 from typing import OrderedDict as TOrderedDict
+from typing import Tuple as TTuple
```
romain-intel, line 18: This can be removed.

metaflow/R.py
```diff
@@ -128,12 +129,13 @@ def run(
     from . import cli
```
romain-intel, line 129: This needs to move into the flow_context below.

metaflow/parameters.py
```diff
@@ -391,6 +419,10 @@ def wrapper(cmd):
         cmd.has_flow_params = True
```
madhur-ob, line 419: `cmd.has_flow_params = True` is used to determine where to dynamically inject user defined parameters...

they are to be injected in commands like `run` and sub-commands such as `trigger` (for `argo-workflows`, etc.)
Note: this is just a comment that I will be referencing later...

PR #1900, opened 2024-06-25
[@parallel][jobsets] bug fixes and refactor

valayDave:
- Abstraction over Jobset Spec
- Create general abstraction and copy style of jobspec for native k8s jobs for jobsets
- simplified the implementation.
- similar pattern/trend to follow for argo
- removed older implementation
- Superseds #1854 

Review commits individually to see core changes. 
- One commit has the entire refactor into a more decoupled implementation. 
- One commit has fixes to the core such as ID minting for control jobs.

### DIFF
--- a/metaflow/plugins/kubernetes/kubernetes.py
+++ b/metaflow/plugins/kubernetes/kubernetes.py
@@ -1,9 +1,9 @@
+import copy
 import json
 import math
 import os
 import re
 import shlex
-import copy
 import time
 from typing import Dict, List, Optional
 from uuid import uuid4
@@ -14,10 +14,11 @@ from metaflow.metaflow_config import (
     ARGO_EVENTS_EVENT,
     ARGO_EVENTS_EVENT_BUS,
     ARGO_EVENTS_EVENT_SOURCE,
-    ARGO_EVENTS_SERVICE_ACCOUNT,
     ARGO_EVENTS_INTERNAL_WEBHOOK_URL,
-    AWS_SECRETS_MANAGER_DEFAULT_REGION,
+    ARGO_EVENTS_SERVICE_ACCOUNT,
     ARGO_EVENTS_WEBHOOK_AUTH,
+    AWS_SECRETS_MANAGER_DEFAULT_REGION,
+    AZURE_KEY_VAULT_PREFIX,
     AZURE_STORAGE_BLOB_SERVICE_ENDPOINT,
     CARD_AZUREROOT,
     CARD_GSROOT,
@@ -31,18 +32,18 @@ from metaflow.metaflow_config import (
     DEFAULT_METADATA,
     DEFAULT_SECRETS_BACKEND_TYPE,
     GCP_SECRET_MANAGER_PREFIX,
-    AZURE_KEY_VAULT_PREFIX,
     KUBERNETES_FETCH_EC2_METADATA,
     KUBERNETES_LABELS,
     KUBERNETES_SANDBOX_INIT_SCRIPT,
-    S3_ENDPOINT_URL,
-    SERVICE_HEADERS,
-    SERVICE_INTERNAL_URL,
-    S3_SERVER_SIDE_ENCRYPTION,
     OTEL_ENDPOINT,
+    S3_ENDPOINT_URL,
+    S3_SERVER_SIDE_ENCRYPTION,
+    SERVICE_HEADERS,
+    KUBERNETES_SECRETS,
+    SERVICE_INTERNAL_URL,
 )
+from metaflow.unbounded_foreach import UBF_CONTROL, UBF_TASK
 from metaflow.metaflow_config_funcs import config_values
-
 from metaflow.mflog import (
     BASH_SAVE_LOGS,
     bash_capture_logs,
@@ -60,6 +61,10 @@ STDERR_FILE = "mflog_stderr"
 STDOUT_PATH = os.path.join(LOGS_DIR, STDOUT_FILE)
 STDERR_PATH = os.path.join(LOGS_DIR, STDERR_FILE)
 
+METAFLOW_PARALLEL_STEP_CLI_OPTIONS_TEMPLATE = (
+    "{METAFLOW_PARALLEL_STEP_CLI_OPTIONS_TEMPLATE}"
+)
+
 
 class KubernetesException(MetaflowException):
     headline = "Kubernetes error"
@@ -67,12 +72,6 @@ class KubernetesException(MetaflowException):
 
 class KubernetesKilledException(MetaflowException):
     headline = "Kubernetes Batch job killed"
-
-
-def _extract_labels_and_annotations_from_job_spec(job_spec):
-    annotations = job_spec.template.metadata.annotations
-    labels = job_spec.template.metadata.labels
-    return copy.copy(annotations), copy.copy(labels)
 
 
 class Kubernetes(object):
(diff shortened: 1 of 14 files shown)


metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -427,15 +426,15 @@ def task_pre_step(
...
             ]
             meta["kubernetes-node-ip"] = os.environ["METAFLOW_KUBERNETES_NODE_IP"]
-            if os.environ.get("METAFLOW_KUBERNETES_JOBSET_NAME"):
-                meta["kubernetes-jobset-name"] = os.environ[
-                    "METAFLOW_KUBERNETES_JOBSET_NAME"
-                ]
+
+            meta["kubernetes-jobset-name"] = os.environ.get(
+                "METAFLOW_KUBERNETES_JOBSET_NAME"
+            )
```
valayDave, line 432: todo: add more metadata like: `UBF_CONTEXT` / `RANK` / `WORLD_SIZE`.

valayDave: If we have already set the `CONTROL_TASK_TAG` in metadata then we don't need the `UBF_CONTEXT` ; Check if the `CONTROL_TASK_TAG` is working as expected.

metaflow/runtime.py
```diff
@@ -1108,56 +1108,45 @@ def mark_resume_done(self):
...
+                MetaDatum(
+                    field="internal_task_type",
+                    value=CONTROL_TASK_TAG,
+                    type="internal_task_type",
+                    tags=["attempt_id:{0}".format(0)],
                 )
+            ]
+            self.metadata.register_metadata(
+                self.run_id, self.step, task_id, task_metadata_list
+            )
```
valayDave, line 1149: todo: ideally move this into the `@parallel` decorator

metaflow/plugins/parallel_decorator.py
```diff
@@ -80,8 +98,6 @@ def _local_multinode_control_task_step_func(flow, env_to_use, step_func, retry_c
     run_id = current.run_id
     step_name = current.step_name
     control_task_id = current.task_id
-
-    (_, split_step_name, split_task_id) = control_task_id.split("-")[1:]
     # UBF handling for multinode case
     top_task_id = control_task_id.replace("control-", "")  # chop "-0"
```
valayDave, line 102: todo : remove `top_task_id`

metaflow/plugins/parallel_decorator.py
valayDave, line 1: todo: make separate PRs for anything in the core for examples files in `parallel_decorator` and `runtime.py`

valayDave: Fixme: ensure that @secrets can work with control tasks too ; Since @parallel has a slightly different control flow than normal tasks, we need to ensure that implementation has parity across local/remote executions.

valayDave: Fixme: remove the `current.parallel` property from the `metaflow_current.py` file and instead dynamically inject the property inside the `@parallel` decorator. [Can be done at a later date. At least add a comment in the code. ]

test/core/tests/tag_catch.py
```diff
@@ -121,7 +121,7 @@ def check_results(self, flow, checker):
                     data = task.data
                     got = sorted(m.value for m in task.metadata if m.type == "attempt")
                     if flow._graph[step.id].parallel_step:
-                        if "control" in task.id:
```
valayDave, line 124: TODO: Check task_metadata instead of checking of checking for the task-id string. the taskmetadata will contain the internal task type which can let us know if it's a control task or not.

metaflow/plugins/kubernetes/kubernetes.py
```diff
@@ -154,57 +153,285 @@ def launch_job(self, **kwargs):
...
             kwargs["name_pattern"] = "t-{uid}-".format(uid=str(uuid4())[:8])
             self._job = self.create_job_object(**kwargs).create().execute()
 
     def create_jobset(
         self,
-        job_spec=None,
-        run_id=None,
-        step_name=None,
-        task_id=None,
+        flow_name,
```
valayDave, line 163: todo: at some point re-unify the code bases for Jobset and native jobs.

metaflow/plugins/kubernetes/kubernetes_cli.py
```diff
@@ -203,19 +204,40 @@ def echo(msg, stream="stderr", job_id=None, **kwargs):
...
         step_args=" ".join(util.dict_to_cli_options(kwargs)),
     )
+    # Since it is a parallel step there are some parts of the step_cli that need to be modified
+    # based on the type of worker in the JobSet. This is why we will create a placeholder string
+    # in the template which will be replaced based on the type of worker.
+
+    if num_parallel:
+        step_cli = "{step_cli} {placeholder_string}".format(
+            step_cli=step_cli,
+            placeholder_string=METAFLOW_PARALLEL_STEP_CLI_OPTIONS_TEMPLATE,
```
valayDave, line 232: replace `{placeholder_string}` with the actual string and not put a format string here.

metaflow/plugins/kubernetes/kubernetes.py
```diff
@@ -154,57 +153,285 @@ def launch_job(self, **kwargs):
...
-            job_spec=job_spec,
-            port=port,
+            task_id=_tskid,
+            attempt=attempt,
+            code_package_url=code_package_url,
+            step_cmds=[
+                step_cli.replace(
+                    METAFLOW_PARALLEL_STEP_CLI_OPTIONS_TEMPLATE,
+                    "--ubf-context $UBF_CONTEXT --split-index %s --task-id %s"
+                    % (index, _tskid),
```
valayDave, line 401: todo: try of `.format` instead of `.replace`

metaflow/plugins/kubernetes/kubernetes.py
```diff
@@ -154,57 +153,285 @@ def launch_job(self, **kwargs):
...
+        # and also set the task-id/spit-index for the control/worker job
+        # appropirately.
+        jobset.control.command(_get_command("0", str(task_id)))
+        jobset.worker.command(
+            _get_command(
+                "`expr $[MF_WORKER_REPLICA_INDEX] + 1`",
+                "-".join(
+                    [
+                        str(task_id),
+                        "wrkr",
```
valayDave, line 418: change to `worker` ;

metaflow/plugins/kubernetes/kubernetes.py
valayDave: @savingoyal : 

`.format` doesnt work as stated earlier for the following reason: 

The point in time we are running the `.format`, we already have constructed the step_cli with more format strings; For example at this point the step_cli looks like : 

```
python -u fail_flow.py --quiet --metadata service --environment local --datastore s3 --event-logger nullSidecarLogger --monitor nullSidecarMonitor --datastore-root s3://mybcket/metaflow/ --package-suffixes .py,.R,.RDS --pylint step train --run-id 5075 --input-paths ${METAFLOW_INPUT_PATHS_0} --retry-count 0 --max-user-code-retries 2 --namespace valay {METAFLOW_PARALLEL_STEP_CLI_OPTIONS_TEMPLATE}
```

At this point the step_cli already contains the `${METAFLOW_INPUT_PATHS_0}` which is a shell variable that will be replaced at runtime. At this point if we use `.format` then python will crash complaining that we haven't given any value for format string `METAFLOW_INPUT_PATHS_0`. So we use replace instead of format.

PR #1902, opened 2024-06-26
feature: Conda env extension hooks

saikonen:
We need some additional extension capabilities to be able to implement #1796 as an extension, these being:

- add 'docker' to list of supported environments for `@conda` and `@pypi`
- ability to **fallback** to the core Conda environment initialization for a subset of steps only.

Note: the changes to `conda_environment.py` would at the moment result in any plugin relying on the new `only_steps=` to break with the bleeding edge nflx-extension conda implementation

### DIFF
--- a/metaflow/metaflow_config.py
+++ b/metaflow/metaflow_config.py
@@ -503,20 +503,6 @@ def get_pinned_conda_libs(python_version, datastore_type):
     return pins
 
 
-# List of virtual environments that support @conda/@pypi
-# extensible via extensions
-def conda_supported_virtual_envs():
-    """
-    Returns a list of supported environments that @conda/@pypi use for validation.
-    """
-    # @conda uses a conda environment to create a virtual environment.
-    # The conda environment can be created through micromamba.
-
-    # To placate people who don't want to see a shred of conda in UX, we symlink
-    # --environment=pypi to --environment=conda
-    return ["conda", "pypi"]
-
-
 # Check if there are extensions to Metaflow to load and override everything
 try:
     from metaflow.extension_support import get_modules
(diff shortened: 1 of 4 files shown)


metaflow/metaflow_config.py
```diff
@@ -503,6 +503,20 @@ def get_pinned_conda_libs(python_version, datastore_type):
...
+def conda_supported_virtual_envs():
+    """
+    Returns a list of supported environments that @conda/@pypi use for validation.
+    """
+    # @conda uses a conda environment to create a virtual environment.
+    # The conda environment can be created through micromamba.
+
+    # To placate people who don't want to see a shred of conda in UX, we symlink
+    # --environment=pypi to --environment=conda
+    return ["conda", "pypi"]
```
savingoyal, line 517: Would hardcoding `docker` for now or relying on a property in `docker_environment` to ensure compatibility work better? Managing it through config seems a bit _anti-pattern_-ish for the long-term since it is not a user exposed config

saikonen: I thought on this for a bit and it felt a bit more natural to implement a way for the extensions to extend a supported list of environments, rather than them implementing a method for the decorator to call.

so the alternative approach here would be to implement something like the following, but only in environments that we support pypi/conda on?
```python
def _is_supported_deps_deco(self, deco_name):
    return deco_name in ["conda", "pypi"]
```

and in the decorators env validation
```python
try:
    if not environment._is_supported_deps_deco(conda_or_pypi_deco.name):
        raise InvalidEnvironmentException("...")
except Exception:
    # env does not even implement the method for checking
    echo("environment does not support pypi/conda")
    raise
```

The main drawback with this approach is that we are now missing a list of environments that _do_ support conda/pypi, so we can't display a helpful list to the user

saikonen: Let's go with the hardcoded option for the time being. Left a TODO regarding future improvements for extending the supported venv list.

savingoyal (approved):

PR #1911, opened 2024-07-02
fix: do not escape html in Argo Events payload

saikonen:
do not unnecessarily escape html characters in argo events payload.


savingoyal:
linked to https://github.com/WGBH-MLA/chowda/issues/204

PR #1913, opened 2024-07-04
[parallel-fixes] core + test changes

valayDave:
- fix tests for ubf based on new changes to core.
- [feedback] register metadata in parallel decorator
- [feedback]@parallel inject in current: - move `current.parallel` from `metaflow_current` to `parallel_decorator`
- [feedback] appropariately setting task-metadata for parallel stuff 
- The 'world size' metadata will be set in the @parallel decorator. 
- The 'node-index' metadata, however, varies depending on the type of computing environment executing the task so it will be set in the appropriate compute decorators. 
- One reason to specify 'node-index' within compute decorators is that the parallel implementation in the compute decorator might not directly set the required environment variables (`MF_PARALLEL_*`). Instead, these environment variables (`MF_PARALLEL_*`) may be set during the `task_pre_step` callback of the compute decorator using decorator implementation specific logic. 
- [feedback] reflecting changes to aws batch
- [feedback] safety check for _parallel_buf_iter in task_pre_step for @parallel
- [feedback] set `is_parallel` in current to denote a step is running under an `@parallel` decorator.

### TODO
- [x] Linter changes for join followed by @parallel

### DIFF
--- a/metaflow/lint.py
+++ b/metaflow/lint.py
@@ -295,6 +295,19 @@ def check_parallel_step_after_next(graph):
 
 @linter.ensure_static_graph
 @linter.check
+def check_join_followed_by_parallel_step(graph):
+    msg = (
+        "An @parallel step should be followed by a join step. Step *{0}* is called "
+        "after an @parallel step but is not a join step. Please add an extra `inputs` "
+        "argument to the step."
+    )
+    for node in graph:
+        if node.parallel_step and not graph[node.out_funcs[0]].type == "join":
+            raise LintWarn(msg.format(node.out_funcs[0]))
+
+
+@linter.ensure_static_graph
+@linter.check
 def check_parallel_foreach_calls_parallel_step(graph):
     msg = (
         "Step *{0.name}* has a @parallel decorator, but is not called "
--- a/metaflow/metaflow_current.py
+++ b/metaflow/metaflow_current.py
@@ -261,14 +261,6 @@ class Current(object):
         return self._username
 
     @property
-    def parallel(self):
-        return Parallel(
-            main_ip=os.environ.get("MF_PARALLEL_MAIN_IP", "127.0.0.1"),
-            num_nodes=int(os.environ.get("MF_PARALLEL_NUM_NODES", "1")),
-            node_index=int(os.environ.get("MF_PARALLEL_NODE_INDEX", "0")),
-        )
-
-    @property
     def tags(self):
         """
         [Legacy function - do not use]
--- a/metaflow/plugins/aws/batch/batch_decorator.py
+++ b/metaflow/plugins/aws/batch/batch_decorator.py
@@ -261,8 +261,8 @@ class BatchDecorator(StepDecorator):
         # metadata. A rudimentary way to detect non-local execution is to
         # check for the existence of AWS_BATCH_JOB_ID environment variable.
 
+        meta = {}
         if "AWS_BATCH_JOB_ID" in os.environ:
-            meta = {}
             meta["aws-batch-job-id"] = os.environ["AWS_BATCH_JOB_ID"]
             meta["aws-batch-job-attempt"] = os.environ["AWS_BATCH_JOB_ATTEMPT"]
             meta["aws-batch-ce-name"] = os.environ["AWS_BATCH_CE_NAME"]
@@ -290,18 +290,6 @@ class BatchDecorator(StepDecorator):
             instance_meta = get_ec2_instance_metadata()
             meta.update(instance_meta)
 
-            entries = [
-                MetaDatum(
-                    field=k,
-                    value=v,
-                    type=k,
-                    tags=["attempt_id:{0}".format(retry_count)],
-                )
-                for k, v in meta.items()
-            ]
-            # Register book-keeping metadata for debugging.
-            metadata.register_metadata(run_id, step_name, task_id, entries)
-
             self._save_logs_sidecar = Sidecar("save_logs_periodically")
             self._save_logs_sidecar.start()
 
@@ -322,6 +310,21 @@ class BatchDecorator(StepDecorator):
 
         if num_parallel >= 1:
             _setup_multinode_environment()
+            # current.parallel.node_index will be correctly available over here.
+            meta.update({"parallel-node-index": current.parallel.node_index})
+
+        if len(meta) > 0:
+            entries = [
+                MetaDatum(
+                    field=k,
+                    value=v,
+                    type=k,
+                    tags=["attempt_id:{0}".format(retry_count)],
+                )
+                for k, v in meta.items()
+            ]
+            # Register book-keeping metadata for debugging.
+            metadata.register_metadata(run_id, step_name, task_id, entries)
 
     def task_finished(
         self, step_name, flow, graph, is_task_ok, retry_count, max_retries
--- a/metaflow/plugins/parallel_decorator.py
+++ b/metaflow/plugins/parallel_decorator.py
@@ -1,11 +1,34 @@
+from collections import namedtuple
 from metaflow.decorators import StepDecorator
-from metaflow.unbounded_foreach import UBF_CONTROL
+from metaflow.unbounded_foreach import UBF_CONTROL, CONTROL_TASK_TAG
 from metaflow.exception import MetaflowException
+from metaflow.metadata import MetaDatum
+from metaflow.metaflow_current import current, Parallel
 import os
 import sys
 
 
 class ParallelDecorator(StepDecorator):
+    """
+    MF Add To Current
+    -----------------
+    parallel -> metaflow.metaflow_current.Parallel
+
+        @@ Returns
+        -------
+        Parallel
+            `namedtuple` with the following fields:
+                - main_ip : str
+                    The IP address of the control task.
+                - num_nodes : int
+                    The total number of tasks created by @parallel
+                - node_index : int
+                    The index of the current task in all the @parallel tasks.
+
+    is_parallel -> bool
+        True if the current step is a @parallel step.
+    """
+
     name = "parallel"
     defaults = {}
     IS_PARALLEL = True
@@ -16,7 +39,6 @@ class ParallelDecorator(StepDecorator):
     def runtime_step_cli(
         self, cli_args, retry_count, max_user_code_retries, ubf_context
     ):
-
         if ubf_context == UBF_CONTROL:
             num_parallel = cli_args.task.ubf_iter.num_parallel
             cli_args.command_options["num-parallel"] = str(num_parallel)
@@ -25,6 +47,75 @@ class ParallelDecorator(StepDecorator):
         self, flow, graph, step_name, decorators, environment, flow_datastore, logger
     ):
         self.environment = environment
+        # We choose `setattr` instead of the default `current._update_env` for two reasons: first, to preserve
+        # the datatype of `current.parallel` as a namedtuple, and second, to ensure the contents of the namedtuple
+        # are accurately set based on when they are accessed in the Metaflow lifecycle. The values for `main_ip`,
+        # `num_nodes`, and `node_index` are determined by environment variables which maybe set in `task_pre_step`.
+        # Thus, we use a `property` to lazily evaluate these namedtuple values.
+        setattr(
+            current.__class__,
+            "parallel",
+            property(
+                fget=lambda _: Parallel(
+                    main_ip=os.environ.get("MF_PARALLEL_MAIN_IP", "127.0.0.1"),
+                    num_nodes=int(os.environ.get("MF_PARALLEL_NUM_NODES", "1")),
+                    node_index=int(os.environ.get("MF_PARALLEL_NODE_INDEX", "0")),
+                )
+            ),
+        )
+
+    def task_pre_step(
+        self,
+        step_name,
+        task_datastore,
+        metadata,
+        run_id,
+        task_id,
+        flow,
+        graph,
+        retry_count,
+        max_user_code_retries,
+        ubf_context,
+        inputs,
+    ):
+        from metaflow import current
+
+        # Set `is_parallel` to `True` in `current` just like we
+        # with `is_production` in the project decorator.
+        current._update_env(
+            {
+                "is_parallel": True,
+            }
+        )
+
+        self.input_paths = [obj.pathspec for obj in inputs]
+        task_metadata_list = [
+            MetaDatum(
+                field="parallel-world-size",
+                value=flow._parallel_ubf_iter.num_parallel,
+                type="parallel-world-size",
+                tags=["attempt_id:{0}".format(0)],
+            )
+        ]
+        if ubf_context == UBF_CONTROL:
+            # A Task's tags are now those of its ancestral Run, so we are not able
+            # to rely on a task's tags to indicate the presence of a control task
+            # so, on top of adding the tags above, we also add a task metadata
+            # entry indicating that this is a "control task".
+            #
+            # Here we will also add a task metadata entry to indicate "control
+            # task". Within the metaflow repo, the only dependency of such a
+            # "control task" indicator is in the integration test suite (see
+            # Step.control_tasks() in client API).
+            task_metadata_list += [
+                MetaDatum(
+                    field="internal_task_type",
+                    value=CONTROL_TASK_TAG,
+                    type="internal_task_type",
+                    tags=["attempt_id:{0}".format(0)],
+                )
+            ]
+        metadata.register_metadata(run_id, step_name, task_id, task_metadata_list)
 
     def task_decorate(
         self, step_func, flow, graph, retry_count, max_user_code_retries, ubf_context
@@ -47,6 +138,7 @@ class ParallelDecorator(StepDecorator):
                 env_to_use,
                 _step_func_with_setup,
                 retry_count,
+                ",".join(self.input_paths),
             )
         else:
             return _step_func_with_setup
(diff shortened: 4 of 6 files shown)


metaflow/plugins/aws/batch/batch_decorator.py
```diff
@@ -322,6 +310,21 @@ def task_pre_step(
...
+                MetaDatum(
+                    field=k,
+                    value=v,
+                    type=k,
+                    tags=["attempt_id:{0}".format(retry_count)],
+                )
+                for k, v in meta.items()
+            ]
+            # Register book-keeping metadata for debugging.
+            metadata.register_metadata(run_id, step_name, task_id, entries)
```
valayDave, line 327: `parallel` related metadata set in the decorator.

metaflow/plugins/parallel_decorator.py
```diff
@@ -25,6 +43,79 @@ def step_init(
...
+            # Step.control_tasks() in client API).
+            task_metadata_list += [
+                MetaDatum(
+                    field="internal_task_type",
+                    value=CONTROL_TASK_TAG,
+                    type="internal_task_type",
+                    tags=["attempt_id:{0}".format(0)],
+                )
+            ]
+        metadata.register_metadata(run_id, step_name, task_id, task_metadata_list)
```
valayDave, line 118: Setting metadata for parallel tasks.

metaflow/plugins/parallel_decorator.py
```diff
@@ -93,12 +184,13 @@ def _local_multinode_control_task_step_func(flow, env_to_use, step_func, retry_c
     script = sys.argv[0]
 
     # start workers
+    # TODO: Logs for worker processes are assigned to control process as of today, which
+    #       should be fixed at some point
     subprocesses = []
     for node_index in range(1, num_parallel):
-        task_id = "%s_node_%d" % (top_task_id, node_index)
+        task_id = "%s_node_%d" % (control_task_id, node_index)
```
valayDave, line 191: nothing functionally changed.

metaflow/plugins/secrets/secrets_decorator.py
```diff
@@ -210,8 +210,17 @@ def task_pre_step(
...
+        # will be a env var clash with UBF_WORKER tasks which will fail the flow in the @secrets decorator's `task_pre_step`.
+        if (
+            ubf_context == UBF_CONTROL
+            and os.environ.get("METAFLOW_RUNTIME_ENVIRONMENT", "local") == "local"
+        ):
+            # When we "locally" run @parallel tasks, the control task will create the worker tasks and the environment variables
+            # of the control task are inherited by the worker tasks. If we don't skip setting secrets in the control task then the
+            # worker tasks will already have the secrets set as environment variables, causing the @secrets' `task_pre_step` to fail.
+            # In remote settings, (e.g. AWS Batch/Kubernetes), the worker task and control task are independently created
+            # so there is no chances of an env var clash.
```
valayDave, line 223: Allow @secrets in control tasks and all other forms of tasks.

test/core/tests/tag_catch.py
```diff
@@ -121,7 +121,9 @@ def check_results(self, flow, checker):
                     data = task.data
                     got = sorted(m.value for m in task.metadata if m.type == "attempt")
                     if flow._graph[step.id].parallel_step:
-                        if "control" in task.id:
+                        if task.metadata_dict.get(
+                            "internal_task_type", None
+                        ):  # Only control tasks have internal_task_type set
```
valayDave, line 126: more reliable way to checking control tasks in tests.

test/core/tests/tag_catch.py
```diff
@@ -121,7 +121,9 @@ def check_results(self, flow, checker):
                     data = task.data
                     got = sorted(m.value for m in task.metadata if m.type == "attempt")
                     if flow._graph[step.id].parallel_step:
-                        if "control" in task.id:
+                        if task.metadata_dict.get(
```
savingoyal, line 124: can you pull this out as a separate PR - this change is likely non-contentious.

test/core/tests/secrets_decorator.py
```diff
@@ -26,7 +26,15 @@ def step_all(self):
         import os
         from metaflow import current
 
-        if current.task_id.startswith("control-"):
+        if (
```
savingoyal, line 29: can you pull this out as a separate PR - this change is likely non-contentious.

metaflow/plugins/test_unbounded_foreach_decorator.py
```diff
@@ -60,13 +66,43 @@ def step_init(
...
+        self.input_paths = [obj.pathspec for obj in inputs]
+
     def control_task_step_func(self, flow, graph, retry_count):
         from metaflow import current
 
         run_id = current.run_id
         step_name = current.step_name
         control_task_id = current.task_id
-        (_, split_step_name, split_task_id) = control_task_id.split("-")[1:]
+        # (_, split_step_name, split_task_id) = control_task_id.split("-")[1:]
```
savingoyal, line 105: this line can be deleted

metaflow/plugins/secrets/secrets_decorator.py
```diff
@@ -210,8 +210,17 @@ def task_pre_step(
         ubf_context,
         inputs,
     ):
-        if ubf_context == UBF_CONTROL:
-            """control tasks (as used in "unbounded for each") don't need secrets"""
+        # We will skip the secret injection for "locally" launched UBF_CONTROL tasks because otherwise there
```
savingoyal, line 213: can you add a FIXME for the future that `@parallel` implementation needs to change so that @secrets can work locally for control node in the future.

metaflow/plugins/secrets/secrets_decorator.py
savingoyal, line 223: can we change the behavior of `@secrets` such that it is a no-op for worker tasks locally, instead of it being a no-op for control tasks? that ensures parity for remote and local execution

metaflow/plugins/parallel_decorator.py
```diff
@@ -25,6 +43,79 @@ def step_init(
...
+        current._update_env(
+            {
+                "is_parallel": True,
+            }
+        )
+
+        self.input_paths = [obj.pathspec for obj in inputs]
+        if not hasattr(flow, "_parallel_ubf_iter"):
+            raise MetaflowException(
+                "Parallel decorator is only supported in unbounded foreach steps."
```
savingoyal, line 90: curious - how is this error triggered? shouldn't the linter throw this error upstream?

metaflow/plugins/aws/batch/batch_decorator.py
```diff
@@ -322,6 +310,21 @@ def task_pre_step(
 
         if num_parallel >= 1:
             _setup_multinode_environment()
+            # current.parallel.node_index will be correctly available over here.
+            meta.update({"parallel-node-index": current.parallel.node_index})
```
savingoyal, line 314: same change would be needed in the kubernetes decorator too - maybe it is in the other PR

savingoyal (changes requested):
lgtm - added a few comments

metaflow/plugins/aws/batch/batch_decorator.py
valayDave: its in another PR which changes k8s decorator.

metaflow/metaflow_current.py
```diff
@@ -260,14 +260,6 @@ def username(self) -> Optional[str]:
...
         return self._username
 
-    @property
-    def parallel(self):
-        return Parallel(
-            main_ip=os.environ.get("MF_PARALLEL_MAIN_IP", "127.0.0.1"),
-            num_nodes=int(os.environ.get("MF_PARALLEL_NUM_NODES", "1")),
-            node_index=int(os.environ.get("MF_PARALLEL_NODE_INDEX", "0")),
-        )
-
```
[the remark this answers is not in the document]
romain-intel: OK -- we can have it documented as such too in the docstrings.

metaflow/plugins/parallel_decorator.py
```diff
@@ -1,11 +1,30 @@
...
+from metaflow.metaflow_current import current, Parallel
 import os
 import sys
 
 
 class ParallelDecorator(StepDecorator):
+    """
+    MF Add To Current
+    -----------------
+    parallel -> metaflow.metaflow_current.Parallel
```
romain-intel, line 15: awesome!

metaflow/plugins/test_unbounded_foreach_decorator.py
```diff
@@ -60,13 +66,43 @@ def step_init(
...
+                [
+                    MetaDatum(
+                        field="internal_task_type",
+                        value=CONTROL_TASK_TAG,
+                        type="internal_task_type",
+                        tags=["attempt_id:{0}".format(0)],
+                    )
+                ],
+            )
+        self.input_paths = [obj.pathspec for obj in inputs]
```
romain-intel, line 97: This line can probably move into the ubf_context if block as it is then only used for the control function.

metaflow/plugins/test_unbounded_foreach_decorator.py
```diff
@@ -97,7 +133,7 @@ def control_task_step_func(self, flow, graph, retry_count):
             task_id = "%s-%d" % (control_task_id.replace("control-", "test-ubf-"), i)
             pathspec = "%s/%s/%s" % (run_id, step_name, task_id)
             mapper_tasks.append(to_unicode(pathspec))
-            input_paths = "%s/%s/%s" % (run_id, split_step_name, split_task_id)
+            input_paths = ",".join(self.input_paths)
```
romain-intel, line 136: ok, so this removes the need to name the control task a particular way yes? I see that it now also has possibly more than one? What is the use case here?

metaflow/plugins/test_unbounded_foreach_decorator.py
```diff
@@ -97,7 +133,7 @@ def control_task_step_func(self, flow, graph, retry_count):
             task_id = "%s-%d" % (control_task_id.replace("control-", "test-ubf-"), i)
```
romain-intel, line 133: If this is to remove the need to name the control task something, shouldn't this go away as well?

test/core/tests/secrets_decorator.py
romain-intel: ya, this is specific to the secrets decorator. Fine leaving it here but totally fine with this change :)

metaflow/runtime.py
```diff
@@ -1108,56 +1108,24 @@ def mark_resume_done(self):
 
     def _get_task_id(self, task_id):
         already_existed = True
+        tags = []
```
romain-intel, line 1111: We need to check these changes with whatever @darinyu is also doing in the #1884 . I like the reduction in lines of code though :)

metaflow/plugins/test_unbounded_foreach_decorator.py
valayDave: instead of constructing the input_paths dynamically, we are just passing them as passed down by the lifecycle sine the task_pre_step gets the input paths. More than this being usecase-specific, this is a simplification of the code since this is a more "natural" way of retrieving the input_paths (instead of dynamically constructing them using the control task id). 

Control tasks will have special metadata that will help mark if the task is a control task. The worker tasks will generate their own task-ids based on task-id passed by the control task.

metaflow/plugins/test_unbounded_foreach_decorator.py
valayDave: Yes will remove

metaflow/plugins/secrets/secrets_decorator.py
valayDave: If I make this change then we need the subsequent changes to the testing of the secrets deco with parallel in this PR otherwise the tests would fail.

metaflow/plugins/parallel_decorator.py
valayDave: It shouldn't be; Added because I didn't know if any code paths could lead to this. Will remove it.

test/core/tests/tag_catch.py
valayDave: Coming from the PR https://github.com/Netflix/metaflow/pull/1917

test/core/tests/tag_catch.py
valayDave: Done https://github.com/Netflix/metaflow/pull/1917

test/core/tests/secrets_decorator.py
valayDave: I didn't pull it in a separate PR because the change to the decorator also affects the changes in the tests.

metaflow/plugins/secrets/secrets_decorator.py
valayDave: Done.

metaflow/plugins/test_unbounded_foreach_decorator.py
valayDave: done

metaflow/plugins/test_unbounded_foreach_decorator.py
valayDave: Done

metaflow/lint.py
```diff
@@ -293,6 +293,19 @@ def check_parallel_step_after_next(graph):
...
+    msg = (
+        "An @parallel step should be followed by a join step. Step *{0}* is called "
+        "after an @parallel step but is not a join step. Please add an extra `inputs` "
+        "argument to the step."
+    )
+    for node in graph:
+        if node.parallel_step and not graph[node.out_funcs[0]].type == "join":
+            raise LintWarn(msg.format(node.out_funcs[0]))
+
+
```
valayDave, line 308: Ensure join follow @parallel

test/core/tests/secrets_decorator.py
valayDave: NVM I pulled it in #1917

metaflow/runtime.py
```diff
@@ -1108,56 +1108,24 @@ def mark_resume_done(self):
...
-                    self.run_id, self.step, task_id, 0
-                )
+            # task_id is preset only by persist_constants().
+            already_existed = not self.metadata.register_task_id(
+                self.run_id,
+                self.step,
+                task_id,
+                0,
+                sys_tags=tags,
+            )
```
romain-intel, line 1128: You seem to no longer be setting internal_task_type but I just merged #1917 which relies on it so a bit confused. Is this done somewhere else now?

metaflow/metaflow_current.py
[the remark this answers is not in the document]
romain-intel: ^

metaflow/plugins/parallel_decorator.py
```diff
@@ -1,11 +1,34 @@
...
+        Parallel
+            `namedtuple` with the following fields:
+                - main_ip : str
+                    The IP address of the control task.
+                - num_nodes : int
+                    The total number of tasks created by @parallel
+                - node_index : int
+                    The index of the current task in all the @parallel tasks.
+
+    is_parallel -> bool
```
romain-intel, line 28: This does not seem to be the case

metaflow/plugins/parallel_decorator.py
```diff
@@ -25,6 +47,75 @@ def step_init(
         self, flow, graph, step_name, decorators, environment, flow_datastore, logger
     ):
         self.environment = environment
+        # We choose `setattr` instead of the default `current._update_env` for two reasons: first, to preserve
```
romain-intel, line 50: Could we then not do this in task_pre_step if that is when the values are available. Not strongly opposed to this but the _update_env basically does what this does so it feels like as long as we do it prior to the user code starting (so task_pre_step is early enough), we should be good too? Or am I missing another case.

metaflow/runtime.py
valayDave: We do set it!. We set it in the task_pre_step of the @parallel decorator.

metaflow/plugins/parallel_decorator.py
valayDave: Should I remove it?  It's fair to say that this will only be available when @parallel is there.

metaflow/plugins/parallel_decorator.py
valayDave: Well overall, I can keep the parallel property in the `current` object and nothing has to change here. 

This became a little essential because: 
- the `Parallel` namedtuple returned by `current.parallel` sets it's attributes using env vars. 
- The earlier implementation had the `Parallel` namedtuple returned via `current.parallel` property (the property was statically set). 
- Now we are moving to an implementation where `current.parallel` is dynamically set. 
- In this new world also want to retain the data structure of `Parallel` so that we have backwards compatibility in terms of types. 
- Now if we set this named tuple using `current._update_env` we will be _statically setting the namedtuple_. This doesn't work for us because the values inside the named tuple are derived using env vars. These env vars are set in `task_pre_step` based on the compute decorator. Since order of calling `task_pre_step` can vary, we need ensure that the namedtuple can have all the right data set to it and it is dynamically retrieving those values. Doing a `current._update_env` won't help us gaurentee that the values will always be correctly set in task_pre_step. 
- Hence setting this as an explicit property via setattr instead of passing it down via `current._update_env`

metaflow/plugins/parallel_decorator.py
```diff
@@ -25,6 +47,75 @@ def step_init(
...
+            # "control task" indicator is in the integration test suite (see
+            # Step.control_tasks() in client API).
+            task_metadata_list += [
+                MetaDatum(
+                    field="internal_task_type",
+                    value=CONTROL_TASK_TAG,
+                    type="internal_task_type",
+                    tags=["attempt_id:{0}".format(0)],
+                )
+            ]
```
valayDave, line 117: @romain-intel setting the internal_task_type here.

PR #1914, opened 2024-07-04
[refactor-jobsets] refactor to new implementation

valayDave:
- Some changes stemming from Netflix/metaflow#1854
- reorder imports
- changes to env var prefixes + id generation
- Abstraction over Jobset Spec
- Create general abstraction and copy style of how we create jobspec for native k8s jobs.
- simplified the implementation.
- similar pattern/trend to follow for argo
- remove older implementation
- @parallel metadata from k8s decorator - added attempt to MetaDatum tags - We didn't have this earlier

### DIFF
--- a/metaflow/plugins/kubernetes/kubernetes.py
+++ b/metaflow/plugins/kubernetes/kubernetes.py
@@ -1,9 +1,9 @@
+import copy
 import json
 import math
 import os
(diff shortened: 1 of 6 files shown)


romain-intel:
This is all in a plugin so all good for me.

PR #1916, opened 2024-07-08
feature: run heartbeat with Argo daemon containers

saikonen:
Introducing run heartbeat updates as an argo daemon container to decouple workflow status from the active workflow pod health.


metaflow/plugins/argo/argo_workflows.py
```diff
@@ -1916,6 +1931,107 @@ def _slack_success_template(self):
...
+        heartbeat_cmds = "%s %s %s" % (
+            " ".join(entrypoint),
+            self.flow.name,
+            run_id_template,
+        )
+
+        # TODO: we do not really need MFLOG logging for the daemon at the moment, but might be good for the future.
+        # Consider if we can do without this setup.
+        # Configure log capture.
+        mflog_expr = export_mflog_env_vars(
```
savingoyal, line 1949: Yeah we can do without mflog for now. Given that this is an underlying implementation detail at the moment that shouldn’t concern the user.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -1916,6 +1931,107 @@ def _slack_success_template(self):
             Http("POST").url(self.notify_slack_webhook_url).body(json.dumps(payload))
         )
 
+    def _heartbeat_daemon_template(self):
+        # Use any step that is known to exist, as we only care
+        # about being able to retrieve the code package for the utility scripts.
+        executable = self.environment.executable("_parameters")
+        run_id_template = "argo-{{workflow.name}}"
+        entrypoint = [executable, "-m metaflow.plugins.argo.heartbeat_daemon "]
```
savingoyal, line 1939: Minor nit - maybe just metaflow.plugins.argo.daemon - it is likely that we would want to use this container for more work besides heartbeats in the near future

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -846,6 +846,8 @@ def _compile_workflow_template(self):
                 .templates(self._container_templates())
                 # Exit hook template(s)
                 .templates(self._exit_hook_templates())
+                # Sidecar templates (Daemon Containers)
+                .templates(self._daemon_container_templates())
```
savingoyal, line 850: Can we make this optional through the cli - —with-daemon-container/—no-with-daemon-container ? That way folks can opt out if needed

savingoyal (changes requested):
LGTM!

metaflow/plugins/argo/argo_workflows.py
saikonen: can't actually do much about this while still relying on environment internals for package download, as `mflog()` bash function setup is part of the `get_package_commands` which is littered with calls to `mflog`.

the env vars exported by `export_mflog_env_vars` are required for the logging, otherwise things break. I can either keep the existing setup, which actually sets up working env vars. Alternatively I can opt for completely dummy values. easy way forward is to not touch the setup which guarantees it works

metaflow/plugins/argo/argo_workflows.py
saikonen: done. One thought on the wording for the toggle though: Currently we only run heartbeats, and in the future some other run-scope-processes in the default container.

what happens with the toggle though, if we introduce a way for users to launch their own daemon containers alongside tasks? should these also not deploy if `--no-with-daemon-container` is specified, or should it not apply to user defined ones?

savingoyal (approved):

PR #1925, opened 2024-07-17
[parallel-fixes] core + test changes

valayDave:
- fix tests for ubf based on new changes to core.
- [feedback] register metadata in parallel decorator
- [feedback]@parallel inject in current: - move `current.parallel` from `metaflow_current` to `parallel_decorator`
- [feedback] appropariately setting task-metadata for parallel stuff - The 'world size' metadata will be set in the @parallel decorator. - The 'node-index' metadata, however, varies depending on the type of computing environment executing the task so it will be set in the appropriate compute decorators. - One reason to specify 'node-index' within compute decorators is that the parallel implementation in the compute decorator might not directly set the required environment variables (`MF_PARALLEL_*`). Instead, these values may be established during the `task_pre_step` phase of the compute decorator using other environment variables set during the implementation. - adding some aws batch changes
- [feedback] set `is_parallel` in current to denote a step is running under an `@parallel` decorator. - users can use `if "is_parallel" in current`
- [feedback] add linter to ensure join after parallel

Re-opening #1913

### DIFF
--- a/metaflow/plugins/parallel_decorator.py
+++ b/metaflow/plugins/parallel_decorator.py
@@ -1,11 +1,34 @@
+from collections import namedtuple
 from metaflow.decorators import StepDecorator
-from metaflow.unbounded_foreach import UBF_CONTROL
+from metaflow.unbounded_foreach import UBF_CONTROL, CONTROL_TASK_TAG
 from metaflow.exception import MetaflowException
+from metaflow.metadata import MetaDatum
+from metaflow.metaflow_current import current, Parallel
 import os
 import sys
 
 
 class ParallelDecorator(StepDecorator):
+    """
+    MF Add To Current
+    -----------------
+    parallel -> metaflow.metaflow_current.Parallel
+
+        @@ Returns
+        -------
+        Parallel
+            `namedtuple` with the following fields:
+                - main_ip : str
+                    The IP address of the control task.
+                - num_nodes : int
+                    The total number of tasks created by @parallel
+                - node_index : int
+                    The index of the current task in all the @parallel tasks.
+
+    is_parallel -> bool
+        True if the current step is a @parallel step.
+    """
+
     name = "parallel"
     defaults = {}
     IS_PARALLEL = True
@@ -16,7 +39,6 @@ class ParallelDecorator(StepDecorator):
     def runtime_step_cli(
         self, cli_args, retry_count, max_user_code_retries, ubf_context
     ):
-
         if ubf_context == UBF_CONTROL:
             num_parallel = cli_args.task.ubf_iter.num_parallel
             cli_args.command_options["num-parallel"] = str(num_parallel)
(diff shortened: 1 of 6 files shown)


metaflow/plugins/parallel_decorator.py
```diff
@@ -1,11 +1,34 @@
...
 
 
 class ParallelDecorator(StepDecorator):
+    """
+    MF Add To Current
+    -----------------
+    parallel -> metaflow.metaflow_current.Parallel
+
+        @@ Returns
+        -------
```
romain-intel, line 18: NIt, not sure if you don't need to fully extend the --- for it to be pickedu p.

metaflow/plugins/parallel_decorator.py
```diff
@@ -25,6 +47,75 @@ def step_init(
...
+        self.input_paths = [obj.pathspec for obj in inputs]
+        task_metadata_list = [
+            MetaDatum(
+                field="parallel-world-size",
+                value=flow._parallel_ubf_iter.num_parallel,
+                type="parallel-world-size",
+                tags=["attempt_id:{0}".format(0)],
+            )
+        ]
+        if ubf_context == UBF_CONTROL:
```
romain-intel, line 100: Are we adding this here so it works even if the runtime isn't used (ie through argo/airflow/whatever)?

romain-intel: If so, that makes sense and we just have to change our internal code as well for the UBFs we have (HPO).

metaflow/plugins/parallel_decorator.py
```diff
@@ -25,6 +47,75 @@ def step_init(
         self, flow, graph, step_name, decorators, environment, flow_datastore, logger
     ):
         self.environment = environment
+        # We choose `setattr` instead of the default `current._update_env` for two reasons: first, to preserve
+        # the datatype of `current.parallel` as a namedtuple, and second, to ensure the contents of the namedtuple
+        # are accurately set based on when they are accessed in the Metaflow lifecycle. The values for `main_ip`,
+        # `num_nodes`, and `node_index` are determined by environment variables which maybe set in `task_pre_step`.
+        # Thus, we use a `property` to lazily evaluate these namedtuple values.
+        setattr(
```
romain-intel, line 55: I'll still ask a question on this but not strongly opposed to the change.

metaflow/runtime.py
```diff
@@ -1108,56 +1108,24 @@ def mark_resume_done(self):
 
     def _get_task_id(self, task_id):
         already_existed = True
+        tags = []
```
romain-intel, line 1111: as I mentioned, we just have to check with @darinyu for this but I don't anticipate any issues.

romain-intel (commented):
A few small things but I think this is good to go. Just have to check with the other resume PR to make sure they both play nice (but I think they will).

metaflow/plugins/parallel_decorator.py
valayDave: I copied this from the cards thing

metaflow/plugins/parallel_decorator.py
valayDave: Fixed the comment

romain-intel: LGTM

romain-intel (approved):

PR #1926, opened 2024-07-18
slurm decorator

madhur-ob:


madhur-ob:
released and can be installed via `pip install metaflow-slurm`

PR #1929, opened 2024-07-22
[flowspec] add artifacts to exclude in `merge_artifacts`

valayDave:


savingoyal (dismissed):

metaflow/flowspec.py
```diff
@@ -28,7 +28,9 @@
 
 from .datastore.inputs import Inputs
 
-INTERNAL_ARTIFACTS_SET = set(["_foreach_values", "_unbounded_foreach"])
+INTERNAL_ARTIFACTS_SET = set(
```
romain-intel, line 31: Should we also exclude `_control_task_is_mapper_zero`?

valayDave: Done

romain-intel (approved):
Not tested but lgtm. If this fixes the issue you were having, I think we should be good :)

PR #1947, opened 2024-08-08
Add `--only-yaml` argo-workflows cli option.

trhodeos:
This supports outputting _all_ k8s templates into stdout (rather than just the workflowtemplate one via --only-json).

Fixes 1940, 1730

### DIFF
--- a/metaflow/plugins/argo/argo_client.py
+++ b/metaflow/plugins/argo/argo_client.py
@@ -1,6 +1,4 @@
 import json
-import os
-import sys
 
 from metaflow.exception import MetaflowException
 from metaflow.plugins.kubernetes.kubernetes_client import KubernetesClient
@@ -283,23 +281,17 @@ class ArgoClient(object):
                 json.loads(e.body)["message"] if e.body is not None else e.reason
             )
 
-    def schedule_workflow_template(self, name, schedule=None, timezone=None):
+    def schedule_workflow_template(self, name, cron_workflow=None):
         # Unfortunately, Kubernetes client does not handle optimistic
         # concurrency control by itself unlike kubectl
         client = self._client.get()
-        body = {
-            "apiVersion": "argoproj.io/v1alpha1",
-            "kind": "CronWorkflow",
-            "metadata": {"name": name},
-            "spec": {
-                "suspend": schedule is None,
-                "schedule": schedule,
-                "timezone": timezone,
-                "workflowSpec": {"workflowTemplateRef": {"name": name}},
-            },
-        }
+        if cron_workflow is None:
+            cron_workflow = {}
+        if not cron_workflow:
+            cron_workflow["metadata"] = {}
+
         try:
-            body["metadata"][
+            cron_workflow["metadata"][
                 "resourceVersion"
             ] = client.CustomObjectsApi().get_namespaced_custom_object(
                 group=self._group,
@@ -315,7 +307,7 @@ class ArgoClient(object):
         except client.rest.ApiException as e:
             # Scheduled workflow does not exist and we want to schedule a workflow
             if e.status == 404:
-                if schedule is None:
+                if cron_workflow["spec"]["schedule"] is None:
                     return
                 try:
                     return client.CustomObjectsApi().create_namespaced_custom_object(
@@ -323,7 +315,7 @@ class ArgoClient(object):
                         version=self._version,
                         namespace=self._namespace,
                         plural="cronworkflows",
-                        body=body,
+                        body=cron_workflow,
                     )
                 except client.rest.ApiException as e:
                     raise ArgoClientException(
@@ -341,7 +333,7 @@ class ArgoClient(object):
                 version=self._version,
                 namespace=self._namespace,
                 plural="cronworkflows",
-                body=body,
+                body=cron_workflow,
                 name=name,
             )
         except client.rest.ApiException as e:
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -7,12 +7,11 @@ import sys
 from collections import defaultdict
 from hashlib import sha1
 from math import inf
-from typing import List, Tuple
 
 from metaflow import JSONType, current
 from metaflow.decorators import flow_decorators
 from metaflow.exception import MetaflowException
-from metaflow.graph import DAGNode, FlowGraph
+from metaflow.graph import FlowGraph
 from metaflow.includefile import FilePathClass
 from metaflow.metaflow_config import (
     ARGO_EVENTS_EVENT,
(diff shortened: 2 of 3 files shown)


metaflow/plugins/argo/argo_client.py
```diff
@@ -1,6 +1,4 @@
 import json
-import os
```
trhodeos, line 2: Looks like my autoformatter went to town on some things... I'll revert those changes.

metaflow/plugins/argo/argo_client.py
```diff
@@ -283,23 +281,17 @@ def trigger_workflow_template(self, name, parameters={}):
...
-            "kind": "CronWorkflow",
-            "metadata": {"name": name},
-            "spec": {
-                "suspend": schedule is None,
-                "schedule": schedule,
-                "timezone": timezone,
-                "workflowSpec": {"workflowTemplateRef": {"name": name}},
-            },
-        }
+        if cron_workflow is None:
```
trhodeos, line 288: note: this mirrors how the "sensor" flow works

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -123,7 +124,13 @@ def argo_workflows(obj, name=None):
     "--only-json",
     is_flag=True,
     default=False,
-    help="Only print out JSON sent to Argo Workflows. Do not deploy anything.",
+    help="Only print out JSON sent to Argo Workflows. Do not deploy anything. DEPRECATED, prefer --output-yaml instead",
```
savingoyal, line 127: Any reason to mark this as deprecated? We haven't deprecated any functionality in Metaflow this far in the project's history

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -123,7 +124,13 @@ def argo_workflows(obj, name=None):
...
+)
+@click.option(
+    "--only-yaml",
+    is_flag=True,
+    default=False,
+    help="Only print out yaml sent to Argo Workflows. Do not deploy anything. Includes all objects",
```
savingoyal, line 133: Unfortunately - this is a bit incorrect. The description for `--only-json` also needs to be fixed (or the implementation updated). Currently, using `--only-json` updates the sensors - which needs to be addressed. 

My preference would be to fix the behavior of `--only-json` so that it is side-effect-free, introduce additional flags to emit other objects (sensors etc.) as json objects so that there is no need for introducing a pyyaml dependency. the output of `--only-json` can then be piped to a local script that can convert the json payload into a yaml.

savingoyal (changes requested):

metaflow/plugins/argo/argo_workflows_cli.py
trhodeos: Mainly because it's "broken" in that it doesn't output _all_ objects and my thought was to have `--only-yaml` replace it. But I can remove this note.

metaflow/plugins/argo/argo_workflows_cli.py
trhodeos: Ack. Let me see if I can fix --only-json in another PR.

As for json vs yaml, see my above comment.

setup.py
```diff
@@ -49,7 +49,7 @@
         [console_scripts]
         metaflow=metaflow.cmd.main_cli:start
       """,
-    install_requires=["requests", "boto3"],
+    install_requires=["requests", "boto3", "pyyaml"],
```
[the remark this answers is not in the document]
trhodeos: Ok, I decided to output all of the templates to a directory (using json), instead of pyyaml... @savingoyal WDYT?

metaflow/plugins/argo/argo_workflows_cli.py
trhodeos: Removed.

setup.py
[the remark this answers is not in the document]
savingoyal: this would break the existing users. we could introduce new options - `--only-workflow-template-json`, `--only-event-sensor-json`, `--only-cron-workflow-template-json` etc. where `--only-json` is aliased to `--only-workflow-template-json`.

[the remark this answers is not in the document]
savingoyal: ah - `--only-json` is not replaced by `--only-json-directory`. i misread your comment - my apologies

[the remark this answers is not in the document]
trhodeos: Ok, just getting back around to this. I've made the changes outlined here https://github.com/Netflix/metaflow/pull/1947#discussion_r1715434964... PTAL!

metaflow/plugins/argo/argo_workflows_cli.py
trhodeos: @savingoyal is this true? Where do the sensors get updated? I traced the code and couldn't find where this'd happen.

trhodeos:
@savingoyal any chance you can take another look? Thanks in advance!

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -3754,8 +3806,7 @@ class ArgoWorkflowTrigger(object):
     # https://github.com/argoproj/argo-events/blob/master/api/sensor.md#argoproj.io/v1alpha1.ArgoWorkflowTrigger
 
     def __init__(self):
-        tree = lambda: defaultdict(tree)
-        self.payload = tree()
+        self.payload = _tree()
```
savingoyal, line 3809: this change seems unrelated to the PR. can you consider pulling this out as a different PR or skipping it?

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -2731,7 +2745,7 @@ def _compile_sensor(self):
             )
 
         # Useful to paint the UI
-        trigger_annotations = {
+        {
```
savingoyal, line 2748: @trhodeos we have impending plans to move away from `one sensor -> one workflow template` mapping (to save resource costs) to `one sensor -> multiple workflow templates` which will break the expectations in this PR for `--only-event-sensor-json`. can you help us with the timelines you are shooting for this PR? happy to chat over zoom if that's helpful.

savingoyal: depending on your use case, an immediate workaround could also be to override argo_client.py through the extensions such that rather than submitting the CRDs to kubernetes, it just prints them out to stderr.

savingoyal (changes requested):

metaflow/plugins/argo/argo_workflows.py
savingoyal: but regardless this change is in the right direction. can you help us with details on how you have tested this change? we should be able to test this PR ourselves some point next week.

metaflow/plugins/argo/argo_workflows.py
trhodeos: Yeah, let me revert

metaflow/plugins/argo/argo_workflows.py
trhodeos: One potential solution is to only add a CLI arg `--output-json-directory-only` and just push all json blobs to that directory. This would essentially allow us to not be tied to the number of resources we generate (as in, adding more sensors is not a problem).

I like the argo_client.py idea, _though_ we'd still like to push directly to argo while we migrate off of using metaflow directly. Still possible, but may require a bit more thinking.

For context: we are trying to use metaflow to generate the resources, spit them out to a gitops repo which has ArgoCD hooked up to it, for better staging of deploys.

As for testing, I'll test locally tomorrow!

PR #1948, opened 2024-08-08
add k8s namespace to pod/job

madhur-ob:


metaflow/plugins/kubernetes/kubernetes.py
```diff
@@ -565,6 +567,7 @@ def create_job_object(
                 "METAFLOW_INIT_SCRIPT", KUBERNETES_SANDBOX_INIT_SCRIPT
             )
             .environment_variable("METAFLOW_OTEL_ENDPOINT", OTEL_ENDPOINT)
+            .environment_variable("METAFLOW_KUBERNETES_NAMESPACE", KUBERNETES_NAMESPACE)
```
valayDave, line 570: Don't need to set it twice ?

metaflow/plugins/kubernetes/kubernetes.py
```diff
@@ -281,6 +282,7 @@ def create_jobset(
...
```
valayDave, line 285: Instead of importing that, we can rely on setting 
```suggestion
            .environment_variable("METAFLOW_KUBERNETES_NAMESPACE", namespace)
```
since namespace is passed down to the function.

valayDave (changes requested):

metaflow/plugins/kubernetes/kubernetes.py
valayDave, line 570: same comment as one for jobsets. You can use `namespace` from env selectors.

metaflow/plugins/kubernetes/kubernetes.py
valayDave: Actually wait! Rather we will need selectors for namespace. Thats a much cleaner and k8s native approach of doing it.

metaflow/plugins/kubernetes/kubernetes.py
```diff
@@ -565,6 +566,7 @@ def create_job_object(
                 "METAFLOW_INIT_SCRIPT", KUBERNETES_SANDBOX_INIT_SCRIPT
             )
             .environment_variable("METAFLOW_OTEL_ENDPOINT", OTEL_ENDPOINT)
+            .environment_variable("METAFLOW_KUBERNETES_NAMESPACE", namespace)
```
savingoyal, line 569: this isn't needed anymore

savingoyal (approved):

valayDave (approved):

PR #1953, opened 2024-08-09
enable heart beat daemon and error tracking exit hook by default

savingoyal:


metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -186,7 +186,7 @@ def argo_workflows(obj, name=None):
 )
 @click.option(
     "--enable-error-msg-capture/--no-enable-error-msg-capture",
-    default=False,
+    default=True,
```
saikonen, line 189: function defaults for this option down the code path could also be changed to keep things consistent

iamsgarg-ob (approved):

PR #1956, opened 2024-08-12
[Ready for Review] Fix issue where resuming on successful run will fail.

darinyu:
when running resume with `python flow.py resume {step_name}`, user may want to rerun the "step" and continue with the execution. In current code, we will blindly copy everything as long as the step was successful.

The proposed change will look at step in topological order and skip cloning the ones on and after the specified steps (and then rerun everything if possible).

### DIFF
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -650,7 +650,7 @@ def resume(
             )
 
     if step_to_rerun is None:
-        rerun_steps = set()
+        steps_to_rerun = set()
     else:
         # validate step name
         if step_to_rerun not in obj.graph.nodes:
@@ -660,7 +660,7 @@ def resume(
                     step_to_rerun, ",".join(list(obj.graph.nodes.keys()))
                 )
             )
-        rerun_steps = {step_to_rerun}
+        steps_to_rerun = {step_to_rerun}
 
     if run_id:
         # Run-ids that are provided by the metadata service are always integers.
@@ -688,7 +688,7 @@ def resume(
         clone_run_id=origin_run_id,
         clone_only=clone_only,
         reentrant=reentrant,
-        rerun_steps=rerun_steps,
+        steps_to_rerun=steps_to_rerun,
         max_workers=max_workers,
         max_num_splits=max_num_splits,
         max_log_size=max_log_size * 1024 * 1024,
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -650,7 +650,7 @@ def resume(
             )
 
     if step_to_rerun is None:
-        clone_steps = set()
+        rerun_steps = set()
     else:
         # validate step name
         if step_to_rerun not in obj.graph.nodes:
@@ -660,7 +660,7 @@ def resume(
                     step_to_rerun, ",".join(list(obj.graph.nodes.keys()))
                 )
             )
-        clone_steps = {step_to_rerun}
+        rerun_steps = {step_to_rerun}
 
     if run_id:
         # Run-ids that are provided by the metadata service are always integers.
@@ -688,7 +688,7 @@ def resume(
         clone_run_id=origin_run_id,
         clone_only=clone_only,
         reentrant=reentrant,
-        clone_steps=clone_steps,
+        rerun_steps=rerun_steps,
         max_workers=max_workers,
         max_num_splits=max_num_splits,
         max_log_size=max_log_size * 1024 * 1024,
--- a/metaflow/runtime.py
+++ b/metaflow/runtime.py
@@ -76,7 +76,7 @@ class NativeRuntime(object):
         clone_run_id=None,
         clone_only=False,
         reentrant=False,
-        clone_steps=None,
+        rerun_steps=None,
         max_workers=MAX_WORKERS,
         max_num_splits=MAX_NUM_SPLITS,
         max_log_size=MAX_LOG_SIZE,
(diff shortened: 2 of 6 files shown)


metaflow/runtime.py
```diff
@@ -160,13 +163,24 @@ def __init__(
             for deco in step.decorators:
                 deco.runtime_init(flow, graph, package, self._run_id)
 
+    def _update_rerun_steps(self):
+        """
+        Any steps following steps to be rerun should also be included as
+        "rerun" steps. This is to ensure that the flow is executed correctly.
+        """
+        for step_name in self._graph.sorted_nodes:
+            if step_name in self._rerun_steps:
```
romain-intel, line 172: we probably need to run this until stability right?

metaflow/runtime.py
```diff
@@ -160,13 +163,24 @@ def __init__(
...
+
     def _new_task(self, step, input_paths=None, **kwargs):
         if input_paths is None:
             may_clone = True
         else:
             may_clone = all(self._is_cloned[path] for path in input_paths)
 
-        if step in self._clone_steps:
+        if step in self._rerun_steps:
             may_clone = False
```
romain-intel, line 184: In general, I think with this new way, we can clean out all the may_clone, etc flags taht are being set. I think it may be confusing because I don't think they are used anymore so that would simplify the code a bit and make it clearer that all the resume logic is now in one place (the clone functions) as opposed to scattered around the runtime code.

romain-intel (commented):
Let's also add a test so we make sure this is caught in the future (using the core tests maybe).

Also, I commented as well somewhere but high level, I think we can simplify the runtime code to remove all resume operations from places outside the clone functions. We currently update some state when things are resuming that I think we could simplify and get rid of.

metaflow/runtime.py
darinyu: I agree, but the code refactor should be in a different PR (I want this fix PR to be concise).

metaflow/runtime.py
darinyu: not sure what you mean "stability" here, the sorted nodes are in topological order already.

metaflow/runtime.py
romain-intel: OK. Let's do a separate PR to clean that up then.

metaflow/cli.py
```diff
@@ -650,7 +650,7 @@ def resume(
             )
 
     if step_to_rerun is None:
-        clone_steps = set()
+        rerun_steps = set()
```
savingoyal, line 653: minor nit - could we do `steps_to_rerun` - reads a bit better

metaflow/runtime.py
```diff
@@ -110,12 +110,15 @@ def __init__(
...
         self._clone_run_id = clone_run_id
         self._clone_only = clone_only
-        self._clone_steps = {} if clone_steps is None else clone_steps
         self._cloned_tasks = []
         self._cloned_task_index = set()
         self._reentrant = reentrant
         self._run_url = None
 
+        # If rerun_steps is specified, we will not clone them in resume mode.
+        self._rerun_steps = {} if rerun_steps is None else rerun_steps
```
savingoyal, line 119: `self._steps_to_rerun = steps_to_rerun or {}`

metaflow/runtime.py
```diff
@@ -160,13 +163,27 @@ def __init__(
             for deco in step.decorators:
                 deco.runtime_init(flow, graph, package, self._run_id)
 
+    def _update_rerun_steps(self):
+        """
+        Any steps following steps to be rerun should also be included as
+        "rerun" steps. This is to ensure that the flow is executed correctly.
```
savingoyal, line 169: maybe i am not following the comment here - but `rerun` steps are the ones that need to be executed during `resume`? if that is the case, why is this update needed?

savingoyal (commented):
left a few comments - mostly for me to understand the nature of the issue here

metaflow/runtime.py
darinyu: for a linear graph a -> b -> c -> d (and b, c succeeded) , when user `resume b`,  we want b and the successor of b (which is c) to rerun as well. Here this `update` function will include all the descendants.

In old resume, I think after this step-to-rerun, it goes back to normal execution mode. 

In the new resume, we need to avoid copying the successful successor tasks (c here) as well, and hence update this "steps-to-rerun" to rerun everything after b.

savingoyal: I would recommend not creating a function would just lifting this logic alongside L:119 so that the code reads a bit easy. Not a whole lot of complexity is embedded in this method.

test/core/tests/resume_succeeded_step.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class ResumeSucceededStepTest(MetaflowTest):
+    """
+    Resuming from the succeeded end step should work
+    """
+
+    RESUME = True
+    # resuming on a successful step.
+    RESUME_STEP = "start"
```
savingoyal, line 11: it might be better to resume a "non-start" step in the test so that we can confirm we are able to skip steps successfully.

savingoyal: in that scenario, we can perhaps test for the existence of an artifact generated in the start step in the resumed end step.

savingoyal (changes requested):

savingoyal:
added a tiny window dressing PR related to this in #1963

metaflow/cli.py
darinyu: done

metaflow/runtime.py
darinyu: done

metaflow/runtime.py
darinyu: done

test/core/tests/resume_succeeded_step.py
darinyu: done

savingoyal (approved):

PR #1962, opened 2024-08-15
Metaflow Configs

romain-intel:
This is not yet fully ready but putting out so people can get a general sense of the direction.


bhogan-bdai:
As a user interested config file instead of using different sets of parameters, what would a config file look like?

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -135,8 +135,8 @@ class KubernetesDecorator(StepDecorator):
     package_sha = None
     run_time_limit = None
 
-    def __init__(self, attributes=None, statically_defined=False):
-        super(KubernetesDecorator, self).__init__(attributes, statically_defined)
+    def init(self):
```
saikonen, line 138: something with this recent setup is causing kubernetes execution to fail:

``` 
in step_init
    if self.attributes["gpu_vendor"].lower() not in ("amd", "nvidia"):
AttributeError: 'NoneType' object has no attribute 'lower'
```

romain-intel: This should be fixed now.

metaflow/decorators.py
```diff
@@ -217,7 +248,13 @@ def add_decorator_options(cmd):
                     % (deco.name, option, seen[option])
                 )
                 raise MetaflowInternalError(msg)
+            elif deco.name.lower() in existing_params:
+                raise MetaflowInternalError(
+                    "Flow decorator '%s' uses an option '%s' which is a reserved "
+                    "keyword. Please use a different option name." % (deco.name, option)
+                )
             else:
+                kwargs["envvar"] = "METAFLOW_FLOW_%s" % option.upper()
```
npow, line 257: there can only be one `envvar` in kwargs?

romain-intel: This is click's way of indicating which env-var maps to that option. So yes, there is only one env-var that maps to the option.

PR #1967, opened 2024-08-17
fix: Metaflow local config init

saikonen:
fixes issue with metaflow local config initialization


romain-intel (approved):
LGTM. Yes, you could have a config but not a local config so this makes sense. Good catch.

PR #1969, opened 2024-08-19
enhance the functionality of _update_env

madhur-ob:


savingoyal (approved):

metaflow/metaflow_current.py
```diff
@@ -62,7 +62,12 @@ def _set_env(
 
     def _update_env(self, env):
         for k, v in env.items():
-            setattr(self.__class__, k, property(fget=lambda _, v=v: v))
+            if isinstance(v, property):
+                setattr(self.__class__, k, v)
```
romain-intel, line 66: nit: I would make sure that the property does not have a setter. I know with the callable, it could technically be broken too but still may be a good idea to enforce it a bit.

metaflow/metaflow_current.py
```diff
@@ -62,7 +62,12 @@ def _set_env(
...
+            elif callable(v):
+                setattr(self, k, functools.partial(v, self))
+            else:
+                setattr(self.__class__, k, property(fget=lambda _, v=v: v))
 
     def __contains__(self, key: str):
```
romain-intel, line 72: With the change above, I don't know how we want to treat __contains__ and get. When things were just properties, it was kind of an easy choice but with callables not 100% sure. Do we need the callables or can we just stick to adding properties?

romain-intel (changes requested):
I am not sure of the use cases and also am a little concerned about adding functions directly (and how that changes the semantics of the current object. Could we stick to just properties?

romain-intel:
I get that @madhur-ob -- it's just that current has particular semantics and adding functiosn to it somewhat expands/breaks it so want to make sure we think carefully about it instead of just adding it. I don't question the technical aspect of it but just don't understand the specific use case for it nor how it plays with the other functions that are there.

savingoyal:
@madhur-ob how about we introduce current.snowflake.session() instead?

PR #1971, opened 2024-08-20
fix(MetaflowData): Raise AttributeError on failed `__getattr__` access

### LINKED ISSUE #1970: `MetaflowData.__getattr__` should raise `AttributeError` on failure
The following code raises a `KeyError` (raised by `dict.__getitem__`) for non-existing names:

https://github.com/Netflix/metaflow/blob/b44f325713a5c6516011beea3259d54ee522c0d6/metaflow/client/core.py#L1029-L1030

According to [Python's documentation](https://docs.python.org/3.10/reference/datamodel.html#object.__getattr__), `__getattr__` should either return the attribute value or raise an `AttributeError`. Otherwise, it will break the [built-in `hasattr()` function](https://docs.python.org/3.10/library/functions.html#hasattr), which relies on checking `AttributeError`.

mutongx:
Python's custom attribute access protocol requires `AttributeError` to be raised on access failure ([docs link](https://docs.python.org/3.10/reference/datamodel.html#object.__getattr__)). The current implement propagates `KeyError` from `dict`, which breaks `hasattr()` operations on `MetaflowData`.

This PR checks for key existence and raise the correct exception in `MetaflowData.__getattr__`.

Fixes #1970.

### DIFF
--- a/metaflow/client/core.py
+++ b/metaflow/client/core.py
@@ -1027,6 +1027,8 @@ class MetaflowData(object):
         self._artifacts = dict((art.id, art) for art in artifacts)
 
     def __getattr__(self, name: str):
+        if name not in self._artifacts:
+            raise AttributeError(name)
         return self._artifacts[name].data
 
     def __contains__(self, var):


romain-intel:
LGTM. We can merge shortly

PR #1972, opened 2024-08-20
feature: better bootstrap dependencies

saikonen:
gets rid of awscli as a dependency, opting to instead use boto3 directly with inline python to download the codepackage. Supersedes #1778 

also introduces an environment variable to allow skipping installing bootstrap dependencies completely. Motivated by #1957

### DIFF
--- a/metaflow/metaflow_environment.py
+++ b/metaflow/metaflow_environment.py
@@ -89,10 +89,21 @@ class MetaflowEnvironment(object):
         It should work silently if everything goes well.
         """
         if datastore_type == "s3":
+            from .plugins.aws.aws_utils import parse_s3_full_path
+
+            bucket, s3_object = parse_s3_full_path(code_package_url)
+            # Explicitly _don't_ use awscli here, because of version incompatibilities
+            # between awscli v1 (installed via pypi) and awscli v2 (installed _not_
+            # through pypi).
+            # Instead, just use boto3 which we already have installed.
             return (
-                '%s -m awscli ${METAFLOW_S3_ENDPOINT_URL:+--endpoint-url=\\"${METAFLOW_S3_ENDPOINT_URL}\\"} '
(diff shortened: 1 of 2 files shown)


metaflow/metaflow_environment.py
```diff
@@ -119,25 +125,34 @@ def _get_download_code_package_cmd(self, code_package_url, datastore_type):
...
-                "%s -m pip install azure-identity azure-storage-blob azure-keyvault-secrets simple-azure-blob-downloader -qqq"
-                % self._python()
-            )
-        elif datastore_type == "gs":
-            cmds.append(
-                "%s -m pip install google-cloud-storage google-auth simple-gcp-object-downloader google-cloud-secret-manager -qqq"
-                % self._python()
-            )
-        else:
+        base_cmd = "{} -m pip install".format(self._python())
```
savingoyal, line 128: we are missing the `-qqq`

metaflow/metaflow_environment.py
```diff
@@ -119,25 +125,34 @@ def _get_download_code_package_cmd(self, code_package_url, datastore_type):
...
-                % datastore_type
+                "Unknown datastore type: {}".format(datastore_type)
             )
-        return " && ".join(cmds)
+
+        cmd = "{} {}".format(
+            base_cmd, " ".join(datastore_packages[datastore_type] + ["requests"])
+        )
+        # skip pip installs if we know that packages might already be available
+        return "if [ -z $SKIP_INSTALL_DEPENDENCIES ]; then {}; fi".format(cmd)
```
romain-intel, line 155: nit: We typically use METAFLOW_ for env vars so we could stick to that?

romain-intel: Not sure how escaping may/may not work though.

romain-intel (commented):
Minor comments but otherwise does not affect us so good to go either ways.

PR #1981, opened 2024-08-22
Change top-level flow decorator options to be prefixed by METAFLOW_FLOW_

romain-intel:
Previously, options like `branch` and `name` (injected by the project decorator for example) could be set using `METAFLOW_BRANCH`. They now need to be set using `METAFLOW_FLOW_BRANCH`.

This change is made to prevent clashes between regular metaflow configuration settings and decorator level options.

No other changes are made so `METAFLOW_RUN_MAX_WORKERS` still works as expected and `METAFLOW_PYLINT` as well.

### DIFF
--- a/metaflow/decorators.py
+++ b/metaflow/decorators.py
@@ -218,6 +218,7 @@ def add_decorator_options(cmd):
                 )
                 raise MetaflowInternalError(msg)
             else:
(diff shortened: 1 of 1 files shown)


romain-intel:
Included in #1962

PR #1982, opened 2024-08-22
POC for package decorator

talsperre:
Adds a new package decorator to package all metaflow runs, including local ones. Additionally, this POC also demonstrates how we can move the logic for packaging/uploading packages outside the critical path of execution, and make it non-blocking by executing it in a thread.


metaflow/cli.py
```diff
@@ -969,6 +975,10 @@ def start(
     ctx.obj.monitor.start()
     _system_monitor.init_system_monitor(ctx.obj.flow.name, ctx.obj.monitor)
 
+    # Add package suffixes and echo to the flow object as they are used by decorators
```
romain-intel, line 978: This seems not needed anymore since you pass it to the MetaflowPackage constructor.

metaflow/cli.py
```diff
@@ -830,8 +830,14 @@ def before_run(obj, tags, decospecs):
     # Package working directory only once per run.
     # We explicitly avoid doing this in `start` since it is invoked for every
     # step in the run.
+    # Update package to be None. The package is now created in the @package decorator in a separate thread
```
romain-intel, line 833: Comment needs updating.

metaflow/package.py
```diff
@@ -59,19 +60,60 @@ def _recurse(root):
...
+        self._init_thread.start()
+
+    def _prepare_and_upload_package(self, flow, environment, flow_datastore, echo):
+        self.logger(f"Creating package for flow: {flow.name}")
         environment.init_environment(echo)
         for step in flow:
             for deco in step.decorators:
                 deco.package_init(flow, step.__name__, environment)
-        self.blob = self._make()
+        self.logger(f"Initalized environment and packages for flow: {flow.name}")
```
romain-intel, line 100: nit: Initialized.

metaflow/runtime.py
```diff
@@ -1552,6 +1552,9 @@ def _launch(self):
                     self.task.ubf_context,
                 )
         env.update(args.get_env())
+        # We add another environment variable that tells us whether we are executing in the main process or
```
romain-intel, line 1555: Is this still needed?

metaflow/plugins/package/package_decorator.py
```diff
@@ -0,0 +1,70 @@
+import os
```
romain-intel, line 1: This now seems to be an empty hull and a lot of code can go away. I wonder if we can also move the env-vars part directly to runtime.py and get rid of this extra decorator.

metaflow/package.py
```diff
@@ -59,19 +60,60 @@ def _recurse(root):
...
+
+        # Make package creation and upload asynchronous
+        self._init_thread = threading.Thread(
+            target=self._prepare_and_upload_package,
+            args=(flow, environment, flow_datastore, echo),
+        )
+        self._init_thread.daemon = True
+        self._init_thread.start()
+
+    def _prepare_and_upload_package(self, flow, environment, flow_datastore, echo):
```
romain-intel, line 94: hardening: wrap in a try/except block to deal with issues like this: https://www.joeshaw.org/python-daemon-threads-considered-harmful/

metaflow/package.py
```diff
@@ -181,6 +223,32 @@ def no_mtime(tarinfo):
         blob[4:8] = [0] * 4  # Reset 4 bytes from offset 4 to account for ts
         return blob
 
+    def wait(self, timeout=DEFAULT_PACKAGE_TIMEOUT):
```
romain-intel, line 226: I would return True/False. Also, quick path on checking on the bool you have about package readiness. Ideally the triple state:
- None: no state so we wait for the thread
- True: done -- no need to wait on anything
- False: some error happened, no need to wait but no package will be available.

We ideally also have to find a way to report that back with wait (this three state)

metaflow/plugins/aws/batch/batch_decorator.py
```diff
@@ -213,11 +213,12 @@ def runtime_init(self, flow, graph, package, run_id):
         self.package = package
         self.run_id = run_id
 
-    def runtime_task_created(
-        self, task_datastore, task_id, split_index, input_paths, is_cloned, ubf_context
-    ):
-        if not is_cloned:
-            self._save_package_once(self.flow_datastore, self.package)
+    # Saving of the package is now done in the package decorator
```
romain-intel, line 216: will be able to clean up save_package_once code as well.

metaflow/plugins/package/package_decorator.py
```diff
@@ -0,0 +1,70 @@
...
+
+from metaflow.decorators import StepDecorator
+from metaflow.exception import MetaflowException
+from metaflow.package import MetaflowPackage
+
+
+class PackageDecorator(StepDecorator):
+    name = "package"
+
+    def set_remote_step(self, decorators):
```
romain-intel, line 14: We will still need something like this. Maybe the "global decorator state" is a good palce to start with this.

romain-intel (commented):
Overall looks good. Haven't played with it but I think it can be cleaned up a tad as well.

PR #1996, opened 2024-08-27
POC modification for Differentiated Artifacts Framework

romain-intel:
This includes both serialization/deserialization customization as well as directed artifact processing.

Examples coming soon.

### DIFF
--- a/metaflow/datastore/content_addressed_store.py
+++ b/metaflow/datastore/content_addressed_store.py
@@ -60,10 +60,13 @@ class ContentAddressedStore(object):
 
         Parameters
         ----------
-        blob_iter : Iterator over bytes objects to save
+        blob_iter : Iterator[Union[bytes, Tuple[bytes, str]]]
+            Iterator over bytes objects to save. If the object is a tuple, the first
+            element is the bytes to save and the second is a compression method or "raw"
         raw : bool, optional
             Whether to save the bytes directly or process them, by default False
-        len_hint : Hint of the number of blobs that will be produced by the
+        len_hint : int, default 0
+            Hint of the number of blobs that will be produced by the
             iterator, by default 0
 
         Returns
@@ -76,11 +79,25 @@ class ContentAddressedStore(object):
 
         def packing_iter():
             for blob in blob_iter:
+                raw_setting = raw
+                pack_version = 1
+                if isinstance(blob, tuple):
+                    blob, compression = blob
+                    if compression == "raw":
+                        raw_setting = True
+                    elif compression == "gzip":
+                        raw_setting = False
+                        pack_version = 1  # Redundant for now
+                    else:
+                        raise MetaflowInternalError(
+                            "Unknown compression method '%s'" % compression
+                        )
+
                 sha = sha1(blob).hexdigest()
                 path = self._storage_impl.path_join(self._prefix, sha[:2], sha)
                 results.append(
                     self.save_blobs_result(
-                        uri=self._storage_impl.full_uri(path) if raw else None,
+                        uri=self._storage_impl.full_uri(path) if raw_setting else None,
                         key=sha,
                     )
                 )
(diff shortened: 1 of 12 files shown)


metaflow/datastore/content_addressed_store.py
```diff
@@ -60,10 +60,13 @@ def save_blobs(self, blob_iter, raw=False, len_hint=0):
 
         Parameters
         ----------
-        blob_iter : Iterator over bytes objects to save
+        blob_iter : Iterator[Union[bytes, Tuple[bytes, str]]]
```
romain-intel, line 63: Added documentation for the specific raw flag

metaflow/datastore/content_addressed_store.py
```diff
@@ -60,10 +60,13 @@ def save_blobs(self, blob_iter, raw=False, len_hint=0):
...
+            Iterator over bytes objects to save. If the object is a tuple, the first
+            element is the bytes to save and the second is a compression method or "raw"
         raw : bool, optional
             Whether to save the bytes directly or process them, by default False
-        len_hint : Hint of the number of blobs that will be produced by the
+        len_hint : int, default 0
```
romain-intel, line 68: clearly need to fix the comment (it seems borked before too)

metaflow/datastore/content_addressed_store.py
```diff
@@ -76,23 +79,40 @@ def save_blobs(self, blob_iter, raw=False, len_hint=0):
 
         def packing_iter():
             for blob in blob_iter:
+                raw_setting = raw
```
romain-intel, line 82: logic is to use raw setting but allow each blob to override. Compression methods would be fixed for now (not sure if there is a lot of value in allowing a huge amount of customization here) but we could add another one ourselves.

metaflow/datastore/datastore_set.py
```diff
@@ -30,7 +27,12 @@ def __init__(
             # produce a set of SHA keys to prefetch based on artifact names
             prefetch = set()
             for ds in self.task_datastores:
-                prefetch.update(ds.keys_for_artifacts(prefetch_data_artifacts))
+                l = [
```
romain-intel, line 30: chage is because there can now be more than one key for each artifact so this is now a list of list.

metaflow/flowspec.py
```diff
@@ -240,15 +243,35 @@ def __iter__(self):
...
+        # Late import to prevent circular deps
+        from .datastore.artifacts import MetaflowArtifact
+
         if self._datastore and name in self._datastore:
             # load the attribute from the datastore...
             x = self._datastore[name]
             # ...and cache it in the object for faster access
-            setattr(self, name, x)
-            return x
+            # For MetaflowArtifact, we extract the representation of it to present to
```
romain-intel, line 253: store representation for user but keep original artifact as well.

metaflow/datastore/task_datastore.py
```diff
@@ -129,8 +149,15 @@ def __init__(
             self._objects = {}
             self._info = {}
         elif self._mode == "r":
-            if data_metadata is not None:
-                # We already loaded the data metadata so just use that
+            if data_metadata is not None and not any(
```
romain-intel, line 152: logic to fetch metadata from S3 if the one from the metadata is incomplete. This way, we don't change metadata view at all. This is only used through client.

metaflow/datastore/task_datastore.py
```diff
@@ -740,8 +801,19 @@ def artifacts_iter():
                     # '_' as they are used by the Metaflow runtime.
                     delattr(flow, var)
                 yield var, val
+                art = flow._orig_artifacts.get(var)
```
romain-intel, line 804: change in behavior possible here since we keep things we serialize if they require it. For others, no change. Note that the idea would be that `post_serialize` could hollow out the MetaflowArtifact as well to keep memory consumption low.

romain-intel (commented):
Added a few clarifying comments.

PR #1998, opened 2024-08-28
feature: Kubernetes kill and list commands

### LINKED ISSUE #1631: Add list and kill support for kubernetes CLI
The metaflow batch CLI supports `list` and `kill` for identifying running jobs on a flow and killing them if needed.  This is useful if you get disconnected from the terminal that started the job and need to stop it.  The kubernetes CLI doesn't have any similar functionality.  Our current workaround is to kill the pods manually (and then watch for and kill the retries) which is tedious and error-prone.  A much better solution would be to add support for `list` and `kill` to the kubernetes CLI.

saikonen:
first draft for introducing `kubernetes list-runs` and `kubernetes kill RUN_ID` commands

list of major caveats:

***label selectors***
Due to Kubernetes being limited to label selectors for filtering resources returned by the API, we need to introduce a new label for flow objects that encodes the flow name for later retrieval. *This makes the changes not backwards compatible*.

Even with the flow hash, we still need to do some in-memory filtering of results for the `kill` command in order to only select jobs for a specific run. Introducing the run_id as a label would introduce an unnecessarily large amount of new labels

***run status***
Displaying a `status` for a run with the `list` command, or filtering by status would require introducing a client-side state machine that would need to go through all the jobs of a run in order to ascertain the current status. The main problem here is that unlike with Argo Workflows, with client-driven Kubernetes we do not have a central status location to query on the cluster side.

closes #1631

### DIFF
--- a/metaflow/plugins/kubernetes/kubernetes_cli.py
+++ b/metaflow/plugins/kubernetes/kubernetes_cli.py
@@ -1,12 +1,18 @@
+import datetime
 import os
 import sys
 import time
 import traceback
 
+from metaflow.plugins.kubernetes.kubernetes_client import KubernetesClient
 import metaflow.tracing as tracing
 from metaflow import JSONTypeClass, util
 from metaflow._vendor import click
-from metaflow.exception import METAFLOW_EXIT_DISALLOW_RETRY, CommandException
+from metaflow.exception import (
+    METAFLOW_EXIT_DISALLOW_RETRY,
+    CommandException,
+    MetaflowException,
+)
 from metaflow.metadata.util import sync_local_metadata_from_datastore
 from metaflow.metaflow_config import DATASTORE_LOCAL_DIR, KUBERNETES_LABELS
 from metaflow.mflog import TASK_LOG_SOURCE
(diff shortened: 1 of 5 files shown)


metaflow/plugins/kubernetes/kubernetes_cli.py
```diff
@@ -305,3 +310,46 @@ def _sync_metadata():
         sys.exit(METAFLOW_EXIT_DISALLOW_RETRY)
     finally:
         _sync_metadata()
+
+
+@kubernetes.command(help="List all runs of the flow on Kubernetes.")
+@click.pass_obj
+def list_runs(obj):
```
savingoyal, line 317: `list` can be congruent to `batch list`

metaflow/plugins/kubernetes/kubernetes_cli.py
```diff
@@ -305,3 +310,46 @@ def _sync_metadata():
...
+        )
+
+    if not executions:
+        obj.echo("No executions for *%s* found on Kubernetes." % (obj.flow.name))
+
+
+@kubernetes.command(help="Kill flow execution on Kubernetes.")
+@click.argument("run-id", required=True, type=str)
+@click.pass_obj
+def kill(obj, run_id):
```
savingoyal, line 344: `kill` can be congruent to `batch kill`

metaflow/plugins/kubernetes/kubernetes_client.py
```diff
@@ -60,6 +64,48 @@ def get(self):
...
+        results = self._find_active_jobs(flow_name, run_id, user)
+
+        return list(results)
+
+    def kill_jobs(self, flow_name, run_id, user, echo):
+        jobs = self._find_active_jobs(
+            flow_name, run_id, user, field_selector="status.successful==0"
+        )
+
+        def _kill_job(job):
```
savingoyal, line 96: does this work for argo-workflows?

saikonen: as-is, no this does not work. I was leveraging the existing `RunningJob` for the kill logic, but from what I understood we do not create Kubernetes Jobs for runs on Argo Workflows, as the workflow dag wraps the pods directly? As such, the lookup for executing jobs doesn't return anything to be terminated.

I would maybe leave argo workflows out of scope for this PR, as there is a more direct way to issue termination to a workflow on that side instead of individually killing pods.

If we want this to apply to argo-workflows as well then I can change the lookup to search for active pods instead of jobs, and introduce the kill logic separately.

saikonen:
closing for the time being in favour of #2023

PR #2002, opened 2024-08-28
[core-change] Top level options from step decorators.

valayDave:
- CLI options injected by step decorators in the Top level CLI
- Step decorators exposing an additional hooks in the lifecycle to accept options passed down from top level
- top level option injection is only done by step decorators that are statically set in the code

### DIFF
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -843,6 +843,7 @@ def version(obj):
 
 @tracing.cli_entrypoint("cli/start")
 @decorators.add_decorator_options
+@decorators.add_step_decorator_options
 @click.command(
     cls=click.CommandCollection,
     sources=[cli] + plugins.get_plugin_cli(),
@@ -1008,6 +1009,8 @@ def start(
         echo,
         deco_options,
     )
+
+    decorators._inject_step_decorator_options(ctx.obj.flow, deco_options)
 
     # In the case of run/resume, we will want to apply the TL decospecs
     # *after* the run decospecs so that they don't take precedence. In other
--- a/metaflow/decorators.py
+++ b/metaflow/decorators.py
@@ -207,6 +207,7 @@ def add_decorator_options(cmd):
     flow_cls = getattr(current_flow, "flow_cls", None)
     if flow_cls is None:
         return cmd
+
     for deco in flow_decorators(flow_cls):
         for option, kwargs in deco.options.items():
             if option in seen:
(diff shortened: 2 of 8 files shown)


metaflow/decorators.py
```diff
@@ -257,6 +281,34 @@ class MyDecorator(StepDecorator):
...
+        Its a class method to ensure that any decorator even attached via `--with`
+        (given its statically present in the code too) can access the options set in the cli.
+        """
+        pass
+
+    def get_top_level_options(self):
+        """
+        Return a list of option-value pairs that correspond to top-level
+        options that should be passed to subprocesses (tasks). The option
+        names should be a subset of the keys in self.options.
```
valayDave, line 305: todo: fix me

metaflow/parameters.py
valayDave, line 1: Needed all of this so that `click` can set step decorator based  Options in cli.py

metaflow/cli.py
```diff
@@ -843,6 +843,7 @@ def version(obj):
 
 @tracing.cli_entrypoint("cli/start")
 @decorators.add_decorator_options
+@decorators.add_step_decorator_options
```
valayDave, line 846: Top level CLI Options added for *only* statically set decorators in code.

metaflow/cli.py
```diff
@@ -1009,6 +1010,8 @@ def start(
         deco_options,
     )
 
+    decorators._inject_step_decorator_options(ctx.obj.flow, deco_options)
```
valayDave, line 1013: todo: fixme : give better name

metaflow/decorators.py
```diff
@@ -257,6 +281,34 @@ class MyDecorator(StepDecorator):
                   pass them around with every lifecycle call.
     """
 
+    cli_options = {}
```
valayDave, line 284: named it `cli_options` because `options` seemed a like a fairly common attributes that user defined step decorators could also have

metaflow/runner/click_api.py
```diff
@@ -186,7 +186,7 @@ def from_cli(cls, flow_file: str, cli_collection: Callable) -> Callable:
         flow_cls = extract_flow_class_from_file(flow_file)
         flow_parameters = [p for _, p in flow_cls._get_parameters()]
         with flow_context(flow_cls) as _:
-            add_decorator_options(cli_collection)
+            add_step_decorator_options(add_decorator_options(cli_collection))
```
valayDave, line 189: added this here because i assumed that low level API will also need it .

metaflow/decorators.py
```diff
@@ -220,6 +221,29 @@ def add_decorator_options(cmd):
...
+    step_deco_dict = _get_all_step_decos()
+    seen = set()
+    step_deco_names = getattr(current_flow, "unique_step_decos_in_flow", None)
+    if step_deco_names is None:
+        return cmd
+
+    for deco_name in step_deco_names:
+        deco = step_deco_dict[deco_name]
+        for option, kwargs in deco.cli_options.items():
+            if option in seen:
```
romain-intel, line 238: this would need to be global (across step and flow level decorators)

metaflow/decorators.py
```diff
@@ -257,6 +281,34 @@ class MyDecorator(StepDecorator):
                   pass them around with every lifecycle call.
     """
 
+    cli_options = {}
+    # `cli_options` is similar to the one the flow decorator. It will be used to
+    # pass options to the step decorator from the cli.
+
+    @classmethod
+    def step_options_init(cls, flow, options_dict):
```
romain-intel, line 289: This can probably be combined with the config thing (ie: we probably don't need to add two new hooks)

romain-intel (commented):
Minor nits. I am also not convinced with the UX but maybe I just need to see a use-case you are thinking. I know this argument could also be made for flow-level decorators but it feels like this could be passed using configs. With step decorators in multiple places, something feels a bit weird providing a top level option to control the setting of *all* the decorators. If there is such a setting, could it not be placed at the flow level? Again, don't have a concrete use case example so hard for me to speculate but the "single setting" "multiple places" feels a bit weird to me.

Issue #2005, opened 2024-08-30
GPU constraint of 0 added to Kubernetes step spec, making it unsatisfiable


dmcguire81:
Here's a short example flow to illustrate the issue:

```python
from metaflow import FlowSpec, step, resources


class NonGPUFlow(FlowSpec):
    @step
    def start(self):
        print("Starting flow")
        self.next(self.end)

    # this step hangs indefinitely without any feedback,
    # and no indication of error such as PodUnschedulable,
    # because nothing can statisfy the constraint gpu=0
    @resources(cpu=2, memory=2048)
    @step
    def end(self):
        print("Flow completed")


if __name__ == "__main__":
    NonGPUFlow()
```

When this is run on Kubernetes, as follows, the `end` step can never be allocated resources:

```sh
$ python NonGPUFlow.py run --with kubernetes
Metaflow 2.12.10+netflix-ext(1.2.1) executing NonGPUFlow for user:davidmcguire
Validating your flow...
    The graph looks good!
Running pylint...
    Pylint not found, so extra checks are disabled.
2024-08-29 22:35:30.205 Workflow starting (run-id 49):
2024-08-29 22:35:34.857 [49/start/240 (pid 6056)] Task is starting.
2024-08-29 22:35:37.488 [49/start/240 (pid 6056)] [pod t-592d481e-nhznc-2j5cx] Task is starting (Pod is pending)...
2024-08-29 22:37:06.836 [49/start/240 (pid 6056)] [pod t-592d481e-nhznc-2j5cx] Setting up task environment.
2024-08-29 22:37:32.401 [49/start/240 (pid 6056)] [pod t-592d481e-nhznc-2j5cx] Downloading code package...
2024-08-29 22:37:33.310 [49/start/240 (pid 6056)] [pod t-592d481e-nhznc-2j5cx] Code package downloaded.
2024-08-29 22:37:33.411 [49/start/240 (pid 6056)] [pod t-592d481e-nhznc-2j5cx] Task is starting.
2024-08-29 22:37:35.124 [49/start/240 (pid 6056)] [pod t-592d481e-nhznc-2j5cx] Starting flow
2024-08-29 22:37:40.147 [49/start/240 (pid 6056)] [pod t-592d481e-nhznc-2j5cx] Task finished with exit code 0.
2024-08-29 22:37:41.789 [49/start/240 (pid 6056)] Task finished successfully.
[...]

PR #2006, opened 2024-08-30
Only use a GPU value for Kubernetes that is non-null and non-zero

dmcguire81:
Fix for #2005

### DIFF
--- a/metaflow/plugins/kubernetes/kubernetes_decorator.py
+++ b/metaflow/plugins/kubernetes/kubernetes_decorator.py
@@ -271,12 +271,11 @@ class KubernetesDecorator(StepDecorator):
 
         for deco in decos:
             if isinstance(deco, ResourcesDecorator):
+                self.attributes["gpu"] = None
                 for k, v in deco.attributes.items():
-                    # If GPU count is specified, explicitly set it in self.attributes.
+                    # If GPU count is specified and non-zero, explicitly set it in self.attributes.
                     if k == "gpu" and v:
                         self.attributes["gpu"] = v
-                    else:
-                        self.attributes["gpu"] = None
 
                     if k in self.attributes:
                         if self.defaults[k] is None:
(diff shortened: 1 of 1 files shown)


metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -273,8 +273,10 @@ def step_init(self, flow, graph, step, decos, environment, flow_datastore, logge
             if isinstance(deco, ResourcesDecorator):
                 for k, v in deco.attributes.items():
                     # If GPU count is specified, explicitly set it in self.attributes.
-                    if k == "gpu" and v != None:
+                    if k == "gpu" and v:
                         self.attributes["gpu"] = v
+                    else:
+                        self.attributes["gpu"] = None
```
dmcguire81, line 279: Realizing this works locally by coincidence of `gpu` being the last attribute (late night) - will fix.

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -271,9 +271,10 @@ def step_init(self, flow, graph, step, decos, environment, flow_datastore, logge
 
         for deco in decos:
             if isinstance(deco, ResourcesDecorator):
+                self.attributes["gpu"] = None
                 for k, v in deco.attributes.items():
-                    # If GPU count is specified, explicitly set it in self.attributes.
-                    if k == "gpu" and v != None:
+                    # If GPU count is specified and non-zero, explicitly set it in self.attributes.
+                    if k == "gpu" and v:
```
saikonen, line 277: looking into alternatives for this at the moment. Note that the default from resources deco is actually `"0"` and not int, so it will pass this check.

dmcguire81: Hmm, I wonder if it's cast before it ends up here, because the [type in the comments in `int`](https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/resources_decorator.py#L26). Also, this seems to fix the problem, which is my biggest priority for removing the footgun; setting the default to `None` would also expedite the fix, so that works, as well.

I can add some logging to capture the default values, here, to demonstrate.

dmcguire81:
Fixed by #2100

PR #2009, opened 2024-08-31
Modify getting Metaflow version

romain-intel:
Favors reading the INFO file if present to be able to have the most accurate version of Metaflow when executing remotely (especially in the presence of extensions).

Also limit reading the INFO file to once per process (as opposed to possibly twice).

Finally, gets the version of the source of Metaflow (and not the current directory)

### DIFF
--- a/metaflow/__init__.py
+++ b/metaflow/__init__.py
@@ -151,6 +151,7 @@ if sys.version_info >= (3, 7):
     from .runner.nbdeploy import NBDeployer
 
 __version_addl__ = []
+__version_override__ = None
 _ext_debug("Loading top-level modules")
 for m in _tl_modules:
     extension_module = load_module(m)
@@ -164,11 +165,15 @@ for m in _tl_modules:
         ext_version = _format_git_describe(
             _call_git_describe(cwd=os.path.dirname(extension_module.__file__))
         )
+        ext_override = None
         if ext_version is None:
-            ext_version = extension_module.__version__
-        if ext_version:
-            version_info = "%s(%s)" % (version_info, ext_version)
-        __version_addl__.append(version_info)
+            ext_version = getattr(extension_module, "__version__", "<unk>")
+            ext_override = getattr(extension_module, "__name_override__", None)
+        if ext_override:
+            __version_override__ = "%s %s" % (ext_override, ext_version)
+            __version_addl__.clear()
+        else:
+            __version_addl__.append("%s(%s)" % (version_info, ext_version))
 
 if __version_addl__:
     __version_addl__ = ";".join(__version_addl__)
--- a/metaflow/__init__.py
+++ b/metaflow/__init__.py
@@ -42,14 +42,9 @@ If you have any questions, feel free to post a bug report/question on the
 Metaflow GitHub page.
 """
 
-import importlib
+import os
 import sys
-import types
 
-from os import path
-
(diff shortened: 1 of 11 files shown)


metaflow/__init__.py
```diff
@@ -164,8 +161,13 @@ class and related decorators.
             alias_submodules(extension_module, tl_package, None, extra_indent=True)
         )
         version_info = getattr(extension_module, "__mf_extensions__", "<unk>")
-        if extension_module.__version__:
-            version_info = "%s(%s)" % (version_info, extension_module.__version__)
+        ext_version = _format_git_describe(
+            _call_git_describe(cwd=os.path.dirname(extension_module.__file__))
+        )
+        if ext_version is None:
```
savingoyal, line 167: shouldn't this be the other way around?

also, i am getting `Metaflow 2.12.18.post5-git2201f4e-dirty+ob(4.3.19) executing...` - the interesting thing is that we don't have version 4.3.19 for any extensions internally. we pin the extension version explicitly to v1. it's likely that `git describe --tags` is the offender here.

metaflow/metaflow_version.py
```diff
@@ -57,87 +56,110 @@ def find_git_on_windows():
...
 
     """
 
-    version = format_git_describe(call_git_describe(), pep440=pep440)
-    version_addl = None
-    if version is None:  # not a git repository
-        import metaflow
-
+    # To get the version we do the following:
+    #  - First check if we have an INFO file with it. If so, use that as it is
```
savingoyal, line 138: what is _it_ here?

metaflow/metaflow_version.py
```diff
@@ -57,87 +56,110 @@ def find_git_on_windows():
...
+    #    the most reliable way to get the version. In particular, when running remotely,
+    #    metaflow is installed in a directory and if any extension using distutils,
```
savingoyal, line 140: grammar nit

metaflow/__init__.py
savingoyal: @madhur-ob do you happen to know how ob releases are git tagged?

savingoyal (changes requested):
seems like this change doesn't quite work for our current set of extensions. @madhur-ob can you help triage what might be going wrong?

metaflow/metaflow_version.py
```diff
@@ -57,87 +56,110 @@ def find_git_on_windows():
...
+            # first, make sure we are actually in a Metaflow repo,
+            # not some other repo
             arguments = [GIT_COMMAND, "rev-parse", "--show-toplevel"]
             reponame = (
-                check_output(arguments, cwd=CURRENT_DIRECTORY, stderr=fnull)
-                .decode("ascii")
-                .strip()
+                check_output(arguments, cwd=cwd, stderr=DEVNULL).decode("ascii").strip()
             )
+            print("For %s got reponame: %s" % (cwd, reponame))
```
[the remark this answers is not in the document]
romain-intel: Right -- that was a debugging leftover -- I'll remove it.

metaflow/metaflow_version.py
romain-intel: I'll rephrase: we have an INFO file present

metaflow/metaflow_version.py
romain-intel: rephrase: if any extension is using distutils to return its version

saikonen (approved):

PR #2010, opened 2024-09-03
[Ready for Review] Update deployer API to support more orchestrators

talsperre:


metaflow/plugins/argo/argo_workflows_deployer.py
```diff
@@ -257,6 +259,7 @@ class ArgoWorkflowsDeployer(DeployerImpl):
     """
 
     TYPE: ClassVar[Optional[str]] = "argo-workflows"
+    TIMEOUT: ClassVar[Optional[int]] = 10
```
savingoyal, line 262: this may not be needed anymore given #1980 is already merged and released

talsperre: Rebased PR to pull in latest changes.

metaflow/runner/deployer.py
```diff
@@ -397,6 +399,7 @@ def create(self, **kwargs) -> DeployedFlow:
             self.name = content.get("name")
             self.flow_name = content.get("flow_name")
             self.metadata = content.get("metadata")
+            self.additional_metadata = content.get("additional_metadata", {})
```
savingoyal, line 402: minor nit - can you consider renaming this to `additional_meta` instead of `metadata` - so that it is apparent that this field has no relationship to metadata. also, if you can add a note above this line that this attribute is used in extensions - that will ensure that nobody considers this dead code in the future. otherwise it's good to go - please feel free to merge once the changes are in.

PR #2015, opened 2024-09-04
fix kill main pid runner + vendor psutil

madhur-ob:


savingoyal (approved):

romain-intel:
To discuss, I am not a huge fan of adding .so like this. The benefit seems small as well over subprocess.

PR #2016, opened 2024-09-04
changing display URLs for argo

iamsgarg-ob:


metaflow/plugins/argo/argo_workflows.py
```diff
@@ -2397,11 +2397,12 @@ def _pager_duty_notification_links(self):
         if ARGO_WORKFLOWS_UI_URL:
             links.append(
                 {
-                    "href": "%s/workflows/%s/%s"
-                    % (
-                        ARGO_WORKFLOWS_UI_URL.rstrip("/"),
-                        "{{workflow.namespace}}",
-                        "{{workflow.name}}",
+                    "href": ARGO_WORKFLOWS_UI_URL.rstrip("/").format(
```
iamsgarg-ob, line 2400: ARGO_WORKFLOWS_UI_URL should be set as:

`https://metaflow.foo.argo.com/{uri}/{namespace}/{workflow_name}`

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -291,17 +291,23 @@ def create(
...
+                }
+                obj.echo("See the deployed workflow here:", bold=True)
+                argo_workflowtemplate_link = ARGO_WORKFLOWS_UI_URL.rstrip("/").format(
+                    **workflow_template_url_values
+                )
+        except:
             obj.echo(
-                "%s/%s\n\n" % (argo_workflowtemplate_link, obj.workflow_name),
-                indent=True,
+                "workflow template was created - errored in formatting workflow template URL"
```
saikonen, line 308: error message could be more specific that `METAFLOW_ARGO_WORKFLOWS_UI_URL` could not be used to format the url.

This exception is never raised though, if you try it out with a template url that is missing the required keys
```
METAFLOW_ARGO_WORKFLOWS_UI_URL=http://example.com/
```

the only exception you will get is if the template string contains keys that are not present in `workflow_template_url_values`, such as
```
METAFLOW_ARGO_WORKFLOWS_UI_URL=http://example.com/{not_present}
```

saikonen (changes requested):

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -667,16 +674,26 @@ def _convert_value(param):
         "(run-id *{run_id}*).".format(name=obj.workflow_name, run_id=run_id),
         bold=True,
     )
+    try:
 
-    run_url = (
-        "%s/%s/%s" % (UI_URL.rstrip("/"), obj.flow.name, run_id) if UI_URL else None
-    )
-
-    if run_url:
```
saikonen, line 675: we do not want to remove the url to Metaflow UI in favour of Argo UI though? preferably we keep both, with tidy messaging as introduced in https://github.com/Netflix/metaflow/pull/1934/files#diff-27da6198697d250e8ed1cea087795ecbbf2a0e8e8e0861dd51c1f0062c2f0c41R653
which is a somewhat overlapping PR to this one.

saikonen (changes requested):

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -667,16 +674,26 @@ def _convert_value(param):
...
+            workflow_run_url_values = {
+                "uri": "workflows",
+                "namespace": KUBERNETES_NAMESPACE,
+                "workflow_name": obj.workflow_name,
+                "suffix": argo_workflow_id,
+            }
+            obj.echo(json.dumps(workflow_run_url_values))
+            obj.echo("See the run in the UI at\n:", bold=True)
+            run_url = ARGO_WORKFLOWS_UI_URL.rstrip("/").format(
+                **workflow_run_url_values
```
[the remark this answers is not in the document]
iamsgarg-ob: Yep that is the intended behavior.. the exception gets raised if the template has a variable that is missing from the dictionary that is supplied. If the supplied dictionary has extra  vars they will simply be ignored

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -667,16 +674,26 @@ def _convert_value(param):
...
-            bold=True,
-        )
+        if ARGO_WORKFLOWS_UI_URL:
+            workflow_run_url_values = {
+                "uri": "workflows",
+                "namespace": KUBERNETES_NAMESPACE,
+                "workflow_name": obj.workflow_name,
+                "suffix": argo_workflow_id,
+            }
+            obj.echo(json.dumps(workflow_run_url_values))
```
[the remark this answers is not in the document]
iamsgarg-ob: Ha.. good catch.. will remove

metaflow/plugins/argo/argo_workflows_cli.py
iamsgarg-ob: Yes that is the intended behavior. Since it's an existing config var in an effort to be backwards compatible this was a little neat solution

Issue #1974, opened 2024-08-20
Request: Allow shadowing in metaflow_extensions


JettJones:
This is kind of an elaborate use case, but my team is frequently running into issues while using `metaflow_extensions` where our internal definitions are found multiple times, causing the program to fail with the `Conflicts` RuntimeError while checking for shadowed extensions:

 > https://github.com/Netflix/metaflow/blob/master/metaflow/extension_support/__init__.py#L761
    
 The path where we'd hit this error is:
1. Running on an external platform, we install our source as a python package.
2. Then, we also copy a subset of our source, and place it earlier in sys.path - this gets us ready to live edit some sources, while having all dependencies (and third party code) to fall back on.
3. But then at runtime - an import to metaflow sources triggers the extension logic init - and two copies of `metaflow_extensions` raises an error.

I'd like to be able to tell metaflow (either in general or for our extension path ), that the conflict is ok and expected - maybe a `METAFLOW_EXTENSIONS_BUT_CHILL` environment variable, that turns the error into a warning.  Stated differently - the python path is defined correctly in our case, the version that is found first is the one we intend to use, so the extra check in `extension_support` is all that blocks our workflow. 

I also may be missing some use case for `_AliasLoader` so if there's an easier work-around I'm excited to take pointers. Thanks!

romain-intel:
Hey @JettJones -- thanks for the feedback. Let me make sure I understand properly:
- you package all your source (including metaflow and extensions) in a python package
- you install that package
- you also copy a portion of your code (including extensions) earlier in your sys.path to effectively "shadow" the code that was installed by the python package?

Is that correct?

If so, yes, you are right that the current extensions mechanism will complain when it can't figure out what extensions you are trying to import. The rationale behind it is that the `metaflow_extensions` package is actually a namespace package and it therefore looks for all "participants" to this namespace package.

Could you let me know what exact error message is being printed by the line you point to? Depending on that, there may be a way of doing it by, for example, ignoring packages that are "in the path" and only looking at installed packages (although I don't know if that is what you want -- it would remove the ability to "edit" the files that you put earlier in your path).

We could also add a flag as you mention but am a bit weary of adding something that explicitly breaks functionality. We could call it `METAFLOW_I_KNOW_THE_RISKS_DISABLE_EXTENSIONS_CHECK` or something like that but would rather scope it as narrowly as possible. The loading mechanism for extensions is already a bit complicated and want to avoid making it even more obscure.

[...]

PR #2018, opened 2024-09-05
Add installed_extensions information

romain-intel:
This adds information to the system tags about the metaflow extensions that are installed. This makes it easier to reproduce the exact same environment. Information about whether the package information is complete or not is also included.

This PR also fixes some issues with extension loading and allows extension information to be gathered in a programatic fashion thereby enabling the discovery of extensions.


romain-intel:
Fixes #1974

PR #2023, opened 2024-09-10
feature: backwards compatible Kubernetes list and kill

### LINKED ISSUE #1631: Add list and kill support for kubernetes CLI

saikonen:
supersedes #1998 as the first attempt, as this approach is backwards compatible.

closes #1631 

**caveat**
this approach iterates through all active pods in a namespace indiscriminately (with a limit of `1000` per request), only filtering on the client side for relevant ones. As a result, in deployments with enormous amounts of active pods, the operation can take quite long.

### DIFF
--- a/metaflow/plugins/kubernetes/kubernetes_cli.py
+++ b/metaflow/plugins/kubernetes/kubernetes_cli.py
@@ -8,7 +8,7 @@ from metaflow.plugins.kubernetes.kubernetes_client import KubernetesClient
 import metaflow.tracing as tracing
 from metaflow import JSONTypeClass, util
 from metaflow._vendor import click
-from metaflow.exception import METAFLOW_EXIT_DISALLOW_RETRY, CommandException
+from metaflow.exception import METAFLOW_EXIT_DISALLOW_RETRY, MetaflowException
 from metaflow.metadata.util import sync_local_metadata_from_datastore
 from metaflow.metaflow_config import DATASTORE_LOCAL_DIR, KUBERNETES_LABELS
 from metaflow.mflog import TASK_LOG_SOURCE
@@ -379,5 +379,12 @@ def kill(obj, run_id, user, my_runs):
     flow_name, run_id, user = parse_cli_options(
         obj.flow.name, run_id, user, my_runs, obj.echo
     )
+
+    if run_id is not None and run_id.startswith("argo-") or user == "argo-workflows":
+        raise MetaflowException(
+            "Killing pods launched by Argo Workflows is not supported. "
+            "Use *argo-workflows terminate* instead."
+        )
+
     kube_client = KubernetesClient()
     kube_client.kill_pods(flow_name, run_id, user, obj.echo)
--- a/metaflow/plugins/kubernetes/kubernetes_cli.py
+++ b/metaflow/plugins/kubernetes/kubernetes_cli.py
@@ -309,7 +309,7 @@ def step(
         _sync_metadata()
 
 
-@kubernetes.command(help="List all runs of the flow on Kubernetes.")
+@kubernetes.command(help="List unfinished Kubernetes tasks of this flow.")
 @click.option(
     "--my-runs",
     default=False,
@@ -352,11 +352,11 @@ def list(obj, run_id, user, my_runs):
         )
 
     if not pods:
-        obj.echo("No active jobs found for *%s* on Kubernetes." % (flow_name))
+        obj.echo("No active Kubernetes pods found.")
 
 
 @kubernetes.command(
-    help="Terminate Kubernetes tasks for this flow. Only terminates current tasks, but will not affect future ones from retries."
+    help="Terminate unfinished Kubernetes tasks of this flow. Killed pods may result in newer attempts when using @retry."
 )
 @click.option(
     "--my-runs",
--- a/metaflow/plugins/kubernetes/kubernetes_cli.py
+++ b/metaflow/plugins/kubernetes/kubernetes_cli.py
@@ -8,11 +8,7 @@ from metaflow.plugins.kubernetes.kubernetes_client import KubernetesClient
 import metaflow.tracing as tracing
 from metaflow import JSONTypeClass, util
 from metaflow._vendor import click
-from metaflow.exception import (
-    METAFLOW_EXIT_DISALLOW_RETRY,
-    CommandException,
-    MetaflowException,
-)
+from metaflow.exception import METAFLOW_EXIT_DISALLOW_RETRY, CommandException
 from metaflow.metadata.util import sync_local_metadata_from_datastore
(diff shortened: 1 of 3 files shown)


metaflow/plugins/kubernetes/kubernetes_cli.py
```diff
@@ -305,3 +307,77 @@ def _sync_metadata():
         sys.exit(METAFLOW_EXIT_DISALLOW_RETRY)
     finally:
         _sync_metadata()
+
+
+@kubernetes.command(help="List all runs of the flow on Kubernetes.")
```
savingoyal, line 312: List unfinished Kubernetes tasks of this flow

metaflow/plugins/kubernetes/kubernetes_cli.py
```diff
@@ -305,3 +307,77 @@ def _sync_metadata():
...
+                status=pod.status.phase,
+            )
+        )
+
+    if not pods:
+        obj.echo("No active jobs found for *%s* on Kubernetes." % (flow_name))
+
+
+@kubernetes.command(
+    help="Terminate Kubernetes tasks for this flow. Only terminates current tasks, but will not affect future ones from retries."
```
savingoyal, line 359: Terminate unfinished Kubernetes tasks of this flow. Killed pods may result in newer attempts when using @retry.

metaflow/plugins/kubernetes/kubernetes_cli.py
```diff
@@ -305,3 +307,77 @@ def _sync_metadata():
...
+                    pod.metadata.labels.get("workflows.argoproj.io/workflow"),
+                ),
+                pod_id=pod.metadata.name,
+                startedAt=format_timestamp(pod.status.start_time),
+                status=pod.status.phase,
+            )
+        )
+
+    if not pods:
+        obj.echo("No active jobs found for *%s* on Kubernetes." % (flow_name))
```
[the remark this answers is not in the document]
savingoyal: maybe simply - `No active Kubernetes pods found.`

metaflow/plugins/kubernetes/kubernetes_cli.py
savingoyal: retry with the small r

metaflow/plugins/kubernetes/kubernetes_client.py
```diff
@@ -60,6 +63,91 @@ def get(self):
...
+
+    def kill_pods(self, flow_name, run_id, user, echo):
+        from kubernetes.stream import stream
+
+        api_instance = self._client.CoreV1Api()
+        pods = self._find_active_pods(flow_name, run_id, user)
+
+        def _kill_pod(pod):
+            echo(
+                "Killing Kubernetes pod %s in namespace %s\n"
```
savingoyal, line 125: we can skip namespace - it might confuse users who might not be familiar with kubernetes namespace but might be familiar with metaflow namespace.

metaflow/plugins/kubernetes/kubernetes_client.py
```diff
@@ -60,6 +63,91 @@ def get(self):
...
+                % (pod.metadata.name, pod.metadata.namespace)
+            )
+            try:
+                stream(
```
savingoyal, line 129: we may not be able to kill with this mechanism internally

metaflow/plugins/kubernetes/kubernetes_client.py
```diff
@@ -60,6 +63,91 @@ def get(self):
...
+                        "/sbin/killall5",
+                    ],
+                    stderr=True,
+                    stdin=False,
+                    stdout=True,
+                    tty=False,
+                )
+            except Exception as ex:
+                # best effort kill for pod can fail.
+                echo("failed to kill pod %s: %s" % (pod.metadata.name, str(ex)))
```
savingoyal, line 146: Failed killing Kubernetes pod %s - %s

metaflow/plugins/kubernetes/kubernetes_client.py
```diff
@@ -1,8 +1,10 @@
+from concurrent.futures import ThreadPoolExecutor
```
savingoyal, line 1: kubernetes_client is overloaded internally so we will have to ensure a symmetric change in the extensions

savingoyal (approved):
lgtm! just need to enhance the killing logic so that it works internally.

metaflow/plugins/kubernetes/kubernetes_client.py
```diff
@@ -60,6 +63,98 @@ def get(self):
...
+                    stderr=True,
+                    stdin=False,
+                    stdout=True,
+                    tty=False,
+                )
+            except Exception:
+                # best effort kill for pod can fail.
+                try:
+                    # Metaflows Kubernetes job names are always a subset of the task pods name.
+                    job_name = pod.metadata.name[:-6]
```
savingoyal, line 145: is this always true?

saikonen: the pods get their name from the jobs `generateName` which comes from [here](https://github.com/Netflix/metaflow/blob/feature/backwards-compatible-kubernetes-kill/metaflow/plugins/kubernetes/kubernetes.py#L161)

For the time being this holds true, as the entropy from the `generateName` suffix is quite low.

saikonen: changed this a bit so now it drops the last suffix regardless of length (in case Kubernetes decides to increase their entropy in the future). Also added a comment to clarify why the job name can be inferred from the pod name

metaflow/plugins/kubernetes/kubernetes_client.py
```diff
@@ -60,6 +63,100 @@ def get(self):
...
+                    stdout=True,
+                    tty=False,
+                )
+            except Exception:
+                # best effort kill for pod can fail.
+                try:
+                    # Metaflows Kubernetes job names are of the format
+                    # t-{uid}-{k8s_job_generate_name_suffix}-{k8s_pod_generate_name_suffix}
+                    # so we can infer the job name from the pod name by dropping the last suffix
+                    job_name = "-".join(pod.metadata.name.split("-")[:-1])
```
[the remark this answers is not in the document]
saikonen: seems so, or upon further inspection its also part of the pod labels as well so I'll go with that as the simplest source

PR #2028, opened 2024-09-13
fix: attach conda decorator in conda_base flow_init

saikonen:
attach `conda` decorator to steps in the `conda_base` decorator when using it. 

This solves an issue where a virtual environment implemented in an extension does not append `conda` as part of environment decospecs, but still needs to know whether a step needs to execute on conda due to the flow decorator.

`decorators._attach_decorators` is a no-op for existing conda decorators on a step so this should not have adverse effects.

### DIFF
--- a/metaflow/plugins/pypi/conda_decorator.py
+++ b/metaflow/plugins/pypi/conda_decorator.py
@@ -353,9 +353,8 @@ class CondaFlowDecorator(FlowDecorator):
     def flow_init(
         self, flow, graph, environment, flow_datastore, metadata, logger, echo, options
     ):
-        # NOTE: This makes sure that using conda_base marks all steps as executing with conda,
-        # without relying on --environment conda setting decospecs.
-        # Important for extensions implementing custom virtual environments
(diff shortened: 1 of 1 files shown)


metaflow/plugins/pypi/conda_decorator.py
```diff
@@ -353,6 +353,13 @@ def is_attribute_user_defined(self, name):
     def flow_init(
         self, flow, graph, environment, flow_datastore, metadata, logger, echo, options
     ):
+        # NOTE: This makes sure that using conda_base marks all steps as executing with conda,
+        # without relying on --environment conda setting decospecs.
+        # Important for extensions implementing custom virtual environments
+        from metaflow import decorators
+
+        decorators._attach_decorators(flow, ["conda"])
```
savingoyal, line 361: do we need a symmetric change for `pypi_base` too? if not, can you add the reasoning as a code comment?

saikonen: pypi_base already does this for itself so no change needed on that side. I'll refine the above comment a bit to clarify.

Issue #245, opened 2020-07-02
Composable Flows and Steps


talebzeghmi:
Large ML projects spanning teams reuse pipelines and models (ex: ensembles, feature engineering, etc).  

There are two aspects of reuse:
1. Reuse a whole Flow, to be able to compose a Flow of other Flows.
2. Reuse a step (imagine it to be a feature engineering step).  Steps currently do not have in parameters and return values making reuse more difficult.

related: https://github.com/Netflix/metaflow/issues/144

seeravikiran:
We have been thinking about (1) [as graph composition] and hopefully will publish more details on the thoughts we have about it. cc @tuulos 
For (2) - you could still get the sharing esp. for feature engineering transform as a library of functions (instead of steps); that can just be imported within your step. Some of our team internal to Netflix employ this route for sharing such business logic.

Also, for relatively common collection of transformations you could still use (1) if you want to even reduce the step boilerplate from being repeated.

dpatschke:
@talebzeghmi Thank you for opening this issue! Your issue has articulated some of the exact metaflow architectural questions that our team has been having around productionizing/pipelining metaflow ... especially around the reusability of feature engineering code within multiple flows.

I don't want to have to copy and paste scikit-learn Transformer code to each new modeling flow especially when there is a lot of boilerplate/utility code that I've written around:
1) leveraging pandas to protect against differing columns being passed in.
2) pulling in a tagged 'production' model from a Run that is then reloaded for just the data 'transform' and not the 'fit' as well.

@seeravikiran Thanks for some of the recommendations regarding structuring and code reusability to address some of items presented in this issue. I will continue to investigate what that would look like on our end. In the meantime, I would like to point you to [this post](https://gitter.im/metaflow_org/community?at=5e2a361201914e3e04392580) made on the metaflow community page that actually proposes a pretty interesting idea to the issue. I'm curious as to your thoughts on this (or something like this).

tuulos:
As @seeravikiran pointed above, we have plans for graph composition. Meanwhile, this form of subclassing is supported https://github.com/Netflix/metaflow/issues/144#issuecomment-592245062

dpatschke:
@tuulos Thanks for the response and the reference. This is extremely helpful and greatly appreciated!

talebzeghmi:
@tuulos, would you be able to share an RFC kind of document on how Metaflow would support composition? In this way we can give feedback from our Applied Scientists on it's usability, before the code is written.

thank you!

[...]

Issue #144, opened 2020-02-27
Support Pythonic Inheritance of Flow Class


jensqin:
Sometimes a couple of steps, parameters or functions can be used by multiple flows. It will be ideal if metaflow supports python standard class inheritance. 

For example, running this script:
```python
from metaflow import FlowSpec, step


class BaseFlow(FlowSpec):
    @step
    def start(self):
        print("this is the start")
        self.next(self.step1)

    @step
    def step1(self):
        print("base step 1")
        self.next(self.end)

    @step
    def end(self):
        print("base step end.")


class SubFlow(BaseFlow):
    @step
    def step1(self):
        print("sub step 1")
        self.next(self.step2)

    @step
    def step2(self):
        print("sub step 2")
        self.next(self.end)


if __name__ == "__main__":
    SubFlow()
```
will get:
```
this is the start
sub step 1
sub step 2
base step end
```

tuulos:
thanks @jensqin for the proposal. Just to be clear, the following type of inheritance is supported already:
```
class MyModel():
    def fit(self):
         self.model = train(self.data)

class MyFlow(FlowSpec, MyModel):
     @step
     def start(self):
          self.fit()
```
However, the graph defined by `FlowSpec` can't be spread across multiple `class`es, i.e. graph composition is not supported.

Most likely we will start supporting graph composition by allowing separate subgraphs, so you will be able to do `self.next(SubFlow)`. This mix-in style approach should be even more flexible than subclassing.

jensqin:
Thank you for responding! @tuulos . Just a few thoughts on it:

1. personally I don't like the current subclassing practice primarily because a data science pipeline is not a subclass of a model intuitively. By 'inheritance' I actually meant I want to reuse a flow as a part of another flow rather than use a fixture for different flows.

2. I like the idea of graph composition very much. However, what if I also want some sort of integrity of my subflow or child data pipeline? For example, suppose SubFlow is used by two data pipelines, when loading the artifacts I would like to get access to only the runs of the first data pipeline. Or in other words, without a 'name' of the whole pipeline or the top level class, how can I figure out other parts of the SubFlow? 

I am just a newbie and trying to understand machine learning workflow. Hope these questions will not bother you!

PR #2035, opened 2024-09-17
Allow Subclassing flows

simonwardjones:
Currently it is not possible to subclass a Flow as in the example snippet below.

**This PR alters the graph construction that enables this flow subclassing.** Instead of using the `ast.NodeVisitor` class to collect the steps and their `ast.FunctionDef`'s.  I have instead iterated on the `dir(flow)` and then extracted the function `ast.FunctionDef`'s.

```


class BaseFlow(FlowSpec):
    @step
    def start(self):
        print("this is the start")
        self.next(self.step1)

    @step
    def step1(self):
        print("base step 1")
        self.next(self.end)

    @step
    def end(self):
        print("base step end.")


class SubFlow(BaseFlow):
    @step
    def step1(self):
        print("sub step 1")
        self.next(self.step2)

    @step
    def step2(self):
        print("sub step 2")
        self.next(self.end)


if __name__ == "__main__":
    SubFlow()
```

As referenced in this issue: https://github.com/Netflix/metaflow/issues/144
and discussed in this issue: https://github.com/Netflix/metaflow/issues/245

I have not yet added tests for this PR as the approach to this might take some discussion. The current Test harness setup uses the `FlowFormatter` class to generate flows on the fly but all of these assume the flow directly inherits off the FlowSpec. How might we go about testing this kind of subclassing?

### DIFF
--- a/metaflow/graph.py
+++ b/metaflow/graph.py
@@ -1,6 +1,7 @@
 import inspect
 import ast
 import re
+import textwrap
 
 
 from .util import to_pod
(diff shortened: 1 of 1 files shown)


simonwardjones:
Closing to re-open from feature branch in forked repository

PR #2036, opened 2024-09-17
WIP: Initial commit for spin steps

talsperre:
This PR aims to add support for spinning up "baby" steps in Metaflow by using the client API as an input datastore.


romain-intel (commented):
Haven't look at all the logic yet but just a few comments. We can discuss in person. Looks quite promising!

PR #2039, opened 2024-09-17
Fix signal handling in SubprocessManager

### LINKED ISSUE #2038: Buggy signal handler impl in CommandManager and SubprocessManager
When a `CommandManager` is created, it sets the interrupt signal handler to `self._handle_sigint` [here](https://github.com/Netflix/metaflow/blob/7a09f2a12350f542b55397399cac19571adfa41d/metaflow/runner/subprocess_manager.py#L175), which replaces any previously set handler (see [docs](https://docs.python.org/3/library/signal.html#signal.signal)).

When running a command synchronously, `self._handle_sigint` is implemented via asyncio even outside of async context. Sending a `kill -INT pid` results in
```
/metaflow/metaflow/runner/subprocess_manager.py:288: RuntimeWarning: coroutine 'CommandManager.kill' was never awaited
  self.cleanup()
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```
The async function never gets called (except on program exit i think). Also due to blocking impl [here](https://github.com/Netflix/metaflow/blob/7a09f2a12350f542b55397399cac19571adfa41d/metaflow/runner/subprocess_manager.py#L280), the `SubprocessManager.commands` doesn't get populated. Due to both of these issues individually, the subprocess doesn't receive any signals. When sending a `ctrl-c`, the terminal dispatches the signal to all subprocesses (tested on iterm2 with zsh and foot with zsh), this still results in the `RuntimeWarning`.

In an async context, calling `asyncio.create_task` without awaiting on it (which is impossible due to a synchronous signal handler), doesn't actually run the task until exit (at least with my testing code). According to [asyncio docs](https://docs.python.org/3/library/asyncio-eventloop.html#unix-signals), only a signal handler registered via `loop.add_signal_handler` can interact with the event loop itself. I assume calling `asyncio.create_task` which internally uses the current event loop is considered an interaction.

Even if that worked, when running multiple commands with `async_run_command`, each call would replace the previous signal handler, leaving only one active (likely for the last call made).

E: According to [posix spec](https://pubs.opengroup.org/onlinepubs/000095399/basedefs/xbd_chap11.html#tag_11_01_09), `ctrl-c` sends a special `INTR` character, which `Generates a SIGINT signal which is sent to all processes in the foreground process group`.

testing scripts:

sleep.py
```py
import time
import signal

def handler(signum, frame):
    print("Signal handler called with signal", signum)
    exit(0)
signal.signal(signal.SIGINT, handler)
signal.signal(signal.SIGTERM, handler)

time.sleep(10)
```

main.py
```py
import time
import asyncio
from subprocess_manager import SubprocessManager

async def async_main():
    async with SubprocessManager() as spm:
        pid1 = await spm.async_run_command(["python", "./sleep.py"])
        pid2 = await spm.async_run_command(["python", "./sleep.py"])
        pid3 = await spm.async_run_command(["python", "./sleep.py"])

        command_obj1 = spm.get(pid1)
        command_obj2 = spm.get(pid2)
        command_obj3 = spm.get(pid3)

        await command_obj3.wait()
        print(f"time='{time.time()}' p3_pid='{command_obj3.process.pid}'")
        async for pos, line in command_obj3.stream_log("stdout"):
            print(f"p3: {line}")

        await command_obj2.wait()
        print(f"time='{time.time()}' p2_pid='{command_obj2.process.pid}'")
        async for pos, line in command_obj2.stream_log("stdout"):
            print(f"p2: {line}")

        await command_obj1.wait()
        print(f"time='{time.time()}' p1_pid='{command_obj1.process.pid}'")
        async for pos, line in command_obj1.stream_log("stdout"):
            print(f"p1: {line}")

def main():
    spm = SubprocessManager()
    spm.run_command(["python", "./sleep.py"], show_output=True)

if __name__ == "__main__":
    print(f"Starting at {time.time()}")
    asyncio.run(async_main())
    # main()
    print(f"Finished at {time.time()}")
```

filipcacky:
Fixes #2038

I've moved signal handling to SubprocessManager and implemented it via asyncio/signal depending on the running context.

### DIFF
--- a/metaflow/runner/subprocess_manager.py
+++ b/metaflow/runner/subprocess_manager.py
@@ -136,9 +136,7 @@ class SubprocessManager(object):
 
         self.commands[pid] = command_obj
 
-        command_obj.process.wait()
-        command_obj.stdout_thread.join()
-        command_obj.stderr_thread.join()
+        command_obj.sync_wait()
 
         return pid
 
@@ -195,7 +193,7 @@ class SubprocessManager(object):
         for v in self.commands.values():
             v.cleanup()
 
-    def kill(self, termination_timeout: float = 5):
+    async def kill(self, termination_timeout: float = 5):
         """
         Kill all managed subprocesses and their descendants.
 
@@ -205,19 +203,31 @@ class SubprocessManager(object):
             The time to wait after sending a SIGTERM to a subprocess and its descendants
             before sending a SIGKILL.
         """
-        pids = [v.process.pid for v in self.commands.values()]
 
+        pids = [v.process.pid for v in self.commands.values() if v.process is not None]
+        await async_kill_processes_and_descendants(pids, 5)
+
+    def sync_kill(self, termination_timeout: float = 5):
+        """
+        Kill all managed subprocesses and their descendants synchronously.
+
+        Parameters
+        ----------
+        termination_timeout : float, default 5
+            The time to wait after sending a SIGTERM to a subprocess and its descendants
+            before sending a SIGKILL.
+        """
+        pids = [v.process.pid for v in self.commands.values() if v.process is not None]
         kill_processes_and_descendants(
             pids,
             termination_timeout,
         )
 
     def _handle_sigint(self, signum, frame):
-        self.kill()
+        self.sync_kill()
 
     async def _async_handle_sigint(self):
-        pids = [v.process.pid for v in self.commands.values()]
-        await async_kill_processes_and_descendants(pids, 5)
+        await self.kill()
 
 
 class CommandManager(object):
@@ -302,6 +312,20 @@ class CommandManager(object):
                     "within %s seconds." % (self.process.pid, command_string, timeout)
                 )
 
+    def sync_wait(self):
+        """
+        Wait for the subprocess to finish synchronously.
+
+        You can only call `sync_wait` if `run` has already been called.
+        """
+
+        if not self.run_called:
+            raise RuntimeError("No command run yet to wait for...")
+
+        self.process.wait()
+        self.stdout_thread.join()
+        self.stderr_thread.join()
+
     def run(self, show_output: bool = False, wait: bool = True) -> int:
         """
         Run the subprocess synchronously. This can only be called once.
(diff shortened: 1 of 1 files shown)


metaflow/runner/subprocess_manager.py
```diff
@@ -10,23 +10,55 @@
 from typing import Callable, Dict, Iterator, List, Optional, Tuple
 
 
+def send_signals(pid, signal):
+    subprocess.call(["pkill", signal, "-P", str(pid)])
```
filipcacky, line 14: Not using `check_call` here on purpose, if the subprocess doesn't have any descendants, `check_call` raises an exception. This (for some arcane reason) slowed the code down substantially.

metaflow/runner/subprocess_manager.py
```diff
@@ -81,8 +125,14 @@ def run_command(
...
 
         command_obj = CommandManager(command, env, cwd)
-        pid = command_obj.run(show_output=show_output)
+        pid = command_obj.run(show_output=show_output, wait=False)
+
         self.commands[pid] = command_obj
+
+        command_obj.process.wait()
+        command_obj.stdout_thread.join()
+        command_obj.stderr_thread.join()
```
filipcacky, line 134: Not happy about this, but its needed in order to populate `self.commands`. `wait` was already taken with an async impl. Not sure if i can change interfaces and potentially break any code that depends on this. Behaviour for any other code won't change due to the new defaulted param `wait` in `run`.

metaflow/runner/subprocess_manager.py
```diff
@@ -42,6 +74,18 @@ class SubprocessManager(object):
...
         self.commands: Dict[int, CommandManager] = {}
 
+        try:
+
+            async def handle_sigint():
+                await self._async_handle_sigint()
+
+            asyncio.get_running_loop().add_signal_handler(
+                signal.SIGINT, lambda: asyncio.create_task(handle_sigint())
+            )
```
filipcacky, line 84: Closure required in order to capture `self`, also asyncio signal handlers can't be async.

metaflow/runner/subprocess_manager.py
```diff
@@ -10,23 +10,58 @@
...
 
-def kill_process_and_descendants(pid, termination_timeout):
+def send_signals(pid, signal):
     # TODO: there's a race condition that new descendants might
     # spawn b/w the invocations of 'pkill' and 'kill'.
     # Needs to be fixed in future.
     try:
-        subprocess.check_call(["pkill", "-TERM", "-P", str(pid)])
-        subprocess.check_call(["kill", "-TERM", str(pid)])
+        subprocess.call(["pkill", signal, "-P", str(pid)])
```
filipcacky, line 18: Not using check_call here on purpose, if the subprocess doesn't have any descendants, check_call raises an exception. This (for some arcane reason) slowed the code down substantially.

savingoyal:
closing in favor of #2053

PR #2041, opened 2024-09-17
include extra attrs of includefile in argo parameter annotations

madhur-ob:


metaflow/plugins/argo/argo_workflows.py
```diff
@@ -488,6 +497,9 @@ def _process_parameters(self):
                 description=param.kwargs.get("help"),
                 is_required=is_required,
             )
+
+            if extra_attrs:
+                parameters[param.name]["extra_attrs"] = extra_attrs
```
savingoyal, line 502: any reason to add the extra_attrs dict and not add the values directly to parameters[param.name]

madhur-ob: hmm, not really..

just collecting all of extra stuff at one place..
that's it..

let me know if this needs to change..

savingoyal: Yeah it will be good to follow the same convention that we currently have for other attributes of parameters

madhur-ob: fixed

saikonen (approved):

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -463,6 +463,15 @@ def _process_parameters(self):
             else:
                 param_type = str(param.kwargs.get("type").__name__)
 
+            extra_attrs = {}
+            if isinstance(param.kwargs.get("type"), FilePathClass):
+                extra_attrs["is_text"] = getattr(
```
iamsgarg-ob, line 468: is there a situation where  the `is_text` ends up as `True` when the File in question is not a text file?

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -463,6 +463,15 @@ def _process_parameters(self):
...
+                    param.kwargs.get("type"), "_is_text", True
+                )
+                extra_attrs["encoding"] = getattr(
```
iamsgarg-ob, line 471: same question here.

metaflow/plugins/argo/argo_workflows.py
madhur-ob: these are the defaults:

```
    def __init__(
        self,
        name: str,
        required: bool = False,
        is_text: bool = True,
        encoding: str = "utf-8",
        help: Optional[str] = None,
        **kwargs: Dict[str, str]
    ):
```

that situation is possible, but then, user should supply `is_text` to `False` explicitly..

madhur-ob: the burden of supplying the correct value of `is_text` and `encoding` is on the user..

metaflow/plugins/argo/argo_workflows.py
madhur-ob: answered above..

metaflow/plugins/argo/argo_workflows.py
madhur-ob: yeah, flow/step will have errors and will fail..

madhur-ob: it will be at runtime IIRC

iamsgarg-ob: oh I see what you mean https://docs.metaflow.org/api/flowspec#includefile.

iamsgarg-ob (approved):

PR #2042, opened 2024-09-17
use extra attrs of includefile

madhur-ob:
should be merged after https://github.com/Netflix/metaflow/pull/2041

PR #1955, opened 2024-08-12
ability to fetch existing DeployedFlow

madhur-ob:

PR #2043, opened 2024-09-19
ability to trigger existing workflows

madhur-ob:
Replaces https://github.com/Netflix/metaflow/pull/1955
Should only be merged after https://github.com/Netflix/metaflow/pull/2041
Also includes changes from https://github.com/Netflix/metaflow/pull/2042

Usage:

```py
with Deployer(
    flow_file="../../try.py",
    branch="hello",
    production=True,
    environment="conda"
).argo_workflows() as ar:
    ar_obj = ar.create()

    # save this somewhere / remember it...
    identifier = ar_obj.name
    metadata = ar_obj.metadata

    # beta is a user defined parameter in flow file '../../try.py'
    df = DeployedFlow.from_deployment(identifier=identifier, metadata=metadata)
    triggered_run = df.trigger(beta={"madhur": "cool"})
```

### DIFF
--- a/metaflow/plugins/argo/argo_workflows_deployer.py
+++ b/metaflow/plugins/argo/argo_workflows_deployer.py
@@ -385,5 +385,3 @@ class ArgoWorkflowsDeployer(DeployerImpl):
                 "delete": delete,
             }
         )
-
-        setattr(DeployedFlow, "from_deployment", staticmethod(from_deployment))
--- a/metaflow/plugins/argo/argo_workflows_deployer.py
+++ b/metaflow/plugins/argo/argo_workflows_deployer.py
@@ -65,6 +65,7 @@ def from_deployment(identifier: str, metadata: Optional[str] = None):
     metadata_annotations = workflow_template.get("metadata", {}).get("annotations", {})
 
     flow_name = metadata_annotations.get("metaflow/flow_name", "")
+    username = metadata_annotations.get("metaflow/owner", "")
     parameters = json.loads(metadata_annotations.get("metaflow/parameters", {}))
 
     # these two only exist if @project decorator is used..
@@ -90,9 +91,13 @@ def from_deployment(identifier: str, metadata: Optional[str] = None):
             fp.write(fake_flow_file_contents)
 
         if branch_name is not None:
-            d = Deployer(fake_flow_file.name, **project_kwargs).argo_workflows()
+            d = Deployer(
+                fake_flow_file.name, env={"METAFLOW_USER": username}, **project_kwargs
+            ).argo_workflows()
         else:
-            d = Deployer(fake_flow_file.name).argo_workflows(name=identifier)
+            d = Deployer(
+                fake_flow_file.name, env={"METAFLOW_USER": username}
+            ).argo_workflows(name=identifier)
 
         d.name = identifier
         d.flow_name = flow_name
--- a/metaflow/plugins/argo/argo_workflows_deployer.py
+++ b/metaflow/plugins/argo/argo_workflows_deployer.py
@@ -1,15 +1,110 @@
 import sys
+import json
 import tempfile
 from typing import Optional, ClassVar
 
+from metaflow.client.core import get_metadata
+from metaflow.exception import MetaflowException
+from metaflow.plugins.argo.argo_client import ArgoClient
+from metaflow.metaflow_config import KUBERNETES_NAMESPACE
 from metaflow.plugins.argo.argo_workflows import ArgoWorkflows
 from metaflow.runner.deployer import (
+    Deployer,
     DeployerImpl,
     DeployedFlow,
     TriggeredRun,
     get_lower_level_group,
     handle_timeout,
 )
+
+
+def generate_fake_flow_file_contents(
+    flow_name: str, param_info: dict, project_name: Optional[str] = None
+):
+    params_code = ""
+    for _, param_details in param_info.items():
+        param_name = param_details["name"]
+        param_type = param_details["type"]
+        param_help = param_details["description"]
+        param_required = param_details["is_required"]
+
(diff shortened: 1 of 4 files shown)


metaflow/runner/deployer.py
```diff
@@ -271,6 +271,21 @@ class DeployedFlow(object):
...
+        self.metadata = self.deployer.metadata
+
+    @staticmethod
+    def from_deployment(
+        identifier: str,
+        metadata: str = None,
+        impl: str = "argo-workflows",
+    ):
+        if impl == "argo-workflows":  # TODO: use a metaflow config variable for `impl`
+            from metaflow.plugins.argo.argo_workflows_deployer import from_deployment
```
saikonen, line 285: one option for getting rid of argo specific impl details in core would be to introduce a new extension point through `metaflow.plugins`, f.ex. 

```
DEPLOYER_DESC = [
    ("argo-workflows", ".argo.argo_workflows_deployer")
]
```
and then just looking up the correct implementation from the list

saikonen: could support extending this through extensions as well but that change can be a different PR. `from_deployment` needs to be well defined at that point too.

metaflow/plugins/argo/argo_workflows_deployer.py
```diff
@@ -12,6 +18,95 @@
 )
 
 
+def generate_fake_flow_file_contents(
```
romain-intel, line 21: This can probably move up right since it is not specific to argo. @talsperre , would this be something we could use?

metaflow/runner/deployer.py
romain-intel: my preference would be inject these methods `from_argo_workflows` etc directly in the `DeployedFlow` class and have `from_deployment` be simply an alias that does something like this: 

```
@classmethod
def from_deployment(cls, identifier: str, metadata: str = None, impl: str = "argo-workflows")
    return getattr(cls, "from_%s" % impl)(identifier, metadata)
```
(modulo a few things like the _ and -).

We could inject these methods exactly like what we do in the `Deployer.__init__` except do it in the metaclass for `DeployedFlow`. For that matter, we should also probably do that in the `Deployer` metaclass too (because right now, we reprocess the deployer impl every time we init the class and set variables on the class itself (which is kind of weird).

metaflow/plugins/argo/argo_workflows_deployer.py
```diff
@@ -12,6 +18,95 @@
...
+
+    project_kwargs = {}
+    if branch_name is not None:
+        if branch_name.startswith("prod."):
+            project_kwargs["production"] = True
+            project_kwargs["branch"] = branch_name[len("prod.") :]
+        elif branch_name.startswith("test."):
+            project_kwargs["branch"] = branch_name[len("test.") :]
+        elif branch_name == "prod":
+            project_kwargs["production"] = True
```
romain-intel, line 82: there will be a problem here if it's a different user than the one who deployed it right?

metaflow/runner/deployer.py
madhur-ob: Please check the simpler way here: https://github.com/Netflix/metaflow/pull/2043/commits/a014447b645c203148d044813642bf03af0aad3b

metaflow/plugins/argo/argo_workflows_deployer.py
madhur-ob: @savingoyal how can I pass a different user?

metaflow/plugins/argo/argo_workflows_deployer.py
madhur-ob: it is kinda specific to argo in the sense that the `param_info` is recovered from the metadata information of the deployed flow on argo workflows..

metaflow/runner/deployer.py
madhur-ob: Hmm, this won't work in all cases...

madhur-ob: https://github.com/Netflix/metaflow/pull/2043/commits/5ee9c544765c60e796eb3572885bee45abd49956 works well..

we can now do stuff like:

```py
identifier = "parameterflowproject.user.madhurtandon23gmail.com.parameterflow"
metadata = "service@https://metadata.dev-latest.outerbounds.xyz/p/default/"

df = DeployedFlow.from_argo_workflows_deployment(identifier=identifier, metadata=metadata)
triggered_run = df.trigger(beta={"madhur": "cool"}, myfile="/Users/madhur/Desktop/output.csv")
```

madhur-ob:
We can now do:

```py
identifier = "parameterflowproject.user.madhurtandon23gmail.com.parameterflow"
metadata = "service@https://metadata.dev-latest.outerbounds.xyz/p/default/"

df = DeployedFlow.from_deployment(identifier=identifier, metadata=metadata)

triggered_run = df.trigger(beta={"madhur": "cool"}, myfile="/Users/madhur/Desktop/output.csv")
```

some variants..

```py
df = DeployedFlow.from_deployment(identifier=identifier, metadata=metadata, impl="xxx")
```
OR
without passing in `impl`, one can also do:
```
METAFLOW_FROM_DEPLOYMENT_IMPL=zzz python demo.py
```

The method `from_argo_workflows_deployment` is now removed and only a single `from_deployment` exists to which one can pass a specific `impl` or perhaps also set the config variable: `METAFLOW_FROM_DEPLOYMENT_IMPL`

metaflow/plugins/argo/argo_workflows_deployer.py
savingoyal: you can test this out by changing the username - take a look at the get_username method.

madhur-ob:
Closing in favour of #2135

PR #2033, opened 2024-09-17
Fix Runner async_run and async_resume

filipcacky:
Fixes #2034

PR #2053, opened 2024-09-23
runner fixes and enhancements

madhur-ob:
This PR is inspired by a few ideas from
- https://github.com/Netflix/metaflow/pull/2039 (fix for SIGINT handling)
- https://github.com/Netflix/metaflow/pull/2033 (`poll()` not available for the `Process` object created async)

It also comes with:
- storing json contents in runner attribute file
- setting status as "timeout" if it occurs

### DIFF
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -1,5 +1,5 @@
 import inspect
-import os
+import json
 import sys
 import traceback
 from datetime import datetime
@@ -7,6 +7,7 @@ from functools import wraps
 
 import metaflow.tracing as tracing
 from metaflow._vendor import click
+from metaflow.client.core import get_metadata
 
 from . import decorators, lint, metaflow_version, namespace, parameters, plugins
 from .cli_args import cli_args
@@ -698,15 +699,17 @@ def resume(
     runtime.print_workflow_info()
 
     runtime.persist_constants()
-    write_file(
-        runner_attribute_file,
-        "%s@%s:%s"
-        % (
-            obj.metadata.__class__.TYPE,
-            obj.metadata.__class__.INFO,
-            "/".join((obj.flow.name, runtime.run_id)),
-        ),
-    )
+
+    if runner_attribute_file:
+        with open(runner_attribute_file, "w") as f:
+            json.dump(
+                {
+                    "run_id": runtime.run_id,
+                    "flow_name": obj.flow.name,
+                    "metadata": get_metadata(),
+                },
+                f,
+            )
 
     # We may skip clone-only resume if this is not a resume leader,
     # and clone is already complete.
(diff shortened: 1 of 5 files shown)


metaflow/runner/subprocess_manager.py
```diff
@@ -42,6 +42,19 @@ class SubprocessManager(object):
     def __init__(self):
         self.commands: Dict[int, CommandManager] = {}
 
+        try:
+            loop = asyncio.get_running_loop()
+            loop.add_signal_handler(
+                signal.SIGINT,
+                lambda: self._handle_sigint(signum=signal.SIGINT, frame=None),
```
filipcacky, line 49: I think this would be better with `asyncio.create_task` and a wrapped async signal handler.

By using a synchronous one with `time.sleep` and `subprocess.check_call`, instead of awaiting on `asyncio.sleep` and `asyncio.create_subprocess_exec`, you're blocking other coroutines from running in the meantime.

madhur-ob: Thanks @filipcacky, would you like to contribute this change?

madhur-ob: If you want, either you can push on top of my branch / fork..
OR we can merge this and you can raise another.. what say?

filipcacky: Hey, sorry, I'm from EU so it's a bit late for me 🙂 I've implemented this in async in my MR. You can copy it over from there.

madhur-ob: No worries, you can contribute tomorrow as discussed in DM 😄

filipcacky: Didn't have permissions to push/create PR to your remote, so I added my changes here https://github.com/filipcacky/metaflow/commit/5dd4aaba43db3f405a1623d05ae25c806c587c9f. Feel free to cherry-pick.

metaflow/runner/utils.py
```diff
@@ -40,14 +41,21 @@ def clear_and_set_os_environ(env: Dict):
     os.environ.update(env)
 
 
+def check_process_status(command_obj: "CommandManager"):
```
saikonen, line 44: are the quotes required on the type?

metaflow/cli.py
```diff
@@ -698,15 +699,17 @@ def resume(
...
-        ),
-    )
+
+    if runner_attribute_file:
+        with open(runner_attribute_file, "w") as f:
+            json.dump(
+                {
+                    "run_id": runtime.run_id,
+                    "flow_name": obj.flow.name,
+                    "metadata": get_metadata(),
```
saikonen, line 709: any issues with backwards compatibility as this changes the contents of the runner_attribute_file significantly?

metaflow/runner/utils.py
madhur-ob: hmm, it is only imported when we do:

```
if TYPE_CHECKING:
    from .subprocess_manager import CommandManager
```

I usually put quotes on custom types / class names etc.

metaflow/cli.py
madhur-ob: it shouldn't be because this file is ONLY used by the runner..

PR #2056, opened 2024-09-24
Async runner improvements

filipcacky:
Follow-up to https://github.com/Netflix/metaflow/pull/2053

Added an async implementation of Runner's signal handling, as discussed here https://github.com/Netflix/metaflow/pull/2053#discussion_r1771641866.

Added an async implementation of Runner's `__get_executing_run`, which is used in `async_run` and `async_resume`, the current implementation relies on `time.sleep`, which blocks other coroutines from running (modified code from my previous PR https://github.com/Netflix/metaflow/pull/2033).

### DIFF
--- a/metaflow/runner/metaflow_runner.py
+++ b/metaflow/runner/metaflow_runner.py
@@ -2,13 +2,16 @@ import importlib
 import os
 import sys
 import json
-import tempfile
 
 from typing import Dict, Iterator, Optional, Tuple
 
 from metaflow import Run
 
-from .utils import handle_timeout, async_handle_timeout
+from .utils import (
+    temporary_fifo,
+    handle_timeout,
+    async_handle_timeout,
+)
 from .subprocess_manager import CommandManager, SubprocessManager
 
 
@@ -267,10 +270,8 @@ class Runner(object):
     async def __aenter__(self) -> "Runner":
         return self
 
-    def __get_executing_run(self, tfp_runner_attribute, command_obj):
-        content = handle_timeout(
-            tfp_runner_attribute, command_obj, self.file_read_timeout
-        )
+    def __get_executing_run(self, attribute_file_fd, command_obj):
+        content = handle_timeout(attribute_file_fd, command_obj, self.file_read_timeout)
         content = json.loads(content)
         pathspec = "%s/%s" % (content.get("flow_name"), content.get("run_id"))
 
@@ -282,9 +283,9 @@ class Runner(object):
         )
         return ExecutingRun(self, command_obj, run_object)
 
-    async def __async_get_executing_run(self, tfp_runner_attribute, command_obj):
+    async def __async_get_executing_run(self, attribute_file_fd, command_obj):
         content = await async_handle_timeout(
-            tfp_runner_attribute, command_obj, self.file_read_timeout
+            attribute_file_fd, command_obj, self.file_read_timeout
         )
         content = json.loads(content)
         pathspec = "%s/%s" % (content.get("flow_name"), content.get("run_id"))
@@ -312,12 +313,9 @@ class Runner(object):
         ExecutingRun
             ExecutingRun containing the results of the run.
         """
-        with tempfile.TemporaryDirectory() as temp_dir:
-            tfp_runner_attribute = tempfile.NamedTemporaryFile(
-                dir=temp_dir, delete=False
-            )
+        with temporary_fifo() as (attribute_file_path, attribute_file_fd):
             command = self.api(**self.top_level_kwargs).run(
-                runner_attribute_file=tfp_runner_attribute.name, **kwargs
+                runner_attribute_file=attribute_file_path, **kwargs
             )
 
             pid = self.spm.run_command(
@@ -328,7 +326,7 @@ class Runner(object):
             )
             command_obj = self.spm.get(pid)
 
-            return self.__get_executing_run(tfp_runner_attribute, command_obj)
+            return self.__get_executing_run(attribute_file_fd, command_obj)
 
     def resume(self, **kwargs):
         """
@@ -346,12 +344,9 @@ class Runner(object):
         ExecutingRun
             ExecutingRun containing the results of the resumed run.
         """
-        with tempfile.TemporaryDirectory() as temp_dir:
-            tfp_runner_attribute = tempfile.NamedTemporaryFile(
-                dir=temp_dir, delete=False
-            )
+        with temporary_fifo() as (attribute_file_path, attribute_file_fd):
             command = self.api(**self.top_level_kwargs).resume(
-                runner_attribute_file=tfp_runner_attribute.name, **kwargs
+                runner_attribute_file=attribute_file_path, **kwargs
             )
 
             pid = self.spm.run_command(
@@ -362,7 +357,7 @@ class Runner(object):
             )
             command_obj = self.spm.get(pid)
 
-            return self.__get_executing_run(tfp_runner_attribute, command_obj)
+            return self.__get_executing_run(attribute_file_fd, command_obj)
 
     async def async_run(self, **kwargs) -> ExecutingRun:
         """
@@ -382,12 +377,9 @@ class Runner(object):
         ExecutingRun
             ExecutingRun representing the run that was started.
         """
-        with tempfile.TemporaryDirectory() as temp_dir:
-            tfp_runner_attribute = tempfile.NamedTemporaryFile(
-                dir=temp_dir, delete=False
-            )
+        with temporary_fifo() as (attribute_file_path, attribute_file_fd):
             command = self.api(**self.top_level_kwargs).run(
-                runner_attribute_file=tfp_runner_attribute.name, **kwargs
+                runner_attribute_file=attribute_file_path, **kwargs
             )
 
             pid = await self.spm.async_run_command(
@@ -397,9 +389,7 @@ class Runner(object):
             )
             command_obj = self.spm.get(pid)
 
-            return await self.__async_get_executing_run(
-                tfp_runner_attribute, command_obj
-            )
+            return await self.__async_get_executing_run(attribute_file_fd, command_obj)
 
     async def async_resume(self, **kwargs):
         """
@@ -419,12 +409,9 @@ class Runner(object):
         ExecutingRun
             ExecutingRun representing the resumed run that was started.
         """
-        with tempfile.TemporaryDirectory() as temp_dir:
-            tfp_runner_attribute = tempfile.NamedTemporaryFile(
-                dir=temp_dir, delete=False
-            )
+        with temporary_fifo() as (attribute_file_path, attribute_file_fd):
             command = self.api(**self.top_level_kwargs).resume(
-                runner_attribute_file=tfp_runner_attribute.name, **kwargs
+                runner_attribute_file=attribute_file_path, **kwargs
             )
 
             pid = await self.spm.async_run_command(
@@ -434,9 +421,7 @@ class Runner(object):
             )
             command_obj = self.spm.get(pid)
 
-            return await self.__async_get_executing_run(
-                tfp_runner_attribute, command_obj
-            )
+            return await self.__async_get_executing_run(attribute_file_fd, command_obj)
 
     def __exit__(self, exc_type, exc_value, traceback):
         self.spm.cleanup()
--- a/metaflow/runner/metaflow_runner.py
+++ b/metaflow/runner/metaflow_runner.py
@@ -8,7 +8,7 @@ from typing import Dict, Iterator, Optional, Tuple
 
 from metaflow import Run
 
-from .utils import handle_timeout
+from .utils import handle_timeout, async_handle_timeout
 from .subprocess_manager import CommandManager, SubprocessManager
 
 
@@ -282,6 +282,20 @@ class Runner(object):
         )
         return ExecutingRun(self, command_obj, run_object)
 
+    async def __async_get_executing_run(self, tfp_runner_attribute, command_obj):
+        content = await async_handle_timeout(
+            tfp_runner_attribute, command_obj, self.file_read_timeout
+        )
+        content = json.loads(content)
+        pathspec = "%s/%s" % (content.get("flow_name"), content.get("run_id"))
+
+        # Set the correct metadata from the runner_attribute file corresponding to this run.
+        metadata_for_flow = content.get("metadata")
+        metadata(metadata_for_flow)
+
+        run_object = Run(pathspec, _namespace_check=False)
+        return ExecutingRun(self, command_obj, run_object)
+
     def run(self, **kwargs) -> ExecutingRun:
         """
         Blocking execution of the run. This method will wait until
@@ -383,7 +397,9 @@ class Runner(object):
             )
             command_obj = self.spm.get(pid)
 
-            return self.__get_executing_run(tfp_runner_attribute, command_obj)
+            return await self.__async_get_executing_run(
+                tfp_runner_attribute, command_obj
+            )
 
     async def async_resume(self, **kwargs):
         """
@@ -418,7 +434,9 @@ class Runner(object):
             )
             command_obj = self.spm.get(pid)
 
-            return self.__get_executing_run(tfp_runner_attribute, command_obj)
+            return await self.__async_get_executing_run(
+                tfp_runner_attribute, command_obj
+            )
 
     def __exit__(self, exc_type, exc_value, traceback):
         self.spm.cleanup()
--- a/metaflow/runner/subprocess_manager.py
+++ b/metaflow/runner/subprocess_manager.py
@@ -29,6 +29,31 @@ def kill_process_and_descendants(pid, termination_timeout):
         pass
 
 
+async def async_send_signals(pids, signal):
+    pkill_processes = [
+        await asyncio.create_subprocess_exec("pkill", signal, "-P", str(pid))
+        for pid in pids
+    ]
+
+    for proc in pkill_processes:
+        await proc.wait()
+
+    kill_processes = [
+        await asyncio.create_subprocess_exec("kill", signal, str(pid)) for pid in pids
+    ]
+
+    for proc in kill_processes:
+        await proc.wait()
+
+
+async def async_kill_processes_and_descendants(pids, termination_timeout):
+    await async_send_signals(pids, "-TERM")
+
+    await asyncio.sleep(termination_timeout)
+
+    await async_send_signals(pids, "-KILL")
+
+
 class LogReadTimeoutError(Exception):
     """Exception raised when reading logs times out."""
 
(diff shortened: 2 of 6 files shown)


metaflow/runner/utils.py
```diff
@@ -74,6 +74,32 @@ def read_from_file_when_ready(
         return content
 
 
+async def async_read_from_file_when_ready(
+    file_path: str, command_obj: "CommandManager", timeout: float = 5
+):
+    await asyncio.wait_for(command_obj.process.wait(), timeout)
```
madhur-ob, line 80: Here, we wait for the process to finish with a timeout..

I am not sure we want that. The file content can be written earlier than the finish time of the process.

And we probably want the file content back as soon as possible..

The synchronous version doesn't wait for the process to finish before checking the file content. It keeps reading the file in a loop, and as soon as there is content, it returns it—even if the process is still running.

metaflow/runner/utils.py
```diff
@@ -74,6 +74,32 @@ def read_from_file_when_ready(
...
+def handle_process_error(command_obj: "CommandManager"):
+    stdout_log = open(command_obj.log_files["stdout"]).read()
+    stderr_log = open(command_obj.log_files["stderr"]).read()
+    command = " ".join(command_obj.command)
+    error_message = "Error executing: '%s':\n" % command
+    if stdout_log.strip():
+        error_message += "\nStdout:\n%s\n" % stdout_log
+    if stderr_log.strip():
+        error_message += "\nStderr:\n%s\n" % stderr_log
+    raise RuntimeError(error_message)
```
madhur-ob, line 100: you can probably return the `error_message` here...

and use that to do `raise RuntimeError(error_message) from e` in other places..

metaflow/runner/utils.py
```diff
@@ -107,12 +133,41 @@ def handle_timeout(
...
-        stderr_log = open(command_obj.log_files["stderr"]).read()
-        command = " ".join(command_obj.command)
-        error_message = "Error executing: '%s':\n" % command
-        if stdout_log.strip():
-            error_message += "\nStdout:\n%s\n" % stdout_log
-        if stderr_log.strip():
-            error_message += "\nStderr:\n%s\n" % stderr_log
-        raise RuntimeError(error_message) from e
+        handle_process_error(command_obj)
+        raise RuntimeError from e
```
madhur-ob, line 137: we are raising `RuntimeError` again here, it was already done in `handle_process_error(command_obj)`

Please see the suggestion to return the error message and use it at once here..

metaflow/runner/utils.py
```diff
@@ -107,12 +133,41 @@ def handle_timeout(
...
+        If a TimeoutError occurs, it raises a RuntimeError with the command's
+        stdout and stderr logs.
+    """
+    try:
+        return await async_read_from_file_when_ready(
+            tfp_runner_attribute.name, command_obj, timeout=file_read_timeout
+        )
+    except (CalledProcessError, TimeoutError) as e:
+        handle_process_error(command_obj)
+        raise RuntimeError from e
```
madhur-ob, line 173: same comment about raising `RuntimeError` again here which is already done inside `handle_process_error`

metaflow/runner/subprocess_manager.py
```diff
@@ -29,6 +29,31 @@ def kill_process_and_descendants(pid, termination_timeout):
...
+
+    kill_processes = [
+        await asyncio.create_subprocess_exec("kill", signal, str(pid)) for pid in pids
+    ]
+
+    for proc in kill_processes:
+        await proc.wait()
+
+
+async def async_kill_processes_and_descendants(pids, termination_timeout):
```
madhur-ob, line 49: can we simplify and kinda merge `async_send_signals` and `async_kill_processes_and_descendants`

also, the combined function could take one pid instead of multiple pids

and then we could do `async_kill` just like we have `kill` called on looping over these pids.

metaflow/runner/utils.py
filipcacky: Thanks, fixed.

metaflow/runner/utils.py
filipcacky: Fixed.

metaflow/runner/utils.py
filipcacky: You're right, this was a little silly of me. Fixed.

metaflow/runner/utils.py
filipcacky: Didn't realize the `poll` call isn't blocking. I rewrote it so that its the same as the sync version except a call to `asyncio.sleep`.

The impl doesn't really sit right with me though, I can imagine a scenario in which a call to `read` doesn't return the entire file content, because another write is issued by the subprocess immediately after. I think we could reimplement this with pipes and poll to see if the write side (subprocess) has closed the pipe and thus indicated it finished the writing. I could take a look at this after.

madhur-ob: So, we for know for a fact that the file being written will only be written once..

This makes things a bit easier..

madhur-ob: And we want to get the content ASAP since that allows us to fetch the `Run` object..
(while the process is still running..)

metaflow/runner/subprocess_manager.py
filipcacky: I think this approach is slightly better, instead of doing it one by one, we can kill all of them at once. Although the impl was a bit silly and wasteful (too many processes), I've simplified it by calling each kill/pkill with an unpacked list of pids.

I've done the same for the sync version, this should be way faster for multiple managed subprocesses. Instead of waiting `n * termination_timeout` we can do only `termination_timeout` with the same effect.

filipcacky: Actually, maybe scratch that. The sync `run_command` interface waits for the command to terminate before returning. Tbh we'd be fine by just calling kill on the current process, if we kept a reference to it somewhere. Looping over all of the commands is just going to end up in most of them failing, while still waiting 2 seconds for each.

metaflow/runner/utils.py
romain-intel: @filipcacky has a point here though -- it is true we write once and in *most* cases it should work fine but it is up to the OS to determine how the writes proceed and there is no guarantee that the content will be written all at once. Another mechanism may be warranted. File locks are always a bit annoying though (and not always reliable). We could also check for the presence of a particular string at the end or something (the OS should ensure writes are at least done sequentially). It may be worth fixing now because these types of bugs are particularly nasty when they do occur (it's rare but not impossible).

filipcacky: It's probably not worth doing in this PR, but when I have a bit more time I can take a look at it).

Reimplementing it for a peace of mind and or better ergonomics on the write side (not being forced to make only a single write) and faster read on the read side (reading until a pipe close, instead of waking up and checking for a non-null content) is I think worth it.

metaflow/runner/subprocess_manager.py
```diff
@@ -9,6 +9,8 @@
 import threading
 from typing import Callable, Dict, Iterator, List, Optional, Tuple
 
+from .utils import check_process_exited
+
 
 def kill_process_and_descendants(pid, termination_timeout):
```
madhur-ob, line 15: I guess we can remove this? If it is not being used, since you replaced it by a plural version..

Actually, it is used in the `kill()` function..
Maybe you can modify `kill` to use your plural version and then get rid of this one perhaps..

If you do it, we can copy over the TODO comments to the plural version..
also, you can use this `kill` in `_handle_sigint` and `_async_handle_sigint`

metaflow/runner/metaflow_runner.py
```diff
@@ -267,6 +267,14 @@ def __enter__(self) -> "Runner":
...
+    def __restore_env_and_metadata(self, content):
+        # Set the environment variables to what they were before the run executed.
+        clear_and_set_os_environ(self.old_env)
+
+        # Set the correct metadata from the runner_attribute file corresponding to this run.
+        metadata_for_flow = content.get("metadata")
+        metadata(metadata_for_flow)
+
     def __get_executing_run(self, tfp_runner_attribute, command_obj):
         # When two 'Runner' executions are done sequentially i.e. one after the other
```
madhur-ob, line 279: maybe you can copy these comments over to the `__restore_env_and_metadata` function..

metaflow/runner/utils.py
madhur-ob: I guess we can resolve this comment for now? But if possible..
@filipcacky can you take a stab at it within this PR itself?

metaflow/runner/subprocess_manager.py
madhur-ob: thanks, resolving..

madhur-ob:
I guess we are close...

@filipcacky Can you probably post an example / code snippet through which we can observe the behaviour before and after merging of this PR? aka what is the exact subtle nuanced behaviour you ran into that this PR solved for you..

I will make sure to verify from my side as well...

metaflow/runner/utils.py
```diff
@@ -74,6 +74,41 @@ def read_from_file_when_ready(
...
+                if content:
+                    break
+                raise CalledProcessError(
+                    command_obj.process.returncode, command_obj.command
+                )
+            if time.time() - start_time > timeout:
+                raise TimeoutError(
+                    "Timeout while waiting for file content from '%s'" % file_path
+                )
+            await asyncio.sleep(0.1)
```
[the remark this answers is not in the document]
filipcacky: Yes. Although now i've just offloaded the sync version to `asyncio.to_thread` to avoid blocking the main thread in filesystem IO.

metaflow/runner/utils.py
filipcacky: Alright, reimplemented it via a FIFO, made the same changes to Deployer.

metaflow/runner/subprocess_manager.py
filipcacky: Reimplemented kill in terms of the plural version to keep the API incase someone depends on it.

> also, you can use this kill in _handle_sigint and _async_handle_sigint
As in calling kill in a loop? I'd rather avoid that since we can just spawn one subprocess to send a signal to all managed processes instead of one for each.

metaflow/runner/utils.py
```diff
@@ -41,50 +44,197 @@ def clear_and_set_os_environ(env: Dict):
...
+    """
+    content = bytearray()
+
+    poll = select.poll()
+    poll.register(fifo_fd, select.POLLIN)
+
+    while True:
+        poll_begin = time.time()
+        poll.poll(timeout)
+        timeout -= 1000 * (time.time() - poll_begin)
```
filipcacky, line 118: `poll` on the fd first in order to wait for the subprocess to open and write. Calling a read before the pipe is opened results in an empty response on a non-blocking fifo.

metaflow/runner/utils.py
```diff
@@ -41,50 +44,197 @@ def clear_and_set_os_environ(env: Dict):
...
+    ------
+    TimeoutError
+        If no event occurs on the FIFO within the timeout.
+    CalledProcessError
+        If the process managed by `command_obj` has exited without writing any
+        content to the FIFO.
+    """
+    return await asyncio.to_thread(
+        read_from_fifo_when_ready, fifo_fd, command_obj, encoding, timeout
+    )
```
filipcacky, line 177: Moved FS IO to a separate thread in order to avoid blocking the main thread with the poll.

metaflow/runner/utils.py
```diff
@@ -41,50 +44,197 @@ def clear_and_set_os_environ(env: Dict):
...
+            raise TimeoutError("Timeout while waiting for the file content")
+
+        try:
+            data = os.read(fifo_fd, 128)
+            while data:
+                content += data
+                data = os.read(fifo_fd, 128)
+
+            # Read from a non-blocking closed FIFO returns an empty byte array
+            break
```
filipcacky, line 130: This code is either going to break when the FIFO write-side has been closed, indicating all bytes have been read. Or it will raise a `BlockingIOError` (EAGAIN) indicating the write-side still has the FIFO opened but hasn't written to it yet.

The `os.read` and using a raw file descriptor is deliberate, the `file.read` from the wrapper returned by `os.fdopen` or simple `open` behaves differently.

filipcacky:
I'll try to come up with a script to demonstrate the differences before and after these changes. The goal is to offload non-asyncio operations and lengthy filesystem I/O from the main thread in the async interface, so it can handle other tasks in the meantime, which may be a bit hard to demonstrate.

filipcacky:
# Scripts demonstrating the difference between old and new behaviour.

async_test_flow_runner.py
```py
import metaflow
import asyncio

async def run_signal_test():
    async with metaflow.Runner(flow_file="./async_test_flow.py") as runner:
        run1 = await runner.async_run()
        run2 = await runner.async_run()
        await run2.wait()
        await run1.wait()
        
if __name__ == "__main__":
    asyncio.run(run_signal_test())
```

async_test_flow.py
```py
import time


class AsyncTestFlow(FlowSpec):
    @step
    def start(self):
        time.sleep(30)
        self.next(self.end)

    @step
    def end(self):
        pass


if __name__ == "__main__":
    AsyncTestFlow()
```

async_test.py
```py
import metaflow
import asyncio
import signal
import time


async def run_signal_test():
    print("Running signal test")
    subprocess = await asyncio.create_subprocess_exec(
        "python", "async_test_flow_runner.py"
    )

    print("\tWaiting 5 seconds for flows to start")
    await asyncio.sleep(5)

    print("\tSending SIGINT to subprocess")
    begin = time.time()
    subprocess.send_signal(signal.SIGINT)
    await subprocess.wait()
    print(f"\tSubprocess exited in {time.time() - begin} seconds\n")


async def foo():
    """
    This function is used to simulate a long running task
    """

    print("\tRunning foo")
    for i in range(20):
        print(f"\tFoo {i}")
        await asyncio.sleep(0.1)


async def run_metaflow():
    async with metaflow.Runner(flow_file="./async_test_flow.py") as runner:
        print("\tRunning flow")
        run1 = await runner.async_run()
        print("\tWaiting for flow to finish")
        await run1.wait()


async def run_async_test():
    print("Running async test")
    t1 = asyncio.create_task(foo())
    t2 = asyncio.create_task(run_metaflow())
    await asyncio.gather(t1, t2)


if __name__ == "__main__":
    asyncio.run(run_signal_test())
    asyncio.run(run_async_test())
```

## Old behaviour tested at cbfaea9 runner fixes and enhancements (#2053)
```
Running signal test
	Waiting 5 seconds for flows to start
	Sending SIGINT to subprocess
	Subprocess exited in 4.178466796875 seconds

Running async test
	Running foo
	Foo 0
	Running flow
	Waiting for flow to finish
	Foo 1
	...
	Foo 19
```

## New behaviour
```
Running signal test
	Waiting 5 seconds for flows to start
	Sending SIGINT to subprocess
	Subprocess exited in 0.06499624252319336 seconds

Running async test
	Running foo
	Foo 0
	Running flow
	Foo 1
        ...
	Foo 10
	Waiting for flow to finish
	Foo 11
        ...
	Foo 19
```

With the new behaviour, other coroutines can run during `await runner.async_run()` - Foo 1 ... Foo 10.

The signal handler sends an INT to both `run1` and `run2`, causing `await run.wait()` to complete for both, and the program terminates immediately. If the subprocesses don't terminate right away, they will be forcefully killed after the `asyncio.sleep` timeout expires.

madhur-ob:
@filipcacky we have some more cleanup stuff here: https://github.com/Netflix/metaflow/pull/2104
maybe you can rebase on top of master to:
- include those changes 
- resolve conflicts that exist..

metaflow/runner/metaflow_runner.py
```diff
@@ -282,6 +283,20 @@ def __get_executing_run(self, tfp_runner_attribute, command_obj):
...
+    async def __async_get_executing_run(self, attribute_file_fd, command_obj):
+        content = await async_handle_timeout(
+            attribute_file_fd, command_obj, self.file_read_timeout
+        )
+        content = json.loads(content)
+        pathspec = "%s/%s" % (content.get("flow_name"), content.get("run_id"))
+
+        # Set the correct metadata from the runner_attribute file corresponding to this run.
+        metadata_for_flow = content.get("metadata")
+        metadata(metadata_for_flow)
```
madhur-ob, line 295: we are no longer setting the `metadata` explicitly.. i.e `metadata(metadata_for_flow)` needs to be removed..

refer to the implementation of `__get_executing_run` and you will see the following..

```py
# Set the correct metadata from the runner_attribute file corresponding to this run.
metadata_for_flow = content.get("metadata")

run_object = Run(
    pathspec, _namespace_check=False, _current_metadata=metadata_for_flow
)
```

metaflow/runner/subprocess_manager.py
```diff
@@ -9,26 +9,49 @@
...
+    await main_term.wait()
+
+    await asyncio.sleep(termination_timeout)
+
+    sub_kill = await asyncio.create_subprocess_exec("pkill", "-KILL", "-P", *pids)
+    await sub_kill.wait()
+
+    main_kill = await asyncio.create_subprocess_exec("kill", "-KILL", *pids)
+    await main_kill.wait()
+
```
madhur-ob, line 53: I think we can encapsulate the SIGTERM part and the SIGKILL part inside try/except each?
just like we did inside the sync version i.e. `kill_processes_and_descendants`

metaflow/runner/subprocess_manager.py
```diff
@@ -46,14 +69,28 @@ def __init__(self):
...
     def _handle_sigint(self, signum, frame):
-        for each_command in self.commands.values():
-            each_command.kill(termination_timeout=2)
+        pids = [
+            str(command.process.pid)
+            for command in self.commands.values()
+            if command.process and not check_process_exited(command)
+        ]
+        if pids:
+            kill_processes_and_descendants(pids, termination_timeout=2)
```
madhur-ob, line 93: small nitpick: earlier, `termination_timeout` was 2 seconds per PID
maybe we should pass `2*len(pids)` here? since this will handle all of them collectively..

madhur-ob: similarly for the async version, if we choose to do this..

madhur-ob (approved):
LGTM, one thing needs to be fixed i.e. https://github.com/Netflix/metaflow/pull/2056#discussion_r1858554612

Other comments are minor / nitpicks but will be good to consider.

metaflow/runner/metaflow_runner.py
filipcacky: Thanks, fixed! Must've missed it while rebasing.

metaflow/runner/subprocess_manager.py
filipcacky: Shouldn't be necessary anymore. All the pids are enumerated before the call and are sent out simultaneously by kill/pkill, so they all get the same amount of time as before.

madhur-ob (approved):

savingoyal (approved):

Issue #2060, opened 2024-09-25
Argo heartbeat-task-daemon using default KSA instead of METAFLOW_KUBERNETES_SERVICE_ACCOUNT


nathan9sanders:
I recently upgraded my `metaflow` version to latest (2.12.22) and ran a flow on Argo workflows (running in GKE). The flow failed due to this error in `heartbeat-task-daemon`:
```
google.api_core.exceptions.Forbidden: 403 GET ...
```

Failure happens when the container tries to download the code [here](https://github.com/Netflix/metaflow/blob/cbfaea9e469093f06eaeeada04919ca0f71170f5/metaflow/metaflow_environment.py#L118). I took a look at the pod using kubectl and saw this:
```
  serviceAccount: default
  serviceAccountName: default
```

The default KSA doesn't have any perms, hence why we get the error. All the other pods use the KSA defined in my config using METAFLOW_KUBERNETES_SERVICE_ACCOUNT, which is bound to an IAM account w/ the right perms using workload identity, so the other pods don't have any issues.

Is this a bug?

I see `serviceAccountName` set on generic Templates, but it is never set on [DaemonTemplate](https://github.com/Netflix/metaflow/blob/cbfaea9e469093f06eaeeada04919ca0f71170f5/metaflow/plugins/argo/argo_workflows.py#L2630), and I do not see it set on [WorkflowSpec](https://github.com/Netflix/metaflow/blob/cbfaea9e469093f06eaeeada04919ca0f71170f5/metaflow/plugins/argo/argo_workflows.py#L725) despite this property setter being declared [here](https://github.com/Netflix/metaflow/blob/cbfaea9e469093f06eaeeada04919ca0f71170f5/metaflow/plugins/argo/argo_workflows.py#L3213).

When I run `argo template get` to check the workflow template generated by metaflow, I can see serviceAccountName is set on most child templates, but not the heartbeat daemon template, and it is not set at the workflow spec level either. This seems to match what I see in the code I referenced above.

Seems like this should either be set on DaemonTemplate or on WorkflowSpec - I'm assuming that if it's set in the workflow spec, then the daemon template will use it if it's undefined on the daemon template. Let me know if you want me to make a PR for this

PR #2061, opened 2024-09-25
Add serviceAccountName to DaemonTemplate

nathan9sanders:
Fixes https://github.com/Netflix/metaflow/issues/2060

The diff is a bit ugly after running `black`; the real diff is just adding these:
```py
def service_account_name(self, service_account_name):
    self.payload["serviceAccountName"] = service_account_name
    return self

...

DaemonTemplate("heartbeat-daemon")
.service_account_name(resources["service_account"])

...
```

The alternative to this fix is to set `serviceAccountName` on WorkflowSpec. See this TODO [here](https://github.com/Netflix/metaflow/blob/cbfaea9e469093f06eaeeada04919ca0f71170f5/metaflow/plugins/argo/argo_workflows.py#L735). I chose to set this on DaemonTemplate instead to minimize the scope of this bugfix, but if adding to WorkflowSpec will be straightforward I can do that.

I've verified this by running
```
pip install ~/path/to/local/metaflow
python flow.py argo workflows create
argo template get myflow -o yaml
```
and now the heartbeat daemon has `serviceAccountName: metaflow-ksa` when it was missing before. Haven't done any testing other than this.

### DIFF
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -2627,50 +2627,54 @@ class ArgoWorkflows(object):
         )
         from kubernetes import client as kubernetes_sdk
 
-        return DaemonTemplate("heartbeat-daemon").container(
-            to_camelcase(
-                kubernetes_sdk.V1Container(
-                    name="main",
-                    # TODO: Make the image configurable
-                    image=resources["image"],
-                    command=cmds,
-                    env=[
-                        kubernetes_sdk.V1EnvVar(name=k, value=str(v))
-                        for k, v in env.items()
-                    ],
-                    env_from=[
-                        kubernetes_sdk.V1EnvFromSource(
-                            secret_ref=kubernetes_sdk.V1SecretEnvSource(
-                                name=str(k),
-                                # optional=True
+        return (
+            DaemonTemplate("heartbeat-daemon")
+            .service_account_name(resources["service_account"])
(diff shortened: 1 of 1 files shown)


savingoyal (approved):

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -2627,50 +2627,54 @@ def _heartbeat_daemon_template(self):
...
-                        for k, v in env.items()
-                    ],
-                    env_from=[
-                        kubernetes_sdk.V1EnvFromSource(
-                            secret_ref=kubernetes_sdk.V1SecretEnvSource(
-                                name=str(k),
-                                # optional=True
+        return (
+            DaemonTemplate("heartbeat-daemon")
+            .service_account_name(resources["service_account"])
```
savingoyal, line 2632: do we need the service account for the hooks too?

nathan9sanders: not sure, but they are also Templates so I am guessing the answer is yes. I can add this to the lifecycle hook templates if desired

saikonen: the existing hooks are only sending basic http requests with an argo built-in and not accessing any resources so hopefully they do not need the service account for now. we can revisit that in another PR if necessary.

saikonen (approved):

PR #2065, opened 2024-09-27
Refactor system event logger and monitor, add new metrics

talsperre:
This PR removes the notion of a separate global context from `system_monitor` and `system_logger`. User's can now implement their event loggers and monitors to make use of the current singleton instead. 

Additionally, it adds the `graph_info` object to the current singleton, but it is not directly exposed to the end user.

### DIFF
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -699,6 +699,15 @@ def resume(
     runtime.print_workflow_info()
 
     runtime.persist_constants()
+    _system_logger.log_event(
+        level="info",
+        module="metaflow.resume",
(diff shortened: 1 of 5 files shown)


romain-intel:
Something seems to have gotten lost -- there is no graph_info anymore.

romain-intel (dismissed):
LGTM

saikonen (approved):
other than my minor comment, looks good to go. Not sure how thorough my testing is but the refactor does not seem that major either.

saikonen (approved):

PR #2076, opened 2024-10-03
[wip] argo daemon improvements

saikonen:
testing out alternatives for argo daemon containers due to their shortcomings with resource management.

best alternative so far seems to be a service&statefulset approach, but this has major caveats

- security concerns with allowing an argo serviceaccount to create services in the namespace
- so far no robust cleanup method for deleting the services associated with a workflow, as exit handlers do not run when terminating a workflow.


saikonen:
closing as the service approach doesn't seem like a viable solution due to risk of dangling resources

PR #2080, opened 2024-10-05
speed up container bootstrap for @pypi/@conda

savingoyal:
roughly ~40% speed up on benchmark flows


metaflow/plugins/pypi/micromamba.py
```diff
@@ -173,10 +203,15 @@ def create(self, id_, packages, python, platform):
...
         self._call(cmd, env)
+        self.logger(
+            f"Created {platform} virtual environment {id_} in {time.time() - start_time:.2f}s!"
+        )
 
+    @functools.lru_cache(maxsize=None)
     def info(self):
         return self._call(["config", "list", "-a"])
 
+    @functools.lru_cache(maxsize=None)
```
savingoyal, line 214: to check if there is any unintended consequence of this

metaflow/plugins/pypi/bootstrap.py
```diff
@@ -64,77 +82,145 @@
...
+
+        cmd = f"""set -e;
+            export PATH=$PATH:$(pwd)/micromamba;
+            export CONDA_PKGS_DIRS=$(pwd)/micromamba/pkgs;
+            micromamba run --prefix {prefix} python -m pip --disable-pip-version-check install --root-user-action=ignore --no-compile --no-index --no-cache-dir --no-deps --prefer-binary --find-links={pypi_pkgs_dir} {pypi_pkgs_dir}/*.whl --no-user"""
+        run_cmd(cmd)
+
+    with concurrent.futures.ThreadPoolExecutor(max_workers=10) as executor:
+        # install micromamba, download conda and pypi packages in parallel
+        future_install_micromamba = executor.submit(install_micromamba, architecture)
```
savingoyal, line 198: check if download from datastore possible

metaflow/cli.py
```diff
@@ -960,7 +1027,7 @@ def start(
     ctx.obj.environment = [
         e for e in ENVIRONMENTS + [MetaflowEnvironment] if e.TYPE == environment
     ][0](ctx.obj.flow)
-    ctx.obj.environment.validate_environment(ctx.obj.logger, datastore)
+    ctx.obj.environment.validate_environment(echo, datastore)
```
savingoyal, line 1030: take care of symmetric changes for fast bakery

metaflow/plugins/pypi/conda_environment.py
```diff
@@ -164,25 +189,54 @@ def _path(url, local_path):
...
+                ]
+                for future in as_completed(solves):
+                    result = future.result()
+                    # sequential downloads
+                    self.solvers[solver].download(*result)
+                    # parallel creates
+                    executor.submit(self.solvers[solver].create, *result)
+                    if storage:
+                        # parallel cache
+                        executor.submit(cache, storage, [result], solver)
```
savingoyal, line 229: pypi execution can start a lot earlier

metaflow/cli.py
```diff
@@ -68,20 +72,81 @@ def echo_dev_null(*args, **kwargs):
...
+            line = INDENT + line
+
+    animation_kwargs = {
+        "fg": kwargs.get("fg"),
+        "bg": kwargs.get("bg"),
+        "bold": kwargs.get("bold", False),
+        "underline": kwargs.get("underline", False),
+    }
+
+    if animate:
```
savingoyal, line 121: check how animate and overwrite shows up in runtime logs and notebooks

metaflow/cli.py
```diff
@@ -68,20 +72,81 @@ def echo_dev_null(*args, **kwargs):
...
+    indent = kwargs.pop("indent", False)
+
+    if _animation_thread is not None:
+        _animation_stop.set()
+        _animation_thread.join()
+        _animation_thread = None
+        _animation_stop.clear()
+        click.echo("\r", nl=False, err=kwargs["err"])
+
+    if indent:
```
romain-intel, line 107: simplify: `if indent and not animate`

metaflow/cli.py
```diff
@@ -68,20 +72,81 @@ def echo_dev_null(*args, **kwargs):
...
+        if animate:
+            # For animated output, we prepend INDENT in the animation function
+            pass
+        else:
+            line = INDENT + line
+
+    animation_kwargs = {
+        "fg": kwargs.get("fg"),
```
romain-intel, line 115: fallbacks needed to be safe.

metaflow/cli.py
romain-intel: Yes please!

metaflow/cli.py
```diff
@@ -68,20 +72,81 @@ def echo_dev_null(*args, **kwargs):
...
+        "bg": kwargs.get("bg"),
+        "bold": kwargs.get("bold", False),
+        "underline": kwargs.get("underline", False),
+    }
+
+    if animate:
+        if animate is True:
+            get_frame = lambda line: f"{next(_default_spinner)} {line}"
```
romain-intel, line 123: f-string not supported in 3.5

metaflow/cli.py
```diff
@@ -68,20 +72,81 @@ def echo_dev_null(*args, **kwargs):
...
+        else:
+            get_frame = animate
```
romain-intel, line 125: confused here -- what is animate? It doesn't seem to be a simple bool here so a bit confused.

metaflow/cli.py
```diff
@@ -68,20 +72,81 @@ def echo_dev_null(*args, **kwargs):
     pass
 
 
+_animation_thread = None
+_animation_stop = threading.Event()
+_default_spinner = cycle(["-", "\\", "|", "/"])
+
+
+def _animate(get_frame, line, err, kwargs, indent):
+    while not _animation_stop.is_set():
```
romain-intel, line 81: this doesn't seem to respect the same arguments as the non animate case.

romain-intel (commented):
Didn't look at non CLI part.

One general comment as well is we have too many echo functions (I was bitten already). It would be nice to clean that up and make all of them take the same arguments.

I like the idea of adding a spinner like that though.

metaflow/cli.py
saikonen: notebooks seem broken still, animate produces new lines for each frame, even though the `\r` echoing by itself works outside of Metaflow runtime.

metaflow/plugins/pypi/conda_environment.py
```diff
@@ -164,25 +189,54 @@ def _path(url, local_path):
...
-            with ThreadPoolExecutor() as executor:
-                _ = list(
-                    executor.map(lambda x: self.solvers[solver].create(*x), results)
-                )
-            if self.datastore_type not in ["local"]:
-                # Cache packages only when a remote datastore is in play.
-                storage = self.datastore(_datastore_packageroot(self.datastore, echo))
-                cache(storage, results, solver)
-        self.logger("Virtual environment(s) bootstrapped!")
+                # parallel solves
```
saikonen, line 216: Can you also rename the PR as the scope of this is beyond simple log changes at this point.

metaflow/plugins/pypi/micromamba.py
saikonen: There is an implicit `None` returned from this call which will be cached, if the environment has not been formed yet. This might cause problems in some cases

metaflow/plugins/pypi/micromamba.py
```diff
@@ -91,32 +108,40 @@ def solve(self, id_, packages, python, platform):
...
+        # already cached. As a perf heuristic, we check if the environment already
+        # exists to short circuit package downloads.
+
+        # cheap check
         if os.path.exists(f"{prefix}/fake.done"):
             return
 
+        # somewhat expensive check
+        # TODO: make this less expensive
+        if self.path_to_environment(id_, platform):
```
saikonen, line 140: during forming a new environment, would this not cache the path to environment as `None` due to it not being ready yet?

metaflow/plugins/pypi/conda_environment.py
saikonen: need to be careful with the timing though, as a lot of pypi steps rely on
`self.micromamba.path_to_environment(id_)` which can get stuck to `None` if called too early.

Issue #1818, opened 2024-04-29
Argo-workflows with name > 52 characters fail if run as CronWorkflows


shrinandj:
If name > 52 characters, the workflow fails to start anyway. The workflow controller fails with:

```
time="2024-04-29T19:07:01.554Z" level=error msg="invalid cron workflow" cronWorkflow=jobs-default/nbcreditscorev4d5mx.prod.main.eventtrainnbcswithfixeddata error="cron workflow name \"abcreditscorev4d5mx.prod.main.eventtrainabcspqrfixederty\" must not be more than 52 characters long (currently 57)"
```
Would be easy if the user got this error (or similar) at the time of creating the workflow itself.

shrinandj:
The user experience is quite tricky here because the permissible name length for regular Argo workflows is greater than for Cron workflows.

So, 
- The user creates a flow and run `python flow.py argo-workflows create`. This creates the workflow template without any error
- The user is able to run this flow either via `python flow.py argo-wokflows trigger` or via the Argo UI and it works just fine.
- The user then adds `@schedule` decorator and run `python argo-workflows create`. This creates the workflow template and the CronWorkflow. No errors.
- But if the name > 52 characters, the actual flows don't run on the cron schedule. The only way to identify the problem is to look at the argo-workflows-controller logs as shown above and see the error.

What's needed here is the following.

When a user creates a flow with `@schedule`:
- Construct the name as it is done today. If the name length < 52 chars, create the template, cronworkflow.
- If the name length > 52 chars:
    - check if there is a previously created cronworkflow object. If so, delete it (since it wouldn't be running anyway).
    - create a name such that it is < 52 characters. e.g. take all characters > 46 and construct a 6 byte has to fit into 52 chars
- Create the new cronworkflow

/cc @savingoyal @saikonen

saikonen:
the name length limitation also affects launching flows via ArgoUI, and deploying flows with `@trigger` or `@trigger_on_finish` which try to create sensors, that will fail due to the naming.

PR #2082, opened 2024-10-06
address name contraction issues in argo-workflows create

savingoyal:
remaining items - 
1. enable `--name` for long names for CLI args except create
2. check if sanitize_for_argo needs fixing
3. moar testing
4. comms


savingoyal:
closes #1818 , #1608

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -290,8 +299,41 @@ def create(
...
+        if obj.workflow_name != obj._v1_workflow_name:
+            # Delete the old workflow if it exists
+            try:
+                ArgoWorkflows.delete(obj._v1_workflow_name)
+                obj.echo("Important!", bold=True, nl=False)
+                obj.echo(
+                    " To comply with new naming restrictions on Argo "
+                    "Workflows, this deployment replaced the\npreviously "
+                    "deployed workflow {v1_workflow_name}.\n".format(
+                        v1_workflow_name=obj._v1_workflow_name
```
saikonen, line 314: we could also check whether `v1_workflow_name` exists or not, and cut this alert out if  we did not replace an old deployment

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -385,48 +511,46 @@ def resolve_workflow_name(obj, name):
...
-        if len(workflow_name) > 253:
+        if len(workflow_name) > limit:
             name_hash = to_unicode(
                 base64.b32encode(sha1(to_bytes(workflow_name)).digest())
-            )[:8].lower()
-            workflow_name = "%s-%s" % (workflow_name[:242], name_hash)
-            obj._is_workflow_name_modified = True
+            )[:5].lower()
+            # TODO: We can create better names that try to preserve the flow name
+            workflow_name = "%s-%s" % (workflow_name[: limit - 6], name_hash)
```
saikonen, line 519: we could use `current.flow_name` as the base and append the hash which comes from the `current.project_flow_name` if the flow name is the most significant info in the template name

saikonen (commented):
an open question still remains how to handle the other argo CLI commands. f.ex. `argo-workflows delete` in the case where a deployment with an old name exists. do we try to delete both old&new every time?

metaflow/plugins/argo/argo_workflows_cli.py
frdeng: Hi, is this an issue?  In `validate_run_id`, it seems that it can extracts the `flow_name`, `project_name` and the `branch_name` properly and the validation would pass.

saikonen:
rebased.

tested that suspend/unsuspend/terminate work as expected even with both old and new deployments running side-by-side (which would only be the case if someone deploys and triggers an old version *after* a new one)

tested that old buggy sensors with too long names in them properly get replaced with the new deployment, and start working as intended.

saikonen:
regarding using `--name` as part of `create`, do we want to raise an exception if the name is too long? or should we instead apply the same truncation as with project based names? this will only affect the argo-workflows side of things, and not the actual flow names.

As I see it, there are two approaches to specifying long names via `--name`
1. truncate the deployment name same as with long project names. flows would deploy no matter what.
2. fail the `argo-workflows --name very-long-name create` and expect the user to adhere to argo-workflows naming standards

approach 1 has the UX trade-off that flows will deploy no matter the name length, but if the user is expecting their very-long-but-descriptive names to be present on the argo side, this will not be the case.

approach 2 is very disruptive, blocking the usage due to platform limitations that may not be of concern to the user.

decided to go with approach 1 for now. see https://github.com/Netflix/metaflow/pull/2082/commits/563a8fd6f881a476f95b296070fb5d16cde6786d for details

---

Edit:
after discussions, we decided that approach 2 is the preferred direction due to users most likely being intentional in their naming, and truncating this would unexpectedly end up losing information intended to be encoded in the deployment.

see https://github.com/Netflix/metaflow/pull/2082/commits/d06ca78cdac0e1f1fdba98f3dafe431d4bad2b04 for implementation.

saikonen:
regarding possible changes to `sanitize_for_argo` and its surjectivity in order to not have collisions between deployed flow names, this seems somewhat of a tricky issue.

1. if we stick to simply replacing offending characters with accepted ones, then the output of `sanitize_for_argo` will be an example of a distinct (in case any replacing was done) flow name that would map to the same output
2. appending a hash from the original name is also a possible solution. this would restrict the problem to hash collisions, but increasing the names length due to sanitizing can end up in the same name-length issues, so we would have to truncate it after the fact in some cases

metaflow/plugins/argo/argo_workflows_cli.py
saikonen: missing workflow delete raises an exception, so the messages wont get printed out. resolving this issue

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -548,18 +559,40 @@ def resolve_workflow_name(obj, name):
...
-                    current.branch_name[:11],
-                    current.flow_name,
-                )
+
+            # Generate a meaningful short name
+            project_name = project
+            branch_name = current.branch_name
+            flow_name = current.flow_name
+            parts = [project_name, branch_name, flow_name]
+            max_name_len = limit - 6
```
saikonen, line 571: why even less than the limit? shouldn't we have enough of an allowance with the 45 characters that we can use up all of them?

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -1102,9 +1342,24 @@ def validate_run_id(
...
+        if ArgoWorkflows.get_existing_deployment(obj._v1_workflow_name) is not None:
+            workflows.append(obj._v1_workflow_name)
+
+    return workflows
+
+
 def sanitize_for_argo(text):
     """
-    Sanitizes a string so it does not contain characters that are not permitted in Argo Workflow resource names.
+    Sanitizes a string so it does not contain characters that are not permitted in
```
saikonen, line 1369: no need for the less collision-prone v2 implementation anymore?

metaflow/plugins/argo/argo_workflows_cli.py
savingoyal: the 6 is for the hash

metaflow/plugins/argo/argo_workflows_cli.py
savingoyal: just to avoid the scenario where every single flow is impacted by this change

test/unit/test_argo_cli.py
saikonen, line 1: this can be removed as we didn't end up going with a v2 sanitization

saikonen (dismissed):

PR #2087, opened 2024-10-08
fix: task status issue with unpicklable artifacts

saikonen:
if a flow contains unpicklable artifacts, the task status for such a step is recorded as `successful` even though pickling fails and leads to a non-zero exit for the task.

example:
```python
from metaflow import FlowSpec, step, kubernetes

class DemoFlow(FlowSpec):

    @kubernetes
    @step
    def start(self):
        self.x = open("hello.txt", "w")
        self.next(self.end)

    @step
    def end(self):
        pass

if __name__ == "__main__":
    DemoFlow()
```

### DIFF
--- a/metaflow/task.py
+++ b/metaflow/task.py
@@ -716,6 +716,11 @@ class MetaflowTask(object):
                     self.flow._task_ok = False
                     raise ex
                 finally:
+                    # The attempt_ok metadata is used to determine task status so it is important
+                    # we ensure that it is written even in case of preceding failures.
+                    # f.ex. failing to serialize artifacts leads to a non-zero exit code for the process,
+                    # even if user code finishes successfully. Flow execution will not continue due to the exit,
+                    # so arguably we should mark the task as failed.
                     attempt_ok = str(bool(self.flow._task_ok))
                     self.metadata.register_metadata(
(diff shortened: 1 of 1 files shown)


metaflow/task.py
```diff
@@ -709,24 +709,29 @@ def run_step(
...
-                            type="internal_attempt_status",
-                            tags=["attempt_id:{0}".format(retry_count)],
-                        )
-                    ],
-                )
+                try:
+                    # persisting might fail due to unpicklable artifacts.
+                    output.persist(self.flow)
+                except Exception as ex:
+                    self.flow._task_ok = False
```
talsperre, line 716: Let's add `self.flow._exception = MetaflowExceptionWrapper(ex)` here as well. Maybe a print statement as well.

saikonen: what is the retained exception used for internally? I'm also not sure if adding a print will help in this case.

I'm re-raising the exception here, but I've noticed that the stderr at this stage is not captured by task logs anymore. Maybe that is by choice, but it leads to the PR being a halfway fix for the issue where a non-zero exit task is correctly marked as failed, but there are no logs to determine why it failed.

PR #2088, opened 2024-10-08
use cwd in nbrunner and nbdeployer

madhur-ob:


romain-intel:
Am fine with this but it does change the UX -- has this been validated?

madhur-ob:
the issue that people were running into -- other python files / directories are not importable if we use a temporary directory.
users had to pass the value of `base_dir` explicitly..

while we can still pass anything really, the `base_dir` will now be the current working directory..
i.e. the directory where the `.ipynb` exists..

any files / other dirs that are sisters i.e. at the same level of the notebook (the `.ipynb` file..)
will now be accessible and importable in the flow being written...

savingoyal:
@romain-intel i don't think there is any change in the UX perse - but this is a bug fix to ensure packaging works as expected.

romain-intel:
oh totally get the change and I agree with it -- I am just saying it is a change in UX exactly for the reason you mention: before we were running in an *isolated* clean environment and now it has a bunch more stuff. I agree that most people want the second behavior which this PR does so I am in favor of the change. I am simply pointing out that this is a behavior change (which I agree most people will view as a "bug fix" but it does change the documented behavior (clearly shown since the doc strings are changing too))

metaflow/runner/nbrun.py
```diff
@@ -43,7 +41,7 @@ class NBRunner(object):
         Additional environment variables to set for the Run. This overrides the
         environment set for this process.
     base_dir : Optional[str], default None
-        The directory to run the subprocess in; if not specified, a temporary
+        The directory to run the subprocess in; if not specified, the current
```
talsperre, line 44: current working directory.

metaflow/runner/nbdeploy.py
```diff
@@ -46,7 +44,7 @@ class NBDeployer(object):
         Additional environment variables to set. This overrides the
         environment set for this process.
     base_dir : Optional[str], default None
-        The directory to run the subprocess in; if not specified, a temporary
+        The directory to run the subprocess in; if not specified, the current
```
talsperre, line 47: current working directory.

savingoyal:
yep -  we are in agreement.

PR #2091, opened 2024-10-09
Don't raise error if metaflow_extension's `Distribution#files` is `None`.

trhodeos:
With the `metaflow-card-html` package, we're seeing the following error:

```
.../metaflow/extension_support/__init__.py", line 414, in _get_extension_packages
    for f in dist.files:
TypeError: 'NoneType' object is not iterable
```

Turns out [`Distribution.files` can return `None`](https://importlib-metadata.readthedocs.io/en/latest/api.html#importlib_metadata.Distribution.files). `metaflow-card-html` doesn't include the appropriate wheel metadata files, and, when included, hits this error.

### DIFF
--- a/metaflow/extension_support/__init__.py
+++ b/metaflow/extension_support/__init__.py
@@ -489,6 +489,8 @@ def _get_extension_packages(ignore_info_file=False, restrict_to_directories=None
             # contributes to. This is to enable multiple namespace packages to contribute
             # to the same extension point (for example, you may have multiple packages
             # that have plugins)
+            if not dist.files:
+                _ext_debug("Invalid package seen '%s' at '%s': no file list found in metadata" % (dist_name, dist_root))
             for f in dist.files or []:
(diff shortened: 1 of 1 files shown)


savingoyal (dismissed):

romain-intel:
The issue though is that this won't load the files and they won't be included when moving to remote execution so I need to find another way; ie: fallback on iterating through the package the way I do for non-distribution packages.

And of course, the package could be fixed too :)

trhodeos:
> The issue though is that this won't load the files and they won't be included when moving to remote execution so I need to find another way; ie: fallback on iterating through the package the way I do for non-distribution packages.
> 
> And of course, the package could be fixed too :)

Yeah, that makes sense. For now, I've added an `_ext_debug` so that metaflow doesn't crash and I've added a TODO.

I'll open an upstream patch for the extension so this metadata gets added as well!

PR #2097, opened 2024-10-13
[Ready for Review] Make TOGGLE_DECOSPECS only add default decorator once.

darinyu:
This is needed when we add value to DECOSPECS (or TOGGLE_DECOPSECS).

Changes:
1. update the METAFLOW_DECOSPECS (from config) to METAFLOW_DEFAULT_DECOSPECS. This will no longer collide with user override METAFLOW_DECOSPECS which translate to `--with`
2. remove `config_merge_cb` from `start` but move it to `run`. This is because "start" is called both in beginning of runtime and start of each task subprocess. Moving it to `run` will also attach decorator to each step.

Before the change:
when we use TOGGLE_DECOSPECS and run a flow, it first inject `X` as decorator at flow level (by this `config_merge_cb`. It then get passed to the subprocess when we run `step`. When `step` starts, it injects this `X` again, **resulting in double decorators**

After the change
when we use TOGGLE_DECOSPECS and run a flow, it no longer inject `X` (because this `config_merge_cb` is no longer in `start` in `cli.py`), When subprocess starts to run `step`, we inject this `X` to decorator (within step command line execution).

To test:

The following should include default card (**once**) in each step.
```
METAFLOW_DEFAULT_DECOSPECS=card python ./helloworld.py run
```

### DIFF
--- a/metaflow/cli_components/run_cmds.py
+++ b/metaflow/cli_components/run_cmds.py
@@ -82,12 +82,9 @@ def config_merge_cb(ctx, param, value):
     # NOTE: Assumes that ctx.auto_envvar_prefix is set to METAFLOW (same as in
     # from_conf)
 
-    # Special case where DEFAULT_DECOSPECS and value are the same. This happens
-    # when there is no --with option at the TL and DEFAULT_DECOSPECS is read from
-    # the env var. In this case, click also passes it as value
+    # Read decospecs options from the environment (METAFLOW_DEFAULT_DECOSPECS=...)
+    # and merge them with the one provided as --with.
     splits = DEFAULT_DECOSPECS.split()
-    if len(splits) == len(value) and all([a == b for (a, b) in zip(splits, value)]):
-        return value
     return tuple(list(value) + splits)
 
 
--- a/metaflow/cli_components/run_cmds.py
+++ b/metaflow/cli_components/run_cmds.py
@@ -88,7 +88,7 @@ def config_merge_cb(ctx, param, value):
     splits = DEFAULT_DECOSPECS.split()
     if len(splits) == len(value) and all([a == b for (a, b) in zip(splits, value)]):
         return value
-    return tuple(list(value) + DEFAULT_DECOSPECS.split())
+    return tuple(list(value) + splits)
 
 
 def common_run_options(func):
--- a/metaflow/cli_components/run_cmds.py
+++ b/metaflow/cli_components/run_cmds.py
@@ -8,6 +8,7 @@ from .. import decorators, namespace, parameters, tracing
 from ..exception import CommandException
 from ..graph import FlowGraph
 from ..metaflow_current import current
+from ..metaflow_config import DECOSPECS
 from ..package import MetaflowPackage
 from ..runtime import NativeRuntime
 from ..system import _system_logger
@@ -70,6 +71,26 @@ def write_file(file_path, content):
             f.write(str(content))
 
 
+def config_merge_cb(ctx, param, value):
+    # Callback to:
+    #  - read  the Click auto_envvar variable from both the
+    #    environment AND the configuration
+    #  - merge that value with the value passed in the command line (value)
+    #  - return the value as a tuple
+    # Note that this function gets called even if there is no option passed on the
+    # command line.
+    # NOTE: Assumes that ctx.auto_envvar_prefix is set to METAFLOW (same as in
+    # from_conf)
+
+    # Special case where DECOSPECS and value are the same. This happens
+    # when there is no --with option at the TL and DECOSPECS is read from
+    # the env var. In this case, click also passes it as value
+    splits = DECOSPECS.split()
+    if len(splits) == len(value) and all([a == b for (a, b) in zip(splits, value)]):
+        return value
+    return tuple(list(value) + DECOSPECS.split())
+
+
 def common_run_options(func):
     @click.option(
         "--tag",
@@ -109,6 +130,7 @@ def common_run_options(func):
         help="Add a decorator to all steps. You can specify this "
         "option multiple times to attach multiple decorators "
         "in steps.",
+        callback=config_merge_cb,
     )
     @click.option(
         "--run-id-file",
--- a/metaflow/cli_components/run_cmds.py
+++ b/metaflow/cli_components/run_cmds.py
@@ -8,7 +8,7 @@ from .. import decorators, namespace, parameters, tracing
 from ..exception import CommandException
 from ..graph import FlowGraph
 from ..metaflow_current import current
-from ..metaflow_config import DECOSPECS
+from ..metaflow_config import DEFAULT_DECOSPECS
 from ..package import MetaflowPackage
 from ..runtime import NativeRuntime
 from ..system import _system_logger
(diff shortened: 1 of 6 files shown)


metaflow/metaflow_config.py
```diff
@@ -285,7 +285,7 @@
 ###
 # Format is a space separated string of decospecs (what is passed
 # using --with)
-DECOSPECS = from_conf("DECOSPECS", "")
+DECOSPECS = from_conf("DEFAULT_DECOSPECS", "")
```
romain-intel, line 288: To be consistent, I would rename this: `DEFAULT_DECOSPECS = from_conf("DEFAULT_DECOSPECS", "")`

metaflow/plugins/cards/card_modules/basic.py
```diff
@@ -739,7 +739,9 @@ class BlankCard(MetaflowCard):
 
     type = "blank"
 
-    def __init__(self, options=dict(title=""), components=[], graph=None, **kwargs):
+    def __init__(
+        self, options=dict(title=""), components=[], graph=None, flow=None, **kwargs
```
romain-intel, line 743: unrelated change? I think it's valid though but jsut checking.

metaflow/cli_components/step_cmd.py
```diff
@@ -7,6 +7,7 @@
 from ..task import MetaflowTask
 from ..unbounded_foreach import UBF_CONTROL, UBF_TASK
 from ..util import decompress_list
+from ..metaflow_config import DECOSPECS
```
romain-intel, line 10: this would therefore be DEFAULT_DECOSPECS

metaflow/cli_components/step_cmd.py
```diff
@@ -127,6 +128,20 @@ def step(
         raise CommandException("Function *%s* is not a step." % step_name)
     echo("Executing a step, *%s*" % step_name, fg="magenta", bold=False)
 
+    decospecs = tuple(DECOSPECS.split())
+    if decospecs:
+        decorators._attach_decorators_to_step(func, decospecs)
```
romain-intel, line 133: I think this needs to be done earlier when we do `_attach_decorators` because here you may be missing out on some of the lifecycle hooks. The issue though is we still need to prevent them from loading the second time around. What I would suggest (and not 100% sure this would work) is that when we launch the subprocess (or any of the commands), we explicitly set `METAFLOW_DEFAULT_DECOSPECS` to an empty value since all decospecs will now be passed using `METAFLOW_DECOSPECS` or on the CL.

metaflow/decorators.py
```diff
@@ -477,15 +477,20 @@ def _base_step_decorator(decotype, *args, **kwargs):
...
+            if not decotype.allow_multiple:
+                # if `allow_multiple` is not `True` then only one
+                # decorator type is allowed per step
+                raise DuplicateStepDecoratorException(decotype.name, func)
+
+            # If multiple decorators of the same type are allowed, we
+            # then check if we see any duplicate attributes. If we do,
+            # we ignore the new decorator and continue.
+            for deco in func.decorators:
+                if deco.name == decotype.name and deco.attributes == kwargs:
```
romain-intel, line 490: I think this is where we can merge with defaults so that we can properly compare.

metaflow/metaflow_config.py
darinyu: updated

metaflow/plugins/cards/card_modules/basic.py
darinyu: its old, removed

metaflow/cli_components/step_cmd.py
darinyu: updated

metaflow/cli_components/step_cmd.py
darinyu: I just realized that we can rely on run_cmd because it also add decorator to downstream. can you help review updated PR?

metaflow/decorators.py
darinyu: weird enough, we dont reach this function now (maybe some changes to the flow or maybe I hallucinated,  I need to take a deeper look on what's going on with the check

metaflow/cli_components/run_cmds.py
```diff
@@ -70,6 +71,26 @@ def write_file(file_path, content):
...
+    # NOTE: Assumes that ctx.auto_envvar_prefix is set to METAFLOW (same as in
+    # from_conf)
+
+    # Special case where DEFAULT_DECOSPECS and value are the same. This happens
+    # when there is no --with option at the TL and DEFAULT_DECOSPECS is read from
+    # the env var. In this case, click also passes it as value
+    splits = DEFAULT_DECOSPECS.split()
+    if len(splits) == len(value) and all([a == b for (a, b) in zip(splits, value)]):
+        return value
+    return tuple(list(value) + DEFAULT_DECOSPECS.split())
```
npow, line 91: nit: you can reuse `splits` here

npow:
how did you test this PR?

metaflow/cli_components/run_cmds.py
darinyu: updated

darinyu:
> how did you test this PR?

added the test command in the description

npow:
Can we add an actual test for this?

darinyu:
> Can we add an actual test for this?
> 
> Edit: just realized this is in the OSS repo. Maybe add a UX test on our end?

yes, I made this change so that we can add default metrics card for runs with GPU, will update UX test for that

metaflow/cli_components/run_cmds.py
```diff
@@ -70,6 +71,26 @@ def write_file(file_path, content):
...
+    #    environment AND the configuration
+    #  - merge that value with the value passed in the command line (value)
+    #  - return the value as a tuple
+    # Note that this function gets called even if there is no option passed on the
+    # command line.
+    # NOTE: Assumes that ctx.auto_envvar_prefix is set to METAFLOW (same as in
+    # from_conf)
+
+    # Special case where DEFAULT_DECOSPECS and value are the same. This happens
+    # when there is no --with option at the TL and DEFAULT_DECOSPECS is read from
```
romain-intel, line 86: nit: it's not the TL but more at the run/resume level.

metaflow/cli_components/run_cmds.py
```diff
@@ -70,6 +71,26 @@ def write_file(file_path, content):
...
+    # the env var. In this case, click also passes it as value
+    splits = DEFAULT_DECOSPECS.split()
+    if len(splits) == len(value) and all([a == b for (a, b) in zip(splits, value)]):
```
romain-intel, line 89: Looking at this: https://click.palletsprojects.com/en/stable/advanced/#callback-evaluation-order, the callback is called even if there is no value passed in but I don't think it will have the value DEFAULT_DECOSPECS since now that is not the same name as before. I think we can get rid this line.

romain-intel (dismissed):
I think this is fine. I think we can further simplify though.

metaflow/cli_components/run_cmds.py
darinyu: updated

metaflow/cli_components/run_cmds.py
darinyu: updated, and also comments.

romain-intel (approved):

PR #2098, opened 2024-10-14
feature: enable extending pypi/conda execution decorator support

saikonen:
The problem is that as new execution environment decorators are being introduced, we need to constantly keep adding these to the list of supported decorators for `@pypi/@conda`.

Proposal is to instead have two flags on step decorators that keep track whether
- decorator is a conda supported execution environment
- what the target platform of said environment is

in the case of the target platform, a decorator could then implement this as a dynamic property if it can be inferred during env resolve time on a per step basis. (f.ex. for the kubernetes decorator through label selectors for `arch`)

### DIFF
--- a/metaflow/plugins/kubernetes/kubernetes_decorator.py
+++ b/metaflow/plugins/kubernetes/kubernetes_decorator.py
@@ -134,7 +134,9 @@ class KubernetesDecorator(StepDecorator):
     package_url = None
     package_sha = None
     run_time_limit = None
-    step_remote_execution = True
+
+    # Conda environment support
+    support_conda = True
     target_platform = "linux-64"
 
     def __init__(self, attributes=None, statically_defined=False):
--- a/metaflow/plugins/kubernetes/kubernetes_decorator.py
+++ b/metaflow/plugins/kubernetes/kubernetes_decorator.py
@@ -136,7 +136,7 @@ class KubernetesDecorator(StepDecorator):
     run_time_limit = None
 
     # Conda environment support
-    support_conda = True
+    supports_conda_environment = True
     target_platform = "linux-64"
 
     def __init__(self, attributes=None, statically_defined=False):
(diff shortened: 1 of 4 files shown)


metaflow/plugins/pypi/conda_environment.py
```diff
@@ -269,11 +269,9 @@ def get_environment(self, step):
         # Resolve `linux-64` Conda environments if @batch or @kubernetes are in play
         target_platform = conda_platform()
         for decorator in step.decorators:
-            # TODO: rather than relying on decorator names, rely on attributes
-            #       to make them extensible.
-            if decorator.name in ["batch", "kubernetes", "nvidia", "snowpark", "slurm"]:
+            if getattr(decorator, "support_conda", None):
```
savingoyal, line 272: for backward compatibility we can have this check as an `or` with the decorator names.

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -135,6 +135,10 @@ class KubernetesDecorator(StepDecorator):
     package_sha = None
     run_time_limit = None
 
+    # Conda environment support
+    support_conda = True
```
savingoyal, line 139: `supports_conda_environment`

savingoyal (commented):
lgtm! minor changes

metaflow/plugins/pypi/conda_environment.py
saikonen: added the fallback for the non-core decorators. or should batch/kubernetes also be on the list?

saikonen: added a note on backward compatibility, and kept the complete list of decorators. Motivation being that some forks/extensions might have their own implementation for batch/kubernetes decorators as well that we do not want to break.

PR #2105, opened 2024-10-17
fix: argo and step-function config param mapping

saikonen:
small change to fix argo/step function handling of config parameters


metaflow/plugins/aws/step_functions/step_functions.py
```diff
@@ -476,6 +476,8 @@ def _process_parameters(self):
         has_schedule = self._cron() is not None
         seen = set()
         for var, param in self.flow._get_parameters():
+            if param.IS_FLOW_PARAMETER:
```
savingoyal, line 479: @saikonen can you add a comment here for why we continue? would be helpful later on

romain-intel:
Included in #1962

PR #2106, opened 2024-10-17
Initial commit for logging all metaflow exceptions

talsperre:
We want to log all exceptions and the corresponding exception messages using the system monitor and loggers.


metaflow/cli.py
```diff
@@ -974,6 +974,13 @@ def start(
     ctx.obj.monitor.start()
     _system_monitor.init_system_monitor(ctx.obj.flow.name, ctx.obj.monitor)
 
+    current._update_env(
+        {
+            "system_logger": _system_logger,
```
savingoyal, line 979: can you help me with an example of how this is consumed? happy to chat on slack too.

PR #2107, opened 2024-10-18
remove setting of metadata in Runner

madhur-ob:


metaflow/client/core.py
```diff
@@ -269,13 +269,19 @@ def __init__(
...
         self._metaflow = Metaflow()
         self._parent = _parent
         self._path_components = None
         self._attempt = attempt
         self._current_namespace = _current_namespace or get_namespace()
+        self._current_metadata = _current_metadata or get_metadata()
         self._namespace_check = _namespace_check
+
+        # Set the metadata being passed..
+        metadata(self._current_metadata)
```
romain-intel, line 283: we don't want to do this at all. Basically, we create a Metaflow object that has self.metadata in it. We want to initialize self.metadata to be the one from self._current_metadata.

metaflow/client/core.py
```diff
@@ -540,16 +548,17 @@ def _unpickle_284(self, data):
         )
 
     def _unpickle_2124(self, data):
```
romain-intel, line 550: need to add another version of course. So we can unpickle all versions of objects.

metaflow/runner/metaflow_runner.py
```diff
@@ -284,11 +284,9 @@ def __get_executing_run(self, tfp_runner_attribute, command_obj):
...
-        metadata_for_flow = content.get("metadata")
-        metadata(metadata_for_flow)
-
-        run_object = Run(pathspec, _namespace_check=False)
+        run_object = Run(
+            pathspec, _namespace_check=False, _current_metadata=content.get("metadata")
+        )
         return ExecutingRun(self, command_obj, run_object)
 
     def run(self, **kwargs) -> ExecutingRun:
```
romain-intel, line 292: I think there is something in deployer which is where I ran into it.

romain-intel:
Superseded by #2113

Issue #2120, opened 2024-10-25
The card viewer server should load data from METAFLOW_CARD_LOCALROOT location


svpino:
We can configure the location of the metadata and cards using the METAFLOW_DATASTORE_SYSROOT_LOCAL and METAFLOW_CARD_LOCALROOT.

Unfortunately, the card viewer server doesn't use these environment variables to determine where the metadata is located. Instead, we have to run the command from the same folder where the metadata is.

Consider relying on the env variables so we don't have to change folders whenever we have to run the card server.

An alternative solution is to add a parameter to control the location of the folder. Something like this:

```
python pipeline.py card server --metadata-dir /.metaflow
```

savingoyal:
thanks for the bug report! we are looking into it

ch-enfuse:
Hi, currently having an issue with this, I saw the changes went in for the latest release (2.12.29) Im still not sure what i need to do to not have the cards create a .metaflow folder

PR #2121, opened 2024-10-28
use METAFLOW_CARD_LOCALROOT for card server

madhur-ob:
Fixes https://github.com/Netflix/metaflow/issues/2120

### DIFF
--- a/metaflow/plugins/cards/card_server.py
+++ b/metaflow/plugins/cards/card_server.py
@@ -45,23 +45,48 @@ class RunWatcher(Thread):
     def __init__(self, flow_name, connection: Connection):
         super().__init__()
 
+        self.daemon = True
+        self._connection = connection
+        self._flow_name = flow_name
+
+        self._watch_file = self._initialize_watch_file()
+        if self._watch_file is None:
+            _ClickLogger(
+                "Warning: Could not initialize watch file location.", fg="yellow"
+            )
+
+        self._current_run_id = self.get_run_id()
+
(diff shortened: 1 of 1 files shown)


metaflow/plugins/cards/card_server.py
```diff
@@ -50,15 +45,20 @@ class RunWatcher(Thread):
...
 
-        self._watch_file = os.path.join(
-            os.getcwd(), DATASTORE_LOCAL_DIR, flow_name, "latest_run"
+        local_root = LocalStorage.datastore_root
+        if local_root is None:
+            local_root = LocalStorage.get_datastore_root_from_config(
+                lambda _: None, create_on_absent=False
+            )
+        self._watch_file = (
+            os.path.join(local_root, flow_name, "latest_run") if local_root else None
```
valayDave, line 54: LGTM but just one quick question : Under what circumstances is the `self._watch_file` set to `None` ie when does `local_root` become none. 

Few things we can do : 

1. We can log to the user that the `local_root` was not found (logging in the init function). 
2. On every call to the `get_run_id` function we can also check if `self._watch_file == None` . One reason for doing this is because users can start the server before they launch the first run of a flow (in which case the file will not be present and `LocalStorage.get_datastore_root_from_config` might return None) . If it's None then we re-run the `local_root` extraction and set the `self._watch_file`.

madhur-ob: Thanks, done all the changes..

valayDave (approved):

PR #2122, opened 2024-10-28
feature: Introduce secrets function

saikonen:
introduces a `get_secrets` function that exposes the same capabilities as the `@secrets` decorator, but returns them as a dictionary instead of setting environment variables.

The use case is to make accessing secrets possible in user-code as well.

example:

```python
@step
def start(self):
    secrets_dict = get_secrets(sources=["secret_source"], role="IAM_role_for_accessing")
    
    for secret_spec, contents_dict in secrets_dict.items():
         pass # do something with the fetched secrets
```


romain-intel:
Need to see why this is not testing -- there is a load problem. May be an ordering issue. I will check (we now use secrets too). Thanks for trying to run the test. I'll take a look.

PR #2124, opened 2024-10-31
Add static and runtime dag info, API to fetch ancestor and successor tasks

talsperre:
Add runtime DAG info so that we can query the ancestor and successor tasks for a given task easily. 

#### Usage
```
from metaflow import Task, namespace
namespace(None)
task = Task('RuntimeDAGFlow/18/step_c/32076012', attempt=0)
```

To get ancestors, progenies, and siblings, use the following API:
```
ancestors = task.ancestors
successors = task.successors
```

The output would be a list of metaflow `Task` objects.


metaflow/metadata/metadata.py
```diff
@@ -672,6 +672,22 @@ def _register_system_metadata(self, run_id, step_name, task_id, attempt):
...
+    @classmethod
+    def filter_tasks_by_metadata(
+        cls, flow_id, run_id, query_step, field_name, field_value
+    ):
+        # TODO: Do we need to do anything wrt to task attempt?
+        task_ids = cls._filter_tasks_by_metadata(
+            flow_id, run_id, query_step, field_name, field_value
+        )
+        return task_ids
+
```
saikonen, line 690: is there a need for the private method, or could this simply be contained in the public-facing one? right now its not doing anything before calling the private one.

saikonen: also, did you have an implementation of this for `service.py` yet?

metaflow/metadata/metadata.py
```diff
@@ -672,6 +672,22 @@ def _register_system_metadata(self, run_id, step_name, task_id, attempt):
...
+    def _filter_tasks_by_metadata(
+        cls, flow_id, run_id, query_step, field_name, field_value
+    ):
+        raise NotImplementedError()
+
+    @classmethod
+    def filter_tasks_by_metadata(
+        cls, flow_id, run_id, query_step, field_name, field_value
+    ):
+        # TODO: Do we need to do anything wrt to task attempt?
```
saikonen, line 685: probably not, as the ancestors for task attempts should be identical, right? What about the `immediate_siblings` though, will they include or exclude attempts of the same task?

metaflow/client/core.py
```diff
@@ -1123,6 +1124,202 @@ def _iter_filter(self, x):
         # exclude private data artifacts
         return x.id[0] != "_"
 
+    def _get_task_for_queried_step(self, flow_id, run_id, query_step):
```
romain-intel, line 1127: Could this function simply be replaced by `Step(flow_id/run_id/step_name).task`? It doesn't seem to bring much more to it.

metaflow/client/core.py
```diff
@@ -1123,6 +1124,202 @@ def _iter_filter(self, x):
...
+            # the same step.
+            query_task = self._get_task_for_queried_step(flow_id, run_id, steps[0])
+            query_foreach_stack_len = len(
+                query_task.metadata_dict.get("foreach-indices", [])
+            )
+
+        if query_foreach_stack_len == cur_foreach_stack_len:
+            # The successor or ancestor tasks belong to the same foreach stack level
+            field_name = "foreach-indices"
+            field_value = self.metadata_dict.get(field_name)
```
romain-intel, line 1187: We don't currently cache metadata_dict so either we could fix that or cache it here to avoid making multiple calls to the metadata service and then sorts. It would need to be cached across _get_related_tasks and this function.

metaflow/client/core.py
```diff
@@ -1123,6 +1124,202 @@ def _iter_filter(self, x):
...
+                # We will compare the foreach-indices-truncated value of ancestor task with the
+                # foreach-indices value of current task
+                field_name = "foreach-indices-truncated"
+                field_value = self.metadata_dict.get("foreach-indices")
+            else:
+                # This is a foreach split
+                # Current Task: foreach-indices = [0, 1, 2], foreach-indices-truncated = [0, 1]
+                # Ancestor Task: foreach-indices = [0, 1], foreach-indices-truncated = [0]
+                # We will compare the foreach-indices value of ancestor task with the
+                # foreach-indices value of current task
```
romain-intel, line 1202: nit: foreach-indices-truncated value of the current task

metaflow/client/core.py
```diff
@@ -1123,6 +1124,202 @@ def _iter_filter(self, x):
...
+                # Current Task: foreach-indices = [0, 1, 2], foreach-indices-truncated = [0, 1]
+                # Successor Task: foreach-indices = [0, 1], foreach-indices-truncated = [0]
+                # We will compare the foreach-indices-truncated value of current task with the
+                # foreach-indices value of successor tasks
+                field_name = "foreach-indices"
+                field_value = self.metadata_dict.get("foreach-indices-truncated")
+        return field_name, field_value
+
+    def _get_related_tasks(self, relation_type: str) -> Dict[str, List[str]]:
+        start_time = time.time()
```
romain-intel, line 1225: not used -- can strip or use.

metaflow/client/core.py
```diff
@@ -1123,6 +1124,202 @@ def _iter_filter(self, x):
...
+
+        field_name, field_value = self._get_metadata_query_vals(
+            flow_id,
+            run_id,
+            len(self.metadata_dict.get("foreach-indices", [])),
+            steps,
+            relation_type,
+        )
+
+        cur_time = time.time()
```
romain-intel, line 1244: not used -- can strip or use.

metaflow/client/core.py
```diff
@@ -1123,6 +1124,202 @@ def _iter_filter(self, x):
...
+            Dictionary of immediate successors of this task. The keys are the
+            names of the successor steps and the values are the corresponding
+            task pathspecs of the successors.
+        """
+        return self._get_related_tasks("successor")
+
+    @property
+    def immediate_siblings(self) -> Dict[str, List[str]]:
+        """
+        Returns a dictionary of closest sibling task pathspecs of this task for each
```
romain-intel, line 1289: closest sibling is confusing: I think you just mean siblings as in: tasks with the same parent.

metaflow/client/core.py
```diff
@@ -1123,6 +1124,202 @@ def _iter_filter(self, x):
...
+        Dict[str, List[str]]
+            Dictionary of closest siblings of this task. The keys are the
+            names of the current step and the values are the corresponding
+            task pathspecs of the siblings.
+        """
+        flow_id, run_id, step_name, _ = self.path_components
+
+        foreach_stack = self.metadata_dict.get("foreach-stack", [])
+        foreach_step_names = self.metadata_dict.get("foreach-step-names", [])
+        if len(foreach_stack) == 0:
```
romain-intel, line 1303: For a user exposed function called immediate_siblings, we should handle the static case too ideally (or just change the name to make it clear it's just for foreach). For example, if I have A -> B/C/D -> E, I would expect that in B, I could get the siblings C and D. Similarly, even if I am not a new foreach split, something like A -> foreach B -> C/D/E -> BJoin -> AJoin, from C, I should get D and E (the right ones). I think this function is excluding those right?

metaflow/metadata_provider/metadata.py
```diff
@@ -672,6 +673,39 @@ def _register_system_metadata(self, run_id, step_name, task_id, attempt):
...
+        query_step: str
+            Step name to query tasks from
+        field_name: str
+            Metadata field name to query
+        field_value: str
+            Metadata field value to query
+
+        Returns
+        -------
+        List[str]
```
romain-intel, line 704: Is this always task_ids? Since we are making this visible to the user, it should be task_name if the user visible task identifier is the name. Basically if the task is known as 123 to the user, it should return that but if it is known as task-123, it should return that.

metaflow/metaflow_config.py
```diff
@@ -248,8 +248,7 @@
 # Default container registry
 DEFAULT_CONTAINER_REGISTRY = from_conf("DEFAULT_CONTAINER_REGISTRY")
 # Controls whether to include foreach stack information in metadata.
-# TODO(Darin, 05/01/24): Remove this flag once we are confident with this feature.
-INCLUDE_FOREACH_STACK = from_conf("INCLUDE_FOREACH_STACK", False)
+INCLUDE_FOREACH_STACK = from_conf("INCLUDE_FOREACH_STACK", True)
```
romain-intel, line 251: We should probably change this at some point and remove it to not make it optional anymore.

metaflow/plugins/metadata_providers/local.py
```diff
@@ -202,6 +203,110 @@ def _optimistically_mutate():
...
+                "step", "task", {}, None, flow_id, run_id, query_step
+            )
+
+            filtered_task_ids = []
+            field_name_prefix = f"sysmeta_{field_name}"
+
+            # Filter tasks based on metadata
+            for task in tasks:
+                task_id = task.get("task_id")
+                if not task_id:
```
romain-intel, line 290: when does this happen? Also, task_id of zero is valid iirc.

metaflow/plugins/metadata_providers/local.py
```diff
@@ -202,6 +203,110 @@ def _optimistically_mutate():
...
+            If metadata file is not found
+        """
+
+        def _get_latest_metadata_file(path: str, field_prefix: str) -> tuple:
+            """Find the most recent metadata file for the given field prefix."""
+            # The metadata is saved as files with the format: sysmeta_<field_name>_<timestamp>.json
+            # and the artifact files are saved as: <attempt>_artifact__<artifact_name>.json
+            # We loop over all the JSON files in the directory and find the latest one
+            # that matches the field prefix.
+            json_files = glob.glob(os.path.join(path, "*.json"))
```
romain-intel, line 250: we should be able to do a more efficient globbing so we don't have to filter by field_prefix later on. SOmething like `f"{field_prefix}*.json"`.

metaflow/task.py
```diff
@@ -493,6 +511,36 @@ def run_step(
...
+                        type="foreach-indices",
+                        tags=metadata_tags,
+                    ),
+                    MetaDatum(
+                        field="foreach-indices-truncated",
+                        value=foreach_indices_truncated,
+                        type="foreach-indices-truncated",
+                        tags=metadata_tags,
+                    ),
+                    MetaDatum(
```
romain-intel, line 535: I believe this is only used in the siblings thing. If that's the case, we may be able to get rid of this when we refactor the siblings thing (if we do that). I am also a little confused as to why this is needed.

metaflow/task.py
```diff
@@ -727,8 +778,20 @@ def run_step(
                                 field="attempt_ok",
                                 value=attempt_ok,
                                 type="internal_attempt_status",
-                                tags=["attempt_id:{0}".format(retry_count)],
-                            )
+                                tags=metadata_tags,
+                            ),
+                            MetaDatum(
+                                field="previous_steps",
```
romain-intel, line 784: nit: consistency here between previous_steps and foreach-indices for example.

romain-intel (commented):
A few comments. I think it's pretty close though. I haven't looked at hte metadata service changes. We may also want to raise a better error message if the service is not new enough?

metaflow/plugins/metadata_providers/service.py
```diff
@@ -304,6 +305,46 @@ def _new_task(
...
+            List of task_ids that satisfy the query
+        """
+        query_params = {
+            "field_name": field_name,
+            "field_value": field_value,
+            "query_step": query_step,
+        }
+        url = ServiceMetadataProvider._obj_path(flow_id, run_id, query_step)
+        url = f"{url}/tasks?{urlencode(query_params)}"
+        return cls._request(cls._monitor, url, "GET")
```
saikonen, line 346: getting an error with this that `cls` does not have `_monitor`. All other calls to `_request` pass in `None`

metaflow/plugins/metadata_providers/service.py
```diff
@@ -304,6 +305,46 @@ def _new_task(
...
+        List[str]
+            List of task_ids that satisfy the query
+        """
+        query_params = {
+            "field_name": field_name,
+            "field_value": field_value,
+            "query_step": query_step,
+        }
+        url = ServiceMetadataProvider._obj_path(flow_id, run_id, query_step)
+        url = f"{url}/tasks?{urlencode(query_params)}"
```
saikonen, line 345: missing import for urlencode. f-strings are probably fine by 2025, as we've gotten rid of the older tests that break with them.

romain-intel: Ya, I was wondering about fstrings but did check and our official minimum version is 3.6 which supports it -- and yes, let's move at least a tad into the future :). I'm going to start using them too and the code will slowly migrate to it (and become infinitesimally faster :) )

metaflow/plugins/metadata_providers/service.py
talsperre: Yes, I will pass in `None` simply.

metaflow/task.py
talsperre: Will refactor siblings function mostly to return siblings irrespective of whether it is in a for each or not.

metaflow/client/core.py
talsperre: Wasn't aware that this returns the first available task.

metaflow/client/core.py
```diff
@@ -1211,9 +1419,12 @@ def metadata_dict(self) -> Dict[str, str]:
...
         # use the newest version of each key, hence sorting
-        return {
-            m.name: m.value for m in sorted(self.metadata, key=lambda m: m.created_at)
-        }
+        if self._metadata_dict is None:
+            self._metadata_dict = {
+                m.name: m.value
+                for m in sorted(self.metadata, key=lambda m: m.created_at)
+            }
+        return self._metadata_dict
```
romain-intel, line 1427: Note: this slightly changes the syntax since now if there is new metadata, the user won't get it. Should check if this impacts other operations. Or scope the caching to just the functions that need it.

metaflow/client/core.py
```diff
@@ -1118,11 +1119,218 @@ class Task(MetaflowObject):
...
+                # Successor Task: foreach-indices = [0, 1], foreach-indices-truncated = [0]
+                # We will compare the foreach-indices-truncated value of current task with the
+                # foreach-indices value of successor tasks
+                field_name = "foreach-indices"
+                field_value = self.metadata_dict.get("foreach-indices-truncated")
+        return field_name, field_value
+
+    def _get_related_tasks(self, is_ancestor: bool) -> Dict[str, List[str]]:
+        flow_id, run_id, _, _ = self.path_components
+        steps = (
```
saikonen, line 1223: huge problem here which leads to the queries not working correctly:
the steps ends up being of type `str` on OSS metadata-service, so using these you end up iterating over characters instead of step names, e.g.:

```
/flows/SplitFlow/runs/63/steps/{/filtered_tasks?metadata_field_name=foreach-indices&metadata_field_value=%7B%7D&query_step=%7B
/flows/SplitFlow/runs/63/steps/e/filtered_tasks?metadata_field_name=foreach-indices&metadata_field_value=%7B%7D&query_step=e
/flows/SplitFlow/runs/63/steps/n/filtered_tasks?metadata_field_name=foreach-indices&metadata_field_value=%7B%7D&query_step=n
/flows/SplitFlow/runs/63/steps/d/filtered_tasks?metadata_field_name=foreach-indices&metadata_field_value=%7B%7D&query_step=d
/flows/SplitFlow/runs/63/steps/}/filtered_tasks?metadata_field_name=foreach-indices&metadata_field_value=%7B%7D&query_step=%7D

```

saikonen (changes requested):

metaflow/client/core.py
```diff
@@ -1118,11 +1119,227 @@ class Task(MetaflowObject):
...
+                f"{flow_id}/{run_id}/{step}/{task_id}"
+                for task_id in self._metaflow.metadata.filter_tasks_by_metadata(
+                    flow_id, run_id, step, field_name, field_value
+                )
+            ]
+            for step in steps
+        }
+
+    @property
+    def immediate_ancestors(self) -> Dict[str, List[str]]:
```
savingoyal, line 1260: can we offer a property `parents` which simply returns a list of task pathspecs. it will return `None`for the `start` step. an open question is if we would want to also offer `parent` - maybe we can cross that bridge as a follow up PR if needed.

metaflow/client/core.py
```diff
@@ -1118,11 +1119,227 @@ class Task(MetaflowObject):
...
+        -------
+        Dict[str, List[str]]
+            Dictionary of immediate ancestors of this task. The keys are the
+            names of the ancestor steps and the values are the corresponding
+            task pathspecs of the ancestors.
+        """
+        return self._get_related_tasks(is_ancestor=True)
+
+    @property
+    def immediate_successors(self) -> Dict[str, List[str]]:
```
savingoyal, line 1275: can we offer a property `children` which simply returns a list of task pathspecs. it will return `None` for the `end` step. similar comment for `child` as for `parent`

metaflow/client/core.py
savingoyal: what are the consistency guarantees offered by this property. do we expect that this property will return an immutable set of parents as soon as this task is registered?

metaflow/client/core.py
savingoyal: what is the expected behavior of this property? as soon as children task are registered, do we start updating the `children` property? also, when do we know that there are not going to be any more children tasks?

metaflow/client/core.py
```diff
@@ -1118,11 +1119,227 @@ class Task(MetaflowObject):
...
+        -------
+        Dict[str, List[str]]
+            Dictionary of immediate successors of this task. The keys are the
+            names of the successor steps and the values are the corresponding
+            task pathspecs of the successors.
+        """
+        return self._get_related_tasks(is_ancestor=False)
+
+    @property
+    def siblings(self) -> Dict[str, List[str]]:
```
savingoyal, line 1290: can this be `siblings(self, ancestors)` - where `ancestors` is a list of task pathspec and defaults to `task.parents`?

metaflow/client/core.py
savingoyal: also - we are assuming that return values here point to the latest successful attempt. might be good to note this in the doc string.

metaflow/task.py
```diff
@@ -493,6 +512,36 @@ def run_step(
                     )
                 )
 
+            # Add runtime dag info - for a nested foreach this may look like:
```
savingoyal, line 515: in terms of metadata that needs to be stored - we could simply stringify the stack and store it - `step1:0,step2:1...` - and that should be it, no? a prefix query should be fast enough

metaflow/client/core.py
```diff
@@ -1118,11 +1120,231 @@ class Task(MetaflowObject):
...
+
+        Returns
+        -------
+        List[str]
+            List of successor step names
+        """
+        _, _, step_name, _ = self.path_components
+        return self._graph_info[step_name]["next"]
+
+    def _get_metadata_query_vals(
```
savingoyal, line 1186:
```   
  def _iter_matching_tasks(self, steps, pattern):
        """
        Yield tasks from specified steps matching a foreach path pattern.

        Parameters
        ----------
        steps : List[str]
            List of step names to search for tasks
        pattern : str
            Regex pattern to match foreach-indices metadata

        Returns
        -------
        Iterator[Task]
            Tasks matching the foreach path pattern
        """
        flow_id, run_id, _, _ = self.path_components
        
        for step in steps:
            task_ids = self._metaflow.metadata.filter_tasks_by_metadata(
                flow_id, run_id, step, "foreach-indices", pattern
            )
            for task_id in task_ids:
                yield Task(
                    pathspec="%s/%s/%s/%s" % (flow_id, run_id, step, task_id),
                    _namespace_check=False
                )
```

savingoyal (changes requested):

metaflow/client/core.py
romain-intel: Why couldn't the filter_task_by_metadata return all the info we need to form the task? It seems it would save a few RT calls right?

romain-intel (commented):
not a review -- just some comments.

metaflow/client/core.py
```diff
@@ -1123,6 +1123,139 @@ def _iter_filter(self, x):
...
+            # Static join - use exact path matching
+            pattern = current_path or ".*"
+            yield from self._iter_matching_tasks(steps, pattern)
+            return
+
+        # Handle single step case
+        target_task = Step(
+            f"{flow_id}/{run_id}/{steps[0]}", _namespace_check=False
+        ).task
+        target_path = target_task.metadata_dict.get("foreach-execution-path", "")
```
savingoyal, line 1179: nit - just `target_task.metadata_dict.get("foreach-execution-path")`

metaflow/client/core.py
```diff
@@ -1123,6 +1123,139 @@ def _iter_filter(self, x):
...
+        -------
+        List["Task"]
+            List of all parent tasks of the current task.
+        """
+        flow_id, run_id, _, _ = self.path_components
+        steps = self.parent.parent_steps
+        if not steps:
+            return
+
+        current_path = self.metadata_dict.get("foreach-execution-path", "")
```
savingoyal, line 1167: nit - just `self.metadata_dict.get("foreach-execution-path")` - just to guard against an eventuality where foreach-execution-path set to empty starts having a meaning.

metaflow/client/core.py
```diff
@@ -1123,6 +1123,139 @@ def _iter_filter(self, x):
...
+
+        for step in steps:
+            task_pathspecs = self._metaflow.metadata.filter_tasks_by_metadata(
+                flow_id, run_id, step, "foreach-execution-path", pattern
+            )
+            for task_pathspec in task_pathspecs:
+                yield Task(pathspec=task_pathspec, _namespace_check=False)
+
+    @property
+    def parent_tasks(self) -> List["Task"]:
```
savingoyal, line 1152: would be good to handle the case where the user is inspecting a task that ran using an old version of metaflow or is using an old version of the service...

metaflow/client/core.py
```diff
@@ -1123,6 +1123,139 @@ def _iter_filter(self, x):
         # exclude private data artifacts
         return x.id[0] != "_"
 
+    def _iter_matching_tasks(self, steps, pattern):
```
savingoyal, line 1126: minor nit - `_iter_matching_tasks(self, steps, metadata_key, metadata_pattern)`

metaflow/client/core.py
```diff
@@ -1123,6 +1123,139 @@ def _iter_filter(self, x):
...
+                flow_id, run_id, step, "foreach-execution-path", pattern
+            )
+            for task_pathspec in task_pathspecs:
+                yield Task(pathspec=task_pathspec, _namespace_check=False)
+
+    @property
+    def parent_tasks(self) -> List["Task"]:
+        """
+        Returns a list of all parent tasks of the current task for the latest successful
+        attempt.
```
savingoyal, line 1155: parents should be the same across attempts?

metaflow/client/core.py
```diff
@@ -1123,6 +1123,139 @@ def _iter_filter(self, x):
...
+                # Pattern: "A:10,B:13"
+                pattern = ",".join(current_path.split(",")[:target_depth])
+
+        yield from self._iter_matching_tasks(steps, pattern)
+
+    @property
+    def child_tasks(self) -> List["Task"]:
+        """
+        Returns a list of all child/children tasks of the current task for the latest successful
+        attempt.
```
savingoyal, line 1209: nit - `if one exists`

metaflow/client/core.py
```diff
@@ -1837,6 +1970,38 @@ def environment_info(self) -> Optional[Dict[str, Any]]:
...
+        """
+        Returns a list of parent step names for the current step.
+
+        Returns
+        -------
+        List[str]
+            List of parent step names
+        """
+        if self.id == "start":
+            return []
```
savingoyal, line 1984: why not an empty return?

metaflow/client/core.py
```diff
@@ -1837,6 +1970,38 @@ def environment_info(self) -> Optional[Dict[str, Any]]:
         for t in self:
             return t.environment_info
 
+    @property
+    def parent_steps(self) -> List[str]:
```
savingoyal, line 1974: we can return the step object instead of a str

metaflow/client/core.py
```diff
@@ -1837,6 +1970,38 @@ def environment_info(self) -> Optional[Dict[str, Any]]:
...
+        List[str]
+            List of parent step names
+        """
+        if self.id == "start":
+            return []
+        graph_info = self.task["_graph_info"].data
+        previous_nodes = []
+        for node_name, attributes in graph_info["steps"].items():
+            if self.id in attributes["next"]:
+                previous_nodes.append(node_name)
```
savingoyal, line 1989: yield to maintain consistency?

metaflow/client/core.py
```diff
@@ -1837,6 +1970,38 @@ def environment_info(self) -> Optional[Dict[str, Any]]:
...
+        return previous_nodes
+
+    @property
+    def child_steps(self) -> List[str]:
```
savingoyal, line 1993: same comments as above

metaflow/metadata_provider/metadata.py
```diff
@@ -672,6 +673,38 @@ def _register_system_metadata(self, run_id, step_name, task_id, attempt):
         if metadata:
             self.register_metadata(run_id, step_name, task_id, metadata)
 
+    @classmethod
+    def filter_tasks_by_metadata(
+        cls,
+        flow_id: str,
+        run_id: str,
+        query_step: str,
```
savingoyal, line 681: nit - `step_name`

metaflow/plugins/metadata_providers/local.py
```diff
@@ -202,6 +204,68 @@ def _optimistically_mutate():
...
+
+        regex = re.compile(pattern)
+        matching_task_pathspecs = []
+
+        for task in tasks:
+            task_id = task.get("task_id")
+            if not task_id:
+                continue
+
+            task_name = task.get("task_name")
```
savingoyal, line 249: curious - why `task_name`?

metaflow/task.py
```diff
@@ -493,6 +504,19 @@ def run_step(
                     )
                 )
 
+            # Add runtime dag information to the metadata of the task
+            foreach_execution_path = self._dynamic_runtime_metadata(foreach_stack)
```
savingoyal, line 508: [this](https://github.com/Netflix/metaflow/pull/2124/files#r1941819183) comment still applies

test/core/tests/client_child_tasks.py
```diff
@@ -0,0 +1,99 @@
+from metaflow_test import MetaflowTest, ExpectationFailed, steps
+
+
+class ChildrenTest(MetaflowTest):
```
savingoyal, line 4: you can combine these two tests together.

test/core/tests/client_child_tasks.py
```diff
@@ -0,0 +1,99 @@
...
+                actual_children_pathspecs_set = set(
+                    [task.pathspec for task in child_tasks]
+                )
+                expected_children_pathspecs_set = set()
+
+                # Get successor steps for the current task
+                child_steps = task.parent.child_steps
+
+                for child_task in child_tasks:
+                    assert task.pathspec in child_task.data.parent_pathspecs, (
```
savingoyal, line 81: can you verify other properties as well?

savingoyal (changes requested):

metaflow/plugins/metadata_providers/local.py
talsperre: Addressed.

metaflow/client/core.py
```diff
@@ -1123,6 +1124,153 @@ def _iter_filter(self, x):
...
+                )
+            except Exception as e:
+                if e.http_code == 404:
+                    # filter_tasks_by_metadata endpoint does not exist in the version of metadata service
+                    # deployed currently. Raise a more informative error message.
+                    raise MetaflowInternalError(
+                        "The version of metadata service deployed currently does not support filtering tasks by metadata. "
+                        "Upgrade to a newer version of Metadata service to use this feature."
+                    ) from e
+
```
savingoyal, line 1158: this code block can be within the service implementation and not in the client. the client shouldn't have to bother about the actual implementation semantics of metadata.

metaflow/client/core.py
```diff
@@ -1837,6 +1985,60 @@ def environment_info(self) -> Optional[Dict[str, Any]]:
...
+        graph_info = self.task["_graph_info"].data
+
+        if self.id != "start":
+            flow, run, _ = self.path_components
+            for node_name, attributes in graph_info["steps"].items():
+                if self.id in attributes["next"]:
+                    yield Step(f"{flow}/{run}/{node_name}", _namespace_check=False)
+
+    @property
+    def parent_steps(self) -> Optional["Step"]:
```
savingoyal, line 2007: the same property is duplicated in L:1989?

savingoyal, line 2007: returns an optional list of steps

metaflow/client/core.py
```diff
@@ -1837,6 +1985,60 @@ def environment_info(self) -> Optional[Dict[str, Any]]:
...
+            return
+        graph_info = self.task["_graph_info"].data
+        flow, run, _ = self.path_components
+        previous_nodes = []
+        for node_name, attributes in graph_info["steps"].items():
+            if self.id in attributes["next"]:
+                yield Step(f"{flow}/{run}/{node_name}", _namespace_check=False)
+
+    @property
+    def child_steps(self) -> "Step":
```
savingoyal, line 2026: returns an optional list of steps

metaflow/exception.py
```diff
@@ -160,6 +160,15 @@ def __init__(self, msg, unhandled):
         self.artifact_names = unhandled
 
 
+class ServiceException(MetaflowException):
```
savingoyal, line 163: why move this here? - this is tied to the service implementation and ideally shouldn't be pulled into the core

metaflow/client/core.py
```diff
@@ -27,6 +27,7 @@
     MetaflowInvalidPathspec,
     MetaflowNamespaceMismatch,
     MetaflowNotFound,
+    ServiceException,
```
savingoyal, line 30: superfluous?

metaflow/metadata_provider/metadata.py
```diff
@@ -672,6 +673,38 @@ def _register_system_metadata(self, run_id, step_name, task_id, attempt):
         if metadata:
             self.register_metadata(run_id, step_name, task_id, metadata)
 
+    @classmethod
+    def filter_tasks_by_metadata(
+        cls,
+        flow_id: str,
```
savingoyal, line 679: nit - we use `flow_name`

metaflow/task.py
```diff
@@ -37,6 +37,17 @@ class MetaflowTask(object):
     MetaflowTask prepares a Flow instance for execution of a single step.
     """
 
+    @staticmethod
+    def _dynamic_runtime_metadata(foreach_stack):
```
savingoyal, line 41: please move this in-line in L:514

test/core/contexts.json
```diff
@@ -24,7 +24,7 @@
                 "--tag=\u523a\u8eab means sashimi",
```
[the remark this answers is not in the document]
savingoyal: you can exclude black from touching files that are not in the git commit

test/core/tests/runtime_dag_tests.py
```diff
@@ -0,0 +1,180 @@
+from metaflow_test import MetaflowTest, ExpectationFailed, steps
```
savingoyal, line 1: nit - file can be named runtime_dag.py to maintain parity with existing convention

test/core/tests/runtime_dag_tests.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+class RuntimeDagTests(MetaflowTest):
```
savingoyal, line 4: nit - just `RuntimeDagTest`

test/core/tests/runtime_dag_tests.py
```diff
@@ -0,0 +1,180 @@
...
+                        f"the actual child task {child_task}."
+                    )
+                except Exception as e:
+                    raise AssertionError(
+                        f"Comparison failed with error: {str(e)}\n"
+                        f"Expected child task: {expected_child_task}\n"
+                        f"Actual child task: {child_task}"
+                    ) from e
+
+        from metaflow import Task
```
savingoyal, line 160: superfluous?

savingoyal (changes requested):

metaflow/metadata_provider/metadata.py
talsperre: Will change it to `flow_name` - I did see `flow_id` in a bunch of different places in `local.py` though.

metaflow/client/core.py
talsperre: If the endpoint does not exist in the metadata service then how do you even return that error message?

savingoyal: this try-except block can be in service.py rather than in the client.

metaflow/client/core.py
```diff
@@ -1837,6 +1974,60 @@ def environment_info(self) -> Optional[Dict[str, Any]]:
         for t in self:
             return t.environment_info
 
+    @property
+    def parent_steps(self) -> Iterator["Step"]:
```
savingoyal, line 1978: this is still duplicated?

metaflow/plugins/metadata_providers/service.py
```diff
@@ -318,6 +320,55 @@ def _new_task(
...
+        url = f"{url}/filtered_tasks?{urlencode(query_params)}"
+        try:
+            resp = cls._request(None, url, "GET")
+        except Exception as e:
+            if e.http_code == 404:
+                # filter_tasks_by_metadata endpoint does not exist in the version of metadata service
+                # deployed currently. Raise a more informative error message.
+                raise MetaflowInternalError(
+                    "The version of metadata service deployed currently does not support filtering tasks by metadata. "
+                    "Upgrade to a newer version of Metadata service to use this feature."
```
savingoyal, line 368: can you add the metaflow service version that they need to upgrade to? basically if the latest is 2.3.4, then, have a message that says upgrade to >=2.4.0

savingoyal: cc @saikonen to bump the version of metaflow-service when the endpoint is released.

metaflow/client/core.py
talsperre: Wait, I didn;t push this change. Let me do that.

PR #2125, opened 2024-10-31
prepare release 2.12.28

darinyu:


darinyu:
dummy [PR ](https://github.com/Netflix/metaflow/actions/runs/11622008667/job/32366763955?pr=2126)(checked out latest released metaflow), and same R=4.4.2 test failed, previously R=4.4.1 worked.

Metaflow doesn't bundle R so this looks fine.

PR #2132, opened 2024-11-05
Update escape hatch sys path when used in jupyter notebook

talsperre:
A small update to the escape hatch stubs. In notebooks, `''` is not necessarily the first element of `sys.path`, so when updating the `sys.path` when the escape hatch client is launched from CWD, we look at all elements of `sys.path` instead of just the first element like previously.


savingoyal (approved):
lgtm!

PR #2133, opened 2024-11-05
Deploy time triggers

KaylaSeeley:
https://docs.google.com/document/d/1-g-hPK5XC7sPmExXqFhxPTVAwnlPYyNKnmWaYIKvxb8/edit?tab=t.0#heading=h.6lt86ol75edp


metaflow/plugins/events_decorator.py
```diff
@@ -155,7 +189,19 @@ def flow_init(
...
+                            param_value, DeployTimeField
+                        ):
+                            new_param_value = DeployTimeField(
+                                "param", [list, dict], None, param_value, False
+                            )
+                            event["parameters"] = new_param_value
                         self.triggers.append(event)
+                    elif callable(event) and not isinstance(event, DeployTimeField):
+                        trig = DeployTimeField(
+                            "event", dict, None, self.attributes["event"], False
```
romain-intel, line 202: can also be str

metaflow/plugins/events_decorator.py
```diff
@@ -178,7 +231,7 @@ def flow_init(
             raise MetaflowException("No event(s) specified in *@trigger* decorator.")
 
         # same event shouldn't occur more than once
-        names = [x["name"] for x in self.triggers]
+        names = [x["name"] for x in self.triggers if not isinstance(x, DeployTimeField)]
```
romain-intel, line 234: not fully complete test, should be something like x not instance of deploytimefield but also x["name"] not a deploy time field.

metaflow/plugins/events_decorator.py
```diff
@@ -188,6 +241,63 @@ def flow_init(
 
         # TODO: Handle scenario for local testing using --trigger.
 
+    def format_deploytime_value(self):
+        for trigger in self.triggers:
+            # Case were trigger is a function that returns a list of events
```
romain-intel, line 246: nit: where

metaflow/plugins/events_decorator.py
```diff
@@ -188,6 +241,63 @@ def flow_init(
...
+            # Need to do this bc we need to iterate over list later
+            if isinstance(trigger, DeployTimeField):
+                if isinstance(deploy_time_eval(trigger), list):
+                    deploy_val = deploy_time_eval(trigger)
+                    self.triggers.remove(trigger)
```
romain-intel, line 251: it's typically dangerous to modify lists like this while they are being iterated on. I'd recommend not touching self.triggers but forming something like `new_triggers` and updating at the very end.

metaflow/plugins/events_decorator.py
```diff
@@ -188,6 +241,63 @@ def flow_init(
 
         # TODO: Handle scenario for local testing using --trigger.
 
+    def format_deploytime_value(self):
+        for trigger in self.triggers:
+            # Case were trigger is a function that returns a list of events
+            # Need to do this bc we need to iterate over list later
+            if isinstance(trigger, DeployTimeField):
+                if isinstance(deploy_time_eval(trigger), list):
```
romain-intel, line 249: Calling `deploy_time_eval` twice. Also, at the top level, something that is a deploy-time field will be:
  - a string (so needs to be handled like line 96)
  - a dictionary (single event in dict form)
  - a list of the two things above (when it is pushed as events)

So for string, that is done (we create the proper entry and done). For dictionary, we may need further processing of the elements (so continue). For list, add each element as above.

metaflow/plugins/events_decorator.py
```diff
@@ -188,6 +241,63 @@ def flow_init(
...
+                    deploy_val = deploy_time_eval(trigger)
+                    self.triggers.remove(trigger)
+                    self.triggers.extend(deploy_val)
+            else:
+                break
+        for trigger in self.triggers:
+            old_trigger = trigger
+            # Entire event is a function (returns either string or dict)
+            if isinstance(trigger, DeployTimeField):
```
romain-intel, line 258: This should not happen since we removed the DeployTimeField in the first step. Now we need to go inside each dict (we either have a fully resolved dict or a dict with some portions that are not yet evaluated).

metaflow/plugins/events_decorator.py
```diff
@@ -188,6 +241,63 @@ def flow_init(
...
+                    pass
+                # Case where just the name is a function (always a str)
+                if isinstance(trigger, str):
+                    trigger = {"name": trigger}
+            trigger_params = trigger.get("parameters", {})
+            # Case where param is a function (can return list or dict)
+            if isinstance(trigger_params, DeployTimeField):
+                trigger_params = deploy_time_eval(trigger_params)
+                try:
+                    trigger_params = json.loads(trigger_params)
```
romain-intel, line 272: we shouldn't need this -- the user needs to return this appropriately.

romain-intel (commented):
Haven't fully finished the main function but I think it needs a bit more work (and some structure to reason about). It is a bit convoluted (not the function, just the logic) so we need to be careful and think clearly (ie: write things out :) ).

metaflow/plugins/events_decorator.py
KaylaSeeley: I think this happens when the user returns a dictionary, not sure why

metaflow/plugins/events_decorator.py
```diff
@@ -312,6 +422,13 @@ def flow_init(
                             "The *project_branch* attribute of the *flow* is not a string"
                         )
                 self.triggers.append(result)
+            elif callable(self.attributes["flow"]) and not isinstance(
```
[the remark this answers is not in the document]
KaylaSeeley: We had this check in our old trigger code in metaflow-custom, so I kept it

metaflow/flowspec.py
```diff
@@ -193,8 +193,11 @@ def _set_constants(self, graph, kwargs):
             setattr(self, var, val)
 
         # We store the DAG information as an artifact called _graph_info
+        for deco in flow_decorators(self):
+            if deco.name == "trigger":
+                for trig in deco.triggers:
+                    print(trig)
```
saikonen, line 199: is this for debugging purposes only? can be removed in that case.
same goes for the changes in `_vendor`, and what looks to be a local vscode config(?)

romain-intel (commented):
Maybe a minor nit to clean it up a tad more but otherwise no concern.

PR #2140, opened 2024-11-13
fix: datastore regression issue

### LINKED ISSUE #2137: BUG: Data store error - AWS batch/step execution
**Environment:**
metaflow version: 2.12.29
Python 3.11 (Docker Image from public.ecr.aws/docker/library/python:3.11)
Running on AWS Batch

**Description:**
Tested with version 2.12.28 and it runs successfully, with this latest version we get:
Data store error: No completed attempts of the task was found for task `MyFlow/sfn-*/_parameters/*-params`.

Maybe worth mentioning that we include a json file into `MyFlow` like:
```

json_config = IncludeFile(
        name="my_config",
        required=True,
        help="The Configuration",
        default=f"./{PARAMS_JSON}",
    )

```

saikonen:
fixes #2137

### DIFF
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -286,27 +286,26 @@ def dump(obj, input_path, private=None, max_value_size=None, include=None, file=
     for ds in ds_list:
         if ds is not None:
             tasks_processed = True
-            echo(
-                "Dumping output of run_id=*{run_id}* "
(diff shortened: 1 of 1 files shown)


saikonen:
does not resolve the original issue as is, closing and opting to rollback problematic PR for now

PR #2145, opened 2024-11-16
Make CLI lazy load

romain-intel:
This will allow parameters to be added when processing the `start` step (for the upcomming config change).


madhur-ob (approved):
tested, can merge but small conflict is to be resolved first

romain-intel:
Included in #1962

PR #2147, opened 2024-11-19
Add generic typing to multicore_utils.py

bryant1410:
This change adds generic typing into the functions from multicore_utils.py. This allows for better type checking.

### DIFF
--- a/metaflow/multicore_utils.py
+++ b/metaflow/multicore_utils.py
@@ -6,7 +6,18 @@ from tempfile import NamedTemporaryFile
 import time
 import metaflow.tracing as tracing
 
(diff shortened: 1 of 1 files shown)


savingoyal (approved):

saikonen (commented):
should/could the types in the docstrings also be updated with the TypeVar types?

bryant1410:
Is the failing test flaky?

romain-intel:
Hey @bryant1410; thanks for the PR!!. The S3 tests here seem flaky (@savingoyal ?). The 7 failures can be checked internally (I'll DM you the link) and they are all flaky so this should be good to go from a testing pov.

PR #2155, opened 2024-11-28
feature: Support choosing Kubernetes QoS class through the decorator

saikonen:
enables choosing the Kubernetes QoS class out of `Guaranteed/Burstable/BestEffort` through the decorator.


metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -109,6 +111,8 @@ class KubernetesDecorator(StepDecorator):
     hostname_resolution_timeout: int, default 10 * 60
         Timeout in seconds for the workers tasks in the gang scheduled cluster to resolve the hostname of control task.
         Only applicable when @parallel is used.
+    qos_class: str, optional, default: None
```
savingoyal, line 114: defaults to `Burstable`?

metaflow/plugins/kubernetes/kube_utils.py
```diff
@@ -23,3 +23,26 @@ def parse_cli_options(flow_name, run_id, user, my_runs, echo):
...
+def qos_requests_and_limits(qos_class: str, cpu: int, memory: int):
+    "return resource requests and limits for the kubernetes pod based on the given QoS Class"
+    # Determine the requests and limits to define chosen QoS class
+    qos_limits = {}
+    qos_requests = {}
+    if qos_class == "Guaranteed":
+        # Guaranteed - has both cpu/memory limits. requests not required, as these will be inferred.
+        qos_limits = {
+            "cpu": str(cpu),
+            "memory": "%sM" % str(memory),
```
savingoyal, line 37: what about storage?

metaflow/plugins/kubernetes/kubernetes_decorator.py
savingoyal, line 114: might be useful to override the default through config as well

metaflow/plugins/kubernetes/kube_utils.py
savingoyal: just ephemeral-storage. gpu's always need to specified with both requests and limits. also, can you verify airflow too?

metaflow/plugins/kubernetes/kube_utils.py
```diff
@@ -23,3 +23,26 @@ def parse_cli_options(flow_name, run_id, user, my_runs, echo):
...
 
     return flow_name, run_id, user
+
+
+def qos_requests_and_limits(qos: str, cpu: int, memory: int):
+    "return resource requests and limits for the kubernetes pod based on the given QoS Class"
+    # Determine the requests and limits to define chosen QoS class
+    qos_limits = {}
+    qos_requests = {}
+    if qos == "Guaranteed":
```
savingoyal, line 33: maybe a case insensitive check?

saikonen: there's validation in the kubernetes decorator that is case-sensitive as well. I can also make all defaults lower case but `besteffort` reads a bit clunky.

metaflow/plugins/kubernetes/kube_utils.py
saikonen: handling storage as well now. For the `BestEffort` case, I figure we keep the ephemeral-storage in the requests, or do we omit it completely?

metaflow/plugins/kubernetes/kube_utils.py
```diff
@@ -27,17 +27,19 @@ def parse_cli_options(flow_name, run_id, user, my_runs, echo):
...
         # Guaranteed - has both cpu/memory limits. requests not required, as these will be inferred.
         qos_limits = {
             "cpu": str(cpu),
             "memory": "%sM" % str(memory),
             "ephemeral-storage": "%sM" % str(storage),
         }
-    elif qos == "BestEffort":
+    elif qos == "besteffort":
         # BestEffort - no limit or requests for cpu/memory
         qos_requests = {"ephemeral-storage": "%sM" % str(storage)}
```
savingoyal, line 44: should we even support best effort at the moment? there doesn't seem to be a use case for it at the moment

metaflow/plugins/kubernetes/kube_utils.py
saikonen: validation changed to case insensitive.

savingoyal:
lgtm! can you verify airflow works as expected?

PR #2156, opened 2024-11-29
fix: S3 paths with fragments not working

saikonen:
S3 paths with url fragments do not work with the Metaflow S3 client, but do work with boto3 directly. The cause is that any key we pass to S3 methods goes through `urlparse`, of which we only use `netloc` and `path`. Fragments get parsed into their own key in the ParseResult.

### DIFF
--- a/metaflow/plugins/datatools/s3/s3.py
+++ b/metaflow/plugins/datatools/s3/s3.py
@@ -600,7 +600,7 @@ class S3(object):
         # returned are Unicode.
         key = getattr(key_value, "key", key_value)
         if self._s3root is None:
-            parsed = urlparse(to_unicode(key))
+            parsed = urlparse(to_unicode(key), allow_fragments=False)
             if parsed.scheme == "s3" and parsed.path:
                 return key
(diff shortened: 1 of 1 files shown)


romain-intel:
I'll check. I suspect it's some flaky tests (we have been having a few that I haven't fixed yet).

metaflow/plugins/datatools/s3/s3.py
```diff
@@ -600,7 +600,7 @@ def _url(self, key_value):
         # returned are Unicode.
         key = getattr(key_value, "key", key_value)
         if self._s3root is None:
-            parsed = urlparse(to_unicode(key))
+            parsed = urlparse(to_unicode(key), allow_fragments=False)
```
savingoyal, line 603: can you add a comment here calling out why `allow_fragments` needs to be set to False for posterity

savingoyal (dismissed):
lgtm with a minor nit

romain-intel (approved):
Test issues are flaky tests.

PR #2157, opened 2024-12-03
Deploy time triggers (v2)

talsperre:


romain-intel:
Ok -- so the only change was the append for events. Was this not initially caught because we had coverage for event but not events? The change looks fine to me.

romain-intel (dismissed):

PR #2161, opened 2024-12-04
Note lack of support for `minutes_between_retries` under AWS Step Functions

dpoznik:
This PR adds one line to the `@retry` decorator's `minutes_between_retries` parameter, clearing up some confusion on why it didn't work as advertised under Step Functions.

From @savingoyal, "when we integrated step functions with aws batch, there wasn't any support for `minutes_between_retries` in step functions - hence that field is ignored. and unfortunately that is still the case with step functions today."


dpoznik:
Closing, as @savingoyal will note this caveat elsewhere.

PR #2167, opened 2024-12-07
Fix case where the decorator's init function could be called twice.

romain-intel:
In some cases, specifically when using argo/airflow/sfn, the decorator's init function could be called twice which caused issues with the conda/pypi decorator.


savingoyal (approved):

romain-intel:
Internal tests pass as well -- merging and releasing.

PR #2169, opened 2024-12-09
Fix issue where run_command blocks when it shouldn't

romain-intel:
In cases where we need to read from a file (for the result of run_command), the current code used to make it sequential (so the timeout was actually never triggered).

This started causing problems on MacOS when writing large amounts of data (8KB FIFO)

### DIFF
--- a/metaflow/runner/metaflow_runner.py
+++ b/metaflow/runner/metaflow_runner.py
@@ -221,7 +221,7 @@ class Runner(object):
         The directory to run the subprocess in; if not specified, the current
         directory is used.
     file_read_timeout : int, default 3600
-        The timeout until which we try to read the runner attribute file.
+        The timeout until which we try to read the runner attribute file (in seconds).
     **kwargs : Any
         Additional arguments that you would pass to `python myflow.py` before
         the `run` command.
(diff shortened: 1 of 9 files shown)


metaflow/runner/metaflow_runner.py
```diff
@@ -326,6 +326,7 @@ def run(self, **kwargs) -> ExecutingRun:
                 show_output=self.show_output,
             )
             command_obj = self.spm.get(pid)
+            command_obj.sync_wait()
 
             return self.__get_executing_run(attribute_file_fd, command_obj)
```
madhur-ob, line 331: I guess `command_obj.sync_wait()` for Runner should happen inside `self.__get_executing_run` since we wanna do it after `content = handle_timeout(...)`

metaflow/runner/metaflow_runner.py
```diff
@@ -357,6 +358,7 @@ def resume(self, **kwargs):
                 show_output=self.show_output,
             )
             command_obj = self.spm.get(pid)
+            command_obj.sync_wait()
```
madhur-ob, line 361: same comment here

PR #2176, opened 2024-12-12
boost local bootstrap latency by 20%

savingoyal:


metaflow/plugins/pypi/micromamba.py
```diff
@@ -78,6 +86,7 @@ def solve(self, id_, packages, python, platform):
                 "--dry-run",
                 "--no-extra-safety-checks",
                 "--repodata-ttl=86400",
+                "--safety-checks=disabled",
```
saikonen, line 89: was there a specific reason for adding this? how does it play out with `--no-extra-safety-checks` ?

savingoyal: yep - it seems that extra-safety-checks runs extra checks alltogether. it wasn't immediately clear if safety-checks=disabled also disables the extra-safety-checks.

PR #2178, opened 2024-12-13
feature: micromamba download from mirror

saikonen:
add support for downloading micromamba from high-throughput mirror

### DIFF
--- a/metaflow/plugins/pypi/micromamba.py
+++ b/metaflow/plugins/pypi/micromamba.py
@@ -3,6 +3,7 @@ import json
 import os
 import subprocess
 import tempfile
+import time
 
 from metaflow.exception import MetaflowException
 from metaflow.util import which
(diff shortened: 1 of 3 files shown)


metaflow/plugins/pypi/micromamba.py
```diff
@@ -338,23 +339,37 @@ def _install_micromamba(installation_location):
     # Unfortunately no 32bit binaries are available for micromamba, which ideally
     # shouldn't be much of a problem in today's world.
     platform = conda_platform()
-    try:
-        subprocess.Popen(f"mkdir -p {installation_location}", shell=True).wait()
```
savingoyal, line 342: do we need to create the install location?

saikonen: I swapped this to `os.makedirs`, can also revert and instead keep the mkdir, or alternatively supply `--create-dirs` for curl.

saikonen: but yes, piping the curl to tar can fail if the install location does not exist, as tar doesn't create missing target directories by itself

savingoyal: ah i missed the os.makedirs

savingoyal (dismissed):

PR #2179, opened 2024-12-14
refactor: rename parse_secret_string_as_json option to json for clarity

isatyamks:
Updated the docstring to reflect the correct option name json instead of the incorrect parse_secret_string_as_jso

        If the secret contains a string payload ("SecretString"):
        - if the `json` option is True (default):
            {SecretString} will be parsed as a JSON. If successfully parsed, AND the JSON contains a
            top-level object, each entry K/V in the object will also be converted to an entry in the result. V will
            always be casted to a string (if not already a string).
        - If `json` option is False:
            {SecretString} will be returned as a single entry in the result, with the key being the secret_id.n.

### DIFF
--- a/metaflow/plugins/aws/secrets_manager/aws_secrets_manager_secrets_provider.py
+++ b/metaflow/plugins/aws/secrets_manager/aws_secrets_manager_secrets_provider.py
@@ -50,11 +50,11 @@ class AwsSecretsManagerSecretsProvider(SecretsProvider):
         The secret payload from AWS is EITHER a string OR a binary blob.
 
         If the secret contains a string payload ("SecretString"):
-        - if the `parse_secret_string_as_json` option is True (default):
+        - if the `json` option is True (default):
             {SecretString} will be parsed as a JSON. If successfully parsed, AND the JSON contains a
             top-level object, each entry K/V in the object will also be converted to an entry in the result. V will
             always be casted to a string (if not already a string).
-        - If `parse_secret_string_as_json` option is False:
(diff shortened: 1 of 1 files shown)


metaflow/plugins/aws/secrets_manager/aws_secrets_manager_secrets_provider.py
```diff
@@ -50,11 +50,11 @@ def get_secret_as_dict(self, secret_id, options={}, role=None):
         The secret payload from AWS is EITHER a string OR a binary blob.
 
         If the secret contains a string payload ("SecretString"):
-        - if the `parse_secret_string_as_json` option is True (default):
+        - if the `json` option is True (default):
             {SecretString} will be parsed as a JSON. If successfully parsed, AND the JSON contains a
             top-level object, each entry K/V in the object will also be converted to an entry in the result. V will
             always be casted to a string (if not already a string).
-        - If `parse_secret_string_as_json` option is False:
+        - If `json` option is False:
```
saikonen, line 57: as we're updating the docstring to match behaviour, could you also add for the `json=False` case that the returned secret key is either the `secret_id` OR the value set by `options={"env_var_name": custom_env_var_name}`

see L149:153 for that code path

isatyamks: hey @saikonen  Now i think the docstring is aligned with the actual behavior and provides a clear explanation of the json=False case.

saikonen (approved):

PR #2181, opened 2024-12-16
wait for terminal state

madhur-ob:


metaflow/plugins/argo/argo_workflows_deployer_objects.py
```diff
@@ -170,6 +171,44 @@ def terminate(self, **kwargs) -> bool:
         command_obj.sync_wait()
         return command_obj.process.returncode == 0
 
+    def wait(self, timeout: Optional[int] = None):
```
savingoyal, line 174: should we make this `wait_for_completion`?

madhur-ob: yes, i asked for good name suggestions, was waiting for one!

PR #2182, opened 2024-12-17
include flowdir in syspath while loading file in Runner API

madhur-ob:
Context:

The `cwd` argument of the Runner is for selecting the directory under which the `subprocess` executes in..

But, it doesn't play a role in loading of the flow file itself -- which is the first step of the lower level API to generate the command that will be executed via `subprocess`..

aka loading of file --> generate command --> run via subprocess (`cwd` is only relevant here)

Thus, in the directory structure like this:

```
├── pipelines/
│   ├── training.py
│   └── common.py
└── tests/
    └── test_sample.py
```

if `training.py` has `from common import packages`
and `test_sample.py` uses the Runner API as follows:

```
with Runner("pipelines/training.py", environment="conda").run() as running:
        return running.run.data
```

the following issue is encountered:
```
>   from common import packages
E   ModuleNotFoundError: No module named 'common'
```

This PR adds the `flow directory` into `sys.path` and also removes it later...

### DIFF
--- a/metaflow/runner/click_api.py
+++ b/metaflow/runner/click_api.py
@@ -188,28 +188,36 @@ loaded_modules = {}
 def extract_flow_class_from_file(flow_file: str) -> FlowSpec:
     if not os.path.exists(flow_file):
         raise FileNotFoundError("Flow file not present at '%s'" % flow_file)
-    # Check if the module has already been loaded
-    if flow_file in loaded_modules:
-        module = loaded_modules[flow_file]
-    else:
-        # Load the module if it's not already loaded
-        spec = importlib.util.spec_from_file_location("module", flow_file)
-        module = importlib.util.module_from_spec(spec)
-        spec.loader.exec_module(module)
-        # Cache the loaded module
-        loaded_modules[flow_file] = module
-    classes = inspect.getmembers(module, inspect.isclass)
 
-    flow_cls = None
-    for _, kls in classes:
-        if kls != FlowSpec and issubclass(kls, FlowSpec):
-            if flow_cls is not None:
-                raise MetaflowException(
-                    "Multiple FlowSpec classes found in %s" % flow_file
-                )
-            flow_cls = kls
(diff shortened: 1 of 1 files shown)


metaflow/runner/click_api.py
```diff
@@ -188,28 +188,42 @@ def get_inspect_param_obj(p: Union[click.Argument, click.Option], kind: str):
...
+                flow_cls = kls
+
+        return flow_cls
+    finally:
+        # Safely remove the flow directory from sys.path if it exists
+        try:
+            # Remove the flow directory from sys.path
+            sys.path.remove(flow_dir)
+        except ValueError:
+            # The path might have been modified during module loading,
```
savingoyal, line 224: what is the user explicitly added flow_dir to sys.path?

madhur-ob: we can only add after checking first if it is not there? and only remove it if we were the ones who added it

madhur-ob: done here: https://github.com/Netflix/metaflow/pull/2182/commits/ae2fee18853e3f10a4e790dc35c57832c3f7c118

metaflow/runner/click_api.py
```diff
@@ -188,28 +188,42 @@ def get_inspect_param_obj(p: Union[click.Argument, click.Option], kind: str):
...
+                    raise MetaflowException(
+                        "Multiple FlowSpec classes found in %s" % flow_file
+                    )
+                flow_cls = kls
+
+        return flow_cls
+    finally:
+        # Only remove from path if we added it
+        if path_was_added:
+            sys.path.remove(flow_dir)
```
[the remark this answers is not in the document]
madhur-ob: would you prefer the safe removal approach in one of the previous commits here: https://github.com/Netflix/metaflow/pull/2182/commits/ec43bc66b1ba4bd33dbc5a33b2dc59129555a0ae

savingoyal (approved):

PR #2183, opened 2024-12-18
Fix/configs update

savingoyal:


metaflow/cli.py
```diff
@@ -354,19 +382,76 @@ def start(
...
+    if (
+        hasattr(ctx, "saved_args")
+        and ctx.saved_args
+        and ctx.saved_args[0] == "resume"
+        and getattr(ctx.obj, "has_config_options", False)
+    ):
+        # In the case of resume, we actually need to load the configurations
+        # from the resumed run to process them. This can be slightly onerous so check
+        # if we need to in the first place
+        if getattr(ctx.obj, "has_cl_config_options", False):
```
savingoyal, line 396: what's cl?

romain-intel: command line. I can make it longer. Basically, this part of the code is only hit if the user explicitly specified a `--config` or `--config-value` option (as opposed to being derived from defaults).

metaflow/cli.py
```diff
@@ -519,10 +591,13 @@ def _check(echo, graph, flow, environment, pylint=True, warnings=False, **kwargs
...
 def print_metaflow_exception(ex):
     echo_always(ex.headline, indent=True, nl=False, bold=True)
+    location = ""
+    if ex.source_file is not None:
+        location += " in file %s" % ex.source_file
     if ex.line_no is None:
-        echo_always(":")
-    else:
-        echo_always(" on line %d:" % ex.line_no, bold=True)
+        location += " on line %d" % ex.line_no
```
saikonen, line 598: This part fails as line_no is None but the string is expecting an integer.

saikonen: should be able to hit this with any pypi/conda flow and failing to apply `--environment conda` on the CLI at least.

saikonen (changes requested):

metaflow/cli.py
romain-intel: well yes, I am an idiot -- the condition is reversed above. Let me fix that.

saikonen (approved):

PR #2193, opened 2025-01-02
[heartbeats] Better error handling

valayDave:
- Ensure better error handling so that any server related flakiness can be handled without crashing the heartbeat thread.
- If the heartbeat thread crashes because of any server issues / DNS issues / Intermittent server flakiness, it will kill the heartbeat thread causing the UI to mark the run as failed. 
- This change ensures that we don't crash the thread an makes it exponentially backoff calling the server.

### DIFF
--- a/metaflow/metadata_provider/heartbeat.py
+++ b/metaflow/metadata_provider/heartbeat.py
@@ -53,7 +53,7 @@ class MetadataHeartBeat(object):
                 retry_counter = 0
             except HeartBeatException as e:
                 retry_counter = retry_counter + 1
-                time.sleep(4**retry_counter)
+                time.sleep(1.5**retry_counter)
 
     def _heartbeat(self):
         if self.hb_url is not None:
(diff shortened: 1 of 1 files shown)


metaflow/metadata_provider/heartbeat.py
```diff
@@ -57,9 +57,23 @@ def _ping(self):
 
     def _heartbeat(self):
         if self.hb_url is not None:
-            response = requests.post(
-                url=self.hb_url, data="{}", headers=self.headers.copy()
-            )
+            try:
```
savingoyal, line 60: should we change `time.sleep(4**retry_counter)` to be less aggressive? It is likely that the UI may still mark the task as failed, given how aggressively the sleep value grows.

valayDave: X^4 is hyper aggressive. Here is what it looks like 

```

x    x^2    x^4    Sum(x^2)    Sum(x^4)
1    1       1       1           1
2    4       16      5           17
3    9       81      14          98
4    16      256     30          354
5    25      625     55          979
6    36      1296    91          2275
7    49      2401    140         4676
8    64      4096    204         8772
9    81      6561    285         15333
10   100     10000   385         25333

```

This means that with x^4 you only hit 5/6 retires. With x^4 you can have way more. Overall, what's the merit of making it too aggressive ?

PR #2196, opened 2025-01-06
[cards] `PythonCode` component

valayDave:
- Metaflow Card component for rendering a syntax highlighed python function
- Driveby: moved `with_component_id` and `create_component_id` functions to `card.py`
- Default cards now have a code section present too
- Passing flow object to MetaflowCard's init function so that the step function can be retrived by the card when required.
- Any external cards that are using the static __init__ signature of the MetaflowCard will need to change it to the new signature. Any MetaflowCard's using **kwargs in the init signature are safe from the blast radius.

- Example flow :

```python
from metaflow import FlowSpec, step, card, current
from metaflow.cards import PythonCode

class PythonCodeDemoFlow(FlowSpec):
    @card
    @step
    def start(self):
        # Example 1: Using a function
        def sample_function():
            x = 1
            y = 2
            return x + y

        current.card.append(PythonCode(sample_function))

        # Example 2: Using a string
        code = """
def another_function():
    return "Hello World"
        """
        current.card.append(PythonCode(code_string=code))

        self.next(self.end)

    @step
    def end(self):
        pass

if __name__ == '__main__':
    PythonCodeDemoFlow()
```


metaflow/plugins/cards/card_modules/components.py
```diff
@@ -823,3 +825,39 @@ def render(self):
...
+            return x + y
+
+        current.card.append(
+            PythonCode(my_function)
+        )
+    ```
+
+    Parameters
+    ----------
+    code_func : function
```
romain-intel, line 851: This is probably just a str; also, please add type hints in definition.

metaflow/cards.py
```diff
@@ -8,6 +8,7 @@
     Markdown,
     VegaChart,
     ProgressBar,
+    PythonCode,
```
romain-intel, line 11: Didn't check but is this just for python code or can it handle any ```<language> type of syntax?

valayDave: No just python code. There is no other code because the `prism` library vendored in our cards package currently only has support for `python`, `log` and `c-like` languages. We would have to vendor other languages for full support (but that was too much work for now and we didn't face any usecases for such)

metaflow/plugins/cards/card_modules/components.py
valayDave: that's a fair point, How about I allow both ? It can take a ``code_func` or a `code_string` parameter.

metaflow/plugins/cards/card_modules/card.py
```diff
@@ -140,3 +141,17 @@ def render(self):
...
+    return type(component).__name__.lower() + "_" + uuid_bit
+
+
+def with_default_component_id(func):
+    def ret_func(self, *args, **kwargs):
+        if self.component_id is None:
+            self.component_id = create_component_id(self)
+        return func(self, *args, **kwargs)
+
+    return ret_func
```
valayDave, line 157: Just moving things around. Nothing has changed overall.

metaflow/plugins/cards/card_modules/components.py
```diff
@@ -7,25 +7,13 @@
...
-    return type(component).__name__.lower() + "_" + uuid_bit
-
-
-def with_default_component_id(func):
-    def ret_func(self, *args, **kwargs):
-        if self.component_id is None:
-            self.component_id = create_component_id(self)
-        return func(self, *args, **kwargs)
-
-    return ret_func
```
valayDave, line 28: Just moved it to a common place so that it can be used in other modules as well.

metaflow/plugins/cards/card_modules/components.py
valayDave: Done!

metaflow/plugins/cards/card_modules/components.py
```diff
@@ -823,3 +811,59 @@ def render(self):
...
+            return "Hello World"
+        '''
+        current.card.append(
+            PythonCode(code_string=code)
+        )
+    ```
+
+    Parameters
+    ----------
+    code_func : function, optional
```
romain-intel, line 846: small nits:
 - the proper type would be something like Callable[..., Any] (not function).
 - For consistency: `code_func : Callable[..., Any], optional, default None` and same for code_string
 - add these to the func definition as type hints.

valayDave: Done! I have also integrated this in the default card and had to make changes accordingly to support that.

metaflow/plugins/cards/card_modules/card.py
```diff
@@ -66,7 +67,7 @@ class MetaflowCard(object):
     # FIXME document runtime_data
     runtime_data = None
 
-    def __init__(self, options={}, components=[], graph=None):
```
valayDave, line 69: Tweeking function signature here so that default card can access the step function and render it .

metaflow/plugins/cards/card_modules/test_cards.py
valayDave, line 1: Altered function signatures here for simplicity.

metaflow/plugins/cards/card_modules/basic.py
```diff
@@ -475,6 +497,16 @@ def render(self):
...
+        step_func = getattr(self.flow, self._task.parent.id)
+        code_table = SectionComponent(
+            title="Task Code",
+            contents=[
+                TableComponent(
+                    data=[[PythonCodeComponent(inspect.getsource(step_func)).render()]]
+                )
+            ],
+        ).render()
+
```
valayDave, line 509: Task code is now a part of the default card.

savingoyal (approved):

PR #2207, opened 2025-01-12
metadata for spot termination for k8s

madhur-ob:
CLI works if I do something like:

```
python spot.py metadata record --run-id 219719 --step-name start --task-id 1346034 --field try yo --tag attempt_id:0
```

This adds the metadata to a running flow, see the screenshot below:

Screenshot 2025-01-12 at 10 03 58 PM

But, how to locate the flow file inside the pod?


metaflow/plugins/kubernetes/spot_monitor.py
```diff
@@ -0,0 +1,112 @@
...
+                    termination_time = response.text
+                    self._emit_termination_metadata(termination_time)
+                    os.kill(os.getppid(), signal.SIGTERM)
+                    break
+            except (requests.exceptions.RequestException, requests.exceptions.Timeout):
+                pass
+            time.sleep(self.POLL_INTERVAL)
+
+    def _emit_termination_metadata(self, termination_time):
+        command = [
```
madhur-ob, line 83: The CLI works when used independently...

but, I wonder how to run it here inside the sidecar..
The main question being: how to locate where the flow file is...
because things are running inside a pod etc.

aka, we need to fix the following:
```
command = [
    sys.executable,
    current.flow_name,
    ....
]
```

madhur-ob: Alternatively, we need to access the `metadata` object, which currently we get by `obj.metadata` (that needs the location of the flow file...)

If we can get / construct the `metadata` object somehow, we no longer need the flow file path...

metaflow/plugins/kubernetes/spot_monitor_sidecar.py
```diff
@@ -0,0 +1,105 @@
...
+                    os.kill(os.getppid(), signal.SIGTERM)
+                    break
+            except (requests.exceptions.RequestException, requests.exceptions.Timeout):
+                pass
+            time.sleep(self.POLL_INTERVAL)
+
+    def _emit_termination_metadata(self, termination_time):
+        command = [
+            sys.executable,
+            f"/metaflow/{os.getenv('FLOW_FILE_PATH')}",
```
savingoyal, line 83: can you set the env to `METAFLOW_FLOW_FILE` - also it seems that it isn't a `PATH` but a `FILE_NAME`?

savingoyal: same for the other envs - i think most of those are already available - you can check using `os.getenv`

metaflow/plugins/kubernetes/kubernetes_cli.py
```diff
@@ -275,6 +275,16 @@ def _sync_metadata():
                 ),
             )
 
+    env.update(
```
savingoyal, line 278: Can you add a note here why these are needed so that they are not wiped by mistake.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -1705,6 +1705,7 @@ def _container_templates(self):
                     },
                     **{
                         # Some optional values for bookkeeping
+                        "MF_FLOW_FILENAME": os.path.basename(sys.argv[0]),
```
savingoyal, line 1708: minor nit - METAFLOW_ instead of MF_ for consistency. some fields come from MF_ but they are neatly scoped to the ones coming from mflog at the moment.

madhur-ob: ok, let me change in both k8s and argo

metaflow/plugins/kubernetes/spot_metadata_cli.py
```diff
@@ -0,0 +1,66 @@
...
+    "--step-name",
+    required=True,
+    help="Step Name for which metadata is to be recorded.",
+)
+@click.option(
+    "--task-id",
+    required=True,
+    help="Task ID for which metadata is to be recorded.",
+)
+@click.option(
```
savingoyal, line 32: ideally, it should take the notice as an input - otherwise, it's too generic. we can leave it as is, but can you add a quick FIXME in comments. we can merge right after.

PR #2208, opened 2025-01-13
fix bug for fq_name

Qianh1225:
The bug cause the issue when user use the @trigger_on_finish. Previously, it is not possible to deploy flows with `@trigger_on_finish` decorator whose event name was a string. 

```python
@trigger_on_finish(flow="example_flow")
class MyFlow(FlowSpec):
        ...
```

Deploying this to our internal scheduler would fail because in `flow_init` we would convert `example_flow` to a fully qualified name (`fq_name`) [here](https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/events_decorator.py#L402-L406) and when parsing the triggers during graph creation, we would read the value as `name` instead of `fq_name`. This PR updates that and the logic for parsing deploy time triggers. 

The code can be further refactored for readability and @Qianh1225 will push a follow up PR for that.

### DIFF
--- a/metaflow/plugins/events_decorator.py
+++ b/metaflow/plugins/events_decorator.py
@@ -598,10 +598,36 @@ class TriggerOnFinishDecorator(FlowDecorator):
             old_trig = trigger
             if isinstance(trigger, DeployTimeField):
                 trigger = deploy_time_eval(trigger)
-                if isinstance(trigger, dict) and "fq_name" not in trigger:
-                    # When a user has a deploy time trigger on finish which returns name,
-                    # project, branch
-                    trigger["fq_name"] = trigger.get("name")
+                if is_stringish(trigger):
+                    trigger = {"fq_name": trigger}
+                elif isinstance(trigger, dict):
+                    if "name" not in trigger:
+                        raise MetaflowException(
+                            "The *flow* attribute for *@trigger_on_finish* is missing the "
+                            "*name* key."
+                        )
+                    flow_name = trigger["name"]
+
+                    if not is_stringish(flow_name) or "." in flow_name:
+                        raise MetaflowException(
+                            "The *name* attribute of the *flow* is not a valid string"
+                        )
+                    result = {"fq_name": flow_name}
+                    if "project" in trigger:
+                        if is_stringish(trigger["project"]):
+                            result["project"] = trigger["project"]
+                        else:
+                            raise MetaflowException(
+                                "The *project* attribute of the *flow* is not a string"
+                            )
+                    if "project_branch" in trigger:
+                        if is_stringish(trigger["project_branch"]):
+                            result["branch"] = trigger["project_branch"]
+                        else:
+                            raise MetaflowException(
+                                "The *project_branch* attribute of the *flow* is not a string"
+                            )
+                    trigger = result
             if isinstance(trigger, dict):
                 trigger["fq_name"] = trigger.get("fq_name")
                 trigger["project"] = trigger.get("project")
--- a/metaflow/plugins/events_decorator.py
+++ b/metaflow/plugins/events_decorator.py
@@ -599,7 +599,7 @@ class TriggerOnFinishDecorator(FlowDecorator):
             if isinstance(trigger, DeployTimeField):
                 trigger = deploy_time_eval(trigger)
             if isinstance(trigger, dict):
-                trigger["fq_name"] = trigger.get("name")
+                trigger["fq_name"] = trigger.get("fq_name")
                 trigger["project"] = trigger.get("project")
                 trigger["branch"] = trigger.get("project_branch")
             # We also added this bc it won't be formatted yet
@@ -607,6 +607,6 @@ class TriggerOnFinishDecorator(FlowDecorator):
                 trigger = {"fq_name": trigger}
                 trigger = self._parse_fq_name(trigger)
             self.triggers[self.triggers.index(old_trig)] = trigger
-
+    
     def get_top_level_options(self):
         return list(self._option_values.items())
--- a/metaflow/plugins/events_decorator.py
+++ b/metaflow/plugins/events_decorator.py
@@ -623,7 +623,7 @@ class TriggerOnFinishDecorator(FlowDecorator):
                             )
                     if "project_branch" in trigger:
                         if is_stringish(trigger["project_branch"]):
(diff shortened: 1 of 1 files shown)


savingoyal:
@Qianh1225 thanks for the PR~! Can you run the pre-commit hook to unblock the merge?

talsperre:
Let's add the new tests internally and wait for them to succeed as well.

Qianh1225:
> Let's add the new tests internally and wait for them to succeed as well.

Do you want to me add test for testing this code? Can you be more specific?

talsperre:
I meant adding of the new UX tests to test deployment of flows with deploy time and static triggers.

saikonen (dismissed):

saikonen (commented):
Change seems good to go. Would like some more details on what the actual issue was that this caused though.

metaflow/plugins/events_decorator.py
```diff
@@ -598,8 +598,38 @@ def format_deploytime_value(self):
             old_trig = trigger
             if isinstance(trigger, DeployTimeField):
                 trigger = deploy_time_eval(trigger)
+                if is_stringish(trigger):
```
romain-intel, line 601: Nit: I would factor this out since it's common with above.

metaflow/plugins/events_decorator.py
```diff
@@ -598,8 +598,38 @@ def format_deploytime_value(self):
...
+                            )
+                    if "project_branch" in trigger:
+                        if is_stringish(trigger["project_branch"]):
+                            result["branch"] = trigger["project_branch"]
+                        else:
+                            raise MetaflowException(
+                                "The *project_branch* attribute of the *flow* is not a string"
+                            )
+                    trigger = result
             if isinstance(trigger, dict):
```
romain-intel, line 631: maybe add a comment on this code: "effect is to set all fields to None if they don't exist.

Qianh1225:
> Change seems good to go. Would like some more details on what the actual issue was that this caused though.
yes. When the user uses @trigger_on_finish with a string as input, e.g., @trigger_on_finish(flow='fakeflow'), it throws an None type error. The direct reason is the code mismatched "name" and "fq_name."

metaflow/plugins/events_decorator.py
```diff
@@ -597,9 +597,41 @@ def format_deploytime_value(self):
             # Entire trigger is a function (returns either string or dict)
             old_trig = trigger
             if isinstance(trigger, DeployTimeField):
+                # convert the trigger to string or dict
                 trigger = deploy_time_eval(trigger)
+                if is_stringish(trigger):
+                    pass
+                elif isinstance(trigger, dict):
```
talsperre, line 604: This logic needs further refactoring in a follow up PR.

talsperre (dismissed):

metaflow/plugins/events_decorator.py
```diff
@@ -597,9 +597,41 @@ def format_deploytime_value(self):
...
+                    if "project" in trigger:
+                        if is_stringish(trigger["project"]):
+                            result["project"] = trigger["project"]
+                        else:
+                            raise MetaflowException(
+                                "The *project* attribute of the *flow* is not a string"
+                            )
+                    if "project_branch" in trigger:
+                        if is_stringish(trigger["project_branch"]):
+                            result["project_branch"] = trigger["project_branch"]
```
npow, line 626: Can you extract out lines 602-630 into a helper function? The code is identical to lines 397-429. Also on line 425 it's setting result to `branch` instead of `project_branch`. Which one is correct?

npow (changes requested):
Is it supposed to be `branch` or `project_branch`? Do we need to update line 425?

metaflow/plugins/events_decorator.py
```diff
@@ -478,7 +481,7 @@ def flow_init(
                                 )
                         if "project_branch" in flow:
                             if is_stringish(flow["project_branch"]):
-                                result["branch"] = flow["project_branch"]
+                                result["project_branch"] = flow["project_branch"]
```
saikonen, line 484: this change seems to actually fix a bug with the argo events implementation that has gone unnoticed. Currently deploying a flow with
```python
@trigger_on_finish(flow={"name": "ProjectEventsTestFlow", "project": "sa_test_project", "project_branch": "user.saikonen"})
```
does not apply a filter on the project_branch though it should.

with the introduced changes this is correctly being applied. The relevant code is in https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/argo/argo_workflows.py#L641-L646
cc @savingoyal

saikonen: I'm not sure if this is a regression or not. have triggers filtered by project&branch been working in the past with argo?

metaflow/plugins/events_decorator.py
```diff
@@ -597,9 +600,41 @@ def format_deploytime_value(self):
             # Entire trigger is a function (returns either string or dict)
             old_trig = trigger
             if isinstance(trigger, DeployTimeField):
+                # convert the trigger to string or dict
                 trigger = deploy_time_eval(trigger)
+                if is_stringish(trigger):
+                    pass
+                elif isinstance(trigger, dict):
+                    if "name" not in trigger:
```
saikonen, line 608: There is an interesting *fix* that this PR introduces. Previously this incorrect usage was also working for whatever reason:

```python
def flow_name_dict_func(ctx):
    # Is 'flow' correct??? docstring says 'name'
    return {"flow": "DeployTimeTriggerParams"}


@trigger_on_finish(flow=flow_name_dict_func)
class DeployTimeTriggerOnFinishFlow4(FlowSpec):
```

With the changes in this PR, Metaflow correctly raises an exception on the missing `name` in the dict

saikonen: There is an issue with the current PR though when trying out with the correct syntax.

```python
def flow_name_dict_func(ctx):
    # Is 'flow' correct??? docstring says 'name'
    return {"name": "DeployTimeTriggerParams"}
    # return {"name": "DeployTimeTriggerParams", "project": "TEST", "project_branch": "test_branch"}


@trigger_on_finish(flow=flow_name_dict_func)
class DeployTimeTriggerOnFinishFlow4(FlowSpec):
```
fails with `KeyError` on argo workflows due to `event["flow"]` not being set.

saikonen:
superseded by #2218

PR #2217, opened 2025-01-17
[RfR]Reload all related parameters in runner

wangchy27:
After Config is introduced, we also add parameters in `cli_components`, we need to reload these parameters in Runner and Deployer


metaflow/runner/deployer_impl.py
```diff
@@ -63,11 +63,20 @@ def __init__(
 
         from metaflow.parameters import flow_context
 
-        if "metaflow.cli" in sys.modules:
-            # Reload the CLI with an "empty" flow -- this will remove any configuration
-            # options. They are re-added in from_cli (called below).
-            with flow_context(None) as _:
-                importlib.reload(sys.modules["metaflow.cli"])
+        # Reload the CLI with an "empty" flow -- this will remove any configuration
```
romain-intel, line 66: Few nits:
 - change comment to say that it is configuration *AND* parameter options
 - it is `init_cmd` (no s). The other one is correct.

romain-intel (approved):
LGTM. Let's make sure all the tests pass but otherwise apart from the nits, looks good.

romain-intel:
As a side note -- the testing number is weird...

metaflow/runner/deployer_impl.py
```diff
@@ -63,11 +63,20 @@ def __init__(
...
+        # options. They are re-added in from_cli (called below).
+        to_be_reload = [
```
savingoyal, line 68: not - `to_reload`

savingoyal (dismissed):
lgtm pending nits

PR #2218, opened 2025-01-21
Fix trigger on finish bugs, and refactor code

talsperre:


saikonen (dismissed):

metaflow/plugins/events_decorator.py
```diff
@@ -593,9 +486,66 @@ def flow_init(
                 run_objs.append(run_obj)
             current._update_env({"trigger": Trigger.from_runs(run_objs)})
 
+    def _parse_static_triggers(self, flows):
```
savingoyal, line 489: is self needed here?

metaflow/plugins/events_decorator.py
```diff
@@ -519,26 +431,7 @@ def flow_init(
...
-            else:
-                raise MetaflowException(
-                    "Incorrect format for *flow* in *@trigger_on_finish* "
-                    "decorator. Specify either just the *flow_name* or a fully "
-                    "qualified name like *project_name.branch_name.flow_name*."
-                )
-            # TODO: Also sanity check project and branch names
-            if not re.match(r"^[A-Za-z0-9_]+$", trigger["flow"]):
-                raise MetaflowException(
-                    "Invalid flow name *%s* in *@trigger_on_finish* "
```
savingoyal, line 538: we are missing this check in the PR?

talsperre: It's still there in the `_parse_fq_name` function [here](https://github.com/Netflix/metaflow/blob/a2e3f3ddfa47d2910c792d67ea47c990368d0970/metaflow/plugins/events_decorator.py#L562).

metaflow/plugins/events_decorator.py
talsperre: It's not needed - I can make it static.

PR #2219, opened 2025-01-21
Properly handle nargs=-1 for Click API

romain-intel:


metaflow/runner/click_api.py
```diff
@@ -319,7 +333,7 @@ def getattr_wrapper(_self, name):
             defaults,
         ) = extract_all_params(cli_collection)
 
-        def _method(_self, **kwargs):
+        def _method(_self, *args, **kwargs):
```
madhur-ob, line 336: do we need to use the `*args` introduced here somewhere?
This is because we do use `**kwargs` inside `_method_sanity_check` later..

metaflow/runner/click_api.py
```diff
@@ -379,6 +393,9 @@ def execute(self) -> List[str]:
                             else:
                                 components.append("--%s" % k)
                                 components.append(str(i))
+                    elif v is None:
+                        continue  # Skip None values -- they are defaults and converting
+                        # them to string will not be what the user wants
```
madhur-ob, line 398: any quick example for cases where we get `None`?

madhur-ob (approved):

metaflow/runner/click_api.py
madhur-ob: also, what about checking for `v is None` when looping over `args.items()`
this check is only for looping over `options.items()`
do we need it? just in case..

metaflow/runner/click_api.py
romain-intel: Ya, without it it didn't work for functions that have args. I think the signature changes or something.

metaflow/runner/click_api.py
romain-intel: we could but I don't think we would hit it. This case happens basically when a default value of None was passed on the command line. It transformed into a None string.

PR #2220, opened 2025-01-21
Update project doc with new possible options

romain-intel:


saikonen (approved):
looks good, though are any of the options actually 'new' ?

still needs some fix for stubs.

metaflow/plugins/project_decorator.py
```diff
@@ -26,6 +26,24 @@ class ProjectDecorator(FlowDecorator):
         projects that use the same production scheduler. The name may
         contain only lowercase alphanumeric characters and underscores.
 
+    branch : Optional[str]
+        The branch to use. If not specified, the branch is set to
+        `user.<username>` unless `production` is set to `True`. This can
+        also be set on the command line using `--branch` as a top-level option.
+        It is an error to specify `branch` in the decorator and on the command line.
```
savingoyal, line 33: at the moment we don't throw an error - i wasn't even aware that we had started to support `branch` and `production` in the decorator

metaflow/plugins/project_decorator.py
```diff
@@ -26,6 +26,24 @@ class ProjectDecorator(FlowDecorator):
         projects that use the same production scheduler. The name may
         contain only lowercase alphanumeric characters and underscores.
 
+    branch : Optional[str]
+        The branch to use. If not specified, the branch is set to
+        `user.<username>` unless `production` is set to `True`. This can
```
savingoyal, line 31: technically the branch is still <username>, the project branch name is user.<username>

talsperre: True, although not sure if we have explicitly documented `project_branch` anywhere in the user docs. For most intents and purposes, the value that's visible/exposed to the users is the `project_branch`.

metaflow/plugins/project_decorator.py
saikonen: had to doublecheck that this is actually a new feature for the decorator and not something old. This was part of #2200 last week: https://github.com/Netflix/metaflow/pull/2200/files#diff-64f160cf84ca8cd829d2bf85bf48c7276d10b030524fe2631c79bc4d35f2c4f3R93

and current `master` does indeed raise an error if trying to specify branch both through top-level, and in the decorator, so the docstring is correct.

romain-intel: Yes -- new feature due to configs. This allows you to be able to configure everything in the project decorator from a config whereas before you could just set the name.

metaflow/plugins/project_decorator.py
romain-intel: Right. From a user's pov, the stuff in between your project name and the flow name is user.<username> I'm happy to change the wording but thought that it would be nice to document somewhere.

talsperre (dismissed):

talsperre (approved):

PR #2226, opened 2025-01-24
feature: fast initializer for conda environments

saikonen:
adds a feature flag to switch between regular micromamba and a custom binary for bootstrapping of environments when executing remotely.

also adds deferring of micromamba install until the binary location is actually requested.

some benchmarking results with `@pypi`

```
@pypi(python="3.10", packages={"torch": "2.5.1"})

Time taken for fast_setup_environment: 11.00 seconds
Time taken for setup_environment: 28.31 seconds
```

```
@pypi(python="3.10", packages={"matplotlib": "3.9.2"})

Time taken for fast_setup_environment: 5.29 seconds
Time taken for setup_environment: 13.53 seconds
```

```
@pypi(python="3.11", packages={
    "seaborn": "0.12",
    "fastapi": "0.103.1",
    "scikit-learn": "1.6.1",
    }
)

Time taken for fast_setup_environment: 6.02 seconds
Time taken for setup_environment: 10.04 seconds

```


metaflow/plugins/pypi/conda_environment.py
```diff
@@ -421,15 +421,16 @@ def bootstrap_commands(self, step_name, datastore_type):
...
             return [
                 "echo 'Bootstrapping virtual environment...'",
                 # We have to prevent the tracing module from loading,
                 # as the bootstrapping process uses the internal S3 client which would fail to import tracing
                 # due to the required dependencies being bundled into the conda environment,
                 # which is yet to be initialized at this point.
-                'DISABLE_TRACING=True python -m metaflow.plugins.pypi.bootstrap "%s" %s "%s" linux-64'
-                % (self.flow.name, id_, self.datastore_type),
+                'DISABLE_TRACING=True python -m metaflow.plugins.pypi.%s "%s" %s "%s" linux-64'
+                % (bootstrap_module_name, self.flow.name, id_, self.datastore_type),
```
savingoyal, line 433: nit - can we maintain the singular - `metaflow.plugins.pypi.bootstrap` module and change the bootstrapping logic within it? easier to benchmark changes that way.

metaflow/plugins/pypi/fast_bootstrap.py
```diff
@@ -88,6 +58,9 @@ def install_fast_initializer(architecture):
             return fast_initializer_path
 
         # TODO: take architecture into account
+        url = CONDA_FAST_INIT_BIN_URL
```
aochagavia, line 61: With this change, does the TODO above still make sense? Or are we going to handle the architecture at a different level, by setting `CONDA_FAST_INIT_BIN_URL` to an architecture-specific url?

saikonen: Fair point. The client needs to have knowledge of the target architecture either way as it is the one doing the solving for packages. The client also provides the architecture input for the bootstrap script, so we might as well keep the URL defining logic completely client-side.

metaflow/plugins/pypi/conda_environment.py
saikonen: done.

metaflow/metaflow_config.py
```diff
@@ -433,6 +433,12 @@
 # should result in an appreciable speedup in flow environment initialization.
 CONDA_DEPENDENCY_RESOLVER = from_conf("CONDA_DEPENDENCY_RESOLVER", "conda")
 
+# Conda Fast init binary url
+CONDA_FAST_INIT_BIN_URL = from_conf("CONDA_FAST_INIT_BINARY_URL", None)
```
savingoyal, line 437: can you remove this config and introduce the path to actual binary in the `install_fast_initializer` method instead? also, if you can note some latency improvements in the PR, that will be great!

savingoyal:
looks good! can you verify that this works with dev-gcp and dev-azure too before merging?

PR #2233, opened 2025-01-30
Fix an issue where processing MutableStep decos was skipped

romain-intel:
If a flow didn't have a MutableFlow decorator but had MutableStep decorators, the decorators were not processed.


romain-intel:
Duplicate

PR #2242, opened 2025-02-03
[kubernetes] extended resources support

valayDave:
- Support for extended resources outside things like CPU/memory/GPU etc. 
- https://kubernetes.io/docs/tasks/administer-cluster/extended-resource-node/


ob-uk (approved):

metaflow/plugins/kubernetes/kubernetes_job.py
```diff
@@ -85,6 +85,9 @@ def create_job_spec(self):
             self._kwargs["memory"],
             self._kwargs["disk"],
         )
+        extended_resources = self._kwargs.get("extended_resources", {})
+        qos_requests = {**qos_requests, **extended_resources}
+        qos_limits = {**qos_limits, **extended_resources}
```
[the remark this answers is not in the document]
saikonen: this is a valid point. For a proper guard against `None` (which in the current implementation cannot happen though) consider something like
```python
extended_resources = self._kwargs.get("extended_resources") or {}
```

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -167,6 +170,7 @@ class KubernetesDecorator(StepDecorator):
         "executable": None,
         "hostname_resolution_timeout": 10 * 60,
         "qos": KUBERNETES_QOS,
+        "extended_resources": {},
```
saikonen, line 173: the default `{}` instead of `None` here makes the whole thing work without guards, as it ensures that by default an empty dict is passed through to CLI and constructors, which do have a default `None`.

Either keep the default as a dict and update the docstring, or if you change it to a None, introduce the required guards as well.

saikonen (changes requested):

metaflow/plugins/kubernetes/kubernetes_decorator.py
valayDave: Fixed it by updating the docstring. (Also added guard because its python ; Nones can come out of nowhere)

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -2625,6 +2625,10 @@ def _container_templates(self):
                 resources["disk"],
             )
 
+            extended_resources = resources.get("extended_resources", {})
```
valayDave:
```suggestion
            extended_resources = resources.get("extended_resources", {}) or {}
```

saikonen (approved):

PR #2243, opened 2025-02-03
serialize artifacts with pickle protocol if possible

amerberg:
Update the artifact serialization to to save with protocol 4 if available. Currently, artifacts are serialized with protocol 2, unless protocol 2 fails due to the size of the artifact, in which case protocol 4 is used. This PR updates the `TaskDatastore` to use protocol 4 if available, and otherwise attempt protocol 2.

Protocol 4 was introduced in Python 3.4, which was released more than 10 years ago, and Metaflow's `setup.py` script identifies Python 3.6 as the earliest supported version, so  this change should not cause problems for modern users. This leaves artifact loading unchanged.

Using protocol 4 by default has a number of advantages:

1. Smaller size of artifacts < 2GB
2. Reducing the time to serialize large (>2GB) artifacts, by eliminating the doomed protocol 2 attempt

### DIFF
--- a/metaflow/datastore/task_datastore.py
+++ b/metaflow/datastore/task_datastore.py
@@ -253,7 +253,7 @@ class TaskDataStore(object):
 
     @only_if_not_done
     @require_mode("w")
-    def save_artifacts(self, artifacts_iter, force_v4=False, len_hint=0):
+    def save_artifacts(self, artifacts_iter, len_hint=0):
         """
         Saves Metaflow Artifacts (Python objects) to the datastore and stores
         any relevant metadata needed to retrieve them.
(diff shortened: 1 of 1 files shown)


romain-intel (dismissed):
LGTM. I'll let @saikonen take a look as well but I think we can merge this.

amerberg:
> Testing[970] @ [b41f37c](https://github.com/Netflix/metaflow/commit/b41f37c75e87539898156b0170be3beb48bed9a6) had 2 FAILUREs.

Is this different from the tests in the "checks" section, which all appear to have passed? If so, is there a way to see the failures?

romain-intel:
The tests are internal tests we run (since we also use the OSS version). The 2 failures are flaky tests. This is good to go for me.

romain-intel (approved):

saikonen (approved):
should be good to go from what I can tell.

talsperre (approved):

PR #2246, opened 2025-02-04
add extract method to MetaflowCode

savingoyal:


metaflow/client/core.py
```diff
@@ -878,6 +879,53 @@ def tarball(self) -> tarfile.TarFile:
...
+                os.replace(src, dst)  # Atomic move operation
+        ```
+        Returns
+        -------
+        TemporaryDirectory
+            A temporary directory containing the extracted code files.
+            The directory and its contents are automatically deleted when
+            this object is garbage collected.
+        """
+        EXCLUSIONS = [
```
npow, line 912: i'll also be calling this code in https://github.com/Netflix/metaflow/pull/2238 (will need to factor out into a shared helper method).

apparently the exclusion logic should be

```
    members = [
        m
        for m in tar.getmembers()
        if not any(
            (x.endswith("/") and m.name.startswith(x)) or (m.name == x)
            for x in exclusions
        )
    ]
```

talsperre: Can this be made configurable or put in `metaflow_config` instead? We might have some more files at Netflix, that should be excluded.

savingoyal: ack - fixed.

metaflow/client/core.py
```diff
@@ -878,6 +879,71 @@ def tarball(self) -> tarfile.TarFile:
...
+        ```
+        Returns
+        -------
+        TemporaryDirectory
+            A temporary directory containing the extracted code files.
+            The directory and its contents are automatically deleted when
+            this object is garbage collected.
+        """
+        exclusions = [
+            "metaflow/",
```
romain-intel, line 913: I think we have constants for this so we should use that. +1 to factoring it out as well.

savingoyal: why would we need to factor this out? #2238 can directly invoke this method unless i am missing something else?

metaflow/client/core.py
```diff
@@ -878,6 +879,71 @@ def tarball(self) -> tarfile.TarFile:
...
+                for x in exclusions
+            )
+        ]
+
+        tmp = TemporaryDirectory()
+        self.tarball.extractall(tmp.name, members)
+        return tmp
+
+    @property
+    def script(self) -> str:
```
romain-intel, line 933: maybe: `script_name`. That would leave us the ability to return the actual script (whcih we already extract and have) if we want to without name conflcit.

metaflow/client/core.py
savingoyal: do you have an example? ideally it would be nicer to include exclusions neither in the code or the config but the code package itself

talsperre: We have the `conda*.cnd` files in the code package.

savingoyal: is it `conda.cnd` or something else?

talsperre: It's `condav2-1.cnd ` specifically - I don't see the harm in making it configurable with sane defaults like the ones we already have.

savingoyal: i have currently hard coded it in - ideally if we can avoid adding more to configs that would be my preference since it can become quite tricky to keep track of different versions of metaflow with a singular config. a proper solution would be to add the non-user files to `INFO` and then exclude them.

metaflow/client/core.py
romain-intel: Recommend waiting on #2269 which will make this easier.

metaflow/client/core.py
```diff
@@ -878,6 +879,72 @@ def tarball(self) -> tarfile.TarFile:
         """
         return self._tar
 
+    @property
+    def extract(self) -> TemporaryDirectory:
```
romain-intel, line 883: would rename to `extract_user_code` or something like that. We discussed a bit and we could do:
- one method called extract with a flag that determines WHAT (user code, environment, everything)
- `extract_user_code` which would be this and then the tarball extraction would be the other one.

I think we were leaning towards the second solution and that could work but it feels a little less symetric.

I think we also discussed a slightly different UX than the temporary directory. I guess the question is how will users use this.

talsperre: Make it a function instead of a property.

romain-intel: Final comment from me on this:
- ok with name `extract`
- make it a function

metaflow/client/core.py
romain-intel: So, discussion was:
- release now
- later will be made easier by 2269
- we can add a comment about where condav2-1.cnd comes from (but yes, it should be included)

npow (approved):

talsperre (approved):

PR #2248, opened 2025-02-05
feature: add req parsers for configs

saikonen:
moving from https://github.com/romain-intel/metaflow-dep-parsers

adds `req_parser` for `@pypi` and `yml_parser` for `@conda` to enable passing dependencies as config values.

### DIFF
--- a/metaflow/plugins/pypi/parsers.py
+++ b/metaflow/plugins/pypi/parsers.py
@@ -94,6 +94,20 @@ def yml_parser(config_value: str) -> Dict[str, Any]:
     return result
 
 
+def toml_parser(config_value: str) -> Dict[str, Any]:
+    pypi_deps = {}
+    python_version = parse_toml_value(config_value, pypi_deps)
+    result = {}
+
+    if python_version:
+        result["python"] = python_version
+
+    if pypi_deps:
+        result["packages"] = pypi_deps
+
+    return result
+
+
 def parse_req_value(
     file_content: str,
     extra_args: Dict[str, List[str]],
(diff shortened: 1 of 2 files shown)


metaflow/plugins/pypi/parsers.py
```diff
@@ -0,0 +1,260 @@
...
+
+def parse_req_value(
+    file_content: str,
+    extra_args: Dict[str, List[str]],
+    sources: Dict[str, List[str]],
+    deps: Dict[str, str],
+    np_deps: Dict[str, str],
+    sys_deps: Dict[str, str],
+) -> Optional[str]:
+    from packaging.requirements import InvalidRequirement, Requirement
```
saikonen, line 105: only change to the original here, as `packaging` is not available by default in environments, having the import trickle down to the top-level causes issues.

romain-intel:
Let me make sure we properly override these internally and if so, this should be good to go.

metaflow/plugins/pypi/parsers.py
```diff
@@ -0,0 +1,260 @@
...
+# Allows things like:
+# pkg = <= version
+# pkg <= version
+# pkg = version
+# pkg = ==version or pkg = =version
+# In other words, the = is optional but possible
+YML_SPLIT_LINE = re.compile(r"(?:=\s)?(<=|>=|~=|==|<|>|=)")
+
+
+def req_parser(config_value: str) -> Dict[str, Any]:
```
savingoyal, line 15: is it possible for extensions to define and override these parsers? that way this implementation can be decoupled with metaflow-netflixext

savingoyal (changes requested):
added a comment

metaflow/plugins/pypi/parsers.py
romain-intel: There is no special mechanism right now. If this was something in core, the override we do for things like Parameter would work fine but here this is not. I don't think this will actually work right now (looking at it) because the plugin `pypi` doesn't exist internally (it's not loaded). I'm going to have to play around with it a bit. We could of course add a similar plugin category `PARSERS` (like everything else) and resolve it in the plugins. Bringing it back up here though will be a bit trickier.

saikonen:
closing in favour of #2306

PR #2258, opened 2025-02-11
Introduce connection pooling in metadata service

savingoyal:
quick change to shave a second off of time taken to launch first task.

### DIFF
--- a/metaflow/plugins/metadata_providers/service.py
+++ b/metaflow/plugins/metadata_providers/service.py
@@ -71,7 +71,9 @@ class ServiceMetadataProvider(MetadataProvider):
     def compute_info(cls, val):
         v = val.rstrip("/")
         try:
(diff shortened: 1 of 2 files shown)


metaflow/plugins/metadata_providers/service.py
```diff
@@ -39,6 +34,23 @@ def __init__(self, msg, http_code=None, body=None):
 class ServiceMetadataProvider(MetadataProvider):
     TYPE = "service"
 
+    _session = requests.Session()
+    _session.mount(
+        "http://",
+        requests.adapters.HTTPAdapter(
+            pool_connections=20,
```
npow, line 41: how did you determine this number?

savingoyal: anecdotal -adjusting it higher or lower has no further practical impact.

PR #2267, opened 2025-02-12
Support escape init function

wangchy27:
Support override __init__ function in escape hatch


metaflow/runner/click_api.py
```diff
@@ -222,22 +222,26 @@ def extract_flow_class_from_file(flow_file: str) -> FlowSpec:
     if flow_dir not in sys.path:
         sys.path.insert(0, flow_dir)
         path_was_added = True
+    print("path_was_added", path_was_added)
```
romain-intel, line 225: nit: remove prints.

metaflow/plugins/env_escape/stub.py
```diff
@@ -276,9 +276,21 @@ def __call__(cls, *args, **kwargs):
         if len(args) > 0 and id(args[0]) == id(cls.___class_connection___):
             return super(MetaWithConnection, cls).__call__(*args, **kwargs)
         else:
-            return cls.___class_connection___.stub_request(
-                None, OP_INIT, cls.___class_remote_class_name___, *args, **kwargs
-            )
+            if hasattr(cls, "__init_overriden__"):
```
romain-intel, line 279: This covers the overrides on the client side. We need to do something similar on the server side to cover overrides there as well.

romain-intel (commented):
This half looks good. Needs other half :)

metaflow/plugins/env_escape/stub.py
```diff
@@ -381,7 +394,10 @@ def create_class(
             name = name[7:]
             method_type = CLASS_METHOD
         if name in overriden_methods:
-            if method_type == NORMAL_METHOD:
+            if name == "__init__":
+                class_dict["__init_overriden__"] = overriden_methods["__init__"]
```
npow, line 398: if we're doing this, why not go all the way and allow overriding all the dunder methods?

metaflow/plugins/env_escape/stub.py
```diff
@@ -276,9 +276,22 @@ def __call__(cls, *args, **kwargs):
         if len(args) > 0 and id(args[0]) == id(cls.___class_connection___):
             return super(MetaWithConnection, cls).__call__(*args, **kwargs)
         else:
-            return cls.___class_connection___.stub_request(
-                None, OP_INIT, cls.___class_remote_class_name___, *args, **kwargs
-            )
+            if hasattr(cls, "__init_overriden__"):
```
romain-intel, line 279: nit: `__overriden_init__`

metaflow/plugins/env_escape/server.py
```diff
@@ -264,7 +264,6 @@ def __init__(self, config_dir, max_pickle_version):
     def serve(self, path=None, port=None):
         # Open up a connection
         if path is not None:
```
romain-intel, line 266: nit: can leave the comment for debugging but otherwise ok.

metaflow/plugins/env_escape/stub.py
romain-intel: I think we may actually already be doing it. The only reason init was special is because it doesn't go through call/callattr. I think the other dunders should go with NORMAL_METHOD below. One thing though is that we may want to restrict it to some whitelisted ones but I feel that this is not a huge problem since this is a very internal thing.

npow:
@romain-intel what is the other half you are referring to?

romain-intel (approved):

romain-intel:
@npow -- @wangchy27 addressed it -- it was initially not handling the server override side but that has been addressed now. It looks good to me now.

PR #2277, opened 2025-02-19
[WIP] - Support argo retry

dhpikolo:
Seems like argo retry does not work as expected when a workflow is triggered using metaflow retry enabled.

```
python hello_world.py argo-workflows create
python hello_world.py argo-workflows trigger
```

### Steps to reproduce:

```python
import pandas as pd
from metaflow import (
    FlowSpec,
    Parameter,
    card,
    project,
    step,
    retry,
    environment
)

from metaflow import current

ENV_VARS = {
    "FAIL_FLOW": "YES",
}

@project(name="dummy_project")
class HelloWorld(FlowSpec):
    force_error = Parameter("force-error", type=bool, default=False)

    @card
    @step
    def start(self):
        self.next(self.print_df)

    @card
    @environment(vars=ENV_VARS)
    @step
    def print_df(self):
        import os

        if os.environ["FAIL_FLOW"] == "YES":
            raise Exception(f"Fail now, current retry count: {current.retry_count}")    
        else:
            self.new_df = pd.DataFrame(
            {
                "city": [1, 2, 3],
                "country": ["de", "de", "at"],
                "order_id": [5, 6, 7],
                "score": [0.5, 0.6, 0.7],
            }
            )
        # pass this step on retry
        print(f"count is = {current.retry_count}")
        self.next(self.end)
    
    @card  
    @step
    def end(self):
        print(f"the new df is: {self.new_df}")

if __name__ == "__main__":
    HelloWorld()

```

- Argo workflow with metaflow retry enabled and fails in all retries. Due to error induced via envar

Image

- Node `print_df` passes after a fix in the backend [or updated the workflow with right env-var value], and then hit  `argo retry --node` button.

- `end` step fails with a missing artifact, `new_df` is a artifiact from the parent step, which was created when the parent node was successful on retry.

```
    print(f"the new df is: {self.new_df}")
                            ^^^^^^^^^^^
  File "/app/metaflow/metaflow/flowspec.py", line 250, in __getattr__
    raise AttributeError("Flow %s has no attribute '%s'" % (self.name, name))
AttributeError: Flow HelloWorld has no attribute 'new_df'
```
Image

**Reason:**

The s3 artifacts in the `end` step did not contain `new_df` pointers in `data.json`, this is essentially due to the retried step `print_df` overwrote the `attempt: 0` artifacts. [Look at the timestamps]

Image

## Desired Behaviour

Ideally, hitting argo retry button should create a new attempt rather than overwriting the old attempt.

[Metaflow ChatRoom Discussion](https://outerbounds-community.slack.com/archives/C020U025QJK/p1737483912784969)

## Solution Proposal

Infer the `retry_count` from flow datastore [example: s3]. The `retry_count` will be the `latest_done _attempt number + 1`

### DIFF
--- a/metaflow/cli_components/step_cmd.py
+++ b/metaflow/cli_components/step_cmd.py
@@ -155,6 +155,16 @@ def step(
         ctx.obj.monitor,
         ubf_context,
     )
+
+    latest_done_attempt = task.flow_datastore.get_latest_done_attempt(
+        run_id=run_id, step_name=step_name, task_id=task_id
+    )
+    if latest_done_attempt:
+        retry_count = latest_done_attempt + 1
+    # Not sure what are the side effects to this.
+    if retry_count >= max_user_code_retries:
+        max_user_code_retries = retry_count
+
     if clone_only:
         task.clone_only(
             step_name,
--- a/metaflow/datastore/flow_datastore.py
+++ b/metaflow/datastore/flow_datastore.py
@@ -1,5 +1,6 @@
 import itertools
 import json
+from typing import Optional
 
 from .. import metaflow_config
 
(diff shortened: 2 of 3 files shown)


dhpikolo:
A known limitation for this solution would be:

- Retrying more than `6` times (or in current case, hitting argo retry button twice) times results in metaflow exception

Screenshot 2025-02-14 at 14 18 23


We are planning to address this limitation by making MAX_ATTEMPTS configurable via envvar.

- https://github.com/dhpikolo/metaflow/pull/2

PR #2279, opened 2025-02-19
Allow `MAX_ATTEMPTS` configurable

dhpikolo:


dhpikolo:
I verified this change working for following use cases, with one of the step using `@retry(times=10)` and setting `METAFLOW_MAX_ATTEMPTS` envvar to `12`

- `python flow.py run`
- `python flow.py run --with kubernetes`
- `python flow.py argo-workflows create`

romain-intel:
I'll take a look a bit closer but one thing I can say already is that we probably want to min it with what is already there because otherwise if it is decreased, the client may misbehave. Related to that, I don't know if we can dis-associate it totally from the run in the client because again, otherwise, the client may misbehave.

PR #2290, opened 2025-02-22
Refactor trigger decorator and deploy time evaluation logic

talsperre:
Refactored `@trigger` logic and simplified the deploy time evaluation logic of `@trigger` decorator. The current implementation is buggy and fails when running `trigger4.py` in this gist [here](https://gist.github.com/talsperre/accd7530d9b75bddbaf00f086a1c6b58). 

For testing, refer to all the flows in this gist [here](https://gist.github.com/talsperre/accd7530d9b75bddbaf00f086a1c6b58). We can add few more tests as well if needed.

### DIFF
--- a/metaflow/plugins/events_decorator.py
+++ b/metaflow/plugins/events_decorator.py
@@ -70,7 +70,26 @@ class TriggerDecorator(FlowDecorator):
         "options": {},
     }
 
-    def process_event_name(self, event):
+    def process_event(self, event):
+        """
+        Process a single event and return a dictionary if static trigger and a function
+        if deploy-time trigger.
+
+        Parameters
+        ----------
+        event : Union[str, Dict[str, Any], Callable]
+            Event to process
+
+        Returns
+        -------
+        Union[Dict[str, Union[str, Callable]], Callable]
+            Processed event
+
+        Raises
+        ------
+        MetaflowException
+            If the event is not in the correct format
+        """
         if is_stringish(event):
             return {"name": str(event)}
         elif isinstance(event, dict):
@@ -82,12 +101,26 @@ class TriggerDecorator(FlowDecorator):
                 event["name"], DeployTimeField
             ):
                 event["name"] = DeployTimeField(
-                    "event_name", str, None, event["name"], False
+                    "event_name",
+                    str,
+                    None,
+                    event["name"],
+                    False,
+                    print_representation=str(event["name"]),
                 )
-            event["parameters"] = self.process_parameters(event.get("parameters", {}))
+            event["parameters"] = self.process_parameters(
+                event.get("parameters", {}), event["name"]
+            )
             return event
         elif callable(event) and not isinstance(event, DeployTimeField):
-            return DeployTimeField("event", [str, dict], None, event, False)
+            return DeployTimeField(
+                "event",
+                [str, dict],
+                None,
+                event,
+                False,
+                print_representation=str(event),
+            )
         else:
             raise MetaflowException(
                 "Incorrect format for *event* attribute in *@trigger* decorator. "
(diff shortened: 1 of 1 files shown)


metaflow/plugins/events_decorator.py
```diff
@@ -96,47 +129,69 @@ def process_event_name(self, event):
...
-                        mapping[1], DeployTimeField
-                    ):
-                        mapping[1] = DeployTimeField(
-                            "parameter_val", str, None, mapping[1], False
+                elif isinstance(mapping, tuple) and len(mapping) == 2:
+                    # (param_name, field_name)
+                    param_name, field_name = mapping
+                    if not is_stringish(param_name) or not is_stringish(field_name):
+                        raise MetaflowException(
+                            f"The *parameters* attribute for event {event_name} is invalid. "
```
romain-intel, line 166: this error message does not seem to properly reflect the condition here. You are actually checking the the content of the tuple are also strings. That too, this is a list of strings or tuples.

metaflow/plugins/events_decorator.py
```diff
@@ -96,47 +129,69 @@ def process_event_name(self, event):
...
                     raise MetaflowException(
                         "The *parameters* attribute for event is invalid. "
                         "It should be a list/tuple of strings and lists/tuples of size 2"
                     )
-        elif callable(parameters) and not isinstance(parameters, DeployTimeField):
-            return DeployTimeField(
-                "parameters", [list, dict, tuple], None, parameters, False
-            )
         elif isinstance(parameters, dict):
+            # {Union[func, param_name]: Union[func, param_name]}
```
romain-intel: This also seems like a wrong comment based on the error condition below.

metaflow/plugins/events_decorator.py
```diff
@@ -96,47 +129,69 @@ def process_event_name(self, event):
...
+        ----------
+        Parameters : Union[callable, List[Union[str, Tuple[str, str]]], Dict[str, str]]
+            Parameters to process
+
+        event_name : Union[str, callable]
+            Name of the event
+
+        Returns
+        -------
+        Union[Dict[Union[str, callable], Union[str, callable]], Callable]
```
romain-intel: return type seems too permissive. You are either returning a callable or a Dict[str, str]

romain-intel (commented):
A few minor comments. I haven't tested with the test flows yet but will look.

metaflow/plugins/events_decorator.py
```diff
@@ -82,12 +101,26 @@ def process_event_name(self, event):
...
             return event
         elif callable(event) and not isinstance(event, DeployTimeField):
-            return DeployTimeField("event", [str, dict], None, event, False)
+            return DeployTimeField(
+                "event",
+                [str, dict],
+                None,
+                event,
+                False,
+                print_representation=str(event),
```
saikonen, line 122: is changing to explicit `print_representation` necessary? Where is the effect of this visible?

metaflow/plugins/events_decorator.py
talsperre: Addressed in latest commit.

metaflow/plugins/events_decorator.py
talsperre: Addressed in new commit

metaflow/plugins/events_decorator.py
talsperre: I just preferred being explicit here - we do implicitly set it to `str(obj)` in the constructor of `DeployTimeField`.

metaflow/plugins/events_decorator.py
```diff
@@ -96,47 +129,67 @@ def process_event_name(self, event):
...
+            Processed parameters
+
+        Raises
+        ------
+        MetaflowException
+            If the parameters are not in the correct format
+        """
         new_param_values = {}
-        if isinstance(parameters, (list, tuple)):
+        if isinstance(parameters, list):
```
romain-intel, line 156: why removing tuple. It probably doesn't matter but just curious.

romain-intel (approved):
Minor nit but definitely looks like an easier read :)

romain-intel:
I'm not sure why the dev test breaks. It doesn't even eem to run anything. @mt-ob , any ideas?

metaflow/plugins/events_decorator.py
talsperre: To keep it consistent with the docstrings.

PR #2306, opened 2025-02-26
@pypi/@conda parsers ...

savingoyal:
... for requirements.txt, conda.yaml, pyproject.toml

```
from metaflow import Config, pypi, FlowSpec, requirements_txt_parser, step

class MyFlow(FlowSpec):
    reqs = Config("reqs", default="requirements.txt", parser=requirements_txt_parser)

    @pypi(**reqs)
    @step
    def start(self):
        self.next(self.end)

    @step
    def end(self):
        pass

if __name__ == "__main__":
    MyFlow()
```


metaflow/plugins/pypi/parsers.py
```diff
@@ -0,0 +1,253 @@
...
+            continue
+
+        line = inline_comment_pattern.sub("", line).strip()
+        if not line:
+            continue
+
+        try:
+            req = Requirement(line)
+        except InvalidRequirement:
+            raise ValueError(f"Not a valid PEP 508 requirement: '{line}'")
```
saikonen, line 54: stacked PR Introduces a MetaflowException type exception for these instead, in order to get cleaner formatting.

Is there any specific reason to stick with ValueError? from what I can tell the exception bubbles all the way and is not caught, leading to a full stack trace in stdout

metaflow/plugins/pypi/parsers.py
```diff
@@ -0,0 +1,253 @@
...
+    for line in content.splitlines():
+        line = line.strip()
+
+        if not line or line.startswith("#"):
+            continue
+
+        line = inline_comment_pattern.sub("", line).strip()
+        if not line:
+            continue
+
```
saikonen, line 50: the previous PR had more specific line handling for requirements.txt content, specifically errors for options:
https://pip.pypa.io/en/stable/reference/requirements-file-format/#global-options

should we reintroduce these, or do we want to keep this implementation as simple as possible, and simply error out on 'unparseable' lines in the file? right now the parser throws an exception if the requirements.txt contains any options, be it global or per-requirement

PR #2307, opened 2025-02-26
more realtime log fetching

savingoyal:
log lines (the first few) should now appear a lot more quicker on the terminal


npow (approved):
lgtm

PR #2308, opened 2025-02-27
parser additions

saikonen:
- change errors to a MetaflowException type for better formatting
- support Rye generated lockfiles with requirements_txt_parser
- add unit tests for parsers
- fix yml parser handling of whitespace


metaflow/plugins/pypi/parsers.py
```diff
@@ -28,7 +33,7 @@ def requirements_txt_parser(content: str):
 
     Raises
     ------
-    ValueError
+     ParserValueError
```
savingoyal, line 36: any reason to have a custom error? ideally it's nice to just rely on a built-in error

savingoyal (changes requested):

PR #2309, opened 2025-02-28
Correct memory units from MB to MiB

dpoznik:
The docs currently suggest that the [`@batch`](https://docs.metaflow.org/api/step-decorators/batch) and [`@resources`](https://docs.metaflow.org/api/step-decorators/resources) memory units are MB, but I suspect they may be MiB.

For AWS Batch computes, I believe these parameters are passed directly to `ResourceRequirement`, like so:
```
      "ResourceRequirements": [
        {
          "Value": "1",
          "Type": "VCPU"
        },
        {
          "Value": "4096",
          "Type": "MEMORY"
        }
      ],
```
and the `ResourceRequirement` [docs](https://docs.aws.amazon.com/batch/latest/APIReference/API_ResourceRequirement.html) indicate memory values are in MiB.

Furthermore, the default of 4096 makes sense to me if the units are MiB, since 4096 MiB = 4 GiB. If MB, it's hard to interpret.

This PR originally proposed changes to four of seven instances of "(in MB)".

In response to reviewer feedback, it also:
- Includes the remaining three, Kubernetes-specific, instances of "(in MB)".
- Changes the Kubernetes memory units and disk units, and the Argo memory units, to `Mi`, for consistency. These are behavior changes.

### DIFF
--- a/R/R/decorators-aws.R
+++ b/R/R/decorators-aws.R
@@ -14,7 +14,7 @@
 #'
 #' @param cpu Integer number of CPUs required for this step. Defaults to `1`.
 #' @param gpu Integer number of GPUs required for this step. Defaults to `0`.
-#' @param memory Integer memory size (in MB) required for this step. Defaults to
+#' @param memory Integer memory size (in MiB) required for this step. Defaults to
 #'   `4096`.
 #' @param image Character. Specifies the image to use when launching on AWS
 #'   Batch. If not specified, an appropriate
@@ -48,9 +48,9 @@
 #' @export
 #'
 #' @examples \dontrun{
-#' # This example will generate a large random matrix which takes up roughly
-#' # 48GB of memory, and sums the entries. The `batch` decorator forces this
-#' # step to run in an environment with 60000MB of memory.
+#' # This example will generate a large random matrix that takes up roughly
+#' # 48 GiB of memory and sum the entries. The `batch` decorator forces this
+#' # step to run in an environment with 60,000 MiB of memory.
 #'
 #' start <- function(self) {
 #'   big_matrix <- matrix(rexp(80000*80000), 80000)
--- a/metaflow/plugins/kubernetes/kube_utils.py
+++ b/metaflow/plugins/kubernetes/kube_utils.py
@@ -43,8 +43,8 @@ def qos_requests_and_limits(qos: str, cpu: int, memory: int, storage: int):
         # Guaranteed - has both cpu/memory limits. requests not required, as these will be inferred.
         qos_limits = {
             "cpu": str(cpu),
-            "memory": "%sM" % str(memory),
-            "ephemeral-storage": "%sM" % str(storage),
+            "memory": "%sMi" % str(memory),
+            "ephemeral-storage": "%sMi" % str(storage),
         }
         # NOTE: Even though Kubernetes will produce matching requests for the specified limits, this happens late in the lifecycle.
         # We specify them explicitly here to make some K8S tooling happy, in case they rely on .resources.requests being present at time of submitting the job.
@@ -53,8 +53,8 @@ def qos_requests_and_limits(qos: str, cpu: int, memory: int, storage: int):
         # Burstable - not Guaranteed, and has a memory/cpu limit or request
         qos_requests = {
             "cpu": str(cpu),
-            "memory": "%sM" % str(memory),
-            "ephemeral-storage": "%sM" % str(storage),
+            "memory": "%sMi" % str(memory),
+            "ephemeral-storage": "%sMi" % str(storage),
         }
     # TODO: Add support for BestEffort once there is a use case for it.
     # BestEffort - no limit or requests for cpu/memory
--- a/metaflow/plugins/kubernetes/kubernetes_decorator.py
+++ b/metaflow/plugins/kubernetes/kubernetes_decorator.py
@@ -61,11 +61,11 @@ class KubernetesDecorator(StepDecorator):
         Number of CPUs required for this step. If `@resources` is
         also present, the maximum value from all decorators is used.
     memory : int, default 4096
-        Memory size (in MB) required for this step. If
+        Memory size (in MiB) required for this step. If
         `@resources` is also present, the maximum value from all decorators is
         used.
     disk : int, default 10240
-        Disk size (in MB) required for this step. If
+        Disk size (in MiB) required for this step. If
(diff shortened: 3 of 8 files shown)


metaflow/plugins/resources_decorator.py
```diff
@@ -28,7 +28,7 @@ class ResourcesDecorator(StepDecorator):
     disk : int, optional, default None
         Disk size (in MB) required for this step. Only applies on Kubernetes.
     memory : int, default 4096
-        Memory size (in MB) required for this step.
+        Memory size (in MiB) required for this step.
```
savingoyal, line 31: given that these values are used in @kubernetes as well - would be good to ensure it's the case there too. also what about disk?

dpoznik: I've not used Kubernetes before, so I wasn't sure about the three instances I left unchanged:
- `@kubernetes` `memory`
- `@kubernetes` `disk`
- `@resources` `disk` (used only for Kubernetes)

I am pretty sure this one (`@resources` `memory`) should be MiB for AWS. That may also be the case for Kubernetes, but I don't know.

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -59,11 +59,11 @@ class KubernetesDecorator(StepDecorator):
         Number of CPUs required for this step. If `@resources` is
         also present, the maximum value from all decorators is used.
     memory : int, default 4096
-        Memory size (in MB) required for this step. If
+        Memory size (in MiB) required for this step. If
         `@resources` is also present, the maximum value from all decorators is
         used.
     disk : int, default 10240
-        Disk size (in MB) required for this step. If
+        Disk size (in MiB) required for this step. If
```
saikonen, line 66: I had a peek at the units, and it seems that memory/disk values for kubernetes are treated as MB so the earlier docstring for these was correct.

we could hold off on changing the kubernetes units in the docs for now, as this would require also changing the implementation to match.

saikonen: alternatively changing the units from `M` to `Mi` in `kube_utils.py#qos_requests_and_limits` should be the only additional change required, as this is used both by the kubernetes and argo workflows implementations.

dpoznik: > alternatively changing the units from `M` to `Mi` in `kube_utils.py#qos_requests_and_limits` should be the only additional change required, as this is used both by the kubernetes and argo workflows implementations.

Cool, thanks for taking a look at the Kubernetes units. Since the defaults are multiples of 1024, MiB seems to be the original intent. So I think it makes sense to update the implementation, especially if doing so is as easy as you've described. Accordingly, I've pushed up a new commit: bfaa672a. But, if you prefer, I could revert bfaa672a and 4fc51071 and update the `@resources` docstring to indicate that the memory units are MiB for AWS Batch and MB for Kubernetes. Just let me know your preference. Thanks!

dpoznik: Fwiw, the Kubernetes "Assign Memory Resources to Containers and Pods" [example](https://kubernetes.io/docs/tasks/configure-pod-container/assign-memory-resource/) uses MiB, potentially indicating that MiB is more standard / expected there. Maybe.

saikonen (dismissed):

dpoznik:
> As a heads up I'll be looking into some other inconsistencies of `@resources` as well next week that have been on the backlog for a while now, and will hold off on merging this PR for a bit in case that prompts more changes.
> 
> For now this PR seems good to go as is so no further action required 👌

OK, great. Thanks for the update!

dpoznik:
> As a heads up I'll be looking into some other inconsistencies of `@resources` as well next week that have been on the backlog for a while now, and will hold off on merging this PR for a bit in case that prompts more changes.
> 
> For now this PR seems good to go as is so no further action required 👌

@saikonen, I just wanted to check in on this PR. Is it still of interest, or should I go ahead and close it?
I removed all the import-sorting fixes for a cleaner diff.

saikonen:
> @saikonen, I just wanted to check in on this PR. Is it still of interest, or should I go ahead and close it? I removed all the import-sorting fixes for a cleaner diff.

Sorry that this has dragged on, there have been a lot of other features fighting for my attention.

The fix is still very much of interest and this can be taken up as part of the `@resources` realigning that I mentioned previously. Unfortunately its not a high priority item at this moment, and there was some hesitation regarding the changes to the resource requests units on the kubernetes side as this will have an effect on increasing the memory footprint of running tasks. As such I can not make any promises on a timeline to get this in.

For the time being, I would say lets keep the PR open as still relevant.

dpoznik:
> For the time being, I would say lets keep the PR open as still relevant.

Cool, will do. Thanks for the update!

R/R/decorators-aws.R
```diff
@@ -49,8 +49,8 @@
 #' 
 #' @examples \dontrun{
 #' # This example will generate a large random matrix which takes up roughly 
-#' # 48GB of memory, and sums the entries. The `batch` decorator forces this
-#' # step to run in an environment with 60000MB of memory.
+#' # 49 GiB of memory, and sums the entries. The `batch` decorator forces this
```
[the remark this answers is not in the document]
dpoznik: Whoops; typo. Fixed.

metaflow/plugins/kubernetes/kube_utils.py
```diff
@@ -43,8 +43,8 @@ def qos_requests_and_limits(qos: str, cpu: int, memory: int, storage: int):
         # Guaranteed - has both cpu/memory limits. requests not required, as these will be inferred.
         qos_limits = {
             "cpu": str(cpu),
-            "memory": "%sM" % str(memory),
-            "ephemeral-storage": "%sM" % str(storage),
+            "memory": "%sMi" % str(memory),
+            "ephemeral-storage": "%sMi" % str(storage),
```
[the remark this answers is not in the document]
dpoznik: OK, I have made this change.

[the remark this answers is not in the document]
dpoznik: Updated PR description accordingly

dpoznik:
I'll go ahead and close this for now, to clear it from my inbox. I'd be more than happy to reopen if it's ever of interest.

PR #2310, opened 2025-02-28
adding triggered_by_user annotation to workflow

iamsgarg-ob:
## What is the change in this PR? 
* Adding the metaflow/triggered_by_user annotation on the workflow when a workflowTemplate is triggered by the metaflow cli. This helps in better tracking of who triggered the workflow. 

## How is this tested?
Screenshot 2025-02-27 at 10 37 58 PM


metaflow/plugins/argo/argo_client.py
```diff
@@ -256,12 +257,27 @@ def _patch_workflow(self, name, body):
...
             )
 
+    def _get_identity_parts(self, identity):
+        id_parts = identity.split(":")
+        parts_size = len(id_parts)
+        usertype = id_parts[0] if parts_size > 0 else "unknown"
+        username = id_parts[1] if parts_size > 1 else "unknown"
+        return usertype, username
+
     def trigger_workflow_template(self, name, parameters={}):
```
savingoyal, line 267: can we pass user_identity to this function so that this entire class is self contained and has no coupling to the rest of the codebase?

metaflow/plugins/argo/argo_client.py
```diff
@@ -256,12 +257,27 @@ def _patch_workflow(self, name, body):
...
         client = self._client.get()
+        usertype, username = self._get_identity_parts(resolve_identity())
         body = {
             "apiVersion": "argoproj.io/v1alpha1",
             "kind": "Workflow",
-            "metadata": {"generateName": name + "-"},
+            "metadata": {
+                "generateName": name + "-",
+                "annotations": {
+                    "metaflow/triggered_by_user": json.dumps(
```
savingoyal, line 276: why not just set metaflow/triggered_by_user directly to resolved_identity()

savingoyal (changes requested):

metaflow/plugins/argo/argo_client.py
iamsgarg-ob: 2 reasons:

1. I was trying to follow the same convention as `triggered_ by` which is a k-v pair. 
2. It is also a little more forward looking if we want to add more parameters to triggered_by_user.

savingoyal: the `_user` in annotation key refers to the key in the payload already right? it might be a bit awkward in the future to have `trigger_by_user` as the annotation key, but then the actual thing that triggers is not a user but is still referenced inside this annotation.

iamsgarg-ob: resolved offline.

metaflow/plugins/argo/argo_client.py
iamsgarg-ob: done.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -315,8 +316,16 @@ def trigger(cls, name, parameters=None):
                     "Workflows before proceeding." % name
                 )
         try:
+            id_parts = resolve_identity().split(":")
+            parts_size = len(id_parts)
+            usertype = id_parts[0] if parts_size > 0 else "unknown"
+            username = id_parts[1] if parts_size > 1 else "unknown"
+
```
saikonen, line 323: is there any concern with using `resolve_identity` here? from what I can tell, it will always produce a two-part as this is just a joined tuple in the end, so the length checks are unnecessary.

more importantly, this does not produce a user in all cases, specifically when using a prod token:
```
METAFLOW_PRODUCTION_TOKEN=testingtoken python testflow.py argo-workflows trigger
```
results in
```
"metaflow/triggered_by_user": "{\"type\": \"production\", \"name\": \"testingtoken\"}",
```
which does not seem informative, nor match the expectation with the annotation name.

Is it okay for the prod token to end up in the `triggered_by_user` annotations? it is by no means secret info, as its present in the regular annotations already.

metaflow/plugins/argo/argo_client.py
```diff
@@ -256,12 +256,19 @@ def _patch_workflow(self, name, body):
...
+    def trigger_workflow_template(self, name, usertype, username, parameters={}):
         client = self._client.get()
         body = {
             "apiVersion": "argoproj.io/v1alpha1",
             "kind": "Workflow",
-            "metadata": {"generateName": name + "-"},
+            "metadata": {
+                "generateName": name + "-",
+                "annotations": {
+                    "metaflow/triggered_by_user": json.dumps(
```
saikonen, line 267: flows triggered by events already have a `metaflow/triggered_by`, which is a list of dicts with `type/name`. Could this not have been the same annotation, or is there a reason to keep it separate?

Collisions should not be a concern at least for now, as a flow can not be triggered both by events, and a user at the same time.

metaflow/plugins/argo/argo_workflows.py
iamsgarg-ob: yes I believe that is acceptable. In fact it is entirely likely that will happen . If that does happen it will be at least be known that the prod token was used to kick off the run. Today we just don't know who/what triggered the run.

metaflow/plugins/argo/argo_client.py
iamsgarg-ob: yes I spoke with @savingoyal and he recommended we keep it separate.

saikonen (approved):

PR #2318, opened 2025-03-01
modernize readme

savingoyal:
last time we edited this was when we open-sourced the project. feedback welcome!


savingoyal:
resources section in metaflow docs is woefully behind. i will push a separate commit on metaflow-docs

PR #2322, opened 2025-03-03
wip: better pip crossplatform resolves

saikonen:
⚠️ Proof-of-Concept. Do not merge ⚠️ 

 for correctly resolving transient dependencies for cross-platform package resolves, where dependencies are conditional on platform/os such as

```
"nvidia-cuda-runtime-cu12 (==12.4.127) ; platform_system == \"Linux\" and platform_machine == \"x86_64\"",
```

This intends to solve an issue with packages such as the default `torch` where resolving on a mac, but deploying onto a linux-64 machine will result in a broken environment due to the conditional dependencies not getting bundled up.

Caveats:
`pip` has deprecated script wrappers and is pushing for module calls instead. This will make patching the sys values harder, and the current setup is bound to break in the future.

### DIFF
--- a/metaflow/plugins/pypi/conda_environment.py
+++ b/metaflow/plugins/pypi/conda_environment.py
@@ -55,7 +55,7 @@ class CondaEnvironment(MetaflowEnvironment):
         # Initialize necessary virtual environments for all Metaflow tasks.
         # Use Micromamba for solving conda packages and Pip for solving pypi packages.
         from .micromamba import Micromamba
-        from .pip import Pip
+        from .pip_resolver import Pip
 
         print_lock = threading.Lock()
 
--- a/metaflow/plugins/pypi/pip_patcher.py
+++ b/metaflow/plugins/pypi/pip_patcher.py
@@ -1,0 +1,15 @@
+from unittest.mock import patch
+import os
+import sys
+
+
+@patch("platform.machine", lambda: os.environ.get("PIP_PATCH_MACHINE", "x86_64"))
+@patch("platform.system", lambda: os.environ.get("PIP_PATCH_SYSTEM", "Linux"))
+def _main(args):
+    from pip import main
+
+    main(args)
+
+
+if __name__ == "__main__":
+    _main(sys.argv[1:])
--- a/metaflow/plugins/pypi/pip_patcher.py
+++ b/metaflow/plugins/pypi/pip_patcher.py
@@ -13,6 +13,7 @@ import sys
     "platform.system", lambda: os.environ.get("PIP_PATCH_SYSTEM", os.uname().sysname)
 )
 def _main(args):
+    # TODO: Pip has deprecated using script wrappers for the cli. this will break in the future, and make patching the sys internals much harder
     from pip import main
 
     exitcode = main(args)
--- a/metaflow/plugins/pypi/pip_patcher.py
+++ b/metaflow/plugins/pypi/pip_patcher.py
@@ -15,7 +15,8 @@ import sys
 def _main(args):
     from pip import main
 
-    main(args)
+    exitcode = main(args)
+    sys.exit(exitcode)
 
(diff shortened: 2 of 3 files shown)


saikonen:
note that for testing this, both `.pip/metadata` and `conda.manifest` local files for the involved flow need to most likely be cleared

metaflow/plugins/pypi/pip_patcher/sitecustomize.py
```diff
@@ -0,0 +1,10 @@
+import os
+
+
+# Because Pip does not offer a direct way to set target platform_system and platform_machine values for resolving packages transitive dependencies, we need to instead
+# manually patch the correct target architecture values for pip to be able to resolve the whole dependency tree successfully.
+# This is necessary for packages that have conditional dependencies dependent on machine/system, e.g. Torch
+import platform
+
+platform.system = lambda: os.environ.get("PIP_PATCH_SYSTEM", os.uname().sysname)
+platform.machine = lambda: os.environ.get("PIP_PATCH_MACHINE", os.uname().machine)
```
romain-intel, line 10: I am not sure this is sufficient in all cases. Here is the default_environment that is used: 
```
    return {
        "implementation_name": implementation_name,
        "implementation_version": iver,
        "os_name": os.name,
        "platform_machine": platform.machine(),
        "platform_release": platform.release(),
        "platform_system": platform.system(),
        "platform_version": platform.version(),
        "python_full_version": platform.python_version(),
        "platform_python_implementation": platform.python_implementation(),
        "python_version": ".".join(platform.python_version_tuple()[:2]),
        "sys_platform": sys.platform,
    }
```
I think we need to override:
 - sys.platform
 - platform.version()

as well. The others relate to posix versus non posix (and I think we are always posix) and cpython or not (and I think we also only do cpython)

romain-intel: and platform.release()

romain-intel: Not tested yet (working on it) but maybe something like this:
```
import json
import os
import platform
import sys

_override_mappings = {
    "platform_machine": platform.machine,
    "platform_release": platform.release,
    "platform_system": platform.system,
    "platform_version": platform.version,
    "sys_platform": sys.platform,
}

overrides = os.environ.get("PIP_CUSTOMIZE_OVERRIDES")
if overrides:
    overrides_dict = json.loads(overrides)
    for key, override_value in overrides_dict.items():
        if key in _override_mappings:
            if callable(_override_mappings[key]):
                _override_mappings[key] = lambda _v=override_value: _v
            else:
                _override_mappings[key] = override_value
```

romain-intel: OK, that didn't work and then I realized that if I override `sys.platform`, bad things happen. I still think we need to override it to handle as many markers as possible.

I ended up going with the other approach. Something like this: 

```
import json
import os
import pip._vendor.packaging.markers as _markers


_overrides = os.environ.get("PIP_CUSTOMIZE_OVERRIDES")
if _overrides:
    _orig_default_environment = _markers.default_environment
    _overrides_dict = json.loads(_overrides)

    def _wrap_default_environment():
        result = _orig_default_environment()
        result.update(_overrides_dict)
        return result

    _markers.default_environment = _wrap_default_environment
```

and that seems to work. Just FYI.

saikonen: opened https://github.com/Netflix/metaflow/pull/2675 to address this, along with resolving an issue cross-platform resolving on a mac when using `metaflow-dev`

what are you using as a source of info for `platform.version`/`platform.release`/`sys.platform` when it comes to the target system? I'm not sure if any/all of these are available currently at least on the OSS side

PR #2329, opened 2025-03-03
[RFR] S3 bugfixes

npow:
This PR fixes several issues which caused s3op to be stuck:
1. Need to call `queue.cancel_join_thread()` so that the workers can exit without flushing the queue, otherwise there is a deadlock
2. Catch the correct exceptions in download/upload (they don't actually raise `ClientError`)
3. Handle `InternalError`
4. Handle `SSLError`
5. Optimistically assume all other unhandled exceptions are transient

Additional improvements:
1. Added exponential backoff to `jitter_sleep()`
2. Set default retry config in s3op.py to match that in aws_client.py
3. Fix a bug where the retry setting was not being applied properly if config was missing
4. Fail early on fatal errors
5. Don't restart from scratch when there's no progress


metaflow/plugins/aws/aws_client.py
```diff
@@ -40,7 +41,8 @@ def get_client(
             # Use the adaptive retry strategy by default -- do not set anything if
```
savingoyal, line 41: this comment needs to be moved too

metaflow/plugins/datatools/s3/s3op.py
```diff
@@ -322,22 +328,59 @@ def op_info(url):
...
+                            continue
         except:
             traceback.print_exc()
+            result_file.flush()
             sys.exit(ERROR_WORKER_EXCEPTION)
 
 
+def convert_to_client_error(e):
+    match = re.search(
+        r"An error occurred \((\w+)\) when calling the (\w+) operation.*: (.+)", str(e)
```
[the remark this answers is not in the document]
npow: yes: https://github.com/boto/botocore/blob/68ca78f3097906c9231840a49931ef4382c41eea/botocore/exceptions.py#L521

metaflow/plugins/datatools/s3/s3op.py
```diff
@@ -322,22 +328,59 @@ def op_info(url):
...
+    )
+    assert match, "Failed to parse error message"
```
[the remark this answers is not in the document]
npow: mostly want to know if the parse failed. i can change it to reraise the original exception

metaflow/plugins/datatools/s3/s3.py
```diff
@@ -1712,38 +1723,16 @@ def try_s3_op(last_ok_count, pending_retries, out_lines, inject_failures):
                         _update_out_lines(out_lines, ok_lines, resize=loop_count == 0)
                         return 0, 0, inject_failures, err_out
 
-        while retry_count <= S3_RETRY_COUNT:
+        while transient_retry_count <= S3_TRANSIENT_RETRY_COUNT:
```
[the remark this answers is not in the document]
npow: added to the notes on "what this function does"

npow:
> any testing scripts for me to use to reproduce the issues this PR is trying to address? it's okay if it's clunky etc. also it would be good to run the battery of s3 tests for these changes. are you able to set up the s3 tests against your env?

We run the s3 tests on our end for every commit and have been able to reproduce the issue there. I also have a flow that reproduces the issue by reading and writing data to/from S3 using hundreds of workers, but the data isn't public.

npow:
> Testing for OSS PR [1011] @ commit [00b184d](https://github.com/Netflix/metaflow/commit/00b184d84137a2e1c45b47bf3b1d2e436495611b) had 12 FAILUREs.

Failures are from the card tests and the tag mutation test, which have been flaky.

metaflow/plugins/datatools/s3/s3op.py
```diff
@@ -322,22 +328,60 @@ def op_info(url):
...
+                            continue
         except:
             traceback.print_exc()
+            result_file.flush()
             sys.exit(ERROR_WORKER_EXCEPTION)
 
 
+def convert_to_client_error(e):
+    match = re.search(
+        r"An error occurred \((\w+)\) when calling the (\w+) operation.*: (.+)", str(e)
```
romain-intel, line 351: nit: pre-compile this expression

metaflow/plugins/datatools/s3/s3op.py
```diff
@@ -322,22 +328,60 @@ def op_info(url):
...
+
+def handle_client_error(err, idx, result_file):
+    error_code = normalize_client_error(err)
+    if error_code == 404:
+        result_file.write("%d %d\n" % (idx, -ERROR_URL_NOT_FOUND))
+        result_file.flush()
+    elif error_code == 403:
+        result_file.write("%d %d\n" % (idx, -ERROR_URL_ACCESS_DENIED))
+        result_file.flush()
+    elif error_code == 503:
```
romain-intel, line 375: nit, could we merge the two cases here? Am fine also keeping it separate as a documentation type of thing.

metaflow/plugins/datatools/s3/s3op.py
```diff
@@ -381,6 +425,15 @@ def start_workers(mode, urls, num_workers, inject_failure, s3config):
                 if proc.exitcode is not None:
                     if proc.exitcode != 0:
                         msg = "Worker process failed (exit code %d)" % proc.exitcode
+
+                        # IMPORTANT: if a child process has put items on a queue, then that process will not
+                        # terminate until all buffered items have been flushed to the pipe, causing a deadlock.
+                        # `cancel_join_thread()` allows the subprocess to exit without flushing the queue.
+                        #
+                        # Without this line, the subprocess would hang indefinitely when it did not exit cleanly
```
romain-intel, line 433: It's not the "subprocess", but "this parent process"

romain-intel (dismissed):
The change in logic looks correct.

.github/workflows/metaflow.s3_tests.yml
```diff
@@ -14,7 +14,7 @@ permissions: read-all
 
 jobs:
   test_data:
-    if: (github.event.action == 'labeled' && (github.event.label.name == 'approved' || github.event.label.name == 'ok-to-test')) || (github.event.action != 'labeled' && (contains(github.event.pull_request.labels.*.name, 'ok-to-test') || contains(github.event.pull_request.labels.*.name, 'approved')))
```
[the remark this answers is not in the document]
npow: yes, we need to disable the s3 tests since they will all fail due to 403, and the internal tests don't kick off because of the failures

metaflow/plugins/datatools/s3/s3op.py
npow: prefer to keep it this way (including some of the other branches that can be collapsed) for clarity

romain-intel (dismissed):

metaflow/plugins/datatools/s3/s3op.py
savingoyal: Do we need to do this matching? The error format is not an interface guaranteed by boto3 and can change at any point.

metaflow/metaflow_config.py
```diff
@@ -109,6 +109,12 @@
 # top-level retries)
 S3_TRANSIENT_RETRY_COUNT = from_conf("S3_TRANSIENT_RETRY_COUNT", 20)
 
+# S3 retry configuration used in the aws client
+# Use the adaptive retry strategy by default
+S3_CLIENT_RETRY_CONFIG = from_conf(
```
savingoyal, line 114: what's the rationale for pulling this out?

metaflow/plugins/datatools/s3/s3op.py
```diff
@@ -322,22 +331,58 @@ def op_info(url):
...
         except:
             traceback.print_exc()
+            result_file.flush()
             sys.exit(ERROR_WORKER_EXCEPTION)
 
 
+def convert_to_client_error(e):
+    match = BOTOCORE_MSG_TEMPLATE_MATCH.search(str(e))
+    if not match:
+        raise e
```
savingoyal, line 355: is this correct?

metaflow/metaflow_config.py
npow: so it's shared between aws_client.py and s3op.py

metaflow/plugins/datatools/s3/s3op.py
npow: do you see a problem here?

metaflow/plugins/datatools/s3/s3op.py
npow: I don't see how we can tell if it's a fatal error or a transient error without doing the matching. Do you have some ideas?

romain-intel (dismissed):
Please re-run black but otherwise LGTM.

savingoyal (approved):

PR #2356, opened 2025-03-20
fix: json parameters argo workflows

### LINKED ISSUE #2273: JSONType parameter not converted to object on argo (Metaflow >=2.12.23)
Hey, I noticed that `JSONType` does not convert parameters to objects on argo. Instead the params remain strings.

- With `python testflow.py run --param '["test"]'`, `self.param` is of type list
- With `python testflow.py argo-workflows trigger --param '["test"]'` the param is of type str

This issue might've been introduced with 2.12.23:
- 2.14.0: param is str
- 2.12.23: param is str
- 2.12.22: **param is list**

Test flow:

```python
from metaflow import JSONType, Parameter, kubernetes, project, step, FlowSpec

@project(name="test")
class TestFlow(FlowSpec):
    param = Parameter("param", type=JSONType)

    @kubernetes()
    @step
    def start(self):
        print(self.param)
        # <class 'str'> on argo, <class 'list'> via run
        print(type(self.param)) 
        self.next(self.end)

    @kubernetes()
    @step
    def end(self):
        pass


if __name__ == "__main__":
    TestFlow()
```

saikonen:
fixes #2273

One drawback of this fix is that when triggering flows through Argo UI, the JSONType parameters values will have to now have outermost quotes, essentially being string-serialized json.

### DIFF
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -3166,7 +3166,7 @@ class ArgoWorkflows(object):
                                                             parameter_name
                                                         ]["type"]
                                                         == "JSON"
-                                                        else "| squote"
+                                                        else "| toRawJson"
                                                     ),
                                                 ),
                                                 # Unfortunately the sensor needs to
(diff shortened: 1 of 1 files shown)


chrschorn:
Thank you for the fix!

PR #2372, opened 2025-04-07
introduce git info by default

savingoyal:
Screenshot 2025-04-07 at 3 36 00 pm


phitoduck:
This is so sick!!

metaflow/metaflow_git.py
```diff
@@ -0,0 +1,165 @@
...
+from os import path, name, environ
+
+# Cache for git information to avoid repeated subprocess calls
+_git_info_cache = None
+
+__all__ = ("get_git_info",)
+
+GIT_COMMAND = "git"
+
+if name == "nt":
```
romain-intel, line 20: we don't support windows so can probably strip this.

metaflow/metaflow_git.py
```diff
@@ -0,0 +1,165 @@
...
+        )
+        return result.stdout.strip() if result.returncode == 0 else None
+    except (OSError, subprocess.SubprocessError):
+        return None
+
+
+def _get_commit_sha():
+    """Get the current git commit SHA"""
+    try:
+        result = subprocess.run(
```
romain-intel, line 101: I wonder if there is a way to combine all the subprocess calls into one single call since we call them all at once.

metaflow/metaflow_git.py
```diff
@@ -0,0 +1,165 @@
...
+            text=True,
+            check=False,
+        )
+        # If output is not empty, there are uncommitted changes
+        return result.returncode == 0 and bool(result.stdout.strip())
+    except (OSError, subprocess.SubprocessError):
+        return None
+
+
+def get_git_info():
```
romain-intel, line 141: type hints (and a bit more specific :)

metaflow/metadata_provider/metadata.py
```diff
@@ -630,6 +630,20 @@ def _get_system_info_as_dict(self):
             sys_info["r_version"] = env["r_version_code"]
         return sys_info
 
+    def _get_git_info_as_dict(self):
+        git_info = {}
+        env = self._environment.get_environment_info()
+        for key in [
+            "script",
```
romain-intel, line 637: this is already in the dict for environment. We should probably not record script. It's not emitted anyways.

romain-intel (changes requested):
Some small comments

metaflow/metaflow_git.py
```diff
@@ -0,0 +1,165 @@
...
+            capture_output=True,
+            text=True,
+            check=False,
+        )
+        return result.stdout.strip() if result.returncode == 0 else None
+    except (OSError, subprocess.SubprocessError):
+        return None
+
+
+def _is_in_git_repo():
```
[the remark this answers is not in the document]
saikonen: confirmed. right now this picks the repo from the current working directory, when the intention most likely is to track the repository that the flow file resides in

PR #2375, opened 2025-04-10
feature: add custom metadata option for incident.io alerts configuration

saikonen:


metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -187,6 +187,12 @@ def argo_workflows(obj, name=None):
     default=None,
     help="Incident.io Alert source config ID. Example '01GW2G3V0S59R238FAHPDS1R66'",
 )
+@click.option(
+    "--incident-io-metadata",
+    default=None,
+    type=JSONTypeClass(),
+    help="Incident.io Alert Custom Metadata fields.",
+)
```
savingoyal, line 195: rather than json - can this be `--incident-io-metadata key=value --incident-io-metadata key2=value2 ...`

saikonen: definitely can be changed to that for nicer ux. I was also wondering whether the value should be settable through the Metaflow profile, or is the CLI option enough for now?

savingoyal: CLI option is enough for now

metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -267,6 +275,18 @@ def create(
         obj.is_project,
     )
 
+    # parse key value pairs into a dict for incident.io metadata if given
+    parsed_metadata = None
+    if incident_io_metadata is not None:
+        parsed_metadata = {}
+        for kv in incident_io_metadata:
+            key, value = kv.split("=", 1)
```
savingoyal, line 283: minor nit - not a blocker for release but we can move this logic to argo_workflows.py

PR #2377, opened 2025-04-15
Git info in metadata (take 2)

romain-intel:


metaflow/metaflow_git.py
```diff
@@ -0,0 +1,131 @@
...
+    global _git_info_cache
+
+    if _git_info_cache is not None:
+        return _git_info_cache
+
+    _git_info_cache = {}
+    if _is_in_git_repo(path):
+        _git_info_cache = {
+            "repo_url": _get_repo_url(path),
+            "branch_name": _get_branch_name(path),
```
romain-intel, line 126: I was thinking something like this:
```
echo "$(git rev-parse --abbrev-ref HEAD) $(git rev-parse --short HEAD)$(git diff --quiet || echo ' dirty')"
```
(one subprocess and also one single metadata element -- it doesn't save on the multitude of invocation of git but does only launch a single subprocesse, well two after we check that git is in the right directory but we could also probably skip that because the git command will error out if not in a git repo)

romain-intel (commented):
I'll let @talsperre also comment. Thanks for removing windows support as well. I don't know if we want to also disable it (ie: allow user to disable it with an env var). It may be useful for very large repos or just in case the user doesn't want to record it.

metaflow/metaflow_git.py
```diff
@@ -0,0 +1,131 @@
...
+                    domain = parts[0].replace("git@", "")
+                    repo_path = parts[1]
+                    url = f"https://{domain}/{repo_path}"
+            return url
+        return None
+    except (OSError, subprocess.SubprocessError):
+        return None
+
+
+def _get_branch_name(path: Union[str, os.PathLike]):
```
talsperre, line 46: Add return types.

metaflow/metaflow_git.py
```diff
@@ -0,0 +1,131 @@
...
+
+# Cache for git information to avoid repeated subprocess calls
+_git_info_cache = None
+
+__all__ = ("get_repository_info",)
+
+GIT_COMMAND = "git"
+
+
+def _get_repo_url(path: Union[str, os.PathLike]):
```
talsperre, line 21: Add return types - here as well.

metaflow/metaflow_git.py
```diff
@@ -0,0 +1,131 @@
...
+            capture_output=True,
+            text=True,
+            check=False,
+        )
+        return result.returncode == 0 and result.stdout.strip() == "true"
+    except (OSError, subprocess.SubprocessError):
+        return False
+
+
+def _has_uncommitted_changes(path: Union[str, os.PathLike]):
```
talsperre, line 91: ditto.

metaflow/metaflow_git.py
```diff
@@ -0,0 +1,131 @@
...
+            capture_output=True,
+            text=True,
+            check=False,
+        )
+        return result.stdout.strip() if result.returncode == 0 else None
+    except (OSError, subprocess.SubprocessError):
+        return None
+
+
+def _is_in_git_repo(path: Union[str, os.PathLike]):
```
talsperre, line 76: ditto.

metaflow/metaflow_git.py
```diff
@@ -0,0 +1,131 @@
...
+            capture_output=True,
+            text=True,
+            check=False,
+        )
+        return result.stdout.strip() if result.returncode == 0 else None
+    except (OSError, subprocess.SubprocessError):
+        return None
+
+
+def _get_commit_sha(path: Union[str, os.PathLike]):
```
talsperre, line 61: ditto.

metaflow/metaflow_git.py
```diff
@@ -0,0 +1,131 @@
...
+from os import path, name, environ
+from typing import Dict, Union
+
+# Cache for git information to avoid repeated subprocess calls
+_git_info_cache = None
+
+__all__ = ("get_repository_info",)
+
+GIT_COMMAND = "git"
+
```
talsperre, line 19: We should refactor to have a single `_run_git_command` command that is used by all the other functions.

metaflow/metaflow_git.py
```diff
@@ -0,0 +1,131 @@
...
+            cwd=path,
+            capture_output=True,
+            text=True,
+            check=False,
+        )
+        if result.returncode == 0:
+            url = result.stdout.strip()
+            # Convert SSH URLs to HTTPS for clickable links
+            if url.startswith("git@"):
+                parts = url.split(":")
```
talsperre, line 35: Is there any case where we should consider: `url.split(":", 1)`? Very rare edge case.

metaflow/metaflow_git.py
```diff
@@ -0,0 +1,131 @@
...
+                if len(parts) == 2:
+                    domain = parts[0].replace("git@", "")
+                    repo_path = parts[1]
```
talsperre, line 38: This will keep `.git` in the URL, right? Seems fine to me butt we could remove that as well if it's cleaner.

metaflow/metaflow_git.py
```diff
@@ -0,0 +1,131 @@
...
+    """Get the current git commit SHA"""
+    try:
+        result = subprocess.run(
+            [GIT_COMMAND, "rev-parse", "HEAD"],
+            cwd=path,
+            capture_output=True,
+            text=True,
+            check=False,
+        )
+        return result.stdout.strip() if result.returncode == 0 else None
```
talsperre, line 71: How do we distinguish between an error and non zero return code?

metaflow/metaflow_git.py
talsperre: +1

talsperre:
Should we make it available via the current singleton as well?

metaflow/metaflow_git.py
savingoyal: it just unnecessarily complicates the code without any real performance benefit

metaflow/metaflow_git.py
savingoyal: correct - that is intentional

metaflow/metaflow_git.py
savingoyal: we dont need to block the execution if somehow git metadata collection fails - it is strictly best effort by design.

metaflow/metaflow_git.py
savingoyal: good catch, will add.

savingoyal:
> I'll let @talsperre also comment. Thanks for removing windows support as well. I don't know if we want to also disable it (ie: allow user to disable it with an env var). It may be useful for very large repos or just in case the user doesn't want to record it.

all calls are going against git metadata and are fast by design. we can swap the dirty call with `git diff-index --quiet HEAD --` so that that's fast too.

talsperre (dismissed):

metaflow/metaflow_git.py
romain-intel: If you want separate calls sure (although there is a performance benefit but I agree it's tiny) -- not sure what it complicates but that's more stylistic. For the fewer fields of metadata, I would *definitely* lean in that direction. Metadata calls are *not* cheap (as you previously pointed out) and adding 4 fields seems a bit much.

savingoyal: we can lob everything into a json value to reduce the number of rows - @saikonen can you look into that too?

metaflow/metaflow_git.py
```diff
@@ -17,71 +17,79 @@
...
+    """
+    try:
+        result = subprocess.run(
+            ["git", *args],
+            cwd=path,
+            capture_output=True,
+            text=True,
+            check=False,
+        )
+        return result.stdout.strip(), result.returncode, False
```
savingoyal, line 37: do we need to return Failure @saikonen ?

saikonen: the failure is a separate case here for subprocess issues (e.g. timeout) that is not captured by the process exit code. I introduced this in order to retain the previous logic with the try/except, though for the return values there is no difference and could be omitted from what I can tell.

PR #2379, opened 2025-04-15
Feature/environment uv (take 2)

romain-intel:


metaflow/plugins/uv/uv_environment.py
```diff
@@ -0,0 +1,65 @@
...
+        self.datastore_type = datastore_type
+        self.logger = logger
+
+    def init_environment(self, echo, only_steps=None):
+        self.logger("Bootstrapping uv...")
+
+    def executable(self, step_name, default=None):
+        return "uv run python"
+
+    def interpreter(self, step_name):
```
romain-intel, line 27: This is not needed -- not used anywhere and not part of MetaflowEnvironment (this is only a pypi thing and that too only for the mainline impl).

metaflow/plugins/uv/uv_environment.py
```diff
@@ -0,0 +1,65 @@
...
+                if parent_dir == current_dir:  # Reached root
+                    raise UVException(
+                        f"Could not find {filename} in current directory or any parent directory"
+                    )
+                current_dir = parent_dir
+
+        pyproject_path = _find("pyproject.toml")
+        uv_lock_path = _find("uv.lock")
+        files = [
+            (uv_lock_path, os.path.basename(uv_lock_path)),
```
romain-intel, line 47: nit: this is "uv.lock" (same for below). This makes it a bit clearer.

metaflow/plugins/uv/bootstrap.py
```diff
@@ -0,0 +1,86 @@
...
+
+        return datastore_packages[datastore_type] + ["requests"]
+
+    def sync_uv_project(datastore_type):
+        print("Syncing uv project...")
+        dependencies = " ".join(get_dependencies(datastore_type))
+        cmd = f"""set -e;
+            ls -la;
+            uv sync --frozen --no-install-package metaflow;
+            uv pip install {dependencies} --strict
```
romain-intel, line 72: I am a little confused here: why are the dependencies added here. This makes it non-reproducible (since the packages could change). Could we instead make it so that these dependencies are included before packaging the lock and pyproject.toml files?

romain-intel (commented):
Some comments. Also general comments:
  - what happens if the user uses --environment=uv but has @conda or @pypi decorators. I think it will fail because the decorators themselves will require a specific type of environment.
  - this is explicitly not like the conda/pypi integration. The biggest one is that this is basically like a conda_base/pypi_base with no other step level decorators. It's not a bad or good thing but just pointing out that it is different.

savingoyal:
> Some comments. Also general comments:
> 
> * what happens if the user uses --environment=uv but has @conda or @pypi decorators. I think it will fail because the decorators themselves will require a specific type of environment.
> * this is explicitly not like the conda/pypi integration. The biggest one is that this is basically like a conda_base/pypi_base with no other step level decorators. It's not a bad or good thing but just pointing out that it is different.

correct - that's the intention - the goal is not to be symmetric with existing features of metaflow but to follow the development pattern of users who most likely create a single uv env in their project and want that env for all the steps. the current approach doesn't preclude supporting the current metaflow pattern in the future either.

metaflow/plugins/uv/uv_environment.py
savingoyal: good point - can strip it out.

metaflow/plugins/uv/uv_environment.py
```diff
@@ -0,0 +1,65 @@
...
+    def init_environment(self, echo, only_steps=None):
+        self.logger("Bootstrapping uv...")
+
+    def executable(self, step_name, default=None):
+        return "uv run python"
+
+    def interpreter(self, step_name):
+        return "uv run python"
+
+    def add_to_package(self):
```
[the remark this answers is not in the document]
savingoyal: can you expand on this a bit? i am not entirely familiar with 2269 yet and it seems that supporting 2269 is not necessarily a blocker for this PR just yet - we would need a solution but I would much rather not ship a solution ahead of a WIP PR.

metaflow/plugins/uv/uv_environment.py
savingoyal: sorry i didnt follow here. what's the proposed change?

metaflow/plugins/uv/bootstrap.py
savingoyal: We wouldn't want to modify the lock/pyproject files since the user controls those files and can lead to surprises for them. In most cases, this step will be a no-op, and the expectation always with the dependencies we introduce is that those are safe and validated dependencies.

romain-intel: I agree to not modifying the user level lock file but I feel we could modify it *before* packaging it. I know this is not the same as other decorators but this definitely breaks the "your environment won't change" and is equivalent to doing a "pip install" at runtime which was one of the drivers for the initial conda environment.

metaflow/plugins/uv/uv_environment.py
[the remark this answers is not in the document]
romain-intel: Agreed -- just want to make sure it is compatible. 2269 basically will NOT put these files in the same directory but somewhere else. Even a TODO of what uv would need would help (maybe it's just moving the files back in place?)

metaflow/plugins/uv/uv_environment.py
romain-intel: `os.path.basename(uv_lock_path)` is an obfuscated way of writing the string "uv.lock".

romain-intel: You can make it a constant to use it twice but it's 2 lines apart so not a huge deal either way.

metaflow/plugins/uv/bootstrap.py
savingoyal: we can chat more in person - but the user-specific env actually doesn't change here, so I don't see where the _definitely breaks_ is coming from

PR #2390, opened 2025-04-23
feature: retry decorator improvements

saikonen:
introduces `only_on=` to the retry decorator in order to conditionally retry on different types of failures

### DIFF
--- a/metaflow/plugins/retry_decorator.py
+++ b/metaflow/plugins/retry_decorator.py
@@ -1,6 +1,11 @@
+import os
+import signal
+import sys
+
 from metaflow.decorators import StepDecorator
 from metaflow.exception import MetaflowException
 from metaflow.metaflow_config import MAX_ATTEMPTS
+from metaflow import current
 
 SUPPORTED_RETRY_EVENTS = ["all", "spot-termination"]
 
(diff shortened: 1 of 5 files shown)


metaflow/plugins/retry_decorator.py
```diff
@@ -36,5 +56,60 @@ def step_init(self, flow, graph, step, decos, environment, flow_datastore, logge
...
+            unsupported_events = [
+                event for event in self.attributes["only_on"] if not _known_event(event)
+            ]
+            if unsupported_events:
+                raise MetaflowException(
+                    "The event(s) %s are not supported for only_on="
+                    % ", ".join("*%s*" % event for event in unsupported_events)
+                )
+
+    def task_pre_step(
```
savingoyal, line 79: can this be contained in batch and kubernetes implementations instead?

metaflow/plugins/retry_decorator.py
```diff
@@ -22,10 +38,14 @@ class RetryDecorator(StepDecorator):
         Number of times to retry this task.
     minutes_between_retries : int, default 2
         Number of minutes between retries.
+    only_on : List[str], default None
+        List of failure events to retry on. Not specifying values will retry on all known events.
+        Accepted values are
+        'step', 'instance-preemption'
```
savingoyal, line 44: this can be `all`, `instance-preemption`. `all` can be the default value and is the same as the current behavior.

metaflow/plugins/retry_decorator.py
```diff
@@ -1,6 +1,22 @@
...
+import threading
+from time import sleep
+
 from metaflow.decorators import StepDecorator
 from metaflow.exception import MetaflowException
 from metaflow.metaflow_config import MAX_ATTEMPTS
+from metaflow import current
+
+
+class RetryEvents(Enum):
```
savingoyal, line 14: do we need this enum?

metaflow/plugins/retry_decorator.py
```diff
@@ -1,6 +1,22 @@
...
+    STEP = "step"
+    PREEMPT = "instance-preemption"
+
+
+PLATFORM_EVICTED_EXITCODE = 234
```
savingoyal, line 19: is this needed here - since technically we might have many different exit codes depending on the error being caught. also depending on the error backend, the exit codes can be different for the same conceptual failure mode.

metaflow/plugins/retry_decorator.py
saikonen: its mainly for convenience when building the retry conditions on the executor implemention side, where we need to go through the supported options. Having a literal value to use instead of retyping strings reduces issues with possible typos or renames down the line.

for example this:
```
RetryEvents.PREEMPT.value in retry_conditions
```

metaflow/plugins/retry_decorator.py
saikonen: but do we even need the `all` as a value, as that behaviour should be covered by omitting the attribute completely? is it also not a plausible use-case to retry only on user-code exceptions?

metaflow/plugins/retry_decorator.py
saikonen: move to kubernetes / batch respectively. Kept the exit code the same for now as there shouldn't be any conflicts with the current value, but this can be changed as needed.

PR #2418, opened 2025-05-20
drop support for python 3.5

savingoyal:


metaflow/extension_support/__init__.py
```diff
@@ -104,7 +104,7 @@ def load_module(module_name):
 def get_modules(extension_point):
     modules_to_load = []
     if not _mfext_supported:
```
romain-intel, line 106: This can completely go away. I thought I actually did this in a PR somewhere.

romain-intel: _mfext_supported was only false for 3.4 or below so now that that is never true, anything mentioning mfext_supported can go away. It can now be assumed to be true.

savingoyal: now it's false for less than 3.6 so it will throw a nice error if someone tries to use 3.5

romain-intel: But it's not supported on 3.5. That's my point. The whole metaflow won't be on 3.5 anymore.

savingoyal:
dropping this in lieu of #2424

PR #2423, opened 2025-05-22
fix: UV-environment skip installing metaflow extensions

saikonen:


metaflow/plugins/uv/bootstrap.py
```diff
@@ -78,11 +79,27 @@ def get_dependencies(datastore_type):
...
 
+    def skip_metaflow_dependencies():
+        skip_pkgs = ["metaflow"]
+        info = read_info_file()
+        if info is not None:
+            try:
+                skip_pkgs.extend([ext_name for ext_name in info["ext_info"][0].keys()])
+            except Exception:
+                print(
+                    "Failed to read extension info from the INFO file. Some packages might be installed during runtime as a result."
```
savingoyal, line 90: Failed to read INFO. Metaflow-related packages might get installed during runtime.

PR #2424, opened 2025-05-23
drop support for python 3.5

madhur-ob:


romain-intel:
hun, look at that. We arelady had it updated :)

PR #2425, opened 2025-05-23
archive option in list files inside code package

madhur-ob:


metaflow/plugins/package_cli.py
```diff
@@ -38,14 +38,24 @@ def info(obj):
 
 
 @package.command(help="List files included in the code package.")
+@click.option(
+    "--archive/--no-archive",
+    default=False,
+    help="If True, lists the file paths as present in the tarball. "
```
savingoyal, line 44: If True, lists the file paths as present in the code package archive; otherwise, lists the files on your filesystem included in the code package

PR #2438, opened 2025-05-29
packaging

mt-ob:


metaflow/decorators.py
```diff
@@ -315,15 +311,26 @@ def package_init(self, flow, step_name, environment):
...
+              - arcname: path relative to the code directory in the package
+            - for CODE_MODULE:
+              - content: name of the module
+              - arcame: None (ignored)
+            - for CONFIG_FILE:
+              - content: path to the file to include
+              - arcname: path relative to the config directory in the package
+            - for CONFIG_CONTENT:
+              - content: bytes to include
+              - arcname: path relative to the config directory in the package
```
savingoyal, line 333: this is quite funky - returning one of two things with different types. need to fix this

metaflow/meta_files.py
```diff
@@ -0,0 +1,132 @@
...
+_info_file_content = None
+_info_file_present = None
+_included_dist_info = None
+_included_dist_present = None
+
+# Ideally these would be in package/mfenv.py but that would cause imports to fail so
+# moving here. The reason is that this is needed to read extension information which needs
+# to happen before mfenv gets packaged.
+
+MFENV_DIR = (
```
savingoyal, line 18: change MFENV to MFCODE. we already have --environment and @environment - MFENV will only lead to more confusion here and everywhere.

metaflow/meta_files.py
```diff
@@ -0,0 +1,132 @@
...
+    if os.path.exists(os.path.join(get_metaflow_root(), MFENV_MARKER)):
+        with open(
+            os.path.join(get_metaflow_root(), MFENV_MARKER), "r", encoding="utf-8"
+        ) as f:
+            meta_info = json.load(f)
+        version = meta_info.get("version")
+        if version not in get_filname_map:
+            raise ValueError(
+                "Unsupported packaging version '%s'. Please update Metaflow" % version
+            )
```
savingoyal, line 90: will we actually ever hit this?

metaflow/decorators.py
```diff
@@ -685,12 +692,8 @@ def foo(self):
     f.is_step = True
     f.decorators = []
     f.config_decorators = []
-    try:
-        # python 3
-        f.name = f.__name__
-    except:
-        # python 2
-        f.name = f.__func__.func_name
+    f.wrappers = []
```
romain-intel, line 695: This is actually not related to this PR. It slipped through.

metaflow/meta_files.py
```diff
@@ -0,0 +1,132 @@
...
+MFCONF_DIR = ".mfconf"  # Directory containing Metaflow's configuration files
+MFENV_MARKER = (
+    ".mfenv_install"  # Special file containing metadata about how Metaflow is packaged
+)
+
+
+class MetaFile(Enum):
+    INFO_FILE = "INFO"
+    CONFIG_FILE = "CONFIG_PARAMETERS"
+    INCLUDED_DIST_INFO = "INCLUDED_DIST_INFO"
```
savingoyal, line 30: what is `INCLUDED_DIST_INFO` - hard to understand what this implies - is there an `EXCLUDED_DIST_INFO` too somewhere?

metaflow/meta_files.py
```diff
@@ -0,0 +1,132 @@
...
+MFENV_DIR = (
+    ".mfenv"  # Directory containing "system" code (metaflow and user dependencies)
+)
+MFCONF_DIR = ".mfconf"  # Directory containing Metaflow's configuration files
+MFENV_MARKER = (
+    ".mfenv_install"  # Special file containing metadata about how Metaflow is packaged
+)
+
+
+class MetaFile(Enum):
```
savingoyal, line 27: for future generations looking at the code, it will be immensely useful to call out in detail what is included in these files.

savingoyal, line 27: btw, do we need to have three separate files? any value in merging all/some?

metaflow/meta_files.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+def meta_file_name(name: Union[MetaFile, str]) -> str:
+    if isinstance(name, MetaFile):
+        return name.value
+    return name
+
+
+def generic_get_filename(
+    name: Union[MetaFile, str], is_meta: Optional[bool] = None
```
savingoyal, line 40: is_meta is not actually used. consider dropping

metaflow/meta_files.py
```diff
@@ -0,0 +1,132 @@
...
+    path_to_file = os.path.join(get_metaflow_root(), real_name)
+    if os.path.isfile(path_to_file):
+        return path_to_file
+    return None
+
+
+def v1_get_filename(
+    name: Union[MetaFile, str],
+    meta_info: Dict[str, Any],
+    is_meta: Optional[bool] = None,
```
savingoyal, line 55: how does this work if i am interested in a file that unfortunately clashes with an internal reserved name? also, is_meta actually useful? what if we simply had another method that provided the contents of `INFO` file and it included everything you need?

metaflow/metaflow_environment.py
```diff
@@ -49,8 +51,26 @@ def bootstrap_commands(self, step_name, datastore_type):
 
     def add_to_package(self):
         """
-        A list of tuples (file, arcname) to add to the job package.
-        `arcname` is an alternative name for the file in the job package.
+        Called to add custom files needed for this environment. This hook will be
+        called in the `MetaflowPackage` class where metaflow compiles the code package
+        tarball. This hook can return one of two things:
+          - a generator yielding a tuple of `(file_path, arcname)` to add files to
```
savingoyal, line 57: this is a bit funky, we shouldn't return two types of tuples

metaflow/package/mfenv.py
```diff
@@ -0,0 +1,867 @@
...
+        name = name_normalizer.sub("-", context.name).lower()
+        if name in self._dist_info:
+            yield PackagedDistribution(
+                os.path.join(get_metaflow_root(), context.name),
+                self._dist_info[name],
+            )
+        return None
+
+
+class MFEnv:
```
savingoyal, line 181: can MetaflowPackage instead be the way that rest of the code accesses functionality here? will make it easier for all the decorator implementations which already have access to the package object to get access to info files etc.

metaflow/util.py
```diff
@@ -9,7 +9,6 @@
 from itertools import takewhile
```
savingoyal, line 9: this can be a different PR, cc @madhur-ob

metaflow/package/mfenv.py
savingoyal: reading through the code, MFEnv seems like an internal implementation detail for MetaflowPackage. If we can keep it as such throughout the code, that will be great!

metaflow/meta_files.py
```diff
@@ -0,0 +1,132 @@
...
+_included_dist_present = None
+
+# Ideally these would be in package/mfenv.py but that would cause imports to fail so
+# moving here. The reason is that this is needed to read extension information which needs
+# to happen before mfenv gets packaged.
+
+MFENV_DIR = (
+    ".mfenv"  # Directory containing "system" code (metaflow and user dependencies)
+)
+MFCONF_DIR = ".mfconf"  # Directory containing Metaflow's configuration files
```
savingoyal, line 21: consider changing the name to something else since it is neither metaflow config not the new config feature

metaflow/package/mfenv.py
```diff
@@ -0,0 +1,867 @@
...
+        if version not in cls._mappings:
+            raise MetaflowException(
+                "Invalid package -- unknown version %s in info: %s" % (version, info)
+            )
+
+        return cls._mappings[version]
+
+    @classmethod
+    def _extract_archive_mfenv_info(
+        cls, archive: Any, packaging_backend=TarPackagingBackend
```
savingoyal, line 357: any methods that need access to archive can be safely moved under MetaflowPackage?

metaflow/package/mfenv.py
```diff
@@ -0,0 +1,867 @@
...
+        }
+    return _cached_distributions
+
+
+_ModuleInfo = NamedTuple(
+    "_ModuleInfo", [("name", str), ("root_paths", Set[str]), ("module", ModuleType)]
+)
+
+
+class PackagedDistribution(metadata.Distribution):
```
savingoyal, line 135: consider moving this to a different file so that we can avoid an import from .mfnv in cli.py

metaflow/package/mfenv.py
```diff
@@ -0,0 +1,867 @@
...
+        return None
+
+
+class MFEnv:
+    _cached_mfenv_info = {}
+
+    _mappings = {}
+
+    @classmethod
+    def get_filename(
```
savingoyal, line 187: fair to assume that some methods work against local filesystem and some against the code package?

metaflow/meta_files.py
```diff
@@ -0,0 +1,132 @@
...
+                _info_file_content = json.load(f)
+                _info_file_present = True
+        else:
+            _info_file_present = False
+    if _info_file_present:
+        return _info_file_content
+    return None
+
+
+def read_included_dist_info():
```
savingoyal, line 117: consider renaming to read_dist_info()

metaflow/meta_files.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+def read_info_file():
+    # The info file is a bit special because it needs to be read to determine what
+    # extensions to load. We need to therefore not load anything yet. This explains
+    # the slightly wheird code structure where there is a bit of the file loading logic
+    # here that is then used in MFEnv (it logically belongs in MFEnv but that file can't
+    # be loaded just yet).
+    global _info_file_content
+    global _info_file_present
```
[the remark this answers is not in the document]
romain-intel: I can make that change. Technically there is an extra global too (the `_UNINITIALIZED`) but it can be shared.

metaflow/meta_files.py
romain-intel: I can clarify: it means the information about any distribution that is included. So no, there is no excluded one. I can just hame it DIST_INFO.

metaflow/meta_files.py
romain-intel: let me check on that. The idea of putting stuff in mfconf is that they won't clash with code.

metaflow/metaflow_environment.py
romain-intel: already addressed above with a question.

metaflow/plugins/pypi/conda_decorator.py
```diff
@@ -159,11 +159,11 @@ def runtime_init(self, flow, graph, package, run_id):
             os.path.join(self.metaflow_dir.name, "metaflow"),
         )
 
-        info = os.path.join(get_metaflow_root(), os.path.basename(INFO_FILE))
+        info = MFEnv.get_filename(MetaFile.INFO_FILE)
```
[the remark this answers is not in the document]
romain-intel: I didn't want to specialize too much I guess but I don't mind having separate utility functions `get_info_file_path`, `get_dist_file_path` etc if you prfer that. I was trying to make it easier to add files later.

metaflow/util.py
romain-intel: I don't think that is a change I did. What was this supposed to do?

metaflow/meta_files.py
romain-intel: So we can have `MFCODE_DIR` and `MFINFO_DIR` ?

metaflow/meta_files.py
```diff
@@ -0,0 +1,132 @@
...
+
+# Ideally these would be in package/mfenv.py but that would cause imports to fail so
+# moving here. The reason is that this is needed to read extension information which needs
+# to happen before mfenv gets packaged.
+
+MFENV_DIR = (
+    ".mfenv"  # Directory containing "system" code (metaflow and user dependencies)
+)
+MFCONF_DIR = ".mfconf"  # Directory containing Metaflow's configuration files
+MFENV_MARKER = (
```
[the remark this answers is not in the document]
romain-intel: would make it less flexible. The marker tells you how to access the info file. The info file is not in the same place already.

metaflow/package/mfenv.py
```diff
@@ -0,0 +1,867 @@
...
+class AddToPackageType(IntEnum):
+    CODE_ONLY = 0x100
+    CONFIG_ONLY = 0x200
+    FILES_ONLY = 0x400
+    ALL = 0xFFF
+    CODE_FILE = 0x501
+    CODE_MODULE = 0x502
+    CODE_METAFLOW = 0x504
+    CONFIG_FILE = 0x601
+    CONFIG_CONTENT = 0x202
```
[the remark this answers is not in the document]
romain-intel: I'll have to check. We need to distinguish between files and content I think in some of the things. I can check if I can get rid of that. I think we also nee to know something is a module to extract the dist info.

PR #2441, opened 2025-06-03
fix s3 list_recursive to exclude sibling directories with shared prefixes

mt-ob:


dotKokott:
This fix works for me @madhur-ob, thanks!

Looking forward to upgrade once that is merged :)

metaflow/plugins/datatools/s3/s3op.py
```diff
@@ -582,11 +582,23 @@ def list_prefix(self, prefix_url, delimiter=""):
...
                     for key in page.get("Contents", []):
-                        url = url_base + key["Key"]
+                        key_path = key["Key"]
+                        # filter out sibling directories that share the same prefix
+                        if delimiter == "":  # recursive mode
+                            normalized_prefix = prefix_url.path
+                            if not normalized_prefix.endswith("/"):
+                                normalized_prefix += "/"
+                            # Only include keys that are actually under our directory
+                            if not (
```
npow, line 592: This filters out an object with key exactly equal to `ABC/` when listing prefix `ABC/`. In S3 key names can actually contain `/` so someone could create a file called `ABC/` (it's weird but valid)

romain-intel: so is the recommended change just > to >= ?

npow: i think that would work

dotKokott:
Is there a way I can help to get this in?

PR #2442, opened 2025-06-04
feature: allow rebuilding conda/pypi environments

saikonen:
this PR adds a command for forcing a rebuild of the conda/pypi environment(s) of a flow.

```
# rebuild all environments
HelloConda.py --environment conda environment rebuild

# specify any number of steps to rebuild only their environments
HelloConda.py --environment conda environment rebuild --step start --step end
```

### DIFF
--- a/metaflow/plugins/pypi/environment_cli.py
+++ b/metaflow/plugins/pypi/environment_cli.py
@@ -15,28 +15,36 @@ def environment(ctx):
     echo = ctx.obj.echo
 
 
-@environment.command(help="Rebuild the environment")
+@environment.command(help="Resolve the environment(s)")
 @click.option(
     "--step",
     "steps",
     multiple=True,
     default=[],
-    help="Steps to rebuild the environment",
+    help="Steps to resolve the environment for",
+)
+@click.option(
+    "--force/--no-force",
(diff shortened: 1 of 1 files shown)


romain-intel:
General comment: could we align on the names. The extension already does this: https://github.com/Netflix/metaflow-nflx-extensions/blob/main/metaflow_extensions/netflix_ext/plugins/environment_cli.py and for the limited functionality supported here, we could do the same names (ie: `resolve --force`) which may make things a bit more consistent.

metaflow/plugins/pypi/environment_cli.py
```diff
@@ -0,0 +1,50 @@
...
+
+@cli.group(help="Commands related to managing the conda/pypi environments")
+@click.pass_context
+def environment(ctx):
+    # the logger is configured in cli.py
+    global echo
+    echo = ctx.obj.echo
+
+
+@environment.command(help="Resolve the environment(s)")
```
romain-intel, line 18: nit: the other plugin uses those as argument `resolve start end mid` isntead of `--step start --start end`.

saikonen:
closing in favour of #2443

PR #2443, opened 2025-06-05
feature: alternate approach to environment force-rebuild

saikonen:
adds a top-level option to toggle force-rebuilding the environment

```
python HelloConda.py --environment conda --force-rebuild-environment run
```

the introduction of the new option is not mandatory. we could simply proceed with an internal envvar to enable the functionality for now.

### DIFF
--- a/metaflow/cli.py
+++ b/metaflow/cli.py
@@ -1,5 +1,6 @@
 import functools
 import inspect
+import os
 import sys
 import traceback
 from datetime import datetime
@@ -242,6 +243,14 @@ def version(obj):
     type=click.Choice(["local"] + [m.TYPE for m in ENVIRONMENTS]),
     help="Execution environment type",
 )
+@click.option(
+    "--force-rebuild-environment/--no-force-rebuild-environment",
+    is_flag=True,
+    default=False,
+    hidden=True,
+    type=bool,
+    help="Force rebuild the environment",
+)
 # See comment for --quiet
 @click.option(
     "--datastore",
@@ -300,6 +309,7 @@ def start(
     quiet=False,
     metadata=None,
     environment=None,
+    force_rebuild_environment=False,
     datastore=None,
     datastore_root=None,
     decospecs=None,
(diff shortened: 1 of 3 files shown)


metaflow/cli.py
```diff
@@ -242,6 +243,14 @@ def version(obj):
     type=click.Choice(["local"] + [m.TYPE for m in ENVIRONMENTS]),
     help="Execution environment type",
 )
+@click.option(
+    "--force-rebuild-environment/--no-force-rebuild-environment",
```
savingoyal: can we change this to `--rebuild-environments`

metaflow/cli.py
```diff
@@ -242,6 +243,14 @@ def version(obj):
...
+    is_flag=True,
+    default=False,
+    hidden=True,
+    type=bool,
+    help="Force rebuild the environment",
```
savingoyal: `explicitly rebuild the execution environments`

metaflow/cli.py
savingoyal: actually, nvm. let's keep the force, otherwise `--no-rebuild-environments` takes a different meaning altogether. tiny nit - `--force-rebuild-environments` since there can be more than one.

metaflow/cli.py
```diff
@@ -432,6 +442,9 @@ def start(
 
     ctx.obj.graph = ctx.obj.flow._graph
 
+    # carry the force rebuild as an internal env var for generic environments to use
+    if force_rebuild_environment:
+        os.environ["_MFENV_FORCE_REBUILD"] = "1"
```
savingoyal: _METAFLOW_ENVIRONMENT_FORCE_REBUILD? `MFENV` will very soon have different connotations

metaflow/cli.py
```diff
@@ -432,6 +442,9 @@ def start(
 
     ctx.obj.graph = ctx.obj.flow._graph
 
+    # carry the force rebuild as an internal env var for generic environments to use
+    if force_rebuild_environments:
+        os.environ["_METAFLOW_ENVIRONMENT_FORCE_REBUILD"] = "1"
```
savingoyal: we can avoid using environment variables and use an object instance variable instead

PR #2445, opened 2025-06-07
S3 tests dev stack using minio

madhur-ob:


test/data/s3/test_s3.py
```diff
@@ -487,9 +487,19 @@ def test_info_many(inject_failure_rate, s3root, prefixes, expected):
 def test_get_exceptions(inject_failure_rate, s3root, prefixes, expected):
     # get_many() goes via s3op, get() is a method - test both the code paths
     with S3(inject_failure_rate=inject_failure_rate) as s3:
-        with pytest.raises(MetaflowS3AccessDenied):
+        with pytest.raises(
+            (
+                MetaflowS3AccessDenied,
+                MetaflowS3NotFound,
```
mt-ob, line 493: MinIO gives `MetaflowS3NotFound` once we handle `NoSuchBucket` to return a 404
The test originally had `MetaflowS3AccessDenied` probably because behaviour of S3 and MinIO differs on handling a path which doesn't exist such as `"s3://foobar/foo"`

The other option is to handle `NoSuchBucket` to return a 403

metaflow/plugins/datatools/s3/s3op.py
```diff
@@ -131,7 +131,7 @@ def normalize_client_error(err):
     except ValueError:
         if error_code in ("AccessDenied", "AllAccessDisabled", "InvalidAccessKeyId"):
             return 403
-        if error_code == "NoSuchKey":
+        if error_code in ("NoSuchKey", "NoSuchBucket"):
```
mt-ob, line 134: Should `NoSuchBucket` give a 403 instead?

savingoyal: `NoSuchBucket` should technically be a 404

PR #2446, opened 2025-06-09
feature: support custom exit hooks

saikonen:
implementing support for running custom functions as exit hooks with argo workflows, local, kubernetes and batch runtimes.

also refactors current lifecycle hook setup to use the same primitives.

Suggested UX for defining custom exit hooks in the flow file:

```python
from time import sleep
from typing import Optional
from metaflow import step, FlowSpec, Parameter, exit_hook, Run


def success_print():
    print("Flow completed successfully. Performing some success related notifications.")


def failure_print(run: Optional[Run]):
    print("Flow failed.")

    if run is not None:
        print("No run was registered.")
        return

    print("failed run id:", run.id)
    print(
        "Failed tasks:\n",
        "\n".join(
            [task.pathspec for step in run for task in step if not task.successful]
        ),
    )


@exit_hook(on_success=[success_print])
@exit_hook(on_error=[failure_print])
class NotifyFlow(FlowSpec):
    should_fail = Parameter(name="should_fail", default=False)

    @step
    def start(self):
        print("Starting 👋")
        print("Should fail?", self.should_fail)
        if self.should_fail:
            sleep(60)
            raise Exception("failing as expected")
        self.next(self.end)

    @step
    def end(self):
        print("Done! 🏁")


if __name__ == "__main__":
    NotifyFlow()
```


metaflow/plugins/argo/hook_decorator.py
```diff
@@ -0,0 +1,46 @@
+from metaflow.decorators import FlowDecorator
+import inspect
+
+from metaflow.exception import MetaflowException
+from metaflow.plugins.argo.exit_hooks import ScriptHook
+
+
+class RunOnFinishDecorator(FlowDecorator):
+    name = "run_on_finish"
```
savingoyal: @exit_hook

EvgeniiTitov:
Really looking forward to this getting merged and released guys!

metaflow/plugins/exit_hook/exit_hook_script.py
```diff
@@ -0,0 +1,40 @@
...
+    # hook_fn
+    hook_fn = getattr(tempflow, fn_name)
+
+    argspec = inspect.getfullargspec(hook_fn)
+
+    # Check if fn expects a run object as an arg.
+    if "run" in argspec.args or argspec.varkw is not None:
+        from metaflow import Run, namespace
+
+        namespace(None)
```
romain-intel: use _namespace_check.

metaflow/plugins/exit_hook/exit_hook_script.py
```diff
@@ -0,0 +1,40 @@
...
+import os
+
+
+def main(flow_file, fn_name, run_pathspec):
+    flow_import = flow_file.rstrip(".py")
+
+    tempflow = importlib.import_module(flow_import)
+
+    # hook_fn
+    hook_fn = getattr(tempflow, fn_name)
```
romain-intel: so this means that exit hooks have to be in the same file as the flow?

metaflow/plugins/exit_hook/exit_hook_decorator.py
```diff
@@ -0,0 +1,32 @@
+from metaflow.decorators import FlowDecorator
+from metaflow.exception import MetaflowException
+
+
+class ExitHookDecorator(FlowDecorator):
+    name = "exit_hook"
+    allow_multiple = True
+
+    defaults = {
+        "image": None,
```
romain-intel: I would make this `options` and a kwargs type of thing. We don't know what may be useful and different backends may have different things they need/want. This is not used in the local one for example.

metaflow/plugins/exit_hook/exit_hook_decorator.py
```diff
@@ -0,0 +1,32 @@
...
+        on_error = self.attributes["on_error"]
+
+        if not on_success and not on_error:
+            raise MetaflowException(
+                "Choose at least one of the options on_success/on_error"
+            )
+
+        self.success_hooks = []
+        self.error_hooks = []
+        for success_fn in self.attributes["on_success"]:
```
romain-intel: nit: extracting into `on_success` and then using the attributes again.

metaflow/plugins/exit_hook/exit_hook_decorator.py
```diff
@@ -0,0 +1,32 @@
...
+            self.success_hooks.append(success_fn.__name__)
```
romain-intel: it would be nice to be able to take strings too. I think that would have the benefit of being more easily configurable (config could contain the string name for the hook)

For functions, I would actually store the full path because that would make it compatible (or more compatible) with the new packaging AND you wouldn't have to load your flow file to run the hooks (flow files can contain other stuff that you may not need/want/can use given your mention that the image may be different, etc.). To execute we could then load *just* what is needed.

metaflow/runtime.py
```diff
@@ -591,6 +592,48 @@ def execute(self):
...
+                    subprocess.check_output(
+                        args=[
+                            "python",
+                            "-m",
+                            "metaflow.plugins.exit_hook.exit_hook_script",
+                            flow_file,
+                            fn_name,
+                            pathspec,
+                        ],
+                        env=os.environ,
```
romain-intel: needs to change to be compatibile with new packaging (depending onw hich gets merged first)

metaflow/runtime.py
```diff
@@ -591,6 +592,48 @@ def execute(self):
...
+                print(result)
+
+            # Call the exit hook functions
+            for fn_name in [
+                name
+                for deco in exit_hook_decos
+                for name in (deco.success_hooks if successful else deco.error_hooks)
+            ]:
+                _call(fn_name)
+        except Exception as ex:
```
romain-intel, line 636: we could also try/catch around the check_output and attempt to run all hooks instead of stopping at the first failure?

metaflow/plugins/exit_hook/exit_hook_decorator.py
mt-ob: done

metaflow/plugins/exit_hook/exit_hook_script.py
mt-ob: done

metaflow/runtime.py
mt-ob: this is done

metaflow/runtime.py
mt-ob: I have little context here, will wait for others...

metaflow/plugins/exit_hook/exit_hook_decorator.py
mt-ob: @romain-intel I have pushed, can you check at your end once?

mt-ob:
@romain-intel now we can do:

```py
from time import sleep
from metaflow import step, FlowSpec, Parameter, exit_hook, Config, config_expr


@exit_hook(on_success=[config_expr("config.on_success")])
@exit_hook(on_error=[config_expr("config.on_error")])
class NotifyFlow(FlowSpec):
    config = Config("config", default="myconfig.json")
    should_fail = Parameter(name="should_fail", default=False)

    @step
    def start(self):
        print("Starting 👋")
        print("Should fail?", self.should_fail)
        if self.should_fail:
            sleep(60)
            raise Exception("failing as expected")
        self.next(self.end)

    @step
    def end(self):
        print("Done! 🏁")


if __name__ == "__main__":
    NotifyFlow()
```

where `hooks.py` has:

```py
from typing import Optional
from metaflow import Run


def success_print():
    print("Flow completed successfully. Performing some success related notifications.")


def failure_print(run: Optional[Run]):
    print("Flow failed.")

    if run is not None:
        print("No run was registered.")
        return

    print("failed run id:", run.id)
    print(
        "Failed tasks:\n",
        "\n".join(
            [task.pathspec for step in run for task in step if not task.successful]
        ),
    )
```

and `myconfig.json` has 

```json
{
  "on_success": "hooks.success_print",
  "on_error": "hooks.failure_print"
}
```

metaflow/plugins/exit_hook/exit_hook_script.py
mt-ob: this is fixed now

metaflow/plugins/exit_hook/exit_hook_decorator.py
mt-ob: I am not sure if `image` is even used, I guess we can remove it?

metaflow/plugins/airflow/airflow.py
```diff
@@ -654,6 +654,12 @@ def compile(self):
                 "to Airflow is not supported currently."
             )
 
+        if self.flow._flow_decorators.get("exit_hook"):
+            raise AirflowException(
+                "Deploying flows with the @exit_hook decorator "
+                "to Airflow is not supported currently."
```
romain-intel: nit: currently supported

metaflow/plugins/aws/step_functions/step_functions.py
```diff
@@ -301,6 +301,12 @@ def _compile(self):
                 "to AWS Step Functions is not supported currently."
             )
 
+        if self.flow._flow_decorators.get("exit_hook"):
+            raise StepFunctionsException(
+                "Deploying flows with the @exit_hook decorator "
+                "to AWS Step Functions is not supported currently."
```
romain-intel: same nit

metaflow/runtime.py
```diff
@@ -591,6 +592,50 @@ def execute(self):
...
+            successful = ("end", ()) in self._finished or self._clone_only
+            pathspec = f"{self._graph.name}/{self._run_id}"
+            flow_file = self._environment.get_environment_info()["script"]
+
+            def _call(fn_name):
+                try:
+                    result = (
+                        subprocess.check_output(
+                            args=[
+                                "python",
```
romain-intel: prefer something like `sys.executable`.  In Conda, that would be the conda executable, if that is not what you want, you can use `ENV_ESCAPE_PY` to get the outer python.

PR #2448, opened 2025-06-10
[Docs] Visualization of the metaflow codebase

ivanmilevtues:
This PR contains documents which visualise the codebase of the metaflow project through mermaid diagrams.
You can see how the whole thing renders here: https://github.com/CodeBoarding/GeneratedOnBoardings/blob/main/metaflow/on_boarding.md

The idea is that such documents can give high-level understanding of the codebase to people who are new to it. This way they can focus on interesting parts (components) of the codebase with knowing how they fit in the big picture. I would love ot hear what is your current process of on-boarding in Netflix and do you think that such diagrams can fit into it :).

Any feedback is more than welcome! Just to clarify we are building these diagrams with Static analysis and LLMs, and currently working on a GitHub Action, so that you can have them always up-to-date!

Full disclosure: we're trying to turn this into a startup, but we're still in a very early stage and figuring out what will actually be useful for people.

### DIFF
--- a/.codeboarding/Core Workflow Engine.md
+++ b/.codeboarding/Core Workflow Engine.md
@@ -1,0 +1,127 @@
+```mermaid
+graph LR
+    Flow_Definition_FlowSpec_["Flow Definition (FlowSpec)"]
+    Flow_Graph_Management["Flow Graph Management"]
+    Native_Workflow_Execution["Native Workflow Execution"]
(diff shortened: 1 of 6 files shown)


ivanmilevtues:
A quick update https://github.com/CodeBoarding/CodeBoarding is now open-source! You can check how the generation works in details now :)

PR #2452, opened 2025-06-13
[cards] support writing metadata

valayDave:
- cards can now have metadata about important information created for them like ids/types/ other info in decorators
- This also allows us for book-keeping the attempts as cards are created.
- We write the metadata in the cli so we put something stateful on top to ensure we don't end up writing the metadata multiple times.
- Overall, since cards are being written async, we can only ensure best effort registration from metaflow side since we don't necessarily block when we save cards.
- This commit implementation will save metadata on a per-card basis. Instead of updating all card metadata at once. We do this because we will otherwise have multiple copies of the metadata in the database.
- The metadata writing is optional and can be disabled via the config variable.
- This addition also allows us to update the metaflow-service code for filtering cards based on attempt basis.


metaflow/metaflow_config.py
```diff
@@ -218,6 +218,8 @@
 
 RUNTIME_CARD_RENDER_INTERVAL = from_conf("RUNTIME_CARD_RENDER_INTERVAL", 60)
 
+CARD_WRITE_TO_METADATA = from_conf("CARD_WRITE_TO_METADATA", True)
```
savingoyal: do we need to offer it as a top level config? we can start writing metadata always in the context of flows.

valayDave: @tuulos recommended that we keep the writing optional since it was not there earlier. Personally don't have a problem with removing it since the UI service might still use the same logic that retrieves stuff from the datastore.

savingoyal: We discussed this in the weekly chat that it would be easy to keep things simple and consistent going forward.

metaflow/plugins/cards/card_decorator.py
```diff
@@ -55,11 +75,14 @@ class CardDecorator(StepDecorator):
...
     """
 
+    _GLOBAL_CARD_INFO = {}
+
     name = "card"
     defaults = {
         "type": "default",
         "options": {},
         "scope": "task",
+        "rank": None,  # Can be one of "high", "medium", "low". Can help derive ordering on the UI.
```
saikonen, line 85: is the comment here correct? my expectation for "rank" would've been integer values. There's also no validation for the value entered.

PR #2453, opened 2025-06-13
[cards] remove dual write

valayDave:
Earlier the `save_card` function ended up saving cards at two locations. Thereby doubling the number of cards. (this was a temporary fix; which we never ended up getting rolled back) 

> Why did we do this ? :

When cards were introduced there was an assumption made about task-ids being unique. This assumption was incorrect. Only the pathspec is unique and there are no such guarantees about task-ids. When task-ids are non-unique, card read would result in finding incorrect cards (because cards were stored based on task-ids). If we had immediately switched from storing cards based on `task-ids` to a `step-name/task-id`, then card reading would have crashed for many users. It would especially happen for users who are accessing cards created by a newer MF client from an older version of MF client. It will also easily end up breaking the metaflow-ui (which maybe using a client from an older version). But since now its been over 3 years and there have been multiple newer versions of clients rolled out, we can safely remove this since most cards will be created by clients in the last 3 years. It keep the listing/reading path intact. But only removes the dual write part.

### DIFF
--- a/metaflow/plugins/cards/card_datastore.py
+++ b/metaflow/plugins/cards/card_datastore.py
@@ -231,23 +231,6 @@ class CardDatastore(object):
 
     def save_card(self, uuid, card_type, card_html, card_id=None, overwrite=True):
         card_file_name = card_type
-        # TEMPORARY_WORKAROUND: FIXME (LATER) : Fix the duplication of below block in a few months.
-        # Check file blame to understand the age of this temporary workaround.
-
-        # This function will end up saving cards at two locations.
-        # Thereby doubling the number of cards. (Which is a temporary fix)
-        # Why do this ? :
-        # When cards were introduced there was an assumption made about task-ids being unique.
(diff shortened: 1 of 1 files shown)


metaflow/plugins/cards/card_datastore.py
```diff
@@ -231,48 +231,17 @@ def save_data(self, uuid, card_type, json_data, card_id=None):
...
-        # It will also easily end up breaking the metaflow-ui (which maybe using a client from an older version).
-        # Hence, we are writing cards to both paths so that we can introduce breaking changes later in the future.
         card_path_with_steps = self.get_card_location(
             self._get_card_write_path(),
             card_file_name,
             uuid,
             card_id=card_id,
             suffix=CardNameSuffix.CARD,
         )
-        if SKIP_CARD_DUALWRITE:
```
savingoyal, line 258: can you remove this var from config?

PR #2456, opened 2025-06-13
Support for conditionals

mt-ob:
```py
from metaflow import FlowSpec, step, Parameter

class SwitchTestFlow(FlowSpec):
    
    threshold = Parameter('threshold', default=5, type=int)

    @step
    def start(self):
        import random
        self.value = random.randint(1, 10)
        print(f"Generated value: {self.value}")
        self.next(self.decide)

    @step
    def decide(self):
        # Set the condition value as a string key
        self.path_choice = "high" if self.value > self.threshold else "low"
        print(f"Condition: {self.value} > {self.threshold}, choosing path: {self.path_choice}")
        self.next({"high": self.high_path, "low": self.low_path}, condition='path_choice')

    @step
    def high_path(self):
        print("Taking HIGH path")
        self.result = f"HIGH: {self.value}"
        self.next(self.end)
    
    @step
    def low_path(self):
        print("Taking LOW path") 
        self.result = f"LOW: {self.value}"
        self.next(self.end)
    
    @step
    def end(self):
        print(f"Final result: {self.result}")

if __name__ == '__main__':
    SwitchTestFlow()
```

### DIFF
--- a/metaflow/plugins/aws/aws_utils.py
+++ b/metaflow/plugins/aws/aws_utils.py
@@ -48,7 +48,7 @@ def get_ec2_instance_metadata():
         # Try to get an IMDSv2 token.
         token = requests.put(
             url="http://169.254.169.254/latest/api/token",
-            headers={"X-aws-ec2-metadata-token-ttl-seconds": 100},
+            headers={"X-aws-ec2-metadata-token-ttl-seconds": "100"},
             timeout=timeout,
         ).text
     except:


mt-ob:
can pass config as keys i.e. `self.next({self.config.path: self.high_path, "low": self.low_path}, condition='path_choice')`

metaflow/flowspec.py
```diff
@@ -811,6 +842,92 @@ def next(self, *dsts: Callable[..., None], **kwargs) -> None:
...
+
+            # Validate that all switch case values are step methods
+            funcs = []
+            for case_value, step_method in switch_cases.items():
+                try:
+                    func_name = step_method.__func__.__name__
+                except:
+                    msg = (
+                        "In step *{step}* switch case '{case}', the value is "
+                        "not a function. Make sure all values in the switch dictionary "
```
savingoyal, line 901: nit - the value is not a Metaflow step.

metaflow/flowspec.py
```diff
@@ -811,6 +842,92 @@ def next(self, *dsts: Callable[..., None], **kwargs) -> None:
...
+                        "are methods of the Flow class.".format(
```
savingoyal, line 902: are steps.. adjust accordingly

metaflow/flowspec.py
```diff
@@ -756,6 +756,27 @@ def _is_primitive_type(item):
         value = item if _is_primitive_type(item) else reprlib.Repr().repr(item)
         return basestring(value)[:MAXIMUM_FOREACH_VALUE_CHARS]
 
+    def _validate_switch_cases(self, switch_cases, step):
+        resolved_cases = {}
+        for case_key, step_method in switch_cases.items():
+            if isinstance(case_key, str) and case_key.startswith("config."):
```
savingoyal, line 762: not sure this works? `{self.c.foo: self.step_a, self.c.bar: self.step_b}` where self.c happens to be a Config

metaflow/flowspec.py
```diff
@@ -756,6 +756,27 @@ def _is_primitive_type(item):
         value = item if _is_primitive_type(item) else reprlib.Repr().repr(item)
         return basestring(value)[:MAXIMUM_FOREACH_VALUE_CHARS]
 
+    def _validate_switch_cases(self, switch_cases, step):
```
savingoyal, line 759: you may want to add guards in step functions, airflow and argo-workflows implementation that conditionals are not supported just yet.

test/core/tests/catch_retry.py
```diff
@@ -4,6 +4,7 @@
 
 class CatchRetryTest(MetaflowTest):
     PRIORITY = 2
+    SKIP_GRAPHS = ["simple-switch"]
```
savingoyal, line 7: why skip simple-switch here and in many other tests?

test/core/tests/basic_artifact.py
```diff
@@ -8,6 +8,7 @@ class BasicArtifactTest(MetaflowTest):
     """
 
     PRIORITY = 0
+    SKIP_GRAPHS = ["simple-switch"]
```
savingoyal, line 11: i believe we can specify (or at least it used to exist at some point) which tests are not eligible for a given config within the config itself. that will ensure you don't have to touch so many files in this PR.

mt-ob:
Closing in favour of https://github.com/Netflix/metaflow/pull/2473

PR #2463, opened 2025-06-18
[WIP] Feat/user decorators v2

romain-intel:


metaflow/plugins/cards/card_decorator.py
```diff
@@ -361,11 +361,6 @@ def _create_top_level_args(self, flow):
 
         return list(self._options(top_level_options))
 
-    def task_exception(
-        self, exception, step_name, flow, graph, retry_count, max_user_code_retries
-    ):
-        self._cleanup(step_name)
-
```
saikonen, line 368: why is this being removed?

metaflow/plugins/kubernetes/kubernetes_decorator.py
```diff
@@ -177,8 +179,6 @@ class KubernetesDecorator(StepDecorator):
     target_platform = KUBERNETES_CONDA_ARCH or "linux-64"
 
     def init(self):
-        super(KubernetesDecorator, self).init()
```
saikonen, line 182: removed as unnecessary? same question for all the decorators `init()` changes

metaflow/plugins/cards/card_decorator.py
romain-intel: it was duplicative -- finished is called in all cases and this casued the count to be wrong (we would decrement it twice in case of an error).

metaflow/plugins/kubernetes/kubernetes_decorator.py
romain-intel: why removed? It is necessary. You need to call the base class' init because that is what resolves the attributes.

romain-intel: wait, I removed it? Ok, let me check

romain-intel: Ah right -- I now use this other function called `external_init`

PR #2473, opened 2025-06-27
support for conditionals

madhur-ob:


test_runner
```diff
@@ -19,6 +19,13 @@ run_tests() {
   cd test/core && PYTHONPATH=`pwd`/../../ python3 run_tests.py --num-parallel 8 && cd ../../
 }
 
+run_switch_tests() {
+  cd test/core && PYTHONPATH=`pwd`/../../ python3 run_tests.py --switch --graphs simple_switch --tests BasicSwitchTest && cd ../../
+  cd test/core && PYTHONPATH=`pwd`/../../ python3 run_tests.py --switch --graphs nested_switch --tests NestedSwitchTest && cd ../../
+  cd test/core && PYTHONPATH=`pwd`/../../ python3 run_tests.py --switch --graphs switch_in_branch --tests SwitchInBranchTest && cd ../../
+  cd test/core && PYTHONPATH=`pwd`/../../ python3 run_tests.py --switch --graphs branch_in_switch --tests BranchInSwitchTest && cd ../../
```
mt-ob: need a better way to run a test against a particular graph only

mt-ob: tests fail because they target a graph which is not meant for them, example below:
```
Exception: Test <branch_in_switch.BranchInSwitchTest object at 0x7f4ee851ef90>
doesn't have a match for step start in graph small-parallel
```

metaflow/flowspec.py
```diff
@@ -811,6 +850,92 @@ def next(self, *dsts: Callable[..., None], **kwargs) -> None:
...
+                msg = (
+                    "Step *{step}* has an invalid self.next() transition. "
+                    "Switch statements cannot be combined with foreach or num_parallel.".format(
+                        step=step
+                    )
+                )
+                raise InvalidNextException(msg)
+
+            switch_cases = dsts[0]
+            if not switch_cases:
```
romain-intel: nit: this could be combined in the first check for dict.

metaflow/flowspec.py
```diff
@@ -756,6 +756,35 @@ def _is_primitive_type(item):
         value = item if _is_primitive_type(item) else reprlib.Repr().repr(item)
         return basestring(value)[:MAXIMUM_FOREACH_VALUE_CHARS]
 
+    def _validate_switch_cases(self, switch_cases, step):
```
romain-intel, line 791: is this for configs? Why not allow {self.config.key_value: ...}

metaflow/flowspec.py
```diff
@@ -811,6 +850,92 @@ def next(self, *dsts: Callable[..., None], **kwargs) -> None:
...
+                    "Condition variable *self.{var}* in step *{step}* "
+                    "does not exist. Make sure you set self.{var} in this step.".format(
+                        step=step, var=condition
+                    )
+                )
+                raise InvalidNextException(msg)
+
+            # Validate that all switch case values are step methods
+            funcs = []
+            for case_value, step_method in switch_cases.items():
```
romain-intel: I am not sure we need to validate this pro-actively here. Can we just not check that the selected one is a step? It feels a bit wasteful to do this. Granted, we won't catch as many errors but we would still catch the ones we care about.

romain-intel (commented):
DIdn't fully review but a few quick ones. Another one to think about a bit is the client. We briefly mentioned that right now it would not list any of the skipped steps (they wouldn't even exist as they are not in the metadata). Thoughts on that?

test_runner
mt-ob: this is fixed using `SKIP_GRAPHS` and `ONLY_GRAPHS`

saikonen (commented):
The output of `FlowGraph` needs some work with the graph structure, as currently both MFGUI and default Card visualizations for the DAG are incomplete (they only list the first conditional step, not all of them)

test this with the following pytest file

*test_graph.py*
```python
from metaflow.graph import FlowGraph

# A test flow class
from metaflow import step, FlowSpec, kubernetes, card

class ConditionalFlow(FlowSpec):
    @card
    @step
    def start(self):
        print("Starting 👋")

        self.conditional_value = "false"

        self.next({"true": self.branch_a, "false": self.branch_b}, condition="conditional_value")

    @step
    def branch_a(self):
        print("Now in Branch A")

        self.next(self.end)

    @step
    def branch_b(self):
        print("Now in Branch B")

        self.next(self.end)

    @step
    def end(self):
        print("Done! 🏁")


# testing the flowgraph info outputs for correctness
def test_flowgraph():

    graph = FlowGraph(ConditionalFlow)

    steps_info_dict, steps_list = graph.output_steps()

    assert "branch_a" in steps_info_dict
    assert "branch_b" in steps_info_dict
    
    assert  ["start", [["branch_a"], ["branch_b"]], "end"] == steps_list


if __name__=="__main__":
    ConditionalFlow()
```

```sh
pytest test_graph.py
```

saikonen:
DAG Visualisations checklist for conditional step flows:

- [x] MFGUI DAG works?
- [ ] Default Card DAG component works?

metaflow/flowspec.py
```diff
@@ -843,6 +882,92 @@ def next(self, *dsts: Callable[..., None], **kwargs) -> None:
...
+                    "Step *{step}* has an invalid self.next() transition. "
+                    "When using 'condition', provide exactly one dictionary argument "
+                    "mapping condition values to step methods.".format(step=step)
+                )
+                raise InvalidNextException(msg)
+
+            if not isinstance(condition, basestring):
+                msg = (
+                    "Step *{step}* has an invalid self.next() transition. "
+                    "The argument to 'condition' must be a string.".format(step=step)
```
romain-intel, line 900: this could also be a config technically.

metaflow/cli_components/run_cmds.py
```diff
@@ -230,6 +231,16 @@ def resume(
                     step_to_rerun, ",".join(list(obj.graph.nodes.keys()))
                 )
             )
+
+        origin_run = Run(f"{obj.flow.name}/{origin_run_id}")
```
savingoyal: this will likely add some delay to all resume executions. also, you would want to set namespace_check to false

savingoyal: is it feasible to add a check later when actually running the resume instead? we can add a warning instead of throwing an error at that time.

mt-ob: hmm, this would mean that resuming a branch not taken would result in success i.e. the current behaviour without this fix... but just with an added warning...

I think these warnings can be lost in the sea of logs and thus will be misleading if a part of CI / CD etc. / if the decision to deploy depends on this therefore retaining the ambiguous, untrustworthy behaviour

so, i advocate for explicit error instead aka the changes I did

mt-ob: added a TODO for now, to be done in a subsequent PR

mt-ob:
> * [ ]  Default Card DAG component works?

To be done in a subsequent PR
cc: @saikonen

metaflow/flowspec.py
mt-ob: done

metaflow/flowspec.py
mt-ob: right now, we allow only keys of the dictionary to be configs
the condition variable has to be a string i.e.
```py
self.path_choice = "high" if self.value > self.threshold else "low"
self.next({self.c.path: self.high_path, "low": self.low_path}, condition="path_choice")
```

where `self.c.path` is a config value named `"high"`
but condition itself has to be a string i.e. `"path_choice"`

metaflow/flowspec.py
mt-ob: not sure i understand this, we do allow statements of this type:
```py
self.next({self.c.path: self.high_path, "low": self.low_path}, condition="path_choice")
```

metaflow/flowspec.py
mt-ob: @savingoyal thoughts? I think we should still keep it...

metaflow/flowspec.py
romain-intel: The code seems to have things like `case_key.startswith("config:")` which seems un-necessary if you allow the `self.c.path`

metaflow/flowspec.py
romain-intel: That's fair.

metaflow/flowspec.py
romain-intel: I'm just not 100% sure of the benefit to check this here -- it provides no value to the user except in a type of debugging (ie: whatever happens for this condition, I will always have a valid out). We could provide an additional check if needed but it feels really wasteful to check if non-taken paths are valid ones.

metaflow/task.py
```diff
@@ -353,6 +353,27 @@ def lineage():
...
+            if hasattr(self.flow, "_switch_cases"):
+                switch_cases = self.flow._switch_cases
+                if condition_value not in switch_cases:
+                    available_cases = list(switch_cases.keys())
+                    raise RuntimeError(
+                        f"Switch condition variable '{node.condition}' has value '{condition_value}' "
+                        f"which is not in the available cases: {available_cases}"
+                    )
+                chosen_step = switch_cases[condition_value]
+                self.flow._transition = ([chosen_step], None)
```
romain-intel: I'm confused by this -- it seems to be seting `_transition` but this could be done directly in `FlowSpec.next`. I am confused why we have this split approach here. When `self.next` is called, we could do all this stuff then and there and set `self._transition` directly. I am also not 100% sure we need to have `_switch_cases` and `_switch_condition` stored as artifacts. If we do include them, we need to add them to `INTERNAL_ARTIFACTS_SET` else things may not work nicely with `merge_artifacts` (but I don't think we need them at all -- `_foreach_var` is used but only internally and not exposed to the user really).

metaflow/task.py
```diff
@@ -671,30 +692,50 @@ def run_step(
...
+                                "were found" % (step_name, expected_inputs, len(inputs))
+                            )
+                    if join_type == "split-switch":
+                        if len(inputs) > 1:
+                            raise MetaflowInternalError(
+                                f"Step *{step_name}* is a switch join but gets multiple inputs"
+                            )
+                        self.flow._set_datastore(inputs[0])
+                        # Initialize parameters for switch joins
+                        if input_paths:
```
romain-intel: is there a case where we don't have input_paths? The below code also seems to only use inputs[0]. It is identical to the code in the else clause. Is there a reason they are separate like this? If so, could you add a comment as to why this is the way it is.

metaflow/util.py
```diff
@@ -212,6 +212,35 @@ def write_latest_run_id(obj, run_id):
         f.write(str(run_id))
 
 
+def get_split_branch_for_node(graph, node_name, split_node_name):
```
romain-intel, line 215: I need to look a bit more into this but I wonder if we can save state so we don't need to do this. We do save the foreach stack, could we save something similar? It feels like we shouldn't need to figure out where we came from (since we just came from it)

metaflow/flowspec.py
mt-ob: That's because we encode it here first: https://github.com/Netflix/metaflow/pull/2473/files#diff-7a4019f33a31d2dce9a8c7235ccb968808d8a6c9919cd5ac5e270d044290e756R108

but happy to hear suggestions on how to improve this..

metaflow/flowspec.py
mt-ob: resolving...

metaflow/flowspec.py
mt-ob: ok, I only validate the chosen step now

metaflow/task.py
mt-ob: thanks, now I set it directly inside `FlowSpec.next` removing the `_validate_switch_transition` method altogether in `task.py`. Also removed `_switch_cases` and `_switch_condition`

metaflow/task.py
mt-ob: the difference was in value of `passdown`
I have updated the code a bit for better readability

metaflow/util.py
mt-ob: we can perhaps use something like `_split_branch_stack` analogous to `_foreach_stack`
but that will help us with `runtime.py`
however, this function is also used within `lint.py` -- do we remove this check from here?

PR #2491, opened 2025-07-15
fix: uv bootstrapping

saikonen:
`read_info_file` was relocated in the recent release. fix import in uv bootstrap


savingoyal:
let's cut a release too

PR #2506, opened 2025-07-20
Feature: Support spinning a single metaflow step (Rebased)

talsperre:
To test `spin` on a new flow you can do the following:

### Simple case:
```
python <flow_name.py> --environment=conda spin <step_name>
```

### Pass in specific pathspec:
```
python runtime_dag_flow.py --environment=conda spin --spin-pathspec RuntimeDAGFlow/13/step_c/275232971
```

### Pass in custom artifacts via module:
```
python runtime_dag_flow.py spin --spin-pathspec RuntimeDAGFlow/13/step_d/275233082 --artifacts-module ./my_artifacts.py
```

### Skip decorators (including the whitelisted ones):
```
python complex_dag_flow.py --environment=conda spin step_d --skip-decorators
```

### Use with Runner API:
```
    with Runner('complex_dag_flow.py', environment="conda").spin(
        step_name,
        spin_pathspec="<Some Val>",
        artifacts_module='./artifacts/complex_dag_step_d.py',
    ) as spin:
        print("-" * 50)
        print(f"Running test for step: step_a")
        spin_task = spin.task
        print(f"my_output: {spin_task['my_output']}")
        assert spin_task['my_output'].data == [-1]
```

See the tests for more examples on hot to use this command.

### DIFF
--- a/metaflow/client/core.py
+++ b/metaflow/client/core.py
@@ -207,7 +207,7 @@ def default_namespace() -> str:
     return get_namespace()
 
 
-def inspect_spin(datastore_root : str = "."):
+def inspect_spin(datastore_root: str = "."):
     """
     Set metadata provider to spin metadata so that users can inspect spin
     steps, tasks, and artifacts.


metaflow/datastore/task_datastore.py
```diff
@@ -250,6 +254,70 @@ def init_task(self):
         """
         self.save_metadata({self.METADATA_ATTEMPT_SUFFIX: {"time": time.time()}})
 
+    @only_if_not_done
+    @require_mode("w")
+    def transfer_artifacts(self, other_datastore, names=None):
```
mt-ob: minor nit: can we add types here for the args i.e. `other_datastore` and `names`

metaflow/datastore/content_addressed_store.py
```diff
@@ -65,6 +65,9 @@ def save_blobs(self, blob_iter, raw=False, len_hint=0):
             Whether to save the bytes directly or process them, by default False
```
mt-ob, line 66: can we add `default False` for `raw` instead of `optional`

metaflow/datastore/flow_datastore.py
```diff
@@ -265,3 +374,11 @@ def load_data(self, keys, force_raw=False):
         """
         for key, blob in self.ca_store.load_blobs(keys, force_raw=force_raw):
             yield key, blob
+
+
+class MetadataCache(object):
```
mt-ob: can this be made an abstract class instead? with the methods raising `NotImplementedError`?

metaflow/util.py
```diff
@@ -178,6 +179,45 @@ def resolve_identity():
     return "%s:%s" % (identity_type, identity_value)
 
 
+def get_latest_task_pathspec(flow_name: str, step_name: str) -> (str, str):
```
mt-ob: I guess this returns a `Task` instance so the function can be renamed since it doesn't just return the pathspec

metaflow/util.py
```diff
@@ -471,3 +511,38 @@ def to_pod(value):
...
+
+    Raises
+    -------
+    MetaflowInternalError
+        If the file cannot be read or does not contain the ARTIFACTS variable.
+    """
+    import importlib.util
+
+    try:
+        spec = importlib.util.spec_from_file_location("artifacts_module", file_path)
```
mt-ob: `"artifacts_module"` can be replaced with actual name of the file such as

`os.path.splitext(os.path.basename(file_path))[0]`

metaflow/task.py
```diff
@@ -521,25 +546,33 @@ def run_step(
 
         step_func = getattr(self.flow, step_name)
         decorators = step_func.decorators
-
+        if self.orig_flow_datastore:
+            # We filter only the whitelisted decorators in case of spin step.
+            decorators = [
+                deco for deco in decorators if deco.name in whitelist_decorators
```
mt-ob: maybe an additional check for `whitelist_decorators is not None or []` can be helpful here?

metaflow/client/filecache.py
```diff
@@ -63,8 +65,8 @@ def __init__(self, cache_dir=None, max_size=None):
         # when querying for sizes of artifacts. Once we have queried for the size
         # of one artifact in a TaskDatastore, caching this means that any
         # queries on that same TaskDatastore will be quick (since we already
-        # have all the metadata)
-        self._task_metadata_caches = OrderedDict()
+        # have all the metadata). We keep track of this in a file so it persists
```
mt-ob, line 68: I wonder if same should be done for `self._store_caches`? which is right now an `OrderedDict `

metaflow/client/core.py
```diff
@@ -1176,149 +1176,191 @@ class Task(MetaflowObject):
...
             for task_pathspec in task_pathspecs:
-                yield Task(pathspec=task_pathspec, _namespace_check=False)
+                yield task_pathspec
+
+    @staticmethod
+    def _get_previous_steps(graph_info, step_name):
+        # Get the parent steps
+        steps = []
+        for node_name, attributes in graph_info["steps"].items():
+            if step_name in attributes["next"]:
```
mt-ob, line 1242: I wonder if `in_funcs` can be used here somehow...

metaflow/client/core.py
```diff
@@ -1176,149 +1176,191 @@ class Task(MetaflowObject):
...
+                        pattern = ".*"
+                    else:
+                        # (Current task, "A:10,B:13,C:21") and (Parent task, "A:10,B:13")
+                        # (Current task, "A:10,B:13") and (Parent task, "A:10,B:13")
+                        if parent_step_type == "split-foreach":
+                            target_depth = current_depth - 1
+                        pattern = ",".join(current_path.split(",")[:target_depth])
 
-        yield from self._iter_matching_tasks(steps, "foreach-execution-path", pattern)
+        metadata_key = "foreach-execution-path"
```
mt-ob: this can be declared way earlier and then re-used instead of `current_path = metadata_dict.get("foreach-execution-path")`

metaflow/client/core.py
```diff
@@ -1176,149 +1176,191 @@ class Task(MetaflowObject):
...
+                        # (Current task, "A:10") and (Child task, "")
+                        pattern = ".*"
+                    else:
+                        # (Current task, "A:10,B:13,C:21") and (Child task, "A:10,B:13")
+                        # (Current task, "A:10,B:13") and (Child task, "A:10,B:13")
+                        if child_step_type == "join":
+                            target_depth = current_depth - 1
+                        pattern = ",".join(current_path.split(",")[:target_depth])
+
+        metadata_key = "foreach-execution-path"
```
mt-ob: same comment as before about declaring earlier

metaflow/runtime.py
```diff
@@ -74,6 +82,226 @@
...
+        self._environment = environment
+        self._package = package
+        self._logger = logger
+        self._entrypoint = entrypoint
+        self._event_logger = event_logger
+        self._monitor = monitor
+
+        self._step_func = step_func
+
+        # Verify whether the use has provided step-name or spin-pathspec
```
mt-ob: nit: `the use` --> `the user`

metaflow/cli_components/run_cmds.py
```diff
@@ -386,3 +398,113 @@ def run(
...
+    "the artifact names and values as the artifact values. The artifact values will "
+    "overwrite the default values of the artifacts used in the spun step.",
+)
+@click.option(
+    "--persist/--no-persist",
+    "persist",
+    default=True,
+    show_default=True,
+    help="Whether to persist the artifacts in the spun step. If set to False, "
+    "the artifacts will notbe persisted and will not be available in the spun step's "
```
mt-ob: nit: `notbe` --> `not be`

metaflow/cli_components/step_cmd.py
```diff
@@ -176,3 +179,161 @@ def step(
...
+):
+    import time
+
+    if ctx.obj.is_quiet:
+        echo = echo_dev_null
+    else:
+        echo = echo_always
+
+    if opt_namespace is not None:
+        namespace(opt_namespace or None)
```
mt-ob: we already checked `None` in the `if` statement
so can avoid `opt_namespace or None` I guess

metaflow/cli_components/step_cmd.py
```diff
@@ -176,3 +179,161 @@ def step(
...
+    input_paths = decompress_list(input_paths) if input_paths else []
+
+    whitelist_decorators = (
+        decompress_list(whitelist_decorators) if whitelist_decorators else []
+    )
+    from_start("SpinStep: initialized decorators")
+    spin_artifacts = read_artifacts_module(artifacts_module) if artifacts_module else {}
+    from_start("SpinStep: read artifacts module")
+
+    ds_type, ds_root = orig_flow_datastore.split("@")
```
mt-ob, line 289: can this be `None` too? because that's the default value when defining this arg

metaflow/runner/metaflow_runner.py
```diff
@@ -193,6 +198,76 @@ async def stream_log(
...
+    Note that you should use either this object as the context manager or
+    `Runner`, not both in a nested manner.
+    """
+
+    def __init__(
+        self, runner: "Runner", command_obj: CommandManager, task_obj: "metaflow.Task"
+    ) -> None:
+        """
+        Create a new ExecutingTask -- this should not be done by the user directly but
+        instead user Runner.spin()
```
mt-ob: nit: `user` --> `use`

metaflow/runner/metaflow_runner.py
```diff
@@ -193,6 +198,76 @@ async def stream_log(
...
+    Note that you should use either this object as the context manager or
+    `Runner`, not both in a nested manner.
+    """
+
+    def __init__(
+        self, runner: "Runner", command_obj: CommandManager, run_obj: Run
+    ) -> None:
+        """
+        Create a new ExecutingRun -- this should not be done by the user directly but
+        instead user Runner.run()
```
mt-ob: same nit: `user` --> `use`

metaflow/runner/metaflow_runner.py
```diff
@@ -193,6 +198,76 @@ async def stream_log(
...
+    ```python
+    with Runner(...).spin() as running:
+        ...
+    ```
+    Note that you should use either this object as the context manager or
+    `Runner`, not both in a nested manner.
+    """
+
+    def __init__(
+        self, runner: "Runner", command_obj: CommandManager, task_obj: "metaflow.Task"
```
mt-ob: can import `from metaflow import Task` directly and use that as type instead of `"metaflow.Task"`, to be consistent with how it's done for ExecutingRun

metaflow/runner/metaflow_runner.py
```diff
@@ -397,6 +472,73 @@ def run(self, **kwargs) -> ExecutingRun:
...
+        content = handle_timeout(attribute_file_fd, command_obj, self.file_read_timeout)
+
+        command_obj.sync_wait()
+
+        content = json.loads(content)
+        pathspec = f"{content.get('flow_name')}/{content.get('run_id')}/{content.get('step_name')}/{content.get('task_id')}"
+
+        # Set the correct metadata from the runner_attribute file corresponding to this run.
+        metadata_for_flow = content.get("metadata")
+        from metaflow import Task
```
mt-ob: can be moved to the top, and thus the type can also be changed to `Task` instead of `"metaflow.Task"` in the `__init__` of `ExecutingTask`

metaflow/runner/metaflow_runner.py
```diff
@@ -57,7 +63,6 @@ def __init__(
         """
         self.runner = runner
         self.command_obj = command_obj
-        self.run = run_obj
 
     def __enter__(self) -> "ExecutingRun":
```
mt-ob: shouldn't the return type now be `ExecutingProcess` instead?

mt-ob: same for other functions such as `async def wait`

metaflow/runner/metaflow_runner.py
```diff
@@ -397,6 +472,73 @@ def run(self, **kwargs) -> ExecutingRun:
...
+        content = await async_handle_timeout(
+            attribute_file_fd, command_obj, self.file_read_timeout
+        )
+        content = json.loads(content)
+        pathspec = f"{content.get('flow_name')}/{content.get('run_id')}/{content.get('step_name')}/{content.get('task_id')}"
+
+        # Set the correct metadata from the runner_attribute file corresponding to this run.
+        metadata_for_flow = content.get("metadata")
+
+        from metaflow import Task
```
mt-ob: same comment, can be imported at the top

metaflow/runner/metaflow_runner.py
mt-ob: maybe some functions specific to `ExecutingRun` (if there are any) will have to moved to the new subclass, would be good to check

metaflow/cli_components/step_cmd.py
talsperre: Just copied it from the `run` command. Will change it.

metaflow/cli_components/step_cmd.py
talsperre: No it can't.

metaflow/runner/metaflow_runner.py
talsperre: Will that possibly cause any issues wrt initializing metadata etc? If not I can make this change.

metaflow/cli.py
```diff
@@ -318,6 +322,13 @@ def version(obj):
     hidden=True,
     is_eager=True,
 )
+@click.option(
+    "--spin-mode",
+    is_flag=True,
+    default=False,
+    help="Enable spin mode for metaflow cli commands. Setting this flag will result "
```
romain-intel: nit: CLI

metaflow/cli.py
```diff
@@ -318,6 +322,13 @@ def version(obj):
     hidden=True,
     is_eager=True,
 )
+@click.option(
+    "--spin-mode",
```
romain-intel: not sure it is default but other commands we use `--spin-mode/--no-spin-mode`. Wonder if we shorten to just `--spin`. Or maybe `--mode spin` and we can also have a `--mode local` which was something @savingoyal had brought up at some point (maybe scope creep to add that but if we are considering it, we can go that naming for now).

metaflow/cli_components/run_cmds.py
```diff
@@ -36,26 +37,36 @@ def before_run(obj, tags, decospecs):
...
+        # obj.environment.init_environment(obj.logger)
+
+        decorators._init_step_decorators(
+            obj.flow,
+            obj.graph,
+            obj.environment,
+            obj.flow_datastore,
+            obj.logger,
+            obj.is_spin,
+            skip_decorators,
```
romain-intel, line 72: is this different from obj.skip_decorators? If not, can we just use that and avoid adding an extra argument here?

metaflow/cli_components/run_cmds.py
```diff
@@ -401,3 +423,109 @@ def run(
             )
     with runtime.run_heartbeat():
         runtime.execute()
+
+
+@parameters.add_custom_parameters(deploy_mode=True)
+@click.command(help="Spins up a task for a given step from a previous run locally.")
+@tracing.cli("cli/spin")
+@click.argument("pathspec")
+@click.option(
```
romain-intel, line 436: nit: maybe add a comment that the default here matches the direction of the check in saved_args a bit above.

metaflow/cli_components/run_cmds.py
```diff
@@ -401,3 +423,109 @@ def run(
...
+    show_default=True,
+    help="Path to a module that contains artifacts to be used in the spun step. "
+    "The artifacts should be defined as a dictionary called ARTIFACTS with keys as "
+    "the artifact names and values as the artifact values. The artifact values will "
+    "overwrite the default values of the artifacts used in the spun step.",
+)
+@click.option(
+    "--persist/--no-persist",
+    "persist",
+    default=True,
```
romain-intel: did we want to make this configurable? I believe OSS would prefer it to be False by default but we are probably going to set it to True internally.

metaflow/cli_components/step_cmd.py
```diff
@@ -176,3 +180,155 @@ def step(
...
     echo("Success", fg="green", bold=True, indent=True)
+
+
+@click.command(help="Internal command to spin a single task.", hidden=True)
+@click.argument("step-name")
+@click.option(
+    "--run-id",
+    default=None,
+    required=True,
+    help="Run ID for the step that's about to be spun",
```
romain-intel: nit: maybe add "Original run ID for the step that will be spun" or something like that. I suppose that --run-id, --task-id and --orig-flow-datastore all point to the task that is being spun up.

romain-intel: It's an internal command so doesn't matter as much but just for future people reading the code :)

metaflow/client/filecache.py
```diff
@@ -393,32 +400,49 @@ def _get_task_datastore(
...
-        # If we are here, we either have attempt=None or nothing in the cache
-        task_ds = flow_ds.get_task_datastore(
-            run_id, step_name, task_id, attempt=attempt
+
+        return flow_ds.get_task_datastore(run_id, step_name, task_id, attempt=attempt)
+
+
+class TaskMetadataCache(MetadataCache):
+    def __init__(self, filecache, ds_type, ds_root, flow_name):
+        self._ds_type = ds_type
```
romain-intel, line 410: itty bitty nit: initialize in same order as args.

metaflow/client/filecache.py
```diff
@@ -393,32 +400,49 @@ def _get_task_datastore(
...
+        return os.path.join(self._filecache.cache_dir, cache_id, token[:2], token)
+
+    def load_metadata(self, run_id, step_name, task_id, attempt):
+        d = self._filecache.read_file(self._path(run_id, step_name, task_id, attempt))
+        if d:
+            return json.loads(d)
+
+    def store_metadata(self, run_id, step_name, task_id, attempt, metadata_dict):
+        self._filecache.create_file(
+            self._path(run_id, step_name, task_id, attempt),
```
romain-intel, line 446: is attempt always not None here? If not, there could be an issue.

metaflow/datastore/content_addressed_store.py
```diff
@@ -65,6 +65,9 @@ def save_blobs(self, blob_iter, raw=False, len_hint=0):
...
         len_hint : Hint of the number of blobs that will be produced by the
             iterator, by default 0
+        _is_transfer : bool, default False
```
romain-intel: nit: no need for the _ here I guess.

metaflow/client/core.py
romain-intel: +1 -- I believe you could use in_funcs to get the step names of the previous step directly without iterating over the graph.

metaflow/client/core.py
```diff
@@ -1181,149 +1191,191 @@ class Task(MetaflowObject):
...
+        # Get the parent steps
+        steps = []
+        for node_name, attributes in graph_info["steps"].items():
+            if step_name in attributes["next"]:
+                steps.append(node_name)
+        return steps
 
     @property
-    def parent_tasks(self) -> Iterator["Task"]:
+    def parent_task_pathspecs(self) -> Iterator[str]:
```
romain-intel, line 1247: I suspect this may not work with conditionals btw. There may be some corner cases thers. @mt-ob may be able to help there and handle that. Basically graph_info captures the static information but would miss which task was actually chosen in the switch.

metaflow/datastore/content_addressed_store.py
```diff
@@ -111,23 +128,32 @@ def load_blobs(self, keys, force_raw=False):
...
         keys : List of string
             Key describing the object to load
-        force_raw : bool, optional
+        force_raw : bool, default False
             Support for backward compatibility with previous datastores. If
             True, this will force the key to be loaded as is (raw). By default,
             False
+        _is_transfer : bool, default False
+            If True, this indicates we are loading blobs to transfer them directly
+            to another datastore. We will, in this case, also transfer the metdata
```
romain-intel: nit: metadata

metaflow/datastore/flow_datastore.py
```diff
@@ -106,6 +118,16 @@ def get_task_datastores(
             If True, returns all attempts up to and including attempt.
         mode : str, default "r"
             Mode to initialize the returned TaskDataStores in.
+        join_type : str, optional
+            If specified, the join type for the task. This is used to determine
+            the user specified artifacts for the task in case of a spin task.
+        orig_flow_datastore : MetadataProvider, optional
```
romain-intel: nit: default None

metaflow/datastore/flow_datastore.py
```diff
@@ -106,6 +118,16 @@ def get_task_datastores(
...
+            The metadata provider in case of a spin task. If provided, the
+            returned TaskDataStore will be a SpinTaskDatastore instead of a
+            TaskDataStore.
+        spin_artifacts : Dict[str, Any], optional
```
romain-intel: ditto

metaflow/datastore/flow_datastore.py
```diff
@@ -106,6 +118,16 @@ def get_task_datastores(
             If True, returns all attempts up to and including attempt.
         mode : str, default "r"
             Mode to initialize the returned TaskDataStores in.
+        join_type : str, optional
```
romain-intel: nit: default None

metaflow/datastore/flow_datastore.py
```diff
@@ -212,8 +251,64 @@ def get_task_datastore(
...
+            # Get the task datastore for the spun task.
+            orig_datastore = orig_flow_datastore.get_task_datastore(
+                run_id,
+                step_name,
+                task_id,
+                attempt=attempt,
+                data_metadata=data_metadata,
+                mode=mode,
+                allow_not_done=allow_not_done,
+                join_type=join_type,
```
romain-intel: nit: join_type does not need to be passed -- it is never used without orig_flow_datastore.

metaflow/datastore/task_datastore.py
```diff
@@ -253,6 +258,70 @@ def init_task(self):
         """
         self.save_metadata({self.METADATA_ATTEMPT_SUFFIX: {"time": time.time()}})
 
+    @only_if_not_done
+    @require_mode("w")
+    def transfer_artifacts(self, other_datastore : "TaskDataStore", names : List[str] =None):
```
romain-intel: type is wrong here.

metaflow/datastore/content_addressed_store.py
```diff
@@ -76,6 +79,20 @@ def save_blobs(self, blob_iter, raw=False, len_hint=0):
 
         def packing_iter():
             for blob in blob_iter:
+                if _is_transfer:
+                    key, blob_data, meta = blob
+                    path = self._storage_impl.path_join(self._prefix, key[:2], key)
+                    # Transfer data is always raw/decompressed, so mark it as such
+                    meta_corrected = {"cas_raw": True, "cas_version": 1}
```
romain-intel, line 88: I am not quite sure I understand this override.

metaflow/datastore/content_addressed_store.py
```diff
@@ -111,23 +128,32 @@ def load_blobs(self, keys, force_raw=False):
...
         load_paths = []
         for key in keys:
             blob = None
             if self._blob_cache:
                 blob = self._blob_cache.load_key(key)
             if blob is not None:
-                yield key, blob
+                if _is_transfer:
+                    # Cached blobs are decompressed/processed bytes regardless of original format
+                    yield key, blob, {"cas_raw": False, "cas_version": 1}
```
romain-intel, line 156: I think I see the point here -- because we unpack prior to storing in the cache but I think we should store the raw bit so we can pass it back to save_blobs for transfer. I don't think this will cover all cases otherwise.

metaflow/cli_components/run_cmds.py
talsperre: ACK, I will do that.

metaflow/cli_components/run_cmds.py
talsperre: They are supposed to be the same. But we need special logic to check for `skip_decorators` in `cli.py` because other run level commands like run/resume do not have this option.

metaflow/client/core.py
talsperre: ACK - I don't think spin works well with conditionals (at least not yet). We can do this as a follow-up later.

metaflow/client/core.py
talsperre: Will address this and potentially the logic for RuntimeDAG later in a follow up PR.

metaflow/client/filecache.py
talsperre: Will raise an exception.

metaflow/client/core.py
```diff
@@ -207,6 +207,15 @@ def default_namespace() -> str:
     return get_namespace()
 
 
+def inspect_spin(datastore_root):
```
romain-intel: minor nit looking at the blog: we can make the default be "." since that will be the common case I suppose.

talsperre: Updated in latest commit.

PR #2507, opened 2025-07-20
Fix/runner packaging

romain-intel:


romain-intel:
Merging due to no one opposing it :).

PR #2515, opened 2025-07-22
fix: triggers not working with argo parallel

saikonen:
triggers are not working when running with argo workflows in combination with `@parallel`

problem seems to stem from argo having different quotations for parameters stemming from event payloads vs. from regular workflow submissions. this in turn breaks the pod spec in the manifest for jobsets


valayDave:
Closing this in favor of #2516

PR #2525, opened 2025-07-28
vendor pure pythonic pyyaml

savingoyal:


romain-intel (changes requested):
6.0.2 is 3.8+ only -- we support 3.6+

savingoyal:
> 6.0.2 is 3.8+ only -- we support 3.6+

updated to 5.3.1

PR #2527, opened 2025-07-29
fix: metaflow-dev improvements

saikonen:
- enables `metaflow-dev up` to find the relevant Makefile when metaflow installed with `--editable`
- updates both metaflow-service and metaflow_ui image tags to official arm64 supported releases
- add cleanup of `AWS_PROFILE` env var that interferes with the local minio profiles set up by `metaflow-dev shell`

### DIFF
--- a/metaflow/cmd/make_wrapper.py
+++ b/metaflow/cmd/make_wrapper.py
@@ -46,14 +46,17 @@ def find_makefile():
             )
             if makefile_candidate.is_file():
                 return makefile_candidate
+        else:
+            # No dist metadata found. This is tied to the version of pip being used
+            # Do not bother with .egg-link installs due to the handling of the file contents being a headache due to lack of a unified spec.
+            print(
+                "Could not locate an installation of Metaflow. No package metadata found."
+            )
+            print(
+                "If Metaflow is installed as editable, try upgrading the version of pip and reinstalling in order to generate proper package metadata.\n"
+            )
     except Exception:
         return None
-
-    # Do not bother with .egg-link installs produced by Python <=3.8 due to the handling of the file contents being a headache due to lack of a unified spec.
-    if sys.version_info[0] <= 3 and sys.version_info[1] <= 8:
-        print(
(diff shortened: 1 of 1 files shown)


metaflow/cmd/make_wrapper.py
```diff
@@ -28,6 +28,27 @@ def find_makefile():
         if makefile_candidate.is_file():
             return makefile_candidate
 
+    # 4) When developing, mMtaflow might be installed with --editable, which means the devtools will not be located within site-packages.
```
mt-ob: typo: `mMtaflow`

metaflow/cmd/make_wrapper.py
```diff
@@ -28,6 +28,27 @@ def find_makefile():
...
+    # We read the actual location from package metadata in this case, but only do this heavier operation if the above lookups fail.
+    try:
+        import json
+        from importlib.metadata import Distribution
```
savingoyal, line 35: for py < 3.10, you would want to do - from metaflow._vendor.importlib_metadata ..

saikonen: using the vendored importlib won't actually solve the issue for older python versions, as editable installs with python <=3.8 do not install dist files, instead producing an `.egg-link` which is not supported by the importlib metadata.

the difference in install files is most likely due to the setuptools versions that the different python version picks. I can either introduce separate logic for <=3.8 that reads the editable path from the egg-link file, or we leave older versions affected. `3.9` onward works as expected with the default importlib so no need to use vendored one for these

savingoyal: at least a check that fails nicely would be great

saikonen: added a detailed error for the editable lookup failure.

while testing this, some of my virtualenvs were installing metaflow with an egg-link, while others were generating the correct dist metadata for lookup. Tried versions 3.8->3.12
The culprit here seems to be the version of pip. `23.0.1` which is the default for many venvs, installs via egg-link, while `pip install pip --upgrade` bumps this to `25.0.1` which installs dist metadata on all of the tested python versions.

savingoyal (approved):

PR #2534, opened 2025-08-05
feat: add S3 command upload support for AWS Step Functions to handle large commands

nlwstein:
## Overview
This PR adds support for uploading commands to S3 when using AWS Step Functions, preventing AWS Batch command size limitations. When the compression feature is enabled, all commands are automatically uploaded to S3 and replaced with download/execute commands. This intends to at least partially resolve https://github.com/Netflix/metaflow/issues/1482.

## Problem
AWS Batch has command size limitations that can cause deployment failures for complex Metaflow workflows with long command strings. This is particularly problematic for flows with many decorators or complex parameter configurations.

## Solution
- Upload all commands to S3 when compression is enabled (no size threshold)
- Replace batch job commands with S3 download and execute commands
- Graceful fallback to inline commands if S3 upload fails
- Simplified configuration with a single on/off switch

### Environment Variables
- `METAFLOW_SFN_COMPRESS_STATE_MACHINE` (default: `false`)
  - Set to `true` to enable S3 command upload functionality

### Constructor Parameters
- `compress_state_machine` (default: `false`)
  - Boolean flag to enable/disable the feature

### S3 Storage Path
Commands are stored at: `{datastore_root}/commands/{flow_name}/{deployment_id}/{node_name}/command.sh`

## Usage Example
```python
# Enable via environment variable
export METAFLOW_SFN_COMPRESS_STATE_MACHINE=true

# Or enable programmatically
@step_functions(compress_state_machine=True)
def my_flow():
    # ... flow definition
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -72,7 +72,7 @@ jobs:
 
     - name: Install R ${{ matrix.ver }} system dependencies
       if: matrix.os == 'ubuntu-22.04'
-      run: sudo apt-get update; sudo apt-get install -y libcurl4-openssl-dev qpdf libgit2-dev libharfbuzz-dev libfribidi-dev
+      run: sudo apt-get update; sudo apt-get install -y libcurl4-openssl-dev qpdf libgit2-dev libharfbuzz-dev libfribidi-dev libwebp-dev
 
     - name: Install R ${{ matrix.ver }} Rlang dependencies
       run: |
--- a/metaflow/metadata_provider/metadata.py
+++ b/metaflow/metadata_provider/metadata.py
@@ -632,7 +632,12 @@ class MetadataProvider(object):
 
     def _get_git_info_as_dict(self):
         git_info = {}
-        env = self._environment.get_environment_info()
+        # NOTE: For flows executing remotely, we want to read from the INFO file of the code package that contains
+        # information on the original environment that deployed the flow.
+        # Otherwise git related info will be missing, as the repository is not part of the codepackage.
+        from metaflow.packaging_sys import MetaflowCodeContent
+
+        env = MetaflowCodeContent.get_info() or self._environment.get_environment_info()
         for key in [
             "repo_url",
             "branch_name",
--- a/metaflow/version.py
+++ b/metaflow/version.py
@@ -1,1 +1,1 @@
-metaflow_version = "2.18.0"
+metaflow_version = "2.18.2"


savingoyal:
Thanks for the PR! I am yet to spend time with it deeply but let me quickly give some high-level feedback inline

metaflow/plugins/aws/step_functions/step_functions_cli.py
```diff
@@ -122,15 +122,33 @@ def step_functions(obj, name=None):
...
-    "Logs log group.",
+    help="Log AWS Step Functions execution history to AWS CloudWatch Logs log group.",
 )
 @click.option(
     "--use-distributed-map/--no-use-distributed-map",
     is_flag=True,
     help="Use AWS Step Functions Distributed Map instead of Inline Map for "
     "defining foreach tasks in Amazon State Language.",
 )
+@click.option(
```
savingoyal, line 144: it might turn out to be a maintenance foot gun to offer this as an optional behaviour to the end user. we can do this by default instead.

metaflow/plugins/aws/step_functions/step_functions_cli.py
```diff
@@ -122,15 +122,33 @@ def step_functions(obj, name=None):
...
+    "--upload-commands-to-s3/--no-upload-commands-to-s3",
+    is_flag=True,
+    default=False,
+    help="Upload large commands to S3 instead of embedding them in job payload "
+    "to avoid 8K AWS limit. Commands will be downloaded at runtime.",
+)
+@click.option(
+    "--command-s3-path",
```
savingoyal: is there a scenario where the user will care about the location? much of datastore implementation is internal detail. offering this as a control increases the burden on ensuring backwards compatibility going forward.

metaflow/plugins/aws/step_functions/step_functions_cli.py
```diff
@@ -122,15 +122,33 @@ def step_functions(obj, name=None):
...
+    default=None,
+    help="S3 path prefix for uploading commands. If not specified, uses "
+    "the default datastore path with 'commands' suffix.",
+)
+@click.option(
+    "--dump-commands",
```
savingoyal: given that we have the --only-json flag, do we need this capability?

metaflow/plugins/aws/step_functions/step_functions_cli.py
savingoyal: actually it might be useful for [debugging](https://github.com/Netflix/metaflow/pull/2534/files#r2255015438). we can rename it to --compress-state-machine/--no-compress-state-machine so that we can also introduce other optimizations in the future.

metaflow/plugins/aws/step_functions/step_functions_cli.py
```diff
@@ -63,7 +63,7 @@ def step_functions(obj, name=None):
 
 
 @step_functions.command(
-    help="Deploy a new version of this workflow to " "AWS Step Functions."
+    help="Deploy a new version of this workflow to AWS Step Functions."
```
savingoyal: would it be possible to clean out unintended linting changes? helps massively with keeping the review focused.

metaflow/plugins/aws/step_functions/step_functions.py
```diff
@@ -887,6 +909,27 @@ def _batch(self, node):
             .attempts(total_retries + 1)
         )
 
+        # If S3 upload is enabled, we need to modify the command after it's created
```
savingoyal: another optimization worth thinking is packaging all the env vars in a file too

metaflow/plugins/aws/step_functions/step_functions.py
```diff
@@ -887,6 +909,27 @@ def _batch(self, node):
...
+        if self.upload_commands_to_s3:
+            # Get the command that was created
+            command = batch_job.payload["containerOverrides"]["command"]
+            command_str = " ".join(command)
+            if len(command_str) > self.command_size_threshold:
+                # Upload the command to S3 during deployment
+                s3_path = self._get_command_s3_path(node.name, self.deployment_id)
```
savingoyal: there are a few different ways to achieve the outcome - how about storing the commands within the metaflow code package instead? that way it is easy to do any kind of garbage collection in the future (a common ask)

savingoyal (changes requested):

metaflow/plugins/aws/step_functions/step_functions.py
nlwstein: We have some use cases where we introspect on deployed step functions - for example, discovering the url of a MF UI server. If the env vars were to be exposed as a file (and a single env var, say MF_ENV_VARS_FILE) this kind of approach would still be viable, but it could break things. I am happy to implement this if you think it would be broadly useful for others, but on our side we would need it gated by a separate flag. I do think this would be necessary to make hitting the 8K limit nearly impossible - but it has more potential drawbacks than the command which users really shouldn't need to inspect for anything.

metaflow/plugins/aws/step_functions/step_functions.py
nlwstein: I like that! I can refactor to store it in the code package.

metaflow/plugins/aws/step_functions/step_functions_cli.py
nlwstein: This was mostly useful during testing, totally fine w/ removing.

metaflow/plugins/aws/step_functions/step_functions.py
nlwstein: @savingoyal I spent some time looking into this (bear with me, I am new to the MF core!) and it's a bit of an ouroboros-like situation to move the command into the code bundle - the command *downloads* the code bundle. 

Therefore, in my opinion, storing the invocation script as a bare S3 file actually makes sense, since it's more of a bootstrap that grants us access to the code bundle in the first place, vs. a logical part of it.

metaflow/plugins/aws/step_functions/step_functions.py
nlwstein: @savingoyal TL;DR: I am happy to do this feature as well, but I suspect that it will have more potential impact on users than this one. Would it make sense to do this as a separate PR?

metaflow/plugins/aws/step_functions/step_functions_cli.py
nlwstein: @savingoyal Cleaned up my mess 🗑️

metaflow/plugins/aws/step_functions/step_functions_cli.py
nlwstein: @savingoyal Removed dump_commands and the whole concept of a threshold for uploading the command to S3 - the feature's just tied to the boolean option.

metaflow/plugins/aws/step_functions/step_functions_cli.py
nlwstein: Removed

metaflow/plugins/aws/step_functions/step_functions_cli.py
nlwstein: Renamed

nlwstein: @savingoyal Do you think we should enable this by default, then, going back to your original comment - and let it exist as a means to turn off the upload? I can't think of a particular reason why it would break anyone's setup, I just default to feature flagging new behaviors, so that's what I did here.

nlwstein:
@savingoyal Thank you for your feedback earlier! Just wondering if this is likely to be accepted - I have forked MF into our Gitlab so I can use this feature for some flows that are hobbled at the moment (as-is I cannot submit them with all of their parameters). I will update here again once I have it working end to end with the fork.

metaflow/plugins/aws/step_functions/step_functions.py
```diff
@@ -79,6 +80,18 @@ def __init__(
         # https://aws.amazon.com/blogs/aws/step-functions-distributed-map-a-serverless-solution-for-large-scale-parallel-data-processing/
         self.use_distributed_map = use_distributed_map
 
+        # S3 command upload configuration
+        self.upload_commands_to_s3 = (
+            compress_state_machine
+            or os.environ.get("METAFLOW_SFN_COMPRESS_STATE_MACHINE", "false").lower()
```
saikonen: the env var reading can be removed from this part completely. Instead I would suggest moving this into `metaflow_config.py` to
```
SFN_COMPRESS_STATE_MACHINE = from_conf("SFN_COMPRESS_STATE_MACHINE", False)
```
which will do the value casting automatically based on the given default.

Then the default can be set for the cli option directly instead.

metaflow/plugins/aws/step_functions/step_functions_cli.py
```diff
@@ -131,6 +131,14 @@ def step_functions(obj, name=None):
     help="Use AWS Step Functions Distributed Map instead of Inline Map for "
     "defining foreach tasks in Amazon State Language.",
 )
+@click.option(
+    "--compress-state-machine/--no-compress-state-machine",
+    is_flag=True,
+    default=False,
```
saikonen:
```
from metaflow.metaflow_config import SFN_COMPRESS_STATE_MACHINE
...

default=SFN_COMPRESS_STATE_MACHINE
```

metaflow/plugins/aws/step_functions/step_functions.py
```diff
@@ -899,6 +936,59 @@ def _get_retries(self, node):
...
+        # Use deployment-specific ID instead of run ID
+        if deployment_id is None:
+            # Use the instance deployment_id if available, otherwise generate a new one
+            deployment_id = getattr(self, "deployment_id", None)
+            if deployment_id is None:
+                import time
+
+                deployment_id = f"deployment-{int(time.time())}"
+
+        return f"{base_path}/{self.flow.name}/{deployment_id}/{node_name}/command.sh"
```
saikonen: Just making a note here that there might be a slight security concern with now having the entire batch command being pulled from a very guessable S3 location. Depending on the organization the bucket might have more lax permissions for access compared to who/what can deploy to SFN

This should not be a problem as is, as long as we keep the default for compression at false, and call out the implications in the release docs.

gregdes: What about passing a SHA or something similar through and verifying? Essentially a signature?

metaflow/plugins/aws/step_functions/step_functions.py
```diff
@@ -899,6 +936,59 @@ def _get_retries(self, node):
...
+            bucket, key = parse_s3_full_path(f"s3://{s3_path}")
+
+            # Upload directly to S3
+            s3_client = boto3.client("s3")
+            s3_client.upload_fileobj(command_stream, bucket, key)
+
+            # Return a command that downloads and executes the uploaded script
+            full_s3_path = f"s3://{bucket}/{key}"
+            download_cmd = (
+                f"aws s3 cp {full_s3_path} /tmp/step_command.sh && "
```
saikonen: users might have configured a custom `S3_ENDPOINT_URL` which will not be accounted for by the raw boto3 s3 client, but the aws s3 command should work, as access on the remote side should be configured correctly.

instead of boto3, we could use the metaflow S3 client here which should account for all configuration.
Also note comments about using the ContentAddressedStore (flow_datastore) as an alternative.

metaflow/plugins/aws/step_functions/step_functions.py
```diff
@@ -899,6 +936,59 @@ def _get_retries(self, node):
...
+            # Use the instance deployment_id if available, otherwise generate a new one
+            deployment_id = getattr(self, "deployment_id", None)
+            if deployment_id is None:
+                import time
+
+                deployment_id = f"deployment-{int(time.time())}"
+
+        return f"{base_path}/{self.flow.name}/{deployment_id}/{node_name}/command.sh"
+
+    def _upload_command_to_s3(self, command, s3_path):
```
saikonen: instead of manually constructing the s3 path and upload location, could the `ContentAddressedStore` (`flow_datastore`) be used instead? This would construct a hash-path based on the uploaded content, making it unique in the flow scope.

should work something like

```
saved_info = self.flow_datastore.save_data([command_stream], len_hint=1)
stored_url, _key = saved_info[0]
```

saikonen: using the flow datastore should also remove the need to construct a deployment_id

metaflow/plugins/aws/step_functions/step_functions.py
saikonen: another note on the download command: we do not require the aws CLI for bootstrapping anymore, so this command might fail. instead it could be an inlined python script, same as we do in `metaflow_environment.py#_get_download_code_package_cmd()`

saikonen:
@nlwstein I have pushed a commit as a proposal to this branch that addresses all of my previous review comments. Could you take a quick peek at the changes in https://github.com/Netflix/metaflow/pull/2534/commits/3044b19e934ebcb1812315c69c59021d23d065f0
to see if there are any issues with them regarding your use case / environment?

If everything is in order, then this should be good to go

nlwstein:
> @nlwstein I have pushed a commit as a proposal to this branch that addresses all of my previous review comments. Could you take a quick peek at the changes in [3044b19](https://github.com/Netflix/metaflow/commit/3044b19e934ebcb1812315c69c59021d23d065f0) to see if there are any issues with them regarding your use case / environment?
> 
> If everything is in order, then this should be good to go

@saikonen Thank you so much! I am going to test your commit here and ensure it works for us, and will report back today.

nlwstein:
@saikonen This does in fact work in our environment - and solves the original case where it was breaking before (supplying enough parameters to a particular flow)! Thanks again for the feedback and the improvements. 

Maybe a bit of a micro-optimization, but could we use $METAFLOW_DATASTORE_SYSROOT_S3 for the bucket path, and $METAFLOW_FLOW_NAME for the flow name in the command? For reference, our bucket name length + `metaflow` directory is 62 characters. I think these changes would make the command length deterministic again. I did this with the original version, partially because I initially "solved" this issue by shortening the orchestrator class name, which in turn reduced the weight of the command. 

Definitely not required on my end - and there may be some parsing shenanigans / changes that your team may not want to entertain to get this - but figured I'd mention it.

saikonen (dismissed):

metaflow/plugins/aws/step_functions/step_functions.py
```diff
@@ -893,6 +901,42 @@ def _batch(self, node):
...
+                    % (bucket, s3_object),
+                )
+                download_cmd = (
+                    f"{self.environment._get_install_dependencies_cmd('s3')} && "  # required for boto3 due to the original dependencies cmd getting packaged, and not being downloaded in time.
+                    f"{download_script} && "
+                    f"chmod +x /tmp/step_command.sh && "
+                    f"bash /tmp/step_command.sh"
+                )
+                new_cmd = shlex.split('bash -c "%s"' % download_cmd)
+                batch_job.payload["containerOverrides"]["command"] = new_cmd
```
savingoyal: wouldn't this be a leaky abstraction?

metaflow/plugins/aws/step_functions/step_functions_cli.py
```diff
@@ -131,6 +132,13 @@ def step_functions(obj, name=None):
     help="Use AWS Step Functions Distributed Map instead of Inline Map for "
     "defining foreach tasks in Amazon State Language.",
 )
+@click.option(
+    "--compress-state-machine/--no-compress-state-machine",
+    is_flag=True,
+    default=SFN_COMPRESS_STATE_MACHINE,
+    help="Upload large commands to S3 instead of embedding them in job payload "
```
savingoyal: help string can be - `Compress AWS Step Functions state machine to fit within the 8K limit.`

metaflow/plugins/aws/step_functions/step_functions_deployer.py
```diff
@@ -76,6 +76,10 @@ def create(
         use_distributed_map : bool, optional, default False
             Use AWS Step Functions Distributed Map instead of Inline Map for defining foreach
             tasks in Amazon State Language.
+        compress_state_machine : bool, optional, default False
+            Upload large commands to S3 instead of embedding them in job payload
```
savingoyal: same comment as https://github.com/Netflix/metaflow/pull/2534/files#r2303925282

metaflow/plugins/aws/step_functions/step_functions.py
nlwstein: @saikonen Added this commit - I am not entirely sure what you mean by leaky abstraction. Do you have a scenario / example that you are concerned about? 

For context, I tested his commit and it worked in our production environment, including with the complex flow that originally inspired the PR. The only issue I saw w/ this approach as opposed to my original proposal was that it doesn't generate as small of a command, and isn't of a deterministic length.

nlwstein: I also have a weak opinion (that I really may not have enough context to even have!) that it would be reasonable to expect anyone running on StepFunctions to have the AWS CLI, which would make the extra logic here not as necessary @saikonen. Or at least, if you're going to use this feature, I don't think it's an unreasonable thing to ask for.

saikonen: leaky abstraction should be fixed now, as the cmd compressing bits that depended on `Batch()` are now contained in the class.

PR #2549, opened 2025-08-14
recursive conditionals with resume

mt-ob:
```py

class LoopingFlow(FlowSpec):
    @step
    def start(self):
        print("Flow starting. Initializing loop counter.")
        # This variable will track our position in the loop.
        self.count = 0
        # We will loop 5 times.
        self.max_iterations = 5
        self.next(self.loop_step)

    @step
    def loop_step(self):
        print(f"Executing loop iteration number: {self.count}")
        self.count += 1
        self.loop_status = 'continue' if self.count < self.max_iterations else 'exit'
        print(f"Loop status is '{self.loop_status}'. Deciding next step...")
        self.next(
            {
                'continue': self.loop_step,
                'exit': self.exit_loop
            },
            condition='loop_status'
        )

    @step
    def exit_loop(self):
        print(f"Loop finished after {self.count} iterations.")
        self.next(self.end)

    @step
    def end(self):
        print("Flow has finished successfully.")

if __name__ == '__main__':
    LoopingFlow()
```

### DIFF
--- a/metaflow/runtime.py
+++ b/metaflow/runtime.py
@@ -468,6 +468,7 @@ class NativeRuntime(object):
                     pathspec_index,
                     cloned_task_pathspec_index,
                     finished_tuple,
+                    iteration_tuple,
                     ubf_context=ubf_context,
                     generate_task_obj=generate_task_obj and (not is_ubf_mapper_task),
                     verbose=verbose,
@@ -478,6 +479,7 @@ class NativeRuntime(object):
                     pathspec_index,
                     cloned_task_pathspec_index,
                     finished_tuple,
+                    iteration_tuple,
                     is_ubf_mapper_task,
                     ubf_context,
                 ) in inputs
--- a/metaflow/runtime.py
+++ b/metaflow/runtime.py
@@ -698,7 +698,7 @@ class NativeRuntime(object):
 
     # Store the parameters needed for task creation, so that pushing on items
     # onto the run_queue is an inexpensive operation.
-    def _queue_push(self, step, task_kwargs, index=None):
+    def _queue_push(self, step, task_kwargs, index=None, is_self_loop_switch=False):
         # In the case of cloning, we set all the cloned tasks as the
         # finished tasks when pushing tasks using _queue_tasks. This means that we
         # could potentially try to push the same task multiple times (for example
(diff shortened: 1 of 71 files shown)


mt-ob:
here's a summary along with some test cases:
- A cycle is only allowed when a conditional step creates a direct self-reference back to itself (A -> A)
    - The step must use the self.next(..., condition=...) syntax.
    - Any other type of loop (e.g., A -> B -> A or a non-conditional step creating a cycle) is invalid.
- Valid cases:
    - Self-Referential Loops: A single conditional step manages its own state and calls itself to continue iterating.
    - Loop Inside a Branch/Foreach: A branch or foreach can be the outer structure, with valid self-referential loops running inside the parallel paths.
- Invalid cases:
    - Multi-Step Cycles: Any transition from a step to a different ancestor step is forbidden.
    - Branch/Foreach Inside a Loop: A join step cannot transition back to the start of the loop.
    - Nested Loops at the Graph Level: The end of an inner loop cannot transition back to the start of an outer loop.

metaflow/runtime.py
```diff
@@ -505,6 +512,73 @@ def execute(self):
...
+                            is_recursive_step = (
+                                node.type == "split-switch"
+                                and step_name in node.out_funcs
+                            )
+
+                            if is_recursive_step and tasks:
+                                # This is a recursive step. We must find the single task that represents
+                                # the last successfully completed iteration *for each branch*.
+                                def get_loop_index(t):
+                                    # Helper function to extract the iteration index from the task's
```
romain-intel: nit: it's just the length no? They are all sequential.

metaflow/runtime.py
```diff
@@ -698,15 +772,15 @@ def _translate_index(self, task, next_step, type, split_index=None):
 
     # Store the parameters needed for task creation, so that pushing on items
     # onto the run_queue is an inexpensive operation.
-    def _queue_push(self, step, task_kwargs, index=None):
+    def _queue_push(self, step, task_kwargs, index=None, is_self_loop_switch=False):
```
romain-intel: nit: I would stick with one nomenclature: "loop" or "recursive". Personally prefer "recursive" but ok as long as consistent.

romain-intel (commented):
just posting partial comments. Still looking at the fake foreach frame.

PR #2550, opened 2025-08-15
feature: conditionals for Argo Workflows

saikonen:
Adds Conditional support to Argo Workflows

TODO:

- [ ] nesting a conditional inside a foreach is not yet working

### DIFF
--- a/metaflow/graph.py
+++ b/metaflow/graph.py
@@ -81,9 +81,15 @@ class DAGNode(object):
         self.split_parents = []
         self.split_branches = []
         self.matching_join = None
+        # Conditional info, also populated in _traverse_graph
         self.is_conditional = False  # will this node always be executed, or is it in a conditional branch?
-        self.conditional_branch = []
-        self.conditional_join = None  # Node where conditional branches end, and further nodes always execute.
+        self.is_conditional_join = (
+            False  # Does this node 'join' some set of conditional branches?
+        )
+        self.conditional_branch = (
+            []
+        )  # All the steps leading to this node that depends on a condition, starting from the split-switch
+        self.conditional_end_node = None  # Node where conditional branches end, and further nodes always execute.
         # these attributes are populated by _postprocess
         self.is_inside_foreach = False
 
@@ -358,6 +364,7 @@ class FlowGraph(object):
                 )
 
                 if is_conditional_join:
+                    node.is_conditional_join = True
                     conditional_root_nodes = root_nodes
 
                 # we are in a conditional branch if we have conditional root nodes left open, and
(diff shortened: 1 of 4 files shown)


metaflow/graph.py
```diff
@@ -296,8 +307,55 @@ def _postprocess(self):
             if [f for f in foreaches if self.nodes[f].matching_join != node.name]:
                 node.is_inside_foreach = True
 
+            # Fill in conditionals related info.
```
savingoyal: do we need this in graph.py? we can determine this in argo-workflows.py as well as we traverse the graph?

saikonen: this can be moved to the argo graph traversal as well if necessary, but I felt that other executors might benefit from it being in the graph.py as their node generation is probably similar enough.

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -1282,6 +1363,12 @@ def _visit(
                                         "{{tasks.%s.outputs.parameters.split-cardinality}}"
                                         % self._sanitize(node.name)
                                     ),
+                                    # Only pick the output step from the first iteration of the foreach task, as it should be identical for all.
```
saikonen: this assertion is untrue. the conditional step inside a foreach can vary based on each executing task.

as such, the current deterministic task-ids that we have for foreach tasks will not work when a conditional branch is the step being joined, e.g. the last step in a foreach

metaflow/graph.py
savingoyal: that's true, but to get this PR merged, it would be simpler to have no changes outside the plugins directory

metaflow/plugins/argo/argo_workflows.py
```diff
@@ -1568,10 +1633,25 @@ def _container_templates(self):
...
+                    % input_paths
+                )
             elif (
                 node.type == "join"
                 and self.graph[node.split_parents[-1]].type == "foreach"
             ):
+                # foreach-joins straight out of conditional branches are not yet supported
+                if node.is_conditional_join:
+                    raise ArgoWorkflowsException(
+                        "Foreach steps with a conditional step as the last one are not yet supported with Argo Workflows."
```
savingoyal: maybe - `Conditionals steps that transition directly into a join step are not currently supported. As a workaround, you can add a normal step after the conditional step that transitions to a join step.`

metaflow/plugins/argo/argo_workflows.py
saikonen: this type of DAG structure has been blocked for the feature until we have a proper fix for input-paths. #2551 is an attempt at this

PR #2574, opened 2025-08-26
Fixing task ID replacement for MNP jobs on AWS Batch

vymao:
If we aren't using the Metaflow metadata service provider, Metaflow defaults to generating task IDs locally. But these task IDs are just simple integers based on how many tasks/steps there are and are sequentially incremented based on `new_task_id` in `metaflow/plugins/metadata_providers/local.py`. This presents a problem when we're doing AWS Batch MNP, since currently we try and mass replace based on the task ID in the secondary command. If this is a simple integer, this will replace many erroneous places. 

For example, if the task ID is "3", there could be many instances of "3" in the secondary command that then have many replacements with "-node-$AWS_BATCH_JOB_NODE_INDEX" when really we just want to replace the actual task ID. 

Here, I've identified two places - the input task ID via `--task-id` and the task ID in `MF_PATHSPEC`, that should be the only two places in the command that have the actual task ID in them that need replacing. It is better to have more specific regexes this way.

Furthermore, if there is no metadata provider, I've added a new check for control MNP jobs to finish by checking the S3 datastore instead.

### DIFF
--- a/metaflow/plugins/aws/batch/batch_client.py
+++ b/metaflow/plugins/aws/batch/batch_client.py
@@ -432,14 +432,6 @@ class BatchJob(object):
 
         self.num_parallel = num_parallel or 0
         if self.num_parallel >= 1:
-            # Set the ulimit of number of open files to 65536. This is because we cannot set it easily once worker processes start on Batch.
-            # job_definition["containerProperties"]["linuxParameters"]["ulimits"] = [
-            #     {
-            #         "name": "nofile",
-            #         "softLimit": 65536,
-            #         "hardLimit": 65536,
-            #     }
-            # ]
             job_definition["type"] = "multinode"
             job_definition["nodeProperties"] = {
                 "numNodes": self.num_parallel,


metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -104,11 +105,17 @@ def execute(self):
             )
             secondary_commands = self.payload["containerOverrides"]["command"][-1]
             # other tasks do not have control- prefix, and have the split id appended to the task -id
-            secondary_commands = secondary_commands.replace(
-                self._task_id,
-                self._task_id.replace("control-", "")
-                + "-node-$AWS_BATCH_JOB_NODE_INDEX",
-            )
+            # Fix: Only replace task ID in specific arguments, not in environment variables
```
savingoyal: thanks for the fix! @saikonen any suggestions on approaches that might allow us to not lean on string substitution (quite finicky)!

saikonen: Haven't touched this implementation in a bit so maybe there is a reason that the task-id patterns are kept in `execute()` but at first glance this seems (in the original implementation) like quite a late phase to be doing anything with the command.

Could a better place be when the cmd is not yet joined into a string, when we can operate on options separately? f.ex. `batch_cli.py#251` and `step_functions.py#925` ?

vymao: I'm not 100% sure of the flow but I think that is possible if the flag `--task-id` is the only thing that needs to be changed. Looking at the command, I'm not sure if there is something else that requires us to modify the task ID (ex. `MF_PATHSPEC`). If `task-id` is the only thing, then I can definitely make that change.

vymao: I've updated the code. I think it is difficult to fully rely on `batch_cli.py#251` because we don't have access to `$AWS_BATCH_JOB_NODE_INDEX` as this would only be available at runtime for a given worker MNP node, I believe. But I've added more placeholders in `batch_cli.py#251` that should make the regex more reliable. I'm less familiar with the `step_functions` file as I'm not using that right now. 

Is it possible to get approval on this soon?

vymao: @savingoyal @saikonen

vymao: Just bumping this!

saikonen: Sorry its taken so long to get back to this. I'll be taking a look & testing this by Thursday at the latest.

vymao: Thank you!

vymao: Just wanted to check if there was anything I could help with?

saikonen: the direction looks good as is so no further changes should be required. My final testing is still pending due to some unrelated infra issues at my end that I need to look into. Aiming to get this PR wrapped up by this week.

saikonen: were you able to have success with this PR on your end? I was finally able to test things and it seems that it is not working as expected. The parallel jobs do launch, but the main control node never finishes, instead timing out.

I would assume that the issue lies in https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/aws/batch/batch_decorator.py#L412
where we poll the status of mapper tasks via the Metaflow client. This relies on task metadata, either through a metadata service, or in the absence of one, the data stored on local disk.

multi-node parallel jobs on batch are single tenant, so they do not share disk between each other. This would mean that a mapper task recording its finished_at locally (due to --metadata local) will not be accessible by the polling control task.

vymao: Apologies @saikonen, I had pre-maturely terminated the jobs assuming it was ok. I've now run this end-to-end and confirmed that this works on my end! If we don't have a metadata provider, I've resorted to just using the S3 datastore to check MNP job statuses, which has worked smoothly for me.

vymao: Also @savingoyal

vymao: Bumping this!

vymao: Hi, just wanted to check in on this?

vymao: Bump!

saikonen: Sorry again for the long wait. I'm finally able to put some hours into this and will check out the changes today.

metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -408,6 +432,14 @@ def _register_job_definition(
...
         self.num_parallel = num_parallel or 0
         if self.num_parallel >= 1:
+            # Set the ulimit of number of open files to 65536. This is because we cannot set it easily once worker processes start on Batch.
+            # job_definition["containerProperties"]["linuxParameters"]["ulimits"] = [
+            #     {
+            #         "name": "nofile",
+            #         "softLimit": 65536,
+            #         "hardLimit": 65536,
+            #     }
+            # ]
```
saikonen: can this be cleaned up?

metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -19,7 +20,34 @@ class BatchClient(object):
...
+                k: os.environ.pop(k) for k in _aws_env_keys if k in os.environ
+            }
+            try:
+                self._client = get_aws_client("batch")
+            finally:
+                # Restore prior env for the rest of the process
+                for k, v in _saved_env.items():
+                    os.environ[k] = v
+        else:
+            self._client = get_aws_client("batch")
```
saikonen, line 50: is this change relevant to the batch parallel issue, or something different? the PR seems to work fine without this part as well

metaflow/plugins/aws/batch/batch_decorator.py
```diff
@@ -421,6 +421,89 @@ def _wait_for_mapper_tasks(self, flow, step_name):
...
+            completed = 0
+            for ps in pathspecs:
+                try:
+                    parts = ps.split("/")
+                    if len(parts) == 3:
+                        run_id, step, task_id = parts
+                    else:
+                        # Fallback in case of unexpected format
+                        run_id, step, task_id = self.run_id, step_name, parts[-1]
+                    tds = TaskDataStore(
```
saikonen, line 468: `task_pre_step` actually binds a task datastore to `self`. Would that work as well?

metaflow/plugins/aws/batch/batch_decorator.py
```diff
@@ -421,6 +421,89 @@ def _wait_for_mapper_tasks(self, flow, step_name):
...
+            )
+        if md is None and "AWS_BATCH_JOB_ID" in os.environ:
+            return self._wait_for_mapper_tasks_batch_api(
+                flow, step_name, last_completion_timeout
+            )
+        return self._wait_for_mapper_tasks_metadata(
+            flow, step_name, last_completion_timeout
+        )
+
+    def _wait_for_mapper_tasks_batch_api(
```
saikonen: nit: `_wait_for_mapper_tasks_datastore` as the other one is suffixed `_metadata`

metaflow/plugins/aws/batch/batch_decorator.py
```diff
@@ -421,6 +421,89 @@ def _wait_for_mapper_tasks(self, flow, step_name):
...
+                        allow_not_done=True,
+                    )
+                    if tds.has_metadata(TaskDataStore.METADATA_DONE_SUFFIX):
+                        completed += 1
+                except Exception as e:
+                    if os.environ.get("METAFLOW_DEBUG_BATCH_POLL") in (
+                        "1",
+                        "true",
+                        "True",
+                    ):
```
saikonen: is this necessary to have as a debug flag? were there a lot of datastore errors encountered to necessitate this? I'd opt to either printing out the error always, or never depending on how relevant it is for the user.

saikonen: alternatively we can revisit this separately and consider adding a proper debug flag for `batch`

saikonen (commented):
functionally this PR is now working as expected. Had some suggestions for cleanup

metaflow/plugins/aws/batch/batch_client.py
vymao: Indeed it works, this was to cover the instances where particular AWS keys have already been set in the environment, which messed up getting the AWS client. This is relevant for the batch process given that we're using the batch client now.

metaflow/plugins/aws/batch/batch_decorator.py
vymao: I don't think so - the TaskDataStore instances are constructed with a fixed run_id/step_name/task_id so it won't be able to query the mapper tasks

metaflow/plugins/aws/batch/batch_decorator.py
vymao: Done

metaflow/plugins/aws/batch/batch_decorator.py
vymao: Nope! I changed it to a warning instead.

metaflow/plugins/aws/batch/batch_client.py
vymao: Yep! Removed.

metaflow/plugins/aws/batch/batch_decorator.py
```diff
@@ -421,6 +421,86 @@ def _wait_for_mapper_tasks(self, flow, step_name):
...
+                        run_id,
+                        step,
+                        task_id,
+                        mode="r",
+                        allow_not_done=True,
+                    )
+                    if tds.has_metadata(TaskDataStore.METADATA_DONE_SUFFIX):
+                        completed += 1
+                except Exception as e:
+                    self.logger.warning("Datastore wait: error checking %s: %s", ps, e)
```
saikonen, line 479: `self.logger` doesn't actually have any methods, it is just `click.secho` being passed in. This also adds unnecessary (duplicate) timestamps to the log lines so sticking to `print` for now is fine.

saikonen: also note all other instances of `self.logger`

metaflow/plugins/aws/batch/batch_client.py
saikonen: what confuses me with these is that the same environment variables should then also interfere with the task running inside a batch process from getting a working S3 client as well, which would break datastore access.

do you have an example AMI which exhibits this issue, or an example on how to reproduce the issue? I'm not running into any issues with my test setup even without these additions.

saikonen: If the environment modification part is not critical to this fix working for your use case, could you introduce that part as a separate PR?

PR #2581, opened 2025-08-28
fix: Card dag breaking up in loops

rsanteri:
DAG in cards was breaking up when loops were used due to recursive steps. This PR fixes that and renders connection back to recursive step instead

image


rsanteri:
CodeQL triggered some warning. Removing minify from card bundle for a moment to get better info about the errors
UPDATE: warning seem to come up from libraries like svelte and svelte-vega. I wonder why these have not popped earlier. Trying to update
UPDATE2: These errors seem to be from ealier PRs. Investigating further

PR #2600, opened 2025-09-12
[include file] yaml file support.

valayDave:
- introduce abstraction for yaml file support.

Example Flow: 

```python
#!/usr/bin/env python3
"""
Simple test script to demonstrate IncludeFile functionality with YAML files.
"""

from metaflow import FlowSpec, step, yaml_parser, IncludeFile

class TestYAMLFlow(FlowSpec):
    """
    A flow to test the IncludeYAMLFile parameter.
    """
    
    config = IncludeFile(
        'config',
        default='config.yaml',
        help='Configuration file in YAML format',
        parser=yaml_parser
    )

    @step
    def start(self):
        """
        Start step that prints the parsed YAML configuration.
        """
        print("Parsed YAML configuration:")
        print(f"Type: {type(self.config)}")
        print(f"Content: {self.config}")
        
        if isinstance(self.config, dict):
            print("Configuration keys:")
            for key, value in self.config.items():
                print(f"  {key}: {value}")
        
        self.next(self.end)

    @step
    def end(self):
        """
        End step.
        """
        print("Flow completed successfully!")

if __name__ == '__main__':
    TestYAMLFlow()


```

### DIFF
--- a/metaflow/includefile.py
+++ b/metaflow/includefile.py
@@ -285,7 +285,7 @@ class IncludeFile(Parameter):
             self._includefile_overrides["is_text"] = is_text
         if encoding is not None:
             self._includefile_overrides["encoding"] = encoding
-        self.parser = parser
+        self._parser = parser
         # NOTA: Right now, there is an issue where these can't be overridden by config
         # in all circumstances. Ignoring for now.
         super(IncludeFile, self).__init__(
@@ -349,14 +349,14 @@ class IncludeFile(Parameter):
         # Get the raw content from the file
         content = v.decode(self.name, var_type="Parameter")
         # If a parser is specified, use it to parse the content
-        if self.parser is not None:
+        if self._parser is not None:
             try:
-                return ConfigInput._call_parser(self.parser, content)
+                return ConfigInput._call_parser(self._parser, content)
             except Exception as e:
                 raise MetaflowException(
                     "Failed to parse content in parameter '%s' using parser: %s"
                     % (self.name, str(e))
-                )
+                ) from e
 
         return content
 
--- a/metaflow/plugins/parsers.py
+++ b/metaflow/plugins/parsers.py
@@ -1,7 +1,7 @@
 from metaflow._vendor import yaml
 
 
-def yaml_parser(content: str):
+def yaml_parser(content: str) -> dict:
     """
     Parse YAML content to a dictionary.
 


romain-intel:
Why not use Config for this?

valayDave:
Configs stay static once deployed. Parameters can change per instantiation of deployment.

romain-intel:
hum, that's fair. I would rather we use a similar approach to Config though to stay a bit consistent. Instead of introducing another type, could you add something like a `parser` option?

valayDave:
Sure! Will fix.

valayDave:
Ready for review !

metaflow/includefile.py
```diff
@@ -336,7 +345,20 @@ def init(self, ignore_errors=False):
...
-        return v.decode(self.name, var_type="Parameter")
+
+        # Get the raw content from the file
+        content = v.decode(self.name, var_type="Parameter")
+        # If a parser is specified, use it to parse the content
+        if self.parser is not None:
+            try:
+                return ConfigInput._call_parser(self.parser, content)
+            except Exception as e:
+                raise MetaflowException(
```
romain-intel, line 356: nit: raise ... from e

metaflow/plugins/parsers.py
```diff
@@ -0,0 +1,16 @@
+from metaflow._vendor import yaml
+
+
+def yaml_parser(content: str):
```
romain-intel: nit: return type in definition.

metaflow/includefile.py
```diff
@@ -270,13 +277,15 @@ def __init__(
...
         help: Optional[str] = None,
+        parser: Optional[Union[str, Callable[[str], Any]]] = None,
         **kwargs: Dict[str, str]
     ):
         self._includefile_overrides = {}
         if is_text is not None:
             self._includefile_overrides["is_text"] = is_text
         if encoding is not None:
             self._includefile_overrides["encoding"] = encoding
+        self.parser = parser
```
romain-intel: nit `self._parser` to be consistent with other private variables here.

romain-intel (commented):
Some minor comments but otherwise lgtm. Thanks for refactoring inside IncludeFile directly.

metaflow/includefile.py
valayDave: Done !

PR #2608, opened 2025-09-16
fix pip tags to use resolved python from environment

mt-ob:


metaflow/plugins/pypi/parsers.py
```diff
@@ -1,5 +1,32 @@
...
+
+    Returns
+    -------
+    str
+        A concrete version like "3.8.*" that conda can resolve.
+    """
+    if not version_spec or not version_spec.strip():
+        return None
+
+    candidates = ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
```
savingoyal: can you add more elements to the candidates set so that we don't have to worry about keeping track of versions for a while (including some 4.* versions)

mt-ob: no longer needed

mt-ob:
this allows stuff like `@pypi(python=">=3.8,<3.11")` and `@pypi(python=">=3.9")` and `@pypi(python="~=3.8.0")`, etc.

PR #2609, opened 2025-09-17
feat: add GPU support to metaflow-dev minikube setup

### LINKED ISSUE #2606: `metaflow-dev`: consider adding `--gpus all` to `minikube start`
## What

Add `--gpus all` as an option to the `$(MINIKUBE) start` call [under the `setup-minikube` Makefile target](https://github.com/Netflix/metaflow/blame/master/devtools/Makefile#L139) used by `metaflow-dev`.

## Notes

This is required to run jobs with e.g. `@resources(gpu=1)` on a local dev environment.

Support may need to be detected before trying to use the flag and/or should be optional somehow (environment variable, perhaps?).

cnaples79:
## Summary
- Add optional GPU support for minikube in metaflow-dev with intelligent auto-detection
- Provide manual control via `MINIKUBE_ENABLE_GPU` environment variable
- Enable GPU workloads like `@resources(gpu=1)` in local development environments

## Changes Made
- **Auto-detection logic**: Detects NVIDIA (`nvidia-smi`) and AMD (`rocm-smi`) GPUs automatically
- **Environment variable control**: `MINIKUBE_ENABLE_GPU=auto|true|false` (default: auto)
- **User feedback**: Informative messages about GPU detection status during startup
- **Help documentation**: Updated help text with environment variable usage
- **Conditional flag addition**: Adds `--gpus all` to `minikube start` only when appropriate

## Modes of Operation
1. **`auto` (default)**: Automatically detects GPU availability and enables if found
2. **`true`**: Force enables GPU support regardless of detection
3. **`false`**: Explicitly disables GPU support

## Test Plan
- ✅ Verified Makefile syntax with `make help`
- ✅ Tested dry-run with `make -n setup-minikube` (shows no GPU detected message)
- ✅ Tested forced enable with `MINIKUBE_ENABLE_GPU=true` (correctly adds `--gpus all` flag)
- ✅ Confirmed help text displays new environment variable documentation

## Example Usage
```bash
# Auto-detect GPU (default behavior)
make setup-minikube

# Force enable GPU support
MINIKUBE_ENABLE_GPU=true make setup-minikube

# Explicitly disable GPU support  
MINIKUBE_ENABLE_GPU=false make setup-minikube
```

Fixes #2606

### DIFF
--- a/devtools/Makefile
+++ b/devtools/Makefile
@@ -10,6 +10,9 @@ help:
 	@echo "  down          - Stop and clean up the environment"
 	@echo "  all-up        - Start the development environment with all services"
 	@echo "  help          - Show this help message"
+	@echo ""
+	@echo "Environment variables:"
+	@echo "  MINIKUBE_ENABLE_GPU=auto|true|false  - Control GPU support (default: auto)"
 
 HELM_VERSION := v3.14.0
 MINIKUBE_VERSION := v1.32.0
@@ -31,6 +34,7 @@ MAKE_CMD := $(MAKE) -f "$(MKFILE_PATH)"
 MINIKUBE_CPUS ?= 4
 MINIKUBE_MEMORY ?= 6144
 MINIKUBE_DISK_SIZE ?= 20g
+MINIKUBE_ENABLE_GPU ?= auto
 
 ifeq ($(shell uname), Darwin)
 	minikube_os = darwin
@@ -46,6 +50,24 @@ ifeq ($(shell uname -m), x86_64)
 else
 	arch = arm64
 	tilt_arch = arm64
+endif
+
+# GPU detection for minikube
+ifeq ($(MINIKUBE_ENABLE_GPU), auto)
+	# Auto-detect GPU availability
+	ifeq ($(shell command -v nvidia-smi >/dev/null 2>&1 && echo "nvidia"), nvidia)
+		gpu_flags = --gpus all
+	else ifeq ($(shell command -v rocm-smi >/dev/null 2>&1 && echo "amd"), amd)
+		gpu_flags = --gpus all
+	else
+		gpu_flags =
+	endif
+else ifeq ($(MINIKUBE_ENABLE_GPU), true)
+	# Force enable GPU support
+	gpu_flags = --gpus all
+else
+	# Explicitly disabled
+	gpu_flags =
 endif
 
 # TODO: Move scripts to a folder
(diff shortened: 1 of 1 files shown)


devtools/Makefile
```diff
@@ -48,6 +52,24 @@ else
 	tilt_arch = arm64
 endif
 
+# GPU detection for minikube
+ifeq ($(MINIKUBE_ENABLE_GPU), auto)
+	# Auto-detect GPU availability
+	ifeq ($(shell command -v nvidia-smi >/dev/null 2>&1 && echo "nvidia"), nvidia)
+		gpu_flags = --gpus all
```
feltech: Depending on how Docker is configured, `--gpus all` might not work, and instead you need to use `--devices nvidia.com/gpu=all` or similar.

(I had this problem with [Docker on NixOS](https://discourse.nixos.org/t/nvidia-docker-container-runtime-doesnt-detect-my-gpu/51336/2))

cnaples79: Thanks for the feedback, I'll update.

cnaples79:
Thanks for the feedback @feltech! I've updated the implementation to address the Docker compatibility concerns:

## Changes Made

**🔧 Improved Docker Compatibility:**
- Default to `--devices nvidia.com/gpu=all` for NVIDIA GPUs (more compatible with different Docker configurations)
- Keep `--gpus all` for AMD/other GPUs
- This addresses the NixOS Docker issue you mentioned

**⚙️ Enhanced Control Options:**
Added `MINIKUBE_GPU_FLAG` environment variable for explicit control:
- `auto` (default): Smart selection based on GPU type
- `gpus`: Force `--gpus all` format  
- `devices`: Force `--devices nvidia.com/gpu=all` format
- Custom value: User-provided flag (e.g., `--devices nvidia.com/gpu=2`)

## Example Usage

```bash
# Auto-detect best GPU flag (default)
make setup-minikube

# Force devices format (good for Docker compatibility issues)
MINIKUBE_GPU_FLAG=devices make setup-minikube

# Force legacy gpus format
MINIKUBE_GPU_FLAG=gpus make setup-minikube

# Custom GPU specification
MINIKUBE_GPU_FLAG="--devices nvidia.com/gpu=2" make setup-minikube
```

This should resolve the Docker configuration compatibility issues while maintaining flexibility for different setups. Let me know if this addresses your concerns!

devtools/Makefile
feltech: Apologies, I may have misled you here, the problem I referred to was using the `docker` command-line, rather than the `minikube` command-line. They both accept a `--gpus` argument, but I think [minikube is a bit more clever](https://minikube.sigs.k8s.io/docs/tutorials/nvidia/) about it. Indeed, `--devices` doesn't seem to be valid for minikube:
```
$ minikube start --devices nvidia.com/gpu=all
Error: unknown flag: --devices
```

My bad, got my wires totally crossed.

cnaples79: Ahhh I see. No worries, I'll update the PR with the correct fix.

cnaples79:
Thanks for the clarification, and you're absolutely right — `minikube` doesn't support `--devices`. I've updated the PR to remove the `--devices` path and always pass `--gpus all` to `minikube start` when GPU is detected or forced via `MINIKUBE_ENABLE_GPU=true`.

Summary of changes:
- Remove `MINIKUBE_GPU_FLAG` and the `--devices nvidia.com/gpu=all` path
- Keep simple/valid `--gpus all` for `minikube`
- Preserve auto‑detection and `MINIKUBE_ENABLE_GPU` env var controls

If you want me to also document the separate Docker CLI considerations (for folks not using `minikube`), I can add a short note in the devtools help.

PR #2612, opened 2025-09-19
Fix extensions and packaging for some types of editable package installs

romain-intel:
In some cases, a distribution install of an editable package includes just a .pth file and not the full file list included in the extension. In those cases, the extension mechanism failed to fall back to traversing the directory indicated in the .pth file.

This PR fixes this issue while trying to maintain the "usual" case of non editable packages as quick as possible (walking the FS is not ideal in all cases).

Some additional clean-up particularly around handling the case of "getting file from the distribution files list" and "walking a tree". We already used to walk the FS for non-distribution packages but now we do it in both cases (potentially) so the code was refactored to make it clearer.

A similar issue could also happen when packaging METAFLOW_PACKAGE_POLICY packages that were installed in an editable mode. The same type of fixes are applied there as well.


metaflow/extension_support/__init__.py
```diff
@@ -468,144 +678,88 @@ def _get_extension_packages(ignore_info_file=False, restrict_to_directories=None
...
+                "files": [],
+                "full_path_files": [],
+                "meta_module": None,  # Meta information about the package (if applicable)
+            }
+            addl_dirs = []
             # At this point, we check to see what extension points this package
             # contributes to. This is to enable multiple namespace packages to contribute
             # to the same extension point (for example, you may have multiple packages
             # that have plugins)
             for f in dist.files:
```
jklukas: For the case of a wheel that's been repacked by Bazel's rules_python, I believe it will still encounter an error on this line as `dist.files` will return `None`. Can we check for the `None` case and fall back to walking the file tree in that case?

metaflow/extension_support/__init__.py
```diff
@@ -695,132 +847,29 @@ def _get_extension_packages(ignore_info_file=False, restrict_to_directories=None
...
-                                ]
-                                skip_extension = set(dirs).issubset(set(ok_subdirs))
-
-                            if skip_extension:
-                                _ext_debug(
-                                    "    Skipping '%s' as no files/directory of interest"
-                                    % _extension_points[idx]
-                                )
-                                continue
+            for root, _, files in os.walk(package_path):
```
jklukas: Consider whether we can set `followlinks` here. Bazel preps sandboxed directories using a lot of symlinks, so might get tripped up without this.

```suggestion
            for root, _, files in os.walk(package_path, followlinks=True):
```

PR #2618, opened 2025-09-24
Fix/aws secrets manager through container

vymao:
Right now, the secrets decorator injects the secrets into an AWS Batch job after the container has been set up. However, there may be times when secrets are necessary at the container setup entrypoint. This PR allows a user to specify a name and secrets ARN so that these can be passed into the job configuration at construction time, so that the right fields can be set like in [the AWS documentation.](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/secrets-envvar-secrets-manager.html)

### DIFF
--- a/metaflow/plugins/aws/batch/batch.py
+++ b/metaflow/plugins/aws/batch/batch.py
@@ -206,6 +206,7 @@ class Batch(object):
         ephemeral_storage=None,
         log_driver=None,
         log_options=None,
+        container_secrets=None,
     ):
         job_name = self._job_name(
             attrs.get("metaflow.user"),
@@ -262,6 +263,7 @@ class Batch(object):
                 ephemeral_storage=ephemeral_storage,
                 log_driver=log_driver,
                 log_options=log_options,
+                container_secrets=container_secrets,
             )
             .task_id(attrs.get("metaflow.task_id"))
             .environment_variable("AWS_DEFAULT_REGION", self._client.region())
@@ -380,6 +382,7 @@ class Batch(object):
         ephemeral_storage=None,
         log_driver=None,
         log_options=None,
+        container_secrets=None,
     ):
         if queue is None:
             queue = next(self._client.active_job_queues(), None)
@@ -421,6 +424,7 @@ class Batch(object):
             ephemeral_storage=ephemeral_storage,
             log_driver=log_driver,
             log_options=log_options,
+            container_secrets=container_secrets,
         )
         self.num_parallel = num_parallel
         self.job = job.execute()
--- a/metaflow/plugins/aws/batch/batch_cli.py
+++ b/metaflow/plugins/aws/batch/batch_cli.py
@@ -281,6 +281,25 @@ def step(
     if env_deco:
         env.update(env_deco[0].attributes["vars"])
 
+    # Collect ECS-style container secrets from @secrets decorator, if any.
+    # These entries have shape: {"name": <ENV>, "value_from": <SecretsManager ARN>} (snake_case input),
+    # and will be injected at container startup by AWS Batch/ECS via job definition.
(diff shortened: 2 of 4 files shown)


metaflow/plugins/secrets/secrets_decorator.py
```diff
@@ -80,6 +80,12 @@ def task_pre_step(
                     SecretSpec.secret_spec_from_str(secret_spec_str_or_dict, role=role)
                 )
             elif isinstance(secret_spec_str_or_dict, dict):
+                # If the dict is an ECS-style container-start secret spec, skip runtime fetching.
```
savingoyal: this would break workloads running on EKS

vymao: Removed!

vymao: Hi @savingoyal, just checking if this works?

vymao: Bump!

vymao: Hi just wanted to bump this again!

savingoyal:
@saikonen reminder for review. @vymao can you rebase on top of head of master?

metaflow/plugins/aws/batch/batch_cli.py
```diff
@@ -298,6 +298,25 @@ def echo(msg, stream="stderr", batch_id=None, **kwargs):
     if env_deco:
         env.update(env_deco[0].attributes["vars"])
 
+    # Collect ECS-style container secrets from @secrets decorator, if any.
+    # These entries have shape: {"name": <ENV>, "value_from": <SecretsManager ARN>} (snake_case input),
+    # and will be injected at container startup by AWS Batch/ECS via job definition.
+    container_secrets = []
+    secrets_deco = [deco for deco in node.decorators if deco.name == "secrets"]
```
saikonen, line 305: This syntax is a major departure of what `@secrets` currently supports so to me it doesn't seem like a good fit from the user experience perspective
- existing `@secrets` syntax will not carry over to the new container level ones
- the format of `sources=` dictates what level the secrets get applied at in the runtime environment

for the following reasons, it would seem like such configuration is better suited in the `@batch` decorator, for example as a direct option
```
@batch(secrets={"name": "example_env_var, "value_from": "arn::aws::..."})
```

metaflow/plugins/aws/batch/batch.py
```diff
@@ -255,6 +255,7 @@ def create_job(
         ephemeral_storage=None,
         log_driver=None,
         log_options=None,
+        container_secrets=None,
```
saikonen, line 258: can you also make sure these are applied with the Step functions implementation.

metaflow/plugins/aws/batch/batch_cli.py
```diff
@@ -298,6 +298,25 @@ def echo(msg, stream="stderr", batch_id=None, **kwargs):
...
+    if secrets_deco:
+        try:
+            for s in secrets_deco[0].attributes.get("sources", []) or []:
+                if isinstance(s, dict):
+                    name = s.get("name")
+                    value_from = s.get("value_from")
```
saikonen, line 311: just some paper cuts with the current implementation: I tried reusing some existing `@secrets` decos from a flow I had, but was surprised secrets were not being picked up. This was simply due to having the `sources=` be a list of keys.

After changing to the required format, I ran into an error with my flow due to not usually having an executionRoleArn due to not requiring it:

```
botocore.errorfactory.ClientException: An error occurred (ClientException) when calling the RegisterJobDefinition operation: Error executing request, Exception : executionRoleArn cannot be empty when using secrets or secretOptions
```

PR #2635, opened 2025-10-10
fix: argo params issue with char encodings

saikonen:
introduce a script for parsing param values from a base64 encoding instead of passing raw values to the command line, in order to get around some encoding issues

other JSONType parameter fixes
- support both json-strings and proper objects as parameter default values
- support json-stringified values in argo event payloads
- unify param serialisation between cron and manual triggers

### DIFF
--- a/metaflow/plugins/argo/param_val.py
+++ b/metaflow/plugins/argo/param_val.py
@@ -3,17 +3,8 @@ import base64
 import json
 
 
-def parse_parameter_value(base64_value, param_type):
+def parse_parameter_value(base64_value):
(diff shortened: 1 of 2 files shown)


metaflow/plugins/argo/param_val.py
```diff
@@ -0,0 +1,14 @@
+import sys
+import base64
+
+
+def parse_parameter_value(base64_value):
+    decoded = base64.b64decode(base64_value).decode("utf-8")
```
iamsgarg-ob: super nit: return base64.b64decode.....

PR #2636, opened 2025-10-13
update module importer

malder08:
Update the ModuleImporter class to use `find_spec` instead of `find_module` and `create_module` and `exec_module` instead of `load_module`. In Python 3.12, the fallback that looked for a `find_module()` method if a meta_path entry didn't have a `find_spec()` method was removed, so the escape hatch didn't work correctly in this version when running with `--environment=conda.`

### DIFF
--- a/metaflow/plugins/env_escape/client_modules.py
+++ b/metaflow/plugins/env_escape/client_modules.py
@@ -142,6 +142,7 @@ class ModuleImporter(object):
         config_dir: Directory containing configuration for the environment escape
         module_prefixes: List of module name prefixes to handle
     """
+
     def __init__(
         self,
         python_executable,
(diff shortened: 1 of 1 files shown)


metaflow/plugins/env_escape/client_modules.py
```diff
@@ -135,84 +139,90 @@ def __init__(
...
-            raise ImportError
-        sys.modules[fullname] = module
-        return module
+        # Modules are created canonically but we need to handle any of the aliases.
+        wrapped_module = self._handled_modules.get(canonical_fullname)
+        if wrapped_module is None:
+            raise ImportError(f"No module named '{fullname}'")
+
+        # Replace the standard module with the wrapped module in sys.modules
+        sys.modules[fullname] = wrapped_module
```
romain-intel: Looking at the code of how this is used in the import system, I believe we shouldn't be adding stuff to `sys.modules`. Maybe we can try to do more in `create_module` and return `wrapped_module` there and then not do anything in `exec_module`. I think `exec_module` is also for reloading which we do not support much right now.

malder08: That makes sense. Originally I was trying not to change the existing logic too much, but I've updated it to have `create_module` handle everything and it seems much better. Thanks for the help!

PR #2639, opened 2025-10-16
Fix/multi inheritance

romain-intel:


metaflow/flowspec.py
valayDave: Simple inheritance based flows crashing in the _process_config_decorators. trivial Example to repro:
```python
# inheritted.py
from metaflow import (
    FlowSpec,
    Config,
    pyproject_toml_parser,
    FlowMutator,
    pypi_base,
)


class project_pypi(FlowMutator):
    def pre_mutate(self, mutable_flow):
        project_deps = dict(mutable_flow.configs).get("project_deps")
        if project_deps["packages"]:
            mutable_flow.add_decorator(
                pypi_base, deco_kwargs=project_deps, duplicates=mutable_flow.ERROR
            )



@project_pypi
class FBBase(FlowSpec):
    project_deps = Config(
        "project_deps", default="pyproject.toml", parser=pyproject_toml_parser
    )

```

```python
# flow.py 
from inheritted import FBBase
from metaflow import step

class MyFlow(FBBase):
    
    @step
    def start(self):
        print("Hello World!")
        self.next(self.end)

    @step
    def end(self):
        print("Flow completed successfully!")
    
if __name__ == "__main__":
    MyFlow()

```
Crashing with stack trace:
```
  File "/home/ubuntu/metaflow/metaflow/cli.py", line 723, in main
    start(auto_envvar_prefix="METAFLOW", obj=state)
  File "/home/ubuntu/metaflow/metaflow/_vendor/click/core.py", line 829, in __call__
    return self.main(args, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/metaflow/metaflow/_vendor/click/core.py", line 782, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/home/ubuntu/metaflow/metaflow/cli_components/utils.py", line 65, in invoke
    click.Command.invoke(self, ctx)
  File "/home/ubuntu/metaflow/metaflow/_vendor/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/metaflow/metaflow/_vendor/click/core.py", line 610, in invoke
    return callback(args, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/metaflow/metaflow/_vendor/click/decorators.py", line 21, in new_func
    return f(get_current_context(), args, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/metaflow/metaflow/cli.py", line 477, in start
    new_cls = ctx.obj.flow._process_config_decorators(config_options)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/metaflow/metaflow/flowspec.py", line 452, in _process_config_decorators
    if param.IS_CONFIG_PARAMETER:
       ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/metaflow/metaflow/user_configs/config_parameters.py", line 164, in __getattr__
    raise AttributeError(key)
AttributeError: IS_CONFIG_PARAMETER
```

metaflow/flowspec.py
```diff
@@ -284,12 +389,12 @@ def _process_config_decorators(cls, config_options, process_configs=True):
             debug.userconf_exec("Setting config %s to %s" % (var, str(val)))
             setattr(cls, var, val)
 
-        cls._flow_state[_FlowState.SET_CONFIG_PARAMETERS] = to_save_configs
+        cls._flow_state[FlowStateItems.SET_CONFIG_PARAMETERS] = to_save_configs
```
valayDave, line 402: Somehow `cls._flow_state[FlowStateItems.SET_CONFIG_PARAMETERS]` is returning `[]` even when `to_save_configs` has a list of tuples. Weird!

metaflow/flowspec.py
```diff
@@ -76,14 +77,84 @@ def __getitem__(self, item):
...
+        inherited_value = self._inherited.get(key)
+
+        if self_value is not None:
+            self._merged_data[key] = self._merge_value(inherited_value, self_value)
+            return self._merged_data[key]
+        elif key in self._self_data:
+            # Case of CACHED_PARAMETERS; a valid value is None. It is never inherited
+            self._merged_data[key] = None
+            return None
+        raise KeyError(key)
```
valayDave, line 124: The root cause of this bug is here . This bug is triggered because `self._merged_data` can be set when the user does something like in L545 (ie doing a getter like `cls._flow_state[_FlowState.SET_CONFIG_PARAMETERS]` ). 

This pattern is a little funky because based on how `__getitem__` is written, if a key is accessed once then it will be set in the `self._merge_data` and once it is set in the `self._merge_data` then even if we call `__setitem__`, the `__getitem__` will ONLY return the older value since it's in `self._merge_data`

metaflow/flowspec.py
valayDave, line 402:
```suggestion
        del cls._flow_state[FlowStateItems.SET_CONFIG_PARAMETERS]
        cls._flow_state[FlowStateItems.SET_CONFIG_PARAMETERS] = to_save_configs
```

This suggestion fixes the bug [I have mentioned here](https://github.com/Netflix/metaflow/pull/2639/files#r2509310770). The root cause of [the bug is explained here](https://github.com/Netflix/metaflow/pull/2639/files#r2509675838).

valayDave (changes requested):
Okay I went and I test drove this. There are some changes needed since this currently actively breaks current the inheritance. The bug was very subtle but the suggestion I have added in the PR explains the only line of code we need to add to fix it! 

I have currently tested the basic inheritance model. I will add one more review on thorougher testing with deployer / runner APIS alongside inheritance.

valayDave (dismissed):

PR #2643, opened 2025-10-17
[S3] Fix S3 list_recursive filtering with paths containing consecutive slashes

npow:
BEFORE commit 21f29e2c (no filtering logic):

1.   S3ROOT = "s3://bucket/path/" (with trailing slash)
2.   User does: S3(s3root=f"{S3ROOT}/{test_prefix}/log/")
3.   This creates: "s3://bucket/path//test_prefix/log/"
4.   S3.__init__ does rstrip("/") → "s3://bucket/path//test_prefix/log" (double slash still there!)
5.   When listing, S3 API is queried with prefix path//test_prefix/log
6.   S3 DOES return results because S3 ignores the double slash in the middle! S3 treats path//test_prefix same as path/test_prefix
7.   All keys returned by S3 are included without filtering
8.   Result: Everything worked fine!

AFTER commit 21f29e2c (with filtering logic):
  1. Same setup, S3 API returns results (S3 still treats // as /)
  2. BUT now the filtering logic kicks in:
    - prefix_url.path = "path//test_prefix/log"
    - normalized_prefix = "path//test_prefix/log/" (just adds trailing slash)
    - Checks if key_path (e.g., "path/test_prefix/log/test.txt") starts with "path//test_prefix/log/"
    - NO MATCH because path/ ≠ path//
  3. All results are filtered out!
  4. Result: Tests fail with empty results

### DIFF
--- a/metaflow/plugins/datatools/s3/s3.py
+++ b/metaflow/plugins/datatools/s3/s3.py
@@ -18,6 +18,7 @@ from metaflow.metaflow_config import (
     DATATOOLS_S3ROOT,
     S3_RETRY_COUNT,
     S3_TRANSIENT_RETRY_COUNT,
+    S3_LOG_TRANSIENT_RETRIES,
     S3_SERVER_SIDE_ENCRYPTION,
     S3_WORKER_COUNT,
     TEMPDIR,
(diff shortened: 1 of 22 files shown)


metaflow/plugins/datatools/s3/s3.py
```diff
@@ -557,7 +561,9 @@ def __init__(
                 raise MetaflowS3URLException(
                     "s3root needs to be an S3 URL prefixed with s3://."
                 )
-            self._s3root = s3root.rstrip("/")
+            # Normalize the path by collapsing consecutive slashes
```
romain-intel, line 564: To be on the safe side (and since some of the values used elswhere can also be input by the user), could we normalize in all cases?

npow:
deprecated in favor of https://github.com/Netflix/metaflow/pull/2673

PR #2644, opened 2025-10-17
Fix/dashes configs and add `plain` option for Configs

romain-intel:


romain-intel:
Has a purposeful bug to test something :)

metaflow/cli.py
```diff
@@ -444,12 +444,16 @@ def start(
         # We can now set the the CONFIGS value in the flow properly. This will overwrite
         # anything that may have been passed in by default and we will use exactly what
         # the original flow had. Note that these are accessed through the parameter name
+        # We need to save the "plane-ness" flag to carry it over
```
npow: plain?

romain-intel: The term felt a bit flat ... :). I'll fix :)

romain-intel:
All tests pass internally -- including those that are defined here. merging

PR #2653, opened 2025-10-22
Fix json decode error

npow:
## Problem

Intermittent `JSONDecodeError` when multiple environments are resolved concurrently during deployment:

```
json.decoder.JSONDecodeError: Expecting value: line 1 column 86673 (char 86672)
```

### Root Cause

Race condition in FIFO-based IPC between deployer subprocess and parent process:

1. **Writer side**: Subprocess writes JSON to FIFO, but Python's buffered I/O may not flush immediately
2. **Reader side**: Parent process reads from FIFO in non-blocking mode
3. **Race**: When subprocess exits quickly after `close()`, reader detects process exit and breaks on empty read
4. **Problem**: OS kernel may still have buffered data in pipe that hasn't been delivered yet
5. **Result**: Truncated JSON at arbitrary positions (~86KB in the error case)

## Solution

Changed `read_from_fifo_when_ready()` to use a hybrid approach:

1. **Start in non-blocking mode** (existing behavior)
  - Use `select.poll()` to wait for data
  - Can detect subprocess failures early
  - Can timeout if subprocess hangs

2. **Switch to blocking mode** once first data arrives
  - Use `fcntl()` to remove `O_NONBLOCK` flag
  - Continue with blocking `read()` calls
  - **POSIX guarantee**: Blocking `read()` returns EOF (0 bytes) ONLY after writer closes AND all kernel pipe buffers are drained

### DIFF
--- a/metaflow/plugins/aws/step_functions/step_functions_cli.py
+++ b/metaflow/plugins/aws/step_functions/step_functions_cli.py
@@ -191,7 +191,6 @@ def create(
                 },
                 f,
             )
-            f.flush()
 
     obj.echo(
         "Deploying *%s* to AWS Step Functions..." % obj.state_machine_name, bold=True
@@ -560,7 +559,6 @@ def trigger(obj, run_id_file=None, deployer_attribute_file=None, **kwargs):
                 },
                 f,
             )
-            f.flush()
 
     obj.echo(
         "Workflow *{name}* triggered on AWS Step Functions "
--- a/metaflow/plugins/argo/argo_workflows_cli.py
+++ b/metaflow/plugins/argo/argo_workflows_cli.py
@@ -287,7 +287,6 @@ def create(
                 },
                 f,
             )
-            f.flush()
 
     obj.echo("Deploying *%s* to Argo Workflows..." % obj.flow.name, bold=True)
 
@@ -945,7 +944,6 @@ def trigger(obj, run_id_file=None, deployer_attribute_file=None, **kwargs):
                 },
                 f,
             )
-            f.flush()
 
     obj.echo(
         "Workflow *{name}* triggered on Argo Workflows "


metaflow/plugins/aws/step_functions/step_functions_cli.py
```diff
@@ -191,6 +191,7 @@ def create(
                 },
                 f,
             )
+            f.flush()
```
savingoyal: why is this needed? wouldn't f.close() already take care of this?

metaflow/runner/utils.py
```diff
@@ -129,30 +130,32 @@ def read_from_fifo_when_ready(
...
+                while True:
+                    chunk = os.read(fifo_fd, 8192)
+                    if not chunk:
+                        # True EOF - all data drained
+                        break
+                    content += chunk
+                # All data read, exit main loop
+                break
             else:
                 if len(events):
```
romain-intel, line 149: When does this happen? So we got some event (like file close?) and no data?

metaflow/runner/utils.py
```diff
@@ -129,30 +130,32 @@ def read_from_fifo_when_ready(
...
+
+                # Now do blocking reads until true EOF
+                while True:
+                    chunk = os.read(fifo_fd, 8192)
+                    if not chunk:
+                        # True EOF - all data drained
+                        break
+                    content += chunk
+                # All data read, exit main loop
+                break
```
aquarin, line 147: This is a very unit-testable chunk of code. Can we add unit test to:
1. To trigger the scenario that caused the error.
2. Verify the unit test fails before this code change.
3. Verify the unit test passes after this fix.

npow: We could add unit tests for this chunk of code, but the actual bug is difficult to reproduce reliably because of a race condition: the writer has to close while kernel buffers still has data

metaflow/runner/utils.py
npow: If poll() returned an event (any event) AND read() returned 0 bytes then it must be EOF:

  - If it was POLLIN (data ready), then read() would have returned data
  - If read() returned 0 bytes despite an event, it must be POLLHUP (writer closed) or a stale POLLIN that resolved to EOF

PR #2656, opened 2025-10-24
fix: PyPI cross-platform resolves when dependencies have platform markers

saikonen:
The current PyPI implementation can fail to produce a working environment in the case where the resolving machine is of a different architecture than the deployment target. Such a case is when any of the packages requested employs platform markers in their dependencies, as these depending on the value these can be ignored by the machine resolving the environment, but end up being required by the one executing the code.

PyTorch is one such library, where it specifies CUDA dependencies with the following constraints
```
; platform_system == \"Linux\" and platform_machine == \"x86_64\"
```
Resolving the environment on a machine not matching the above, all CUDA dependencies would end up not being shipped in the code package.
During execution though, torch would expect to find these, as the markers match the platform. This would lead to errors during runtime unless CUDA was provided by other means, f.ex. by the Docker base image being used.

### DIFF
--- a/metaflow/plugins/pypi/pip.py
+++ b/metaflow/plugins/pypi/pip.py
@@ -151,7 +151,7 @@ class Pip(object):
                     res["hash"] = vcs_info["commit_id"]
                 return res
 
-            from packaging.requirements import Requirement
+            from metaflow._vendor.packaging.requirements import Requirement
 
             def _extract_platform_specific_deps(
                 pkg_metadata, target_system, target_machine


metaflow/plugins/pypi/pip.py
```diff
@@ -150,6 +152,82 @@ def _format(dl_info):
...
+                    )
+
+                    if plat_system is None and plat_machine is None:
+                        continue
+
+                    if plat_system == target_system or plat_machine == target_machine:
+                        # we must make sure that this dependency gets added to the list,
+                        # as it will not be carried by the default resolve due to platform/machine mismatch.
+                        # TODO: keep the original constraint, as they might not be strict version pins.
+                        p, v = pkg.split("==")
```
savingoyal: i believe this TODO needs to be addressed before merge?

metaflow/plugins/pypi/pip.py
```diff
@@ -150,6 +151,85 @@ def _format(dl_info):
                     res["hash"] = vcs_info["commit_id"]
                 return res
 
+            from packaging.requirements import Requirement
```
saikonen: note implicit dependency on packaging. this is at least limited to only the solve step, but still not desirable

savingoyal: we can vendor this lib

romain-intel: packaging is already vendored.

romain-intel:
So are you also re-exploring the patching approach?

saikonen:
closing in favour of #2322

PR #2659, opened 2025-10-27
fix: extremely wide foreach on argo breaking

saikonen:
an extremely wide foreach on argo can break due to the argument list for the shell command becoming too long as we pass all task id's as input-paths to the step.

This PR changes it so the input-paths are dumped to disk instead and used from the file.


iamsgarg-ob:
lgtm

PR #2670, opened 2025-11-03
Fix: S3 path trailing slash in s3 test code

aquarin:
### Issue
* test_list_recursive_sibling_prefix_filtering is failing
* We found `f"{base_s3root}/{test_prefix}/log/` generated "/" twice.
* This is confirmed through putting that generated path string to `aws s3 ls` command. Double slash returned nothing, while single slash returned a correct list.

### Fix
* This is a short-term fix, by simply stripping out the trailing "/"
* In the long term, we need to figure out a way to normalize the s3 path.

### DIFF
--- a/devtools/Tiltfile
+++ b/devtools/Tiltfile
@@ -630,7 +630,7 @@ if "metadata-service" in enabled_components:
             'metadatadb.database=metaflow',
             'metadatadb.host=postgresql',
             'image.repository=public.ecr.aws/outerbounds/metaflow_metadata_service',
-            'image.tag=2.4.13-2-g70af4ed',
+            'image.tag=2.5.0',
             'resources.requests.cpu=25m',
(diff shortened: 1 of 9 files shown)


test/data/s3/test_s3.py
```diff
@@ -977,7 +977,7 @@ def test_list_recursive_sibling_prefix_filtering(inject_failure_rate):
             s3_setup.put(file_path, f"test content for {file_path}")
 
     with S3(
-        s3root=f"{base_s3root}/{test_prefix}/log/",
```
npow, line 980: so the issue here is because s3root has consecutive slashes in the middle.. keeping it as `log/` should work

npow:
deprecated in favor of https://github.com/Netflix/metaflow/pull/2673

PR #2672, opened 2025-11-04
Fix: JobSet with @retry(times=0) validation error

### LINKED ISSUE #2632: Issue when using Jobset with @retry(times=0)
Hi, I am facing an issue when running metaflow_ray(Jobset) through Argo Workflow. The Template.retry_strategy() method only adds the retryStrategy payload when times > 0. This means when total_retries is 0,{{retries}} becomes None which can cause issues in the jobset template. Does anyone has any idea? Should the condition be >= ?

https://github.com/Netflix/metaflow/blob/9b98f3277030f5246d53b77ad4380c04ad695f57/metaflow/plugins/argo/argo_workflows.py#L4218-L4225

Jobset version : `0.9.1`
Argo Workflow: `3.5.8`
metaflow_ray: `0.1.4`
Metaflow: `3.18.9`

Here is my error

```
time="2025-10-07T05:54:16 UTC" level=info msg="kubectl create -f /tmp/manifest.yaml -o json"
The JobSet "js-82bf73{{retries}}" is invalid: metadata.name: Invalid value: "js-82bf73{{retries}}": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
time="2025-10-07T05:54:17 UTC" level=info msg="sub-process exited" argo=true error="<nil>"
Error: exit status 1
```

Here is my code

```
from metaflow import FlowSpec, project, step, load_config, kubernetes, retry, pip, metaflow_ray

@load_config
@project(name="retryflow")
class RetryFlow(FlowSpec):

    def _do_ray_job(self):
        import ray
        import time
        from counter import Counter

        ray.init()

        memory = ray.cluster_resources().get("memory")
        print("memory: %sGB" % (round(int(memory) / (1024 * 1024 * 1024), 2)))

        c = Counter.remote()

        for _ in range(10):
            time.sleep(1)
            c.incr.remote(1)

        print(ray.get(c.get.remote()))

    @step
    def start(self):
        self.next(self.execute, num_parallel=2)
        
    @retry(times=0)
    @kubernetes
    @metaflow_ray
    @pip(libraries={"ray": "2.49.1","metaflow-ray": "0.1.4"})
    @step
    def execute(self):
        self._do_ray_job()
        self.next(self.join)

    @step
    def join(self, inputs):
        self.next(self.end)

    @step
    def end(self):
        pass

if __name__ == "__main__":
    RetryFlow()

```

Dhanush010:
When using @retry(times=0) with JobSet through Argo Workflows, the workflow fails because {{retries}} is added to the jobset-name even though retryStrategy is not specified in the template.

The issue was that the code checked for the presence of the @retry decorator instead of checking if retries are actually enabled (total_retries > 0). When total_retries = 0, retry_strategy() doesn't add retryStrategy, so {{retries}} is not available in Argo Workflows.

Fix by computing total_retries and only adding retry-related parameters (including {{retries}} in the jobset-name) when total_retries > 0.

Fixes #2632

### DIFF
--- a/metaflow/plugins/argo/argo_events.py
+++ b/metaflow/plugins/argo/argo_events.py
@@ -58,7 +58,7 @@ class ArgoEvent(object):
             Value
         """
 
-        self._payload[key] = str(value)
+        self._payload[key] = value
         return self
 
     def safe_publish(self, payload=None, ignore_errors=True):
--- a/metaflow/plugins/cards/ui/package-lock.json
+++ b/metaflow/plugins/cards/ui/package-lock.json
@@ -3882,9 +3882,9 @@
       "license": "MIT"
     },
     "node_modules/js-yaml": {
-      "version": "4.1.0",
-      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.1.0.tgz",
-      "integrity": "sha512-wpxZs9NoxZaJESJGIZTyDEaYpl0FKSA+FB9aJiyemKhMwkxQg63h4T1KJgUGHpTqPDNRcmmYLugrRjJlBtWvRA==",
+      "version": "4.1.1",
+      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.1.1.tgz",
+      "integrity": "sha512-qQKT4zQxXl8lLwBtHMWwaTcGfFOZviOJet3Oy/xmGk2gZH677CJM9EvtfdSkgWcATZhj/55JZ0rmy3myCT5lsA==",
       "dev": true,
       "license": "MIT",
       "dependencies": {


Dhanush010:
Hi,
I've fixed the pre-commit formatting issue. The `black` formatter required the `max()` call to be split across multiple lines. Applied the formatting fix and pushed the changes - the pre-commit check should pass now.

Dhanush010:
Hi @valayDave,
The test failure is due to a transient infrastructure issue, not related to the code changes. The error shows:
Error: failed to fetch https://github.com/argoproj/argo-helm/releases/download/argo-events-2.4.8/argo-events-2.4.8.tgz : 503 Service Unavailable

This occurred during the "Bring up the environment" step when trying to download the argo-events Helm chart. The failure happened before any actual tests ran. This is a GitHub release download service issue (503 error), not a problem with the code changes.

Could someone please re-run the test? The other 31 checks passed successfully.

PR #2673, opened 2025-11-05
Fix s3 slash shenanigans

npow:
1. updated s3 tests to run everything with/without trailing `/` in `S3ROOT`
2. fixed failing tests


npow:
minIO tests failing due to 


> Applications must assign non-conflicting, unique keys for all objects. This includes avoiding creating objects where the name can collide with that of a parent or sibling object. AIStor returns an empty set for LIST operations at the location of the collision.


specifically, this is a prefix and an object:

> key: {'Key': 'metaflow/prefix/prefix', 'LastModified': datetime.datetime(2025, 11, 6, 10, 3, 44, 234000, tzinfo=tzutc()), 'ETag': '"cde78d0c26d4a660d0f8f077a310d842"', 'Size': 32, 'StorageClass': 'STANDARD'}

references:

1. https://docs.min.io/enterprise/aistor-object-store/reference/aistor-server/thresholds/#conflicting-objects
3. https://github.com/minio/minio/issues/20834

PR #2681, opened 2025-11-07
[CLI] Handle truthy values

npow:
https://github.com/Netflix/metaflow/blob/master/metaflow/runner/click_api.py#L141-L152 is only checking for True/False and doesn't set `cli_name` for other truthy/falsy values


savingoyal (approved):

npow:
Seems to be working now and we're unable to reproduce. Closing for now

PR #2685, opened 2025-11-12
Extend click types

talsperre:
We want to allow extensions to be able to add their own custom parameter types, and to be able to use these custom parameters via the Runner API as well. This PR allows extensions to update the list of "approved" parameter types in the metaflow runner API. 

It follows logic similar to `get_pinned_conda_libs` function.

### DIFF
--- a/metaflow/runner/click_api.py
+++ b/metaflow/runner/click_api.py
@@ -29,18 +29,6 @@ from typing import Union
 
 from metaflow import FlowSpec, Parameter
 from metaflow._vendor import click
-from metaflow._vendor.click.types import (
-    BoolParamType,
-    Choice,
-    DateTime,
-    File,
-    FloatParamType,
-    IntParamType,
-    Path,
-    StringParamType,
-    Tuple,
-    UUIDParameterType,
-)
 from metaflow.decorators import add_decorator_options
 from metaflow.exception import MetaflowException
 from metaflow.flowspec import FlowStateItems
@@ -51,8 +39,6 @@ from metaflow.user_configs.config_options import (
     ConfigValue,
     ConvertDictOrStr,
     ConvertPath,
-    LocalFileInput,
-    MultipleTuple,
     config_options_with_config_input,
 )
 from metaflow.user_decorators.user_flow_decorator import FlowMutator


metaflow/runner/click_api.py
```diff
@@ -57,26 +57,14 @@
 )
 from metaflow.user_decorators.user_flow_decorator import FlowMutator
 
+# Import Click type mappings from config (allows extensions to add custom types)
+from metaflow.metaflow_config import get_click_to_python_types
+
+click_to_python_types = get_click_to_python_types()
+
 # Define a recursive type alias for JSON
 JSON = Union[Dict[str, "JSON"], List["JSON"], str, int, float, bool, None]
```
mt-ob: nit: you can import this also from `metaflow/metaflow_config.py`??
right now it is defined at two places...

mt-ob (approved):
LGTM, but one minor comment...

metaflow/metaflow_config.py
```diff
@@ -615,6 +619,56 @@ def get_pinned_conda_libs(python_version, datastore_type):
...
+        DateTime,
+        File,
+        FloatParamType,
+        IntParamType,
+        Path,
+        StringParamType,
+        Tuple,
+        UUIDParameterType,
+    )
+    from metaflow.parameters import JSONTypeClass
```
mt-ob: imports should be moved to the top??

romain-intel:
This LGTM -- @talsperre , you can address the comment and we can merge?

metaflow/runner/click_api.py
```diff
@@ -57,25 +57,10 @@
...
-    UUIDParameterType: uuid.UUID,
-    Path: str,
-    DateTime: datetime.datetime,
-    Tuple: tuple,
-    Choice: str,
-    File: str,
-    JSONTypeClass: JSON,
-    FilePathClass: str,
-    LocalFileInput: str,
-    MultipleTuple: TTuple[str, Union[JSON, ConfigValue]],
```
mt-ob, line 77: can you check if the imports of `LocalFileInput` and `MultipleTuple` are no longer needed?

mt-ob: similarly for other imports at the top such as `StringParamType` etc.

PR #2688, opened 2025-11-13
fix: pypi with site pip config

saikonen:
site-specific pip configs were not working with `@pypi` as the pip config was fetched using the underlying conda environment, which has a different interpreter (and thus different site-config) than the hosts pip.

### DIFF
--- a/metaflow/plugins/pypi/pip.py
+++ b/metaflow/plugins/pypi/pip.py
@@ -1,5 +1,6 @@
 import json
 import os
+import sys
 import re
 import shutil
 import subprocess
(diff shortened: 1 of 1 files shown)


metaflow/plugins/pypi/pip.py
```diff
@@ -60,6 +61,9 @@ def __init__(self, micromamba=None, logger=None):
         else:
             self.logger = lambda *args, **kwargs: None  # No-op logger if not provided
 
+        # Resolve the host system pip for calls to fetch config
+        self._host_pip = which("pip3") or which("pip")
```
savingoyal: shouldn't we be using `pip` linked to `sys.executable -m pip`?

PR #2701, opened 2025-11-20
feature: add compute info to current singleton

saikonen:


saikonen:
test it out with

```python
from metaflow import step, FlowSpec, current

class HelloFlow(FlowSpec):
    @step
    def start(self):
        print("Starting 👋")
        print("Stats on current singleton:")
        print(f"     runtime: {current.runtime}")
        print(f"     compute: {current.compute}")
        print(f"     is_local: {current.is_local}")
        self.next(self.end)

    @step
    def end(self):
        print("Done! 🏁")


if __name__ == "__main__":
    HelloFlow()
```

PR #2703, opened 2025-11-22
[S3] raise error on s3 retry exhaustion

npow:
When `put_many()` or `put_files()` exhausted all transient retry attempts, the operations would silently return partial results instead of raising an exception.

### DIFF
--- a/metaflow/plugins/datatools/s3/s3.py
+++ b/metaflow/plugins/datatools/s3/s3.py
@@ -1598,7 +1598,7 @@ class S3(object):
         # Use the maximum of both retry counts to ensure consistent behavior
         # between single-op and multi-op cases
         max_retry_count = max(S3_TRANSIENT_RETRY_COUNT, S3_RETRY_COUNT)
-        transient_retry_count = 0  # Number of transient retries (per top-level retry)
+        transient_retry_count = 0  # Number of transient retries
         inject_failures = _inject_failure_rate()
         out_lines = []  # List to contain the lines returned by _s3op_with_retries
         pending_retries = (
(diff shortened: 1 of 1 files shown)


metaflow/plugins/datatools/s3/s3.py
```diff
@@ -1566,9 +1588,16 @@ def _inject_failure_rate():
...
+            # in normal testing scenarios. However, if explicitly set to 100,
+            # respect that for testing retry exhaustion paths.
+            if self._s3_inject_failures >= 100:
+                return self._s3_inject_failures
             return min(90, self._s3_inject_failures)
 
+        # Use the maximum of both retry counts to ensure consistent behavior
+        # between single-op and multi-op cases
+        max_retry_count = max(S3_TRANSIENT_RETRY_COUNT, S3_RETRY_COUNT)
         transient_retry_count = 0  # Number of transient retries (per top-level retry)
```
romain-intel: nit: remove comment: "per top level retry"

metaflow/plugins/datatools/s3/s3.py
```diff
@@ -1691,7 +1709,7 @@ def try_s3_op(last_ok_count, pending_retries, out_lines, inject_failures):
                         # Here we did not have any error -- transient or otherwise.
                         ok_lines = stdout.splitlines()
                         _update_out_lines(out_lines, ok_lines, resize=loop_count == 0)
-                        return (len(ok_lines), 0, inject_failures, None)
+                        return (len(ok_lines), 0, inject_failures, None, None)
```
romain-intel, line 1712: Comment on line 1714 could also be reworded a bit.

PR #2719, opened 2025-12-10
Remove a case where duplicate init functions were called (when deploy…

romain-intel:
…ing)

### DIFF
--- a/metaflow/plugins/pypi/conda_decorator.py
+++ b/metaflow/plugins/pypi/conda_decorator.py
@@ -330,11 +330,11 @@ class CondaFlowDecorator(FlowDecorator):
         decorators._attach_decorators(flow, ["conda"])
         decorators._process_late_attached_decorator(
             ["conda"],
-            obj.flow,
-            obj.graph,
-            obj.environment,
-            obj.flow_datastore,
-            obj.logger,
+            flow,
+            graph,
+            environment,
+            flow_datastore,
+            logger,
         )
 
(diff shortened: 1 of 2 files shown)


metaflow/plugins/pypi/pypi_decorator.py
```diff
@@ -141,7 +141,14 @@ def flow_init(
...
 
         decorators._attach_decorators(flow, ["pypi"])
-        decorators._init(flow)
+        decorators._process_late_attached_decorator(
+            ["pypi"],
+            obj.flow,
+            obj.graph,
+            obj.environment,
+            obj.flow_datastore,
+            obj.logger,
```
saikonen: `obj` is not defined. should this just be the parameters for `flow_init`?

metaflow/plugins/pypi/conda_decorator.py
```diff
@@ -328,6 +328,14 @@ def flow_init(
         from metaflow import decorators
 
         decorators._attach_decorators(flow, ["conda"])
+        decorators._process_late_attached_decorator(
+            ["conda"],
+            obj.flow,
+            obj.graph,
+            obj.environment,
+            obj.flow_datastore,
+            obj.logger,
```
saikonen: same here, there is no `obj`

romain-intel: copy-paste shenanigans :).

PR #2720, opened 2025-12-10
Fix a case where a default none config would get confused with the pl…

romain-intel:
…ain flag

### DIFF
--- a/test/unit/configs/flows/config_plain_flow.py
+++ b/test/unit/configs/flows/config_plain_flow.py
@@ -136,7 +136,9 @@ class ConfigPlainFlow(FlowSpec):
         assert (
             self.plain_none_config is None
         ), f"Expected None, got {self.plain_none_config}"
+        print(f"✓ Plain None config validated")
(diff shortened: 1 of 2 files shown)


test/unit/configs/flows/config_plain_flow.py
```diff
@@ -124,6 +133,12 @@ def start(self):
             f"✓ Plain tuple validated: {self.plain_tuple_value} (type: {self.plain_tuple_type})"
         )
 
+        assert (
+            self.plain_none_config is None
+        ), f"Expected None, got {self.plain_none_config}"
+        print(f"✓ Plain None config validated")
+        assert self.none_config is None, f"Expected None, got {self.none_config}"
+        print(f"✓ Non-plain None config validated")
```
npow, line 141: thanks claude

romain-intel: what claude -- in this case, claude's name is romain :)

PR #2723, opened 2025-12-15
feature: enabling freethreaded python support for pypi/conda

saikonen:
UX and possible corner cases still to be covered. Current approach is to have a `t` suffix in the python version to denote no GIL

### DIFF
--- a/metaflow/plugins/pypi/pip.py
+++ b/metaflow/plugins/pypi/pip.py
@@ -83,12 +83,13 @@ class Pip(object):
             )
 
         debug.conda_exec("Solving packages for PyPI environment %s" % id_)
+        freethreaded = python.endswith("t")
         with tempfile.TemporaryDirectory() as tmp_dir:
             report = "{tmp_dir}/report.json".format(tmp_dir=tmp_dir)
             implementations, platforms, abis = zip(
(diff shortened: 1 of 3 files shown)


saikonen:
easy test flow (swap out `@conda` for `@pypi` to test abi tag support):

```python
from metaflow import step, FlowSpec, conda
import sys

class CondaGILFlow(FlowSpec):
    @conda(python="3.13t")
    @step
    def start(self):
        print("Starting 👋")
        print("Is GIL? ", sys._is_gil_enabled())
        self.next(self.end)
    
    @conda(python="3.14t")
    @step
    def end(self):
        print("Is GIL? ", sys._is_gil_enabled())
        print("Done! 🏁")


if __name__ == "__main__":
    CondaGILFlow()
```

metaflow/plugins/pypi/pip.py
```diff
@@ -83,12 +83,13 @@ def solve(self, id_, packages, python, platform):
             )
 
         debug.conda_exec("Solving packages for PyPI environment %s" % id_)
+        freethreaded = python.endswith("t")
```
savingoyal, line 86: does it work with patch versions?

PR #2726, opened 2025-12-18
kill job based on terminate mode configuration

iamsgarg-ob:
Introduces a metaflow config based on which the user can control the behavior of how metaflow behaves when the user exit's out of a run ( Ctrl +C ) 

```
KUBERNETES_JOB_TERMINATE_MODE   = delete - metaflow will attempt to delete the BatchV1 CR altogether. 
KUBERNETES_JOB_TERMINATE_MODE   = <any other value> - falls back to the legacy behavior of setting job.spec.parallelism -> 0 
```

This moves it closer to how jobsets behave today.


metaflow/plugins/kubernetes/kubernetes_job.py
```diff
@@ -4,10 +4,12 @@
 import time
 
 from metaflow.exception import MetaflowException
-from metaflow.metaflow_config import KUBERNETES_SECRETS
+from metaflow.metaflow_config import KUBERNETES_SECRETS, KUBERNETES_JOB_TERMINATE_MODE
 from metaflow.tracing import inject_tracing_vars
 
 CLIENT_REFRESH_INTERVAL_SECONDS = 300
+DELETE_JOB_PROPAGATION_POLICY = "Background"
```
saikonen, line 11: Is there no reason to make this also configurable in conjunction with the termination mode? could simply be inlined otherwise.

PR #2732, opened 2025-12-23
pre-commit: Trim trailing whitespace

cclauss:
Add a `pre-commit` hook from https://github.com/pre-commit/pre-commit-hooks to trim
* [`trailing-whitespace`](https://github.com/pre-commit/pre-commit-hooks?tab=readme-ov-file#trailing-whitespace)


romain-intel (approved):
I would indeed avoid touching vendored stuff (but I think you updated it and removed that). LGTM otherwise.

PR #2742, opened 2026-01-08
[namespaced events]

valayDave:
- Make @trigger and ArgoEvent support namespaced eventing. This means users can do the following:

Have an upstream frow raise an ArgoEvent which is caught by a downstream flow under the same project and branch. For example, below flow raises the event.

```python
# python pub.py --branch foo run 
from metaflow import FlowSpec, step, project, namespaced_event_name
from metaflow.integrations import ArgoEvent

@project(name="myproject")
class MyFlow(FlowSpec):
    @step
    def start(self):
        # Publish a namespaced event: mfns.myproject.test.foo.data_ready
        ArgoEvent(namespaced_event_name("data_ready")).publish()
        self.next(self.end)

    @step
    def end(self):
        pass
```

And this one catches that event:
```python
# python pub.py --branch foo argo-workflows create 
@project(name="myproject")
@trigger(event=namespaced_event_name("data_ready"))
class DownstreamFlow(FlowSpec):
    ...
```
- Each event name returned by `namespaced_event_name` is blindly prefixed with `mfns.` to distinguish namespaced events (this is configurable via `METAFLOW_NAMESPACED_EVENTS_PREFIX`). this acts as a proxy for knowing that "metaflow" raised the event.
- The intention of this change is to allow users to enforce catching/raising events only within the same project/branch namespace to make experimentation much easier.
- We have a [custom decorator implementation](https://github.com/outerbounds/custom-decorator-examples/tree/main/namespaced_events) of this but it is not as natural as a core change.
- We rely on the same machinery introduced in Netflix/metaflow#2157 which allowed @trigger event/s parameters to be DeployTimeFields. We just lift this information / functionality into the docstring too and make it a little more public.

### DIFF
--- a/metaflow/metaflow_config.py
+++ b/metaflow/metaflow_config.py
@@ -455,6 +455,9 @@ ARGO_EVENTS_SENSOR_NAMESPACE = from_conf(
     "ARGO_EVENTS_SENSOR_NAMESPACE", KUBERNETES_NAMESPACE
 )
 
+# Prefix for namespaced events (used by @trigger with namespaced=True)
+NAMESPACED_EVENTS_PREFIX = from_conf("NAMESPACED_EVENTS_PREFIX", "mfns")
+
 ARGO_WORKFLOWS_UI_URL = from_conf("ARGO_WORKFLOWS_UI_URL")
 
 ##
--- a/metaflow/plugins/namespaced_events.py
+++ b/metaflow/plugins/namespaced_events.py
@@ -1,0 +1,105 @@
+from collections import namedtuple
+import functools
+import re
+
+from metaflow.metaflow_config import NAMESPACED_EVENTS_PREFIX
+from metaflow.util import is_stringish
+
+ParsedEvent = namedtuple(
+    "ParsedEvent",
+    ["full_name", "project", "branch", "logical_name", "is_namespaced"],
+)
+
+
+def namespaced_event_name(event_name):
+    """
+    Creates a project-namespaced event name based on @project settings.
+
+    Use this to automatically prefix event names with your @project
+    namespace, ensuring events are isolated per project and branch.
+
+    The resolved name follows the format:
+        mfns.<project>.<branch>.<event_name>
+
+    If no @project decorator is present, resolves to:
+        mfns.<event_name>
+
+    Parameters
+    ----------
+    event_name : str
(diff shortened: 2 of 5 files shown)


metaflow/metaflow_config.py
```diff
@@ -455,6 +455,9 @@
     "ARGO_EVENTS_SENSOR_NAMESPACE", KUBERNETES_NAMESPACE
 )
 
+# Prefix for namespaced events (used by @trigger with namespaced=True)
+NAMESPACED_EVENTS_PREFIX = from_conf("NAMESPACED_EVENTS_PREFIX", "mfns")
```
saikonen, line 459: is a namespace prefix really required if the main use case is in combination with `@project` ? wouldn't the project+branch be enough of a unique namespace?

metaflow/plugins/namespaced_events.py
```diff
@@ -0,0 +1,174 @@
...
+        Production boolean passed for projects on command line.
+
+    Returns
+    -------
+    str
+        Namespaced event name in format mfns.<project>.<is_prod>.<branch>.<logical_name>
+        or mfns.<logical_name> if no project/branch
+    """
+    if project and branch:
+        prod_prefix = "prod" if is_prod else "test"
```
saikonen: are these needed? isn't this info contained in the project & branch already for `--production` as well?

valayDave: Done!

metaflow/metaflow_config.py
valayDave: The problem is that users might already have been raising events in that format. If we are not introducing something that raises events in that format then we should make sure that we prefix it properly so it doesn't conflict with any events users are raising currently.

metaflow/plugins/namespaced_events.py
```diff
@@ -0,0 +1,163 @@
...
+        from metaflow import current
+
+        project_name = current.get("project_name", None)
+        branch_name = current.get("branch_name", None)
+        return _make_namespaced_event_name(event_name, project_name, branch_name)
+
+    return functools.partial(_delayed_name_generator, event_name=event_name)
+
+
+def _parse_namespaced_event(evt_name):
```
saikonen: this seems unused

saikonen (approved):

Issue #2741, opened 2026-01-08
Allow --only-json to generate Argo templates without requiring cluster access


vivekkhimani:
Hi team! 

We’re trying to generate static Argo Workflow templates from Metaflow without having Metaflow apply them directly to a Kubernetes cluster (i.e., similar to what `create` does, but generation-only).

We looked at using the `--only-json` flag, which is close to what we want. However, that code path still calls `ArgoWorkflows.get_existing_deployment()` to check for an existing workflow template in Kubernetes. This requires cluster access at generation time, which breaks our use case.

In our setup (GitOps-style), we want Metaflow to:

- Generate the Argo Workflow template
- Output it (JSON/YAML)
- Let our own tooling handle applying it to the cluster later

Requiring Kubernetes access during generation makes this difficult, since generation often happens in CI or other environments without cluster credentials. We locally worked around this by modifying the code to skip the existing-deployment check when `--only-json` is passed, and this has worked well for us so far. Before opening a PR, we wanted to ask:
- Would you be open to accepting this behavior upstream?
- Do you have opinions on whether `--only-json` should be strictly generation-only and avoid all cluster interactions?
- Is there a preferred way you’d want this structured (e.g., a new flag vs. tightening the semantics of --only-json)?

Happy to put together a clean PR if this aligns with the project’s direction. Thanks!

PR #2743, opened 2026-01-09
Skip Kubernetes cluster access when using `--only-json` flag in Argo Workflows CLI

vivekkhimani:
## Context

When using Metaflow with GitOps-style workflows, users often need to generate Argo Workflow templates without applying them directly to a Kubernetes cluster. The `--only-json` flag can technically help for this purpose by outputting the workflow for inspecting or consuming in external tooling.

However, the current implementation of `--only-json` still calls `ArgoWorkflows.get_existing_deployment()` (via `resolve_token()`) to check for an existing workflow template in Kubernetes. This call uses the Kubernetes client and requires cluster credentials, which breaks the following use cases:

- **CI/CD pipelines**: Generation often happens in CI environments (at least in our case, and I presume this to be a common industry pattern) that don't have (and shouldn't have) cluster credentials
- **GitOps workflows**: Users want Metaflow to generate the template and let their own tooling (ArgoCD) handle applying it to the cluster
- **Local development**: Developers may want to inspect the generated workflow without configuring cluster access

The cluster access is primarily needed for:
1. `check_metadata_service_version()` - Verifies metadata service compatibility
2. `resolve_token()` → `ArgoWorkflows.get_existing_deployment()` - Checks for existing deployments to handle production token authorization

Neither of these are necessary when the user explicitly indicates they only want JSON output.

## Solution

When `--only-json` is passed, skip all operations that require Kubernetes cluster access:

1. **Skip `check_metadata_service_version()`** - Not needed since we're not deploying
2. **Skip `resolve_token()`** - Not needed since we're not deploying and don't need to check existing deployments
3. **Use a placeholder production token** - Since the JSON is for inspection/external tooling, a placeholder value (`__PLACEHOLDER_PRODUCTION_TOKEN__`) is used instead

This change tightens the semantics of `--only-json` to be strictly generation-only, with no cluster interactions required.

## Trade-offs & Considerations

- **Placeholder token in generated JSON**: The generated workflow template will contain `__PLACEHOLDER_PRODUCTION_TOKEN__` instead of a real production token. This is intentional as users in GitOps workflows will typically inject the correct token through their own deployment pipeline or may not need it at all.

- **Token-related CLI options ignored**: When `--only-json` is used, the `--authorize`, `--generate-new-token`, and `--new-token` options are effectively ignored since token resolution is skipped.

- **Package upload still occurs**: The code package is still uploaded to cloud storage (S3/Azure/GCS) as part of `make_flow()`. This is necessary to generate accurate URLs in the workflow template. Ideal would be a fully offline generation but that's a non-trivial change and fundamentally challenges the core packaging architecture.

## Testing

To verify the change:

```bash
# Before: This would fail without cluster access
python myflow.py argo-workflows create --only-json

# After: Generates JSON without requiring cluster credentials
python myflow.py argo-workflows create --only-json > workflow.json
```

The generated JSON can then be applied to the cluster using external tooling:
```bash
kubectl apply -f workflow.json
```

Fixes issue #2741

### DIFF
--- a/metaflow/plugins/argo/argo_client.py
+++ b/metaflow/plugins/argo/argo_client.py
@@ -86,11 +86,16 @@ class ArgoClient(object):
                 if not continue_token:
                     break
             except client.rest.ApiException as e:
+                error_body = json.loads(e.body) if e.body else {}
+                error_message = error_body.get("message", e.reason)
                 if e.status == 404:
                     return None
-                raise ArgoClientException(
-                    json.loads(e.body)["message"] if e.body is not None else e.reason
-                )
+                elif e.status == 410 and error_body.get("reason") == "Expired":
+                    new_token = error_body.get("metadata", {}).get("continue")
+                    if new_token:
+                        continue_token = new_token
+                        continue
+                raise ArgoClientException(error_message)
 
     def register_workflow_template(self, name, workflow_template):
         # Unfortunately, Kubernetes client does not handle optimistic
--- a/metaflow/version.py
+++ b/metaflow/version.py
@@ -1,1 +1,1 @@
-metaflow_version = "2.19.15"
+metaflow_version = "2.19.16"


metaflow/plugins/argo/argo_workflows_cli.py
```diff
@@ -290,22 +290,29 @@ def create(
...
-        given_token,
-        generate_new_token,
-        obj.is_project,
-    )
+    if only_json:
+        # When only generating JSON, we skip cluster access operations:
+        # - Metadata service version check (requires service access)
+        # - Token resolution (requires Kubernetes cluster access to check existing deployments)
+        # Instead, we use a placeholder token since the JSON is just for inspection.
+        token = "__PLACEHOLDER_PRODUCTION_TOKEN__"
```
saikonen, line 298: issue with this approach is that the kubernetes annotations in the produced template will be incorrect for the flow, e.g. `metaflow/production_token` will have the placeholder value instead of an actual matching prod token.

this can result in issues with re-deploying the flow via CLI later on as the prod token check will fail.

was the intention with the naming such that users using the json-only template would replace the placeholder value with their own if needed?

saikonen: the alternative though is to generate a token, but with this we would end up always replacing existing prod tokens when deploying the json-only template, as there is no offline way of checking for existing deployments and reusing its token.

Leaning towards the current PR being an acceptable approach for this reason.

vivekkhimani: > was the intention with the naming such that users using the json-only template would replace the placeholder value with their own if needed?

That is right. The thinking was that the users who actually want to deploy directly from the CLI itself hopefully won't use the `--only-json` flag to get a template and will rely on the usual `argo-workflow create` and `argo-workflow trigger` commands.

PR #2754, opened 2026-01-19
[kubernetes] add image for task in environment variables

valayDave:


metaflow/plugins/argo/argo_workflows.py
```diff
@@ -2261,6 +2261,7 @@ def _container_templates(self):
                         "METAFLOW_DEFAULT_METADATA": DEFAULT_METADATA,
                         "METAFLOW_CARD_S3ROOT": CARD_S3ROOT,
                         "METAFLOW_KUBERNETES_WORKLOAD": 1,
+                        "METAFLOW_KUBERNETES_CONTAINER_IMAGE": resources["image"],
```
savingoyal, line 2264: it would be great to figure out the actual version of the image used with sha tags somehow

PR #2755, opened 2026-01-20
Fix #524: Ignore --max-value-size when using dump --file and   Fixes #684 double slashes in S3 paths with trailing slash in config

### LINKED ISSUE #524: `--max-value-size` should be ignored when using `dump --file`

### LINKED ISSUE #684: Double slashes in `metaflow.S3` keys

dheerenmohta:
## Summary
  Fixes #524

  When using `metaflow dump --file output.pkl`, the `--max-value-size`
  parameter should be ignored to allow complete artifact data to be saved.

  ## Problem
  Previously, when dumping to a file, large artifacts were being replaced
  with `< artifact too large >` placeholder, even though files can handle
  large data.

  ## Solution
  Modified `cli_components/dump_cmd.py` to set `max_value_size=None` when
  `--file` is provided, while maintaining the size limit for stdout output
  (default 1000 bytes) to prevent terminal flooding.

  ## Changes
  - **File modified:** `cli_components/dump_cmd.py` (line 50)
  - Changed `max_value_size` to be conditionally set based on whether
  `--file` is provided

  ## Testing
  - When dumping to file: all artifacts are saved completely
  - When dumping to stdout: size limit still applies as expected (default
  1000 bytes)


romain-intel:
I don't have a problem with this one (minus the coupling to the other one) but let me check with others too. We may have to change the doc a bit.

metaflow/plugins/datatools/s3/s3.py
```diff
@@ -143,6 +143,39 @@ class MetaflowS3InsufficientDiskSpace(MetaflowException):
     headline = "Insufficient disk space"
 
 
+def _s3_path_join(*parts):
```
npow: this change seems unrelated to the PR title?

dheerenmohta: thanks @npow , updated the PR title, 

Fix #524: Ignore --max-value-size when using dump --file and   Fixes #684 double slashes in S3 paths with trailing slash in config

npow: Sorry my message wasn't clear -- since these are two unrelated changes, you should split the PRs (you already have a separate one for the S3 changes, which is good) and the S3 changes shouldn't be in this one. One PR = one logical change, don't mix unrelated changes

dheerenmohta:
Thanks for the feedback @npow! I've updated this PR to remove the #524 fix (--max-value-size). This PR now contains only the S3 changes for issue #684.

However, after further investigation based on your earlier comment on PR #2750, I discovered that the S3 path normalization approach was incorrect. According to the AWS S3 specification and boto3 implementation, S3 paths should NOT be normalized - double slashes create different objects.

I've reverted the normalization changes and added comprehensive inline documentation explaining why paths are preserved as-is. The recommendation now is for users to configure their S3 root without trailing slashes.

Should I close this PR and keep only PR #2750 (which now has the reverted changes with documentation)?

dheerenmohta:
Closing this PR as it was mixing two unrelated fixes (#524 and #684). 

I've separated them:
- PR #2750 now contains only the S3 changes for #684 (with the revert and documentation)
- I'll create a separate PR for the #524 fix

PR #2757, opened 2026-01-21
Add delete and bulk-delete support to S3 client

### LINKED ISSUE #1010: Add delete calls to S3 client

Shrujanapatil:
Fixes #1010

Adds delete and bulk-delete support to the Metaflow S3 client to enable
proper cleanup of S3 artifacts and prevent orphaned objects.

The implementation follows existing retry and error-handling patterns.

### DIFF
--- a/metaflow/plugins/datastores/s3_storage.py
+++ b/metaflow/plugins/datastores/s3_storage.py
@@ -75,6 +75,15 @@ class S3Storage(DataStoreStorage):
                 for o in results
             ]
 
+    def delete_files(self, paths):
+    with S3(
+        s3root=self.datastore_root,
+        tmproot=ARTIFACT_LOCALROOT,
+        external_client=self.s3_client,
+    ) as s3:
+        s3.delete_many(paths)
+
+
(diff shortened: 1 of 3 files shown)


romain-intel:
There seems to be something wrong with this PR... It looks like you uploaded your whole venv?

Shrujanapatil:
Thanks for catching the venv issue — that’s fixed now

metaflow/plugins/datastores/s3_storage.py
```diff
@@ -75,6 +75,15 @@ def list_content(self, paths):
                 for o in results
             ]
 
+    def delete_files(self, paths):
+    with S3(
```
npow, line 79: formatting is off -- this shouldn't be valid python code. can you also add some tests? that would have caught this issue. 

i'm updating the contributing guide with instructions on testing (and other things) here, you can take a look: https://github.com/Netflix/metaflow/blob/npow/contributing/CONTRIBUTING.md

PR #2765, opened 2026-01-24
fix: S3 list_paths regression

saikonen:
fixes a regression introduced in `2.19.6` that affects listing paths in the bucket root with the `S3` client `list_paths`

### DIFF
--- a/test/data/s3/conftest.py
+++ b/test/data/s3/conftest.py
@@ -3,7 +3,14 @@ from metaflow import current
 from .. import S3ROOT
 
 # S3ROOT variants for testing both with and without trailing slash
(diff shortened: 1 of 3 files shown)


saikonen:
repro:
```python
# test.py
import sys
from metaflow import S3


if __name__=="__main__":
    try:
      bucket = sys.argv[1]
      print("listing bucket:", bucket)
    except IndexError:
        print("Please provide a bucket name.")
        sys.exit(1)

    with S3(s3root=f"s3://{bucket}") as s3:
      objs = s3.list_paths()
      print([o.key for o in objs])
```

```sh
# lists nothing
python test.py your-metaflow-bucket

# lists just fine for subpaths
python test.py your-metaflow-bucket/metaflow
```

npow:
All internal S3 tests passing

Issue #2692, opened 2025-11-17
Expose container-level containerProperties configuration (IPC mode, ulimits, sharedMemorySize, linuxParameters) in @batch decorator


aman5319:
We are running Metaflow on AWS Batch and recently hit a limitation when deploying GPU-based Graph Neural Network workloads (PyG). PyG requires larger shared memory (/dev/shm) and specific ulimits to avoid Bus Errors during large graph allocations.

These require Docker flags such as:

```
--ipc=host
--ulimit memlock=-1
--ulimit stack=67108864

```

The exact warning I get when running. 

**NOTE: The SHMEM allocation limit is set to the default of 64MB. This may be
insufficient for PyG. NVIDIA recommends the use of the following flags:
docker run --gpus all --ipc=host --ulimit memlock=-1 --ulimit stack=67108864 ...**

Currently, the **batch** decorator does not expose any way to configure these container-level properties. Inspecting batch_client.py, only max_swap and swappiness are exposed via linuxParameters, with no escape hatch to modify the full containerProperties.

This makes certain ML workloads (e.g., PyG, DGL, large CUDA graphs, RAPIDS) impossible to run without modifying Metaflow internals.

**Requested Feature**
Expose either of the following:

**Option A: Direct kwargs pass-through**
Allow batch(container_properties=...) to merge into the AWS Batch Job Definition.

**Option B: Escape hatch**
Something like:

```
batch(
    gpu=1,
    image="xyz",
    ecs_overrides={
        "containerProperties": {...}
    }
)

```

**Why this matters**
Graph-based workloads, high-memory CUDA ops, and frameworks like PyG require larger SHMEM and specific ulimits. Without these, Metaflow cannot support a class of real-world ML tasks that rely on Batch GPU compute.

I’m happy to submit a PR once there’s agreement on the API design.

savingoyal:
thanks for opening the issue. option a looks most promising - would you be open to a PR? depending on the options being introduced, we can quickly iterate on the UX after the PR exists - my expectation is that changing the UX shouldn't be much work.

PR #2772, opened 2026-01-29
Expose container-level containerProperties (Ulimit)

aman5319:
Refer to issue - https://github.com/Netflix/metaflow/issues/2692

### DIFF
--- a/metaflow/plugins/aws/batch/batch_cli.py
+++ b/metaflow/plugins/aws/batch/batch_cli.py
@@ -190,6 +190,8 @@ def kill(ctx, run_id, user, my_runs):
     type=int,
     help="Number of parallel nodes to run as a multi-node job.",
 )
+@click.option("--ulimits", help="Ulimits requirement for AWS Batch.")
+
 @click.pass_context
 def step(
     ctx,
@@ -222,6 +224,7 @@ def step(
     log_driver=None,
     log_options=None,
     num_parallel=None,
+    ulimits=None,
     **kwargs
 ):
     def echo(msg, stream="stderr", batch_id=None, **kwargs):
@@ -366,6 +369,7 @@ def step(
                 log_driver=log_driver,
                 log_options=log_options,
                 num_parallel=num_parallel,
+                ulimits=ulimits
             )
     except Exception:
         traceback.print_exc()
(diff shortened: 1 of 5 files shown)


aman5319:
@savingoyal

metaflow/plugins/aws/batch/batch_cli.py
```diff
@@ -190,6 +190,8 @@ def kill(ctx, run_id, user, my_runs):
     type=int,
     help="Number of parallel nodes to run as a multi-node job.",
 )
+@click.option("--ulimits", help="Ulimits requirement for AWS Batch.")
```
saikonen, line 193: as ulimits is presumably a dict and not just a string, this should be of type `JSONTypeClass()` to properly parse.

metaflow/plugins/aws/batch/batch_decorator.py
```diff
@@ -127,6 +127,7 @@ class BatchDecorator(StepDecorator):
         "log_driver": None,
         "log_options": None,
         "executable": None,
+        "ulimits": None,
```
saikonen, line 130: please add a description for the attribute to the docstring as well, what is the value supposed to be like etc.

saikonen: also, as the value is presumably not just a string, this needs to be properly serialized in order to correctly pass to the batch step CLI. See for example how `kubernetes_decorator.py` and `kubernetes_cli.py` handle the `tolerations` attribute.

metaflow/plugins/aws/batch/batch_client.py
```diff
@@ -285,6 +286,40 @@ def _register_job_definition(
...
+                            )
+                        )
+
+                    # Check for softLimit and hardLimit (AWS requirement)
+                    if "softLimit" not in item or "hardLimit" not in item:
+                        raise BatchJobException(
+                            "ulimit '{}' is missing 'softLimit' or 'hardLimit'".format(name)
+                        )  
+                
+                job_definition["containerProperties"]["ulimits"]=ulimits
```
saikonen, line 321: validations of this kind are better done in the batch decorators `step_init` method. It might also be fine to run without any validations (and support AWS possibly changing what they accept), as the Batch API response should error out with something sensible?

saikonen (changes requested):
could you also rebase the PR to solve the conflicts

metaflow/plugins/aws/batch/batch_client.py
aman5319: I think handling from metaflow codebase is better solution and giving a valid reason saves the headache because when we silently let AWS handle the errors could be harder to debug.

PR #2775, opened 2026-01-30
Add support for class attributes and bound methods to escape hatch

malder08:
Update the escape hatch to allow escaping the following components:
- Class-level attributes
   - Also fixed a bug where falsy enum values (e.g. 0 and “”) would throw a RuntimeError since they were considered None 
- Bound methods


metaflow/plugins/env_escape/client.py
```diff
@@ -378,16 +378,16 @@ def name_to_parent_name(name):
         else:
             lookup_name = name
 
-        if name == "function":
-            # Special handling of pickled functions. We create a new class that
+        if name in ("function", "method"):
+            # Special handling of pickled functions and methods. We create a new class that
```
wangchy27, line 382: what is the difference between method and function here? Does `member` means class member function?

metaflow/plugins/env_escape/server.py
```diff
@@ -391,9 +401,9 @@ def pickle_object(self, obj):
     def unpickle_object(self, obj):
         if (not isinstance(obj, ObjReference)) or obj.value_type != VALUE_LOCAL:
             raise ValueError("Invalid transferred object: %s" % str(obj))
-        obj = self._local_objects.get(obj.identifier)
-        if obj:
-            return obj
+        result = self._local_objects.get(obj.identifier)
```
wangchy27, line 404: Good catch!

metaflow/plugins/env_escape/client.py
malder08: Yes `method` is a class instance method. I believe they should be handled the same as standalone functions but because the type is `method` instead of `function`, the escape hatch didn't recognize it and threw an error.

PR #2770, opened 2026-01-28
Feat - Make auto emission of events configurable from argo UI

tfurmston:
This PR adds a parameter to Metaflow Argo Workflow templates that allows users to toggle whether to emit event information to Argo Events when submitting a workflow from the Argo Workflow UI. 

This feature was discussed on the Metaflow Outerbounds Slack workspace [here](https://outerbounds-community.slack.com/archives/C02116BBNTU/p1769435079655379).

PR #2782, opened 2026-02-04
Feat/make auto emit events configurable from argo UI

tfurmston:
This PR is a re-opening of this [PR](https://github.com/Netflix/metaflow/pull/2770), which was merged and then reverted. 

The reason for the original PR being reverted can be found [here](https://github.com/Netflix/metaflow/pull/2770#issuecomment-3836605447). Essentially, because of the following:

* introducing a new always-present parameter to flows on argo that will be confusing to users
* adding argo-specific references to parameters.py is undesirable

### Previous PR Description

It adds a parameter to Metaflow Argo Workflow templates that allows users to toggle whether to emit event information to Argo Events when submitting a workflow from the Argo Workflow UI.

This feature was discussed on the Metaflow Outerbounds Slack workspace [here](https://outerbounds-community.slack.com/archives/C02116BBNTU/p1769435079655379).

### DIFF
--- a/.github/workflows/test-stubs.yml
+++ b/.github/workflows/test-stubs.yml
@@ -42,7 +42,11 @@ jobs:
     - name: Install Python ${{ matrix.ver }} dependencies
       run: |
         python3 -m pip install --upgrade pip setuptools
-        python3 -m pip install pytest build "mypy<1.9" pytest-mypy-plugins
+        if python3 -c "import sys; exit(0 if sys.version_info >= (3,10) else 1)"; then
+          python3 -m pip install pytest build "mypy<1.9" "pytest-mypy-plugins>=4"
+        else
+          python3 -m pip install pytest build "mypy<1.9" "pytest-mypy-plugins<4"
+        fi
 
     - name: Install metaflow
       run: pip install .
@@ -66,4 +70,10 @@ jobs:
         max_attempts: 2
         timeout_minutes: 3
         retry_on: error
-        command: cd ./stubs && pytest --mypy-ini-file test/mypy_${{ matrix.ver }}.cfg --mypy-only-local-stub && cd -
+        command: |
+          cd ./stubs
+          if python3 -c "import sys; exit(0 if sys.version_info >= (3,10) else 1)"; then
+            pytest --mypy-ini-file test/mypy_${{ matrix.ver }}.cfg
+          else
+            pytest --mypy-ini-file test/mypy_${{ matrix.ver }}.cfg --mypy-only-local-stub
+          fi
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -81,7 +81,7 @@ jobs:
         python3 -m venv path/to/venv
         source path/to/venv/bin/activate
         python3 -m pip install .
-        Rscript -e 'install.packages("devtools", repos="https://cloud.r-project.org", Ncpus=8)'
+        Rscript -e 'install.packages(c("devtools", "remotes"), repos="https://cloud.r-project.org", Ncpus=8)'
         Rscript -e 'devtools::install_deps("R", dependencies=TRUE, repos="https://cloud.r-project.org", upgrade="default")'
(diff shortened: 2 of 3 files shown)


tfurmston:
Hey @saikonen 

I've reopened the PR here. 

> I haven't had time to look into this thoroughly yet, but am going to explore some mechanisms of defining parameters that can override CLI options where applicable.
Some kind of way to configure that "this is a CLI option I want to expose in Argo as well" is needed, but not all options can be lifted anyway as some are deploy-time only.

For my understanding, when you are talking about CLI option you are talking about the CLI options in the Argo Workflows plugin, right? So the idea would be to allow certain CLI options in that plugin to become parameters in the workflow somehow?

tfurmston:
@saikonen How about we add a Boolean flag to the Argo Workflow CLI that when set adds this particular option as a parameter to the workflow? 

If the flag is set and the there is a nameclash with an existing parameter in the workflow, which I guess would be very unlikely anyway, we would just fail the workflow creation. If someone wanted this option set, they would have to update the parameter name in the workflow. 

Would be much less intrusive to the main codebase than the original approach.

What do you think?

tfurmston:
> ### Greptile Summary
> 
> This PR re-introduces a change (originally in PR #2770, which was merged and then reverted) that makes the `auto-emit-argo-events` toggle configurable at Argo workflow submission time by exposing it as an Argo workflow-level parameter (`auto_emit_argo_events`). The CLI flag previously baked in the deployment-time boolean is replaced with an Argo template expression `{{workflow.parameters.auto_emit_argo_events}}`, letting users override it from the Argo UI.
> 
> However, the PR does not appear to have addressed either of the two issues cited in the original revert:
> 
>     * **Always-present parameter**: `auto_emit_argo_events` is appended unconditionally to the workflow arguments of every Argo flow, including those that have no Argo Events integration. This will surface a confusing, irrelevant parameter in the Argo UI for the majority of users.
> 
>     * **Argo-specific coupling in `parameters.py`**: Adding `"auto-emit-argo-events"` to the global reserved parameter list in `parameters.py` introduces an Argo-specific concept into a general-purpose module, which is undesirable from an architectural standpoint.
> 
> 
> Both of these issues were explicitly listed as the reasons for the prior revert, and neither seems to be addressed in this re-submission.
> ### Confidence Score: 2/5
> 
>     * This PR re-introduces changes that were previously reverted without addressing the underlying architectural concerns that triggered the revert.
> 
>     * The two issues that caused the original PR to be reverted — unconditional exposure of an Argo-specific parameter for all workflows, and coupling `parameters.py` to Argo — are both still present in this re-submission. The functional change itself is straightforward and likely correct at runtime, but the design concerns are significant enough to warrant addressing before merging.
> 
>     * `metaflow/parameters.py` (Argo-specific reserved name in a general module) and `metaflow/plugins/argo/argo_workflows.py` (unconditional addition of the `auto_emit_argo_events` parameter to all workflows).
> 
> 
> ### Important Files Changed
> Filename 	Overview
> metaflow/plugins/argo/argo_workflows.py 	Replaces the hard-coded `auto-emit-argo-events` CLI flag with an Argo template expression `{{workflow.parameters.auto_emit_argo_events}}` and adds a corresponding workflow-level parameter so the value can be overridden from the Argo UI. The parameter is added unconditionally to all workflows, reintroducing the UX concern that caused the previous PR to be reverted.
> metaflow/parameters.py 	Adds `"auto-emit-argo-events"` to the global reserved parameter name list, coupling a general-purpose module to an Argo-specific concept — the same architectural concern that caused the original PR to be reverted.
> 
> Last reviewed commit: [5ea7c67](https://github.com/Netflix/metaflow/commit/5ea7c67b370ad78a099648da1c941a9b4745c2f0)

This is true, but at present the PR has been reopened as-is in order to discuss alternatives. As such, it is not suggested that this PR has addressed those issues, which I agree are valid points. Once we reach  a consensus I will be happy to make the changes to address these points.

PR #2788, opened 2026-02-11
Add pathspec format validation for Client API objects

### LINKED ISSUE #948: Pathspec to create Flow/Run/Step/Task/DataArtifact is not validated

dheerenmohta:
## Summary

Fixes #948

Added pathspec format validation to improve error messages when creating Flow/Run/Step/Task/DataArtifact objects.

## Problem

Before this change, if you made a typo in a pathspec (like `Run("MyFlow/abc")` where the run ID should be numeric), you'd only get a generic "not found" error from the metadata provider. This made it hard to spot simple mistakes like:
- Typos in names
- Using dashes instead of underscores
- Starting names with numbers
- Extra or missing slashes

## Solution

Added validation that checks:
- **Flow/Step/Artifact names**: Must start with letter or underscore, can contain alphanumeric + underscores (basically valid Python identifiers)
- **Run IDs and Task IDs**: Must be numeric
- **No empty components**: Rejects things like `"Flow//Step"`
- **No leading/trailing slashes**: Rejects `"/Flow"` or `"Flow/"`

The validation happens immediately when you create the object, so you get a clear error right away.

## Examples

**Before:**
```python
Run("MyFlow/abc123")  # Fails with "Run does not exist"
```

**After:**
```python
Run("MyFlow/abc123")
# MetaflowInvalidPathspec: Invalid run ID 'abc123'. Run IDs must be numeric.
```

Other examples:
- `Flow("123Flow")` → "Invalid flow name '123Flow'. Flow names must start with a letter or underscore..."
- `Step("Flow//step")` → "Pathspec 'Flow//step' contains empty components"

## Testing

Added comprehensive unit tests in `test/unit/test_pathspec_validation.py`:
- ✅ Valid pathspecs for all object types
- ✅ Invalid pathspecs with helpful error messages
- ✅ Edge cases (empty strings, special chars, etc.)

Tested manually with 40+ test cases - all passing.

## Backward Compatibility

This is fully backward compatible. All previously valid pathspecs still work - we're only adding validation for cases that would have failed anyway (just with better error messages now).

### DIFF
--- a/metaflow/client/core.py
+++ b/metaflow/client/core.py
@@ -2,6 +2,7 @@ from __future__ import print_function
 
 import json
 import os
+import re
 import tarfile
 from collections import namedtuple
 from datetime import datetime
@@ -59,6 +60,16 @@ filecache = None
 current_namespace = False
 
 current_metadata = False
+
+# Pathspec validation patterns
+# Flow names and step names should be valid Python identifiers (start with letter/underscore)
+_FLOW_NAME_PATTERN = re.compile(r'^[a-zA-Z_][a-zA-Z0-9_]*$')
+# Run IDs and Task IDs are numeric
+_NUMERIC_ID_PATTERN = re.compile(r'^[0-9]+$')
+# Step names should be valid Python identifiers
+_STEP_NAME_PATTERN = re.compile(r'^[a-zA-Z_][a-zA-Z0-9_]*$')
+# Artifact names should be valid Python identifiers
+_ARTIFACT_NAME_PATTERN = re.compile(r'^[a-zA-Z_][a-zA-Z0-9_]*$')
 
 
 def metadata(ms: str) -> str:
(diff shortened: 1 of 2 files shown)


savingoyal:
@dheerenmohta IDs can be non-numeric especially run and task ids

talsperre:
@dheerenmohta closing this PR since tasks/run ids can be non numeric.

PR #2792, opened 2026-02-13
fix tarfile filter compat for python < 3.12 in uv bootstrap

### LINKED ISSUE #2759: Bug: Metaflow uv plugin requires Python 3.12 becuase of use of TarFile filters
Metaflow explicitly supports older versions of Python, but if you use the `--environment=uv` flag, it requires >= 3.12 because the plugin adds a filter:
https://github.com/Netflix/metaflow/blob/f4b2dbb725a5cc3e0a71504097cf8a93ea62ca8f/metaflow/plugins/uv/bootstrap.py#L53

This tar filtering was only added in Python 3.12 [and wasn't backported](https://docs.python.org/3/library/tarfile.html):
> Changed in version 3.12: Archives are extracted using a [filter](https://docs.python.org/3/library/tarfile.html#tarfile-extraction-filter), which makes it possible to either limit surprising/dangerous features, or to acknowledge that they are expected and the archive is fully trusted.
see also [here](https://discuss.python.org/t/should-tarfile-filtering-have-been-backported-to-versions-3-12/94530).

Given uv is often used to bootstrap newer Python versions on older base containers, Metaflow should probably be somewhat tolerant to this and at the very least fall back for older Python versions to the old behaviour.

Mr-Neutr0n:
Fixes #2759

The uv bootstrap uses `tar.extractall(..., filter=_tar_filter)` which was only added in Python 3.12. If you're on an older Python version this blows up with a `TypeError` since `extractall` doesn't recognize the `filter` kwarg.

Since uv is commonly used to bootstrap newer Python versions on older base containers, this is a pretty common scenario.

The fix checks `sys.version_info` and falls back to manual member iteration + `tar.extract()` on Python < 3.12. Same behavior either way — only the `uv` binary gets extracted with a flattened path.

### DIFF
--- a/metaflow/plugins/cards/ui/package-lock.json
+++ b/metaflow/plugins/cards/ui/package-lock.json
@@ -4914,9 +4914,9 @@
       }
     },
     "node_modules/qs": {
-      "version": "6.14.1",
-      "resolved": "https://registry.npmjs.org/qs/-/qs-6.14.1.tgz",
-      "integrity": "sha512-4EK3+xJl8Ts67nLYNwqw/dsFVnCf+qR7RgXSK9jEEm9unao3njwMDdmsdvoKBKHzxd7tCYz5e5M+SnMjdtXGQQ==",
+      "version": "6.14.2",
+      "resolved": "https://registry.npmjs.org/qs/-/qs-6.14.2.tgz",
+      "integrity": "sha512-V/yCWTTF7VJ9hIh18Ugr2zhJMP01MY7c5kh4J870L7imm6/DIzBsNLTXzMwUA3yZ5b/KBqLx8Kp3uRvd7xSe3Q==",
       "dev": true,
       "license": "BSD-3-Clause",
       "dependencies": {


savingoyal (approved):
waiting on @saikonen 's approval before merging

metaflow/plugins/uv/bootstrap.py
```diff
@@ -50,19 +50,31 @@ def install_uv():
...
         }
 
-        def _tar_filter(member: tarfile.TarInfo, path):
-            if os.path.basename(member.name) != "uv":
-                return None  # skip
-            member.path = os.path.basename(member.path)
-            return member
+        def _extract_uv_binary(tar, dest):
+            """Extract just the uv binary from the tarball, flattening the path."""
+            if sys.version_info >= (3, 12):
```
saikonen, line 55: Just a note that `filter=` is also available in `3.11.4`, `3.10.12`, `3.9.17` and `3.8.17`, but the point is valid that this will break on older base images that do not support the parameter.

metaflow/plugins/uv/bootstrap.py
```diff
@@ -50,19 +50,31 @@ def install_uv():
...
+                    member.path = os.path.basename(member.path)
+                    return member
+
+                tar.extractall(dest, filter=_tar_filter)
+            else:
+                for member in tar.getmembers():
+                    if os.path.basename(member.name) == "uv":
+                        member.name = os.path.basename(member.name)
+                        tar.extract(member, dest)
+                        break
```
saikonen, line 69: Is there any direct benefit with using the filter when available? Could we instead have this `tar.extract` implementation for all Python versions for simplicity?

metaflow/plugins/uv/bootstrap.py
```diff
@@ -50,19 +50,31 @@ def install_uv():
...
+
+                def _tar_filter(member, path):
+                    if os.path.basename(member.name) != "uv":
+                        return None
+                    member.path = os.path.basename(member.path)
+                    return member
+
+                tar.extractall(dest, filter=_tar_filter)
+            else:
+                for member in tar.getmembers():
```
saikonen, line 65: there is a subtle issue here that makes the whole thing not work. Using `getmembers()` leads to an error:
```
io.UnsupportedOperation: seek
```

You can replace this with `for member in tar:` to get things to work.

npow:
@greptile

npow:
Duplicate of #2949 which addresses the same Python <3.12 tarfile compatibility issue. Closing in favor of that PR.

PR #2800, opened 2026-02-19
fix: prompt for sudo before backgrounding minikube tunnel

### LINKED ISSUE #2605: `metaflow-dev`: `minikube tunnel` sudo access can't accept password
## What

`metaflow-dev up` executes `minikube tunnel` [as a background process](https://github.com/Netflix/metaflow/blame/a6edb29b0224db324ee1b5df15f1838cd1bde3d0/devtools/Makefile#L192) (`&`), so when `minikube tunnel` requests a password (`[sudo] password for user:`) it's not possible to type it in.

## Workaround

Run the (undocumented) `metaflow-dev setup-minikube`, then (undocumented) `metaflow-dev tunnel`, type in the password, before finally running `metaflow-dev up` in a separate terminal.

## Notes

The [minikube handbook](https://minikube.sigs.k8s.io/docs/handbook/accessing/) calls out this sudo access as an issue, and suggests allowing minikube passwordless sudo access, but I don't think this can work for `metaflow-dev` as a general solution, since it downloads minikube to an unpredictable location.

I tried `minikube tunnel --rootless`, but it still asks for sudo password, so "rootless" must be referring to something else.

I do note that my "Hello World"-style example flow can run even if I comment out the `tunnel`. I'd be curious to know what the use-case is for the tunnel - the minikube docs are a bit unclear to me, and the metaflow git history and original pull request doesn't explain.

vyagh:
`minikube tunnel` needs elevated privileges to configure host networking, but `start.sh` runs it as a background process (`&`), so the `[sudo] password` prompt goes unanswered.

Added `sudo -v` before the `tunnel &` line, this collects the password upfront so the background process can authenticate without a TTY.

Closes #2605

### DIFF
--- a/devtools/Makefile
+++ b/devtools/Makefile
@@ -182,6 +182,8 @@ up: install-brew check-docker install-curl install-gum setup-minikube install-he
 	@echo '    echo "📝 Selecting services..."' >> "$(DEVTOOLS_DIR)/start.sh"
 	@echo '    SERVICES=$$($(PICK_SERVICES))' >> "$(DEVTOOLS_DIR)/start.sh"
 	@echo 'fi' >> "$(DEVTOOLS_DIR)/start.sh"
+	@echo 'echo "🔐 minikube tunnel requires sudo access to configure networking."' >> $(DEVTOOLS_DIR)/start.sh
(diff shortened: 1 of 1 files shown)


npow:
@greptile

PR #2802, opened 2026-02-19
fix: bootstrap pip if missing before installing remote deps

### LINKED ISSUE #2718: Initial dependencies install can fail if `pip` is missing
`pip` is an implicit requirement for the remote environment, as it is used to fetch required libraries such as `requests` and `boto3`. Missing this means environment hydration fails from the start.

A install-if-missing check should be added to ensure that we have the tool for installing dependencies.

This would also be an opportunity to look into if using `uv` instead of `pip` is possible for the installing of dependencies in the remote environment, specifically `_get_install_dependencies_cmd` and alike.

vyagh:
Adds an `ensurepip` fallback when [pip](cci:1://file:///home/vyagh/Work/oss/GSOC/metaflow/test/unit/test_metaflow_environment.py:41:4-45:75) is missing or unavailable in remote task environments.

Fixes #2718 

Minimal Python images (e.g. `python:3.12-slim`) often don't include [pip](cci:1://file:///home/vyagh/Work/oss/GSOC/metaflow/test/unit/test_metaflow_environment.py:41:4-45:75) by default. This causes environment hydration to fail silently or with confusing errors on Kubernetes/Batch because the bootstrap process relies on `python -m pip` to install required libraries like `boto3` and `requests`.

Since `ensurepip` is part of the python standard lib, we can use it to bootstrap [pip](cci:1://file:///home/vyagh/Work/oss/GSOC/metaflow/test/unit/test_metaflow_environment.py:41:4-45:75) when needed without requiring changes to the user's base image.

- Modified [_get_install_dependencies_cmd](cci:1://file:///home/vyagh/Work/oss/GSOC/metaflow/metaflow/metaflow_environment.py:158:4-189:87) in [metaflow_environment.py](cci:7://file:///home/vyagh/Work/oss/GSOC/metaflow/metaflow/metaflow_environment.py:0:0-0:0) to prepend a shell check for [pip](cci:1://file:///home/vyagh/Work/oss/GSOC/metaflow/test/unit/test_metaflow_environment.py:41:4-45:75).
- Added a fallback to `python -m ensurepip` if [pip](cci:1://file:///home/vyagh/Work/oss/GSOC/metaflow/test/unit/test_metaflow_environment.py:41:4-45:75) is absent.
- Added a clear error message if both [pip](cci:1://file:///home/vyagh/Work/oss/GSOC/metaflow/test/unit/test_metaflow_environment.py:41:4-45:75) and `ensurepip` are unavailable (which can happen on some minimal Debian/Ubuntu distros), pointing the user to fix their base image.
- Added unit tests in [test/unit/test_metaflow_environment.py](cci:7://file:///home/vyagh/Work/oss/GSOC/metaflow/test/unit/test_metaflow_environment.py:0:0-0:0) covering the new shell logic and datastore package lists.

### Testing
Verified with unit tests

### DIFF
--- a/metaflow/metaflow_environment.py
+++ b/metaflow/metaflow_environment.py
@@ -157,8 +157,17 @@ class MetaflowEnvironment(object):
             )
 
     def _get_install_dependencies_cmd(self, datastore_type):
+        python = self._python()
+        # bootstrap pip if missing; try ensurepip (may not be available on all distros)
+        ensure_pip_cmd = (
+            "{python} -m pip --version > /dev/null 2>&1"
+            " || {{ {python} -m ensurepip --upgrade --default-pip > /dev/null 2>&1; }}"
+            " || {{ echo 'ERROR: pip is not available and ensurepip failed."
+            " Please use a base image that includes pip.' >&2 && exit 1; }}"
+        ).format(python=python)
+
         base_cmd = "{} -m pip install -qqq --no-compile --no-cache-dir --disable-pip-version-check".format(
-            self._python()
+            python
         )
 
         datastore_packages = {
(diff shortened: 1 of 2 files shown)


npow:
@greptile

PR #2805, opened 2026-02-19
Fix temp file leak in parallel_imap_unordered on child failure

### LINKED ISSUE #2804: Bug: `parallel_imap_unordered` leaks temp files when child processes fail
## Summary

`parallel_imap_unordered` in `metaflow/multicore_utils.py` leaks temporary files when a child process exits with a non-zero exit code.

## Context / Motivation

When a child process fails, `MulticoreException` is raised (line 129) *before* `os.remove(output_file)` (line 133) ever runs. This leaves the failed child's temp file on disk. Additionally, temp files for all other in-flight children are also never cleaned up since the exception aborts the loop without any `finally` cleanup.

Over time — especially in workloads that use `parallel_imap_unordered` heavily (like S3 operations via `s3op.py`) — orphaned temp files prefixed with `parallel_map_` accumulate in the system's temp directory.

## Steps to Reproduce

```python
from metaflow.multicore_utils import parallel_imap_unordered

def failing_func(x):
    if x == 3:
        raise ValueError("intentional failure")
    return x * 2

# This raises MulticoreException but leaks temp files
try:
    list(parallel_imap_unordered(failing_func, range(5), max_parallel=2))
except:
    pass

# Check /tmp for leftover parallel_map_* files
import glob
leaked = glob.glob("/tmp/parallel_map_*")
print(f"Leaked files: {leaked}")
```

## Expected Behavior

All temp files created by `_spawn` should be cleaned up regardless of whether child processes succeed or fail.

## Actual Behavior

- The failed child's temp file is never deleted (it was popped from `pids` but `os.remove` is never reached).
- Any remaining in-flight children's temp files are also leaked since the exception exits the loop entirely.

## Environment

- OS: macOS / Linux (any Unix where `os.fork()` is used)
- Python: 3.7+
- Metaflow: master branch

## Proposed Fix

1. Remove the failed child's temp file before raising `MulticoreException`.
2. Wrap the main loop in a `try/finally` block to ensure temp files for any remaining children are always cleaned up.
3. Add tests covering the failure path (the existing test only covers the happy path).

I'd like to submit a PR for this fix. Happy to iterate on the approach.

ishaan-arora-1:
## Summary

Fix a resource leak in `parallel_imap_unordered` where temporary files are never cleaned up when a child process fails.

## Context / Motivation

When a child process in `parallel_imap_unordered` exits with a non-zero exit code, `MulticoreException` is raised (line 129) *before* `os.remove(output_file)` (line 133) ever executes. This causes two problems:

1. **The failed child's temp file is leaked** — it was already popped from the `pids` list, but `os.remove` is never reached because the exception is raised first.
2. **All remaining in-flight children's temp files are also leaked** — the exception exits the `while pids` loop entirely with no `finally` cleanup, so any other children still tracked in `pids` have their temp files orphaned on disk.

Over time — especially in workloads that use `parallel_imap_unordered` heavily (like S3 operations via `s3op.py`) — orphaned temp files prefixed with `parallel_map_` accumulate in the system's temp directory.

Fixes #2804

## Changes Made

- **`metaflow/multicore_utils.py`**: Wrapped the main `while pids` loop in a `try/finally` block. The `finally` clause iterates over any remaining entries in `pids` and removes their temp files. Additionally, the failed child's temp file (already popped from `pids`) is explicitly removed before raising `MulticoreException`. All removals use `try/except OSError: pass` since the file may have already been cleaned up.
- **`test/unit/test_multicore_utils.py`**: Added three new tests:
  - `test_parallel_map_cleans_temp_files_on_success` — verifies no temp files remain after a successful `parallel_map` run.
  - `test_parallel_imap_unordered_cleans_temp_files_on_child_failure` — verifies no temp files remain after a child failure in `parallel_imap_unordered`.
  - `test_parallel_map_cleans_temp_files_on_child_failure` — verifies no temp files remain after a child failure in `parallel_map`.
- No breaking changes, no new dependencies, fully backward-compatible.

## Testing

- Added `test_parallel_imap_unordered_cleans_temp_files_on_child_failure()` — uses a function that raises on a specific input. Confirms that after `MulticoreException` is raised, zero `parallel_map_*` files remain in the temp directory. **This test fails without the fix and passes with it.**
- Added `test_parallel_map_cleans_temp_files_on_child_failure()` — same verification through the `parallel_map` wrapper.
- Added `test_parallel_map_cleans_temp_files_on_success()` — verifies correct cleanup on the happy path as well.
- All 4 tests (including the original `test_parallel_map`) pass.
- Ran the full `test/unit/` suite — all pre-existing test results unchanged.
- Code formatted with Black via `pre-commit run` — passes cleanly.

- **`try/except OSError: pass` for removals**: The temp file may have already been removed (e.g., by the child process itself or OS cleanup), so we silently ignore `OSError` on removal. This is intentional and follows the existing pattern in the codebase.
- **`finally` block on a generator**: Since `parallel_imap_unordered` is a generator, the `finally` block executes when the generator is either fully consumed, garbage-collected, or when an exception propagates out. This covers all exit paths correctly.
- **No changes to child process behavior**: The fix is entirely in the parent process's cleanup logic. The `_spawn` / child process code is untouched.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,93 @@
+## PR Type
+
+<!-- Check one -->
+
+- [ ] Bug fix
+- [ ] New feature
+- [ ] Core Runtime change (higher bar -- see [CONTRIBUTING.md](../CONTRIBUTING.md#core-runtime-contributions-higher-bar))
+- [ ] Docs / tooling
+- [ ] Refactoring
+
+## Summary
+
+<!-- What user-visible behavior changes? 1-2 sentences. -->
+
+## Issue
+
+<!-- Link to issue. Required for bug fixes, required for Core Runtime. -->
+
+Fixes #
+
+## Reproduction
+
+<!-- Required for bug fixes. Required for Core Runtime changes. -->
+<!-- Provide a minimal reproduction that fails before and succeeds after. -->
+
+**Runtime:** <!-- local / kubernetes / batch / argo / etc. -->
+
+**Commands to run:**
+```bash
+# paste exact commands
+```
+
+**Where evidence shows up:** <!-- parent console / task logs / metadata / UI -->
+
+<details>
+<summary>Before (error / log snippet)</summary>
+
+```
+paste here
+```
+
+</details>
+
+<details>
+<summary>After (evidence that fix works)</summary>
+
+```
+paste here
+```
+
+</details>
+
+## Root Cause
+
+<!-- Required for Core Runtime. Recommended for all bug fixes. -->
+<!-- Explain the causal chain: what invariant was violated, where in the code. -->
+<!-- See PRs #2796, #2751, #2714 for examples of the level of detail we're looking for. -->
+
+## Why This Fix Is Correct
+
+<!-- What invariant is restored? Why is the fix minimal? -->
+
+## Failure Modes Considered
+
+<!-- Required for Core Runtime (at least 2). Recommended for all bug fixes. -->
+<!-- Examples: concurrency/retries, subprocess output propagation, env-var leakage, backward compat -->
+
+1.
+2.
+
+## Tests
+
+- [ ] Unit tests added/updated
+- [ ] Reproduction script provided (required for Core Runtime)
+- [ ] CI passes
+- [ ] If tests are impractical: explain why below and provide manual evidence above
+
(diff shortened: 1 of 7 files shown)


npow:
@greptile

PR #2807, opened 2026-02-19
Surface AWS credential info in MetaflowS3AccessDenied when METAFLOW_DEBUG_S3CLIENT=1

### LINKED ISSUE #545: Helper functions / verbose debugging to triage IAM (S3) issues

blankirigaya:
## Summary

Fixes #545

When you get an S3 access denied error in Metaflow, the error message doesn't tell you which AWS credentials were being used. This makes it really hard to debug IAM permission issues. This PR fixes that by showing credential details when you enable debug mode.

---

## What changed

**`metaflow/plugins/aws/aws_utils.py`**

Added a `get_credential_debug_info()` function that figures out:
- Where your credentials are coming from (environment variables, `~/.aws/credentials`, EC2 role, etc.)
- Your access key ID (only showing first 4 and last 4 characters for security)
- Which AWS region you're using
- Your AWS identity and account ID (by calling STS)

**`metaflow/plugins/datatools/s3/s3op.py`**

Updated the S3 access denied error to show debug info when `METAFLOW_DEBUG_S3CLIENT=1` is set.

---

## What it looks like

**Before:**
```
MetaflowS3AccessDenied: s3://my-bucket/data/artifact
```

**After (normal mode):**
```
MetaflowS3AccessDenied: s3://my-bucket/data/artifact
Tip: Set METAFLOW_DEBUG_S3CLIENT=1 to see which AWS credentials are being used.
```

**After (with `METAFLOW_DEBUG_S3CLIENT=1`):**
```
MetaflowS3AccessDenied: s3://my-bucket/data/artifact

Credential debug info (METAFLOW_DEBUG_S3CLIENT=1):
  Credential source : Environment variables (AWS_ACCESS_KEY_ID / AWS_SECRET_ACCESS_KEY)
  Access key ID     : AKIA...MPLE
  AWS region        : ap-south-1
  Caller ARN        : arn:aws:iam::123456789012:user/aj
  AWS account ID    : 123456789012

Tip: Check that this identity has the right S3 permissions
     (s3:GetObject, s3:PutObject, s3:ListBucket) on your bucket.
```

---

## How it works

- Uses the existing `METAFLOW_DEBUG_S3CLIENT` flag (no new config needed)
- Only runs when an error happens - zero overhead during normal operation
- Handles all the common ways people set up AWS credentials
- Never shows your actual secret key - just the access key ID with most characters hidden
- If the STS call fails, it handles it gracefully and just says so
- Writes to stderr so it works with Metaflow's subprocess setup

---

## Testing

Tested with different credential setups and confirmed the error messages show up correctly.

Show AWS credentials in S3 access denied errors

**Note:** AI tools were used to help write parts of this code and documentation.

### DIFF
--- a/metaflow/plugins/aws/aws_utils.py
+++ b/metaflow/plugins/aws/aws_utils.py
@@ -7,87 +7,137 @@ def get_credential_debug_info(s3config=None):
     """
     Return a human-readable string describing the active AWS credentials.
 
-    This function reflects the *effective* identity used by boto3,
-    including assumed-role contexts (via STS get_caller_identity).
+    When s3config.role is set, mirrors the AssumeRole credential chain used by
+    aws_client.py so the reported identity matches what actually made the S3 call.
+    When no role is configured, reflects ambient boto3 credential resolution.
 
     Never raises — failures are reported inline.
     """
     lines = []
 
     try:
+        # Lazy imports: boto3/botocore are optional dependencies and may not be
+        # present in all Metaflow deployments. Importing here avoids import-time
+        # errors in environments that do not use AWS at all.
         import boto3
         import botocore.config
+        import botocore.credentials
+        import botocore.session as botocore_session_module
 
-        session = boto3.session.Session()
+        # Extract the configured IAM role ARN, if any. getattr guards against
+        # s3config namedtuple fields being absent in older versions.
+        role = s3config and getattr(s3config, "role", None)
 
-        # Note: this reflects ambient credential resolution only.
-        # If s3config.role is set, the actual S3 operations assume a different
-        # IAM role — the STS identity below will show the base credentials,
-        # not the assumed role that triggered the 403.
-        if s3config and getattr(s3config, "role", None):
-            lines.append(
-                "  Note: S3 operations use assumed role: %s\n"
-                "  The credentials below reflect the base session, not the assumed role."
-                % s3config.role
+        # Always start from a plain ambient session. When a role is configured
+        # this becomes the *source* session for AssumeRole; otherwise it is used
+        # directly for credential inspection and the STS call.
+        base_session = boto3.session.Session()
+
+        if role:
+            # Replicate the exact credential chain from aws_client.py so the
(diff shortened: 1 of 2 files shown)


blankirigaya:
Hi @npow, thanks for the detailed review! I've addressed all three points:
1. PR title — Updated to clearly describe what the change does without requiring knowledge of the issue number.
2. Testing & subprocess propagation — You were right that the original implementation wouldn't surface through subprocesses. I've rearchitected it completely:


- get_credential_debug_info() is now called inside the exit() function in s3op.py — directly within the subprocess where the live boto3 session exists

- The message is written to stderr via the existing print("s3op failed:\n%s" % msg, file=sys.stderr) path, which the parent process already captures and surfaces in the exception

- Also fixed a boto3 API bug: replaced .resolve() (not available on all credential types) with reading method first, then calling .get_frozen_credentials() which works across all credential providers

3. Existing debug flag — Removed the new S3_VERBOSE_ERRORS / METAFLOW_S3_VERBOSE_ERRORS flag entirely. The feature now reuses the existing METAFLOW_DEBUG_S3CLIENT=1 flag via debug.s3client, consistent with how other S3 debug behaviour is gated.
Test output with METAFLOW_DEBUG_S3CLIENT=1:
Credential debug info (METAFLOW_DEBUG_S3CLIENT=1):
  Credential source : Environment variables (AWS_ACCESS_KEY_ID / AWS_SECRET_ACCESS_KEY)
  Access key ID     : AKIA...MPLE
  AWS region        : us-east-1
  Caller ARN        : (STS lookup failed: InvalidClientTokenId — expected with fake creds, graceful degradation working correctly)
Screenshot 2026-02-20 150440


## above screenshot / test done after using invaild aws id to simulate access denied error

npow:
@greptile

blankirigaya:
@npow let me know if any other changes are required i have made suggested change and solved issues by Greptile without altering core of the codebase as well without creating contrary changes to the issue and feature being add by the PR 

## thank you and would appreciate your feedback

blankirigaya:
@npow sir please review this

PR #2808, opened 2026-02-19
fix: validate pathspec format and handle trailing slashes

### LINKED ISSUE #948: Pathspec to create Flow/Run/Step/Task/DataArtifact is not validated

danielalanbates:
## Summary
Fix pathspec validation to handle trailing slashes and reject empty path components, preventing silently malformed `Flow`/`Run`/`Step`/`Task`/`DataArtifact` objects.

## Context / Motivation
Fixes #948. Pathspecs with trailing slashes (e.g., `Run('Flow/1/')`) bypass validation because `str.split('/')` creates an empty string component that shifts the count. This causes `Run('Flow/1/')` to have 3 components, silently passing the Step check instead of being rejected. The resulting object behaves unpredictably.

## Changes Made
- Strip trailing slashes from pathspec before splitting (`pathspec.rstrip('/')`)
- Reject pathspecs with empty components (e.g., `Flow//RunID`, `/Flow`) with a clear error message
- The existing component-count validation then works correctly for all cases

## Testing
- Added `test/unit/test_pathspec_validation.py` with 16 tests covering:
  - Too few / too many components for each object type (Flow, Run, Step, Task, DataArtifact)
  - Trailing slashes that previously bypassed validation
  - Empty segments (`Flow//RunID`)
  - Leading slashes (`/Flow`)
- All 16 tests pass: `python -m pytest test/unit/test_pathspec_validation.py -v`

## Before/After

| Pathspec | Before | After |
|---|---|---|
| `Run('Flow/1/')` | Silently accepted (3 components → passed Step check) | Correctly rejected as invalid Run |
| `Step('Flow/1/start/')` | Silently accepted (4 components → passed Task check) | Correctly rejected as invalid Step |
| `Run('Flow//1')` | Undefined behavior | `MetaflowInvalidPathspec: Pathspec contains empty components` |

---

*This PR was created with the assistance of Claude Opus 4.6 by Anthropic. Happy to make any adjustments! Reviewed and submitted by a human.*

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,93 @@
+## PR Type
+
+<!-- Check one -->
+
+- [ ] Bug fix
+- [ ] New feature
+- [ ] Core Runtime change (higher bar -- see [CONTRIBUTING.md](../CONTRIBUTING.md#core-runtime-contributions-higher-bar))
+- [ ] Docs / tooling
+- [ ] Refactoring
+
+## Summary
+
+<!-- What user-visible behavior changes? 1-2 sentences. -->
+
+## Issue
+
+<!-- Link to issue. Required for bug fixes, required for Core Runtime. -->
+
+Fixes #
+
+## Reproduction
+
+<!-- Required for bug fixes. Required for Core Runtime changes. -->
+<!-- Provide a minimal reproduction that fails before and succeeds after. -->
+
+**Runtime:** <!-- local / kubernetes / batch / argo / etc. -->
+
+**Commands to run:**
+```bash
+# paste exact commands
+```
+
+**Where evidence shows up:** <!-- parent console / task logs / metadata / UI -->
+
+<details>
+<summary>Before (error / log snippet)</summary>
+
+```
+paste here
+```
+
+</details>
+
+<details>
+<summary>After (evidence that fix works)</summary>
+
+```
+paste here
+```
+
+</details>
+
+## Root Cause
+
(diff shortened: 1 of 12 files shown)


npow:
@greptile

danielalanbates:
Thanks for the feedback, @npow! I've pushed an update addressing your review comments. Please take another look when you get a chance.

danielalanbates:
Thanks for the feedback, @npow! I've pushed an update addressing your review comments. Please take another look when you get a chance.

talsperre:
@danielalanbates Hi - the fix looks good to me. Can you simplify the tests so that it's easier to review? Also, we prefer to use fixtures for Pytest and have all the tests in the module directly instead of as part of a class. See this test for [example](https://github.com/Netflix/metaflow/blob/master/test/unit/test_config_value.py).

danielalanbates:
Thanks for the feedback, @npow! I've pushed an update addressing your review comments. Please take another look when you get a chance.

PR #2815, opened 2026-02-20
fix: prevent temp file and child process leaks in parallel_imap_unordered

shripad-jyothinath:
### Summary
`parallel_imap_unordered` in `metaflow/multicore_utils.py` leaks temporary files and child processes when a child process exits with a non-zero exit code or the generator exits early.

### Context / Motivation
When a child process fails, a `MulticoreException` is raised before `os.remove(output_file)` ever runs. This leaves the failed child's temp file on disk. Additionally, temp files for all other in-flight children are also never cleaned up since the exception aborts the loop without any `finally` cleanup. Furthermore, the in-flight children are left running as zombie processes. 

Over time — especially in workloads that use `parallel_imap_unordered` heavily (like S3 operations) — orphaned temp files prefixed with `parallel_map_` accumulate in the system's temp directory.

### Steps to Reproduce
```python
from metaflow.multicore_utils import parallel_imap_unordered

def failing_func(x):
    if x == 3:
        raise ValueError("intentional failure")
    return x * 2

# This raises MulticoreException but leaks temp files
try:
    list(parallel_imap_unordered(failing_func, range(5), max_parallel=2))
except:
    pass

# Check /tmp for leftover parallel_map_* files
import glob
leaked = glob.glob("/tmp/parallel_map_*")
print(f"Leaked files: {leaked}")

### DIFF
--- a/metaflow/multicore_utils.py
+++ b/metaflow/multicore_utils.py
@@ -3,6 +3,7 @@ import os
 import traceback
 from itertools import islice
 from tempfile import NamedTemporaryFile
+import signal
 import time
 import metaflow.tracing as tracing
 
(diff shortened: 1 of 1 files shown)


npow:
@greptile

PR #2821, opened 2026-02-20
fix: close leaked file descriptors in CommandManager.async_run

Abhiii47:
## Summary
Fixes a file descriptor leak in `CommandManager.async_run()` where stdout/stderr log file handles were never closed.

## Problem
In `subprocess_manager.py`, the `async_run()` method opens file handles inline and passes them directly to `asyncio.create_subprocess_exec()`:

```python
# Before (leaking file descriptors)
self.process = await asyncio.create_subprocess_exec(
    *self.command,
    stdout=open(stdout_logfile, "w", encoding="utf-8"),  # never closed
    stderr=open(stderr_logfile, "w", encoding="utf-8"),  # never closed
)
```

Since no references to these file objects are stored, they are **never explicitly closed**. Each call to `async_run()` leaks two file descriptors. In long-running services or workflows that spawn many subprocesses, this can eventually hit the OS file descriptor limit and cause `OSError: [Errno 24] Too many open files`.

> **Note:** The synchronous `run()` method does not have this issue — it uses `subprocess.PIPE` and closes pipes explicitly in the streaming threads.

## Fix
- Open file handles **before** the try block and store references
- Save them in `self._log_file_handles` on success
- Close them immediately on exception in `async_run()`
- Close all stored handles in `cleanup()` before removing the temp directory
- Initialize `_log_file_handles = []` in `__init__` for safety

```python
# After (file descriptors properly managed)
stdout_fh = open(stdout_logfile, "w", encoding="utf-8")
stderr_fh = open(stderr_logfile, "w", encoding="utf-8")
try:
    self.process = await asyncio.create_subprocess_exec(
        *self.command,
        stdout=stdout_fh,
        stderr=stderr_fh,
    )
    self._log_file_handles = [stdout_fh, stderr_fh]
    ...
except Exception as e:
    stdout_fh.close()
    stderr_fh.close()
    ...
```

## Changes
| File | Change |
|------|--------|
| `metaflow/runner/subprocess_manager.py` | Store and close file handles properly |

## Testing
- Code review verified the fix is correct and minimal
- No behavioral change — only resource management improvement
- The `cleanup()` method is already called by `SubprocessManager.__aexit__` and `CommandManager.__aexit__`, so handles are closed on context manager exit


npow:
@greptile

metaflow/runner/subprocess_manager.py
```diff
@@ -402,23 +403,29 @@ async def async_run(self):
             stdout_logfile = os.path.join(self.temp_dir, "stdout.log")
             stderr_logfile = os.path.join(self.temp_dir, "stderr.log")
 
+            stdout_fh = open(stdout_logfile, "w", encoding="utf-8")
```
npow, line 406: how come you cant use the standard context manager pattern here?

Abhiii47: Good question! The issue is that async_run() only starts the subprocess and returns immediately the process keeps running and writing to stdout/stderr long after this method returns. If we used with open(...) as fh, the handle would be closed at the end of the block while the subprocess is still actively writing to it, which would break things.
So instead, we store references to the handles and close them later in cleanup(), which gets called by aexit when the 
CommandManager context manager exits. That way the handles stay open for the full lifetime of the subprocess

PR #2823, opened 2026-02-20
Fix/with dir exception safety

Abhiii47:
## Summary

The `with_dir` context manager in `metaflow/runner/utils.py` doesn't restore the working directory when an exception occurs inside the `with` block. This wraps the `yield` in `try/finally` so `os.chdir(current_dir)` always executes.

## Problem

```python
@contextmanager
def with_dir(new_dir):
    current_dir = os.getcwd()
    os.chdir(new_dir)
    yield new_dir
    os.chdir(current_dir)  # skipped if exception raised
```

If any exception propagates out of a `with with_dir(...)` block, the `os.chdir(current_dir)` line is never reached, leaving the process in a corrupted working directory for all subsequent operations. This affects 6+ call sites across `metaflow_runner.py` and `deployer_impl.py`.

## Fix

```python
@contextmanager
def with_dir(new_dir):
    current_dir = os.getcwd()
    os.chdir(new_dir)
    try:
        yield new_dir
    finally:
        os.chdir(current_dir)
```

This is the canonical Python pattern for resource cleanup in context managers (see [contextlib docs](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager)).

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,93 @@
+## PR Type
+
+<!-- Check one -->
+
+- [ ] Bug fix
+- [ ] New feature
+- [ ] Core Runtime change (higher bar -- see [CONTRIBUTING.md](../CONTRIBUTING.md#core-runtime-contributions-higher-bar))
+- [ ] Docs / tooling
+- [ ] Refactoring
+
+## Summary
+
+<!-- What user-visible behavior changes? 1-2 sentences. -->
+
+## Issue
+
+<!-- Link to issue. Required for bug fixes, required for Core Runtime. -->
+
+Fixes #
+
+## Reproduction
+
+<!-- Required for bug fixes. Required for Core Runtime changes. -->
+<!-- Provide a minimal reproduction that fails before and succeeds after. -->
+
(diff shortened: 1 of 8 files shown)


npow:
@greptile

talsperre:
```
metaflow/runner/utils.py
Duplicate with_dir not fixed
There is an identical with_dir context manager in metaflow/packaging_sys/utils.py:53 that still has the same bug (no try/finally around yield). Consider applying the same fix there for consistency, or at least tracking it as a follow-up.
```

@Abhiii47 can you update the PR with this fix that greptile suggested? I will run the tests on it later.

Abhiii47:
> ```
> metaflow/runner/utils.py
> Duplicate with_dir not fixed
> There is an identical with_dir context manager in metaflow/packaging_sys/utils.py:53 that still has the same bug (no try/finally around yield). Consider applying the same fix there for consistency, or at least tracking it as a follow-up.
> ```
> 
> @Abhiii47 can you update the PR with this fix that greptile suggested? I will run the tests on it later.

can u check it talsperre i have updated it

Abhiii47:
any update?

PR #2825, opened 2026-02-20
Add CODEOWNERS for automatic PR reviewer assignment

npow:
Maps repository components to their primary maintainers based on git history analysis, enabling GitHub to auto-assign reviewers on PRs.

### DIFF
--- a/.github/workflows/codeowners.yml
+++ b/.github/workflows/codeowners.yml
@@ -12,7 +12,7 @@ jobs:
   check-codeowners:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


npow:
@greptile

savingoyal (commented):
Thanks! How is this file generated and kept fresh? Also given the small number of reviewers, do we need it at this point? I am okay either way as long as this file auto-updates.

npow:
Switched to distributed (colocated) `CODEOWNERS` files. Those will need to be updated when:
1. a new module is added
2. someone contributes significant code to an existing module

PR #2827, opened 2026-02-20
fix: cap heartbeat retry backoff and add interruptible stop

suhr25:
## Summary

Fixes an issue in `MetadataHeartBeat._ping()` where repeated heartbeat failures could cause exponential backoff to grow without bound, effectively pausing heartbeats for hours (or longer) even after the service recovered.

---

## Context / Motivation

The heartbeat loop runs in a daemon thread inside the sidecar and is responsible for signaling task liveness.
Previously, consecutive failures kept increasing `retry_counter` with no upper limit:

```py
retry_counter += 1
time.sleep(1.5 ** retry_counter)
```

Over time, this led to extremely long sleep intervals. Once the metadata service recovered, the heartbeat wouldn’t resume until the sleep finished — making active tasks appear dead or stuck.
Additionally, `_shutdown()` had no way to interrupt this sleep, so the thread could linger unnecessarily during teardown.

---

## Changes Made

* **Capped exponential backoff** to prevent unbounded growth in retry delays
* Added a **stop `Event`** to control the lifecycle of the ping loop
* Replaced `time.sleep(...)` with **interruptible waits** using `Event.wait(...)`
* Ensured `_shutdown()` cleanly stops the loop and sends one final heartbeat

Key changes (simplified):

```py
MAX_HEARTBEAT_RETRY_COUNT = 10

self._stop = Event()
```

```py
while not self._stop.is_set():
    ...
    self._stop.wait(timeout=frequency_secs)
```

```py
retry_counter = min(retry_counter + 1, MAX_HEARTBEAT_RETRY_COUNT)
self._stop.wait(timeout=1.5 ** retry_counter)
```

```py
def _shutdown(self):
    self._stop.set()
    self._heartbeat()
```

---

## Testing

* Added a **unit test** that simulates consecutive heartbeat failures and verifies:

  * retry backoff is capped
  * wait duration does not grow unbounded
  * the loop exits promptly when shutdown is triggered
* Ran all unit tests locally:

  ```bash
  cd test/unit
  python -m pytest -v
  ```

---

* Kept the fix **minimal and targeted** to avoid changing surrounding architecture
* Chose a cap (`~57s max delay`) that still allows progressive backoff without breaking heartbeat expectations
* Did not introduce broader refactoring (e.g., session pooling) to keep this PR focused on correctness and reliability

---

### DIFF
--- a/metaflow/metadata_provider/heartbeat.py
+++ b/metaflow/metadata_provider/heartbeat.py
@@ -1,6 +1,6 @@
 import json
 import time
-from threading import Thread
+from threading import Event, Thread
 
 import requests
 
@@ -18,9 +18,13 @@ class HeartBeatException(MetaflowException):
         super(HeartBeatException, self).__init__(msg)
 
 
+MAX_HEARTBEAT_RETRY_COUNT = 10  # caps backoff at 1.5**10 ≈ 57 seconds
+
+
 class MetadataHeartBeat(object):
     def __init__(self):
         self.headers = SERVICE_HEADERS
+        self._stop = Event()
         self.req_thread = Thread(target=self._ping)
         self.req_thread.daemon = True
         self.default_frequency_secs = 10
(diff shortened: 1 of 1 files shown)


suhr25:
Hi @npow,
This change bounds the exponential retry backoff and introduces an interruptible wait mechanism, preventing prolonged heartbeat gaps after sustained failures. It also ensures proper shutdown by signaling the loop to terminate and issuing a final heartbeat.

npow:
@greptile

PR #2828, opened 2026-02-20
Prevent assignment to flow-level attributes inside steps with explici…

Amitavoo:
Problem: :

-Flow-level attributes and Parameters can currently be reassigned inside steps.
This leads to:

-Confusing artifact behavior

-Inconsistent runtime state

-Potential internal errors during datastore cleanup

Solution: :

-Convert flow-level attributes into read-only properties during step execution

-Raise a clear AttributeError if reassignment is attempted

-Add a defensive guard in datastore cleanup to prevent crashes when attributes were never created

Result: :

-Clear, explicit user-facing error

-No internal framework crash

-Stable and predictable runtime behavior


npow:
@greptile

PR #2829, opened 2026-02-20
feature: reduce reserved Parameter names from 14 to 5

### LINKED ISSUE #1753: Reduce the number of reserved parameter names
Related to the bug fix in #1736 it would be worth considering to reduce the number of reserved parameter names.

One possible approach is to prefix the internal arguments with `_` to avoid the collisions with user defined parameters. The main concern with this is that it can lead to some odd UX specifically with the case of tags. If `tags` stops being a reserved parameter, a  user can define one and then run a flow with

```
python CustomTagFlow.py run --tag this-is-an-internal-metaflow-tag --tags user-defined-parameter-value
```

which at first glance can be extremely confusing. Other edge cases might also exist.

vagxrth:
## PR Type

  - [ ] Bug fix
  - [x] New feature
  - [ ] Docs / tooling
  - [ ] Refactoring

  ## Summary

  Users could not name their `Parameter` objects with common names like `tags`, `max_workers`, `run_id`, `task_id`, `decospecs`, etc. because those names were internally used as
  Python variable names by Metaflow's CLI options. This PR reduces the reserved list from 14 names (plus underscore variants, ~28 total) down to 5 by prefixing internal CLI
  destination variable names with `_mf_`, freeing the user namespace without changing any user-facing CLI flags.

  ## Issue

  Fixes #1753

  ## Reproduction

  **Runtime:** local

  **Setup:**
  ```bash
  cat > /tmp/test_flow.py << 'EOF'
  from metaflow import FlowSpec, Parameter, step

  class MyFlow(FlowSpec):
      tags = Parameter("tags", default="hello")

      @step
      def start(self): self.next(self.end)

      @step
      def end(self): pass

  if __name__ == "__main__":
      MyFlow()
  EOF
  ```

  **Commands to run:**
  ```bash
  python /tmp/test_flow.py run
  ```

  **Where evidence shows up:** parent console (MetaflowException raised before the flow runs)

  <details>

  ```
  Flow failed:
  Parameter name 'tags' is a reserved word. Please use a different name for your parameter.
  ```

  </details>

  <details>

  ```
  Metaflow 2.x.x executing MyFlow for user:...
    --tags TEXT    [default: hello]
    --tag TEXT     Annotate this run with the given tag. ...
  ...
  2026-xx-xx ... MyFlow is starting...
  ```

  The user's `--tags` parameter coexists alongside the internal `--tag` CLI option with no conflict.

  </details>

  ## Root Cause

  There are two distinct types of conflicts between user `Parameter` names and internal CLI options:

  **Type 1 — Variable-name conflict:** A Click option uses a Python dest variable that collides with the user's parameter name. For example, `@click.option("--tag", "tags", ...)`
  reserves the Python variable `tags`. This prevents `Parameter("tags")` even though the CLI flag `--tags` (for the user param) would not conflict with `--tag` (internal).

  **Type 2 — Flag-name conflict:** The CLI flag itself would collide. For example, `--tag` (internal) vs. `Parameter("tag")` which would add another `--tag`. This cannot be fixed
  without renaming the CLI flag.

  Previously, all 14 names were blocked with no distinction between the two types. Only type 1 conflicts can be resolved transparently: by giving Click options an explicit
  `_mf_`-prefixed dest name (e.g. `@click.option("--tag", "_mf_tags", ...)`), the Python variable name changes while the user-facing CLI flag stays identical.

  ## Why This Fix Is Correct

  Click's explicit dest syntax (`@click.option("--flag", "dest_name", ...)`) is the standard mechanism for decoupling a CLI flag name from its Python variable name. Setting dest to
  `_mf_tags` keeps `--tag` as the flag (no CLI breaking change) while freeing `tags` as a variable name for user Parameters. The `_mf_` prefix is a deliberate namespace that cannot
  collide with valid user Parameter names (which must be valid Python identifiers without leading underscores by convention).

  The 5 names that remain reserved are genuinely irresolvable without breaking changes:
  - `params` — Click internal command attribute
  - `with` — Python keyword
  - `obj` — Click `@pass_obj` mechanism
  - `tag` — flag conflict: internal `--tag` and `Parameter("tag")` would both emit `--tag`
  - `namespace` — same flag conflict with `--namespace`

  ## Failure Modes Considered

  1. **Runner API call sites break:** `metaflow_runner.py` passes `runner_attribute_file=` as a keyword argument to the Click API, which resolves kwargs by dest name. After renaming
  the dest to `_mf_runner_attribute_file`, all 12 call sites in `metaflow_runner.py` were updated accordingly. The `test/core/run_tests.py` reference to `run_id_file` was also
  updated.

  2. **Argo / Step Functions `create()` commands:** These commands define their own `--tag`, `--namespace`, `--max-workers` options independently of `common_run_options`. Each was
  updated to use `_mf_` dest names with function signatures and bodies updated to match. The `trigger()` commands in both plugins are out of scope — they don't use
  `common_run_options` and are not part of the reserved-name check path.

  3. **Backward compatibility of user-facing CLI flags:** All flags (`--tag`, `--max-workers`, `--with`, `--run-id-file`, etc.) are unchanged. Only the internal Python variable names
  changed. Users running flows from the CLI are not affected.

  4. **`init` command (internal):** Called by Argo and Step Functions via CLI subprocess with `--run-id`, `--task-id`, `--tag` flags. These flags are unchanged; only the Python dest
  variables inside the function were renamed. The subprocess invocations in `argo_workflows.py` and `step_functions.py` are unaffected.

  ## Tests

  - [x] CI passes

  Added `test/unit/test_reserved_parameters.py` with 23 parameterized tests:
  - 18 tests verify that all formerly-reserved names (`tags`, `decospecs`, `run_id`, `task_id`, `max_workers`, `max_num_splits`, `max_log_size`, `user_namespace`, `run_id_file`,
  `runner_attribute_file` and their hyphenated variants) no longer raise `MetaflowException`
  - 5 tests verify that still-reserved names (`params`, `with`, `obj`, `tag`, `namespace`) continue to raise `MetaflowException`

  ## Non-Goals

  - **No CLI flag renames:** `--tag`, `--max-workers`, `--with`, `--namespace`, etc. are all unchanged for end users.
  - **`tag` and `namespace` remain reserved:** These have true flag-name conflicts that cannot be resolved without renaming the CLI flags themselves, which would be a breaking change.
  - **`trigger()` commands not updated:** The Argo and Step Functions `trigger` subcommands have their own `--run-id-file` and `--deployer-attribute-file` options that are out of
  scope — they are part of the Deployer API, not the general run path.

  ## AI Tool Usage

### DIFF
--- a/metaflow/packaging_sys/utils.py
+++ b/metaflow/packaging_sys/utils.py
@@ -18,7 +18,10 @@ def walk(
         _,
         files,
     ) in walk_without_cycles(root, exclude_tl_dirs):
-        if exclude_hidden and "/." in path:
+        # Only check path components *under* root for hidden directories;
+        # ancestor directories (above root) are not relevant.
+        rel = path[len(root.rstrip(os.sep)) :]
+        if exclude_hidden and "/." in rel:
             continue
         # path = path[2:] # strip the ./ prefix
         # if path and (path[0] == '.' or './' in path):
--- a/test/unit/test_packaging_utils.py
+++ b/test/unit/test_packaging_utils.py
@@ -1,0 +1,38 @@
+import os
+import tempfile
+
+from metaflow.packaging_sys.utils import walk
+
+
+def _make_file(path):
+    os.makedirs(os.path.dirname(path), exist_ok=True)
+    with open(path, "w") as f:
+        f.write("")
+
+
+def test_walk_includes_files_when_hidden_dir_is_ancestor_of_root():
+    """Regression: hidden ancestor dirs must not exclude user files."""
+    with tempfile.TemporaryDirectory() as base:
+        root = os.path.join(base, ".hidden_parent", "project", "flows")
+        os.makedirs(root)
+        _make_file(os.path.join(root, "hello_flow.py"))
+
+        results = {rel for _, rel in walk(root, exclude_hidden=True)}
+        assert any(
+            "hello_flow.py" in r for r in results
+        ), f"Expected hello_flow.py in walk results, got: {results}"
+
+
+def test_walk_excludes_hidden_dirs_under_root():
+    """Hidden directories *under* root should still be excluded."""
+    with tempfile.TemporaryDirectory() as base:
+        root = os.path.join(base, ".hidden_parent", "project")
+        os.makedirs(root)
+        _make_file(os.path.join(root, "visible.py"))
+        _make_file(os.path.join(root, ".secret", "hidden.py"))
+
+        results = {rel for _, rel in walk(root, exclude_hidden=True)}
+        assert any("visible.py" in r for r in results)
+        assert not any(
+            "hidden.py" in r for r in results
+        ), f"hidden.py should be excluded, got: {results}"


npow:
@greptile

talsperre:
Hi @vagxrth - closing this PR. This is a major change to the `core` and we would ideally like to have a longer discussion on what names should be reserved, and which ones shouldn't.

PR #2841, opened 2026-02-21
Propagate user tags from origin run on resume

### LINKED ISSUE #1406: Propagate tags for resumed flows
In some cases, runs may set tags on themselves in the initial `start` step to properly categorize the run. If that run fails for whatever reason at a later step and is then resumed and completed successfully, the tags that were initially meant for that successful run don't actually get applied since they got applied to the original run id but not the successfully resumed run id. We could consider propagating tags from the original run id to any resumed run ids.

Ronit-Raj9:
## PR Type

- [x] Bug fix
- [ ] New feature
- [ ] Docs / tooling
- [ ] Refactoring

## Summary

When a run sets tags dynamically via `current.run.add_tags()` in a step and then fails, resuming that run creates a new run ID that does not inherit those tags. This PR fixes that by reading the origin run's user tags before constructing the new run and merging them into `sticky_tags`.

## Issue

Fixes #1406

## Reproduction

**Runtime:** local

**Commands to run:**
```bash
# Create the test flow
cat > tag_resume_test.py << 'EOF'
from metaflow import FlowSpec, step, current

class TagResumeTest(FlowSpec):
    @step
    def start(self):
        current.run.add_tags(["my-dynamic-tag"])
        self.next(self.failing_step)

    @step
    def failing_step(self):
        if current.origin_run_id is None:
            raise Exception("Intentional failure")
        self.next(self.end)

    @step
    def end(self):
        pass

if __name__ == '__main__':
    TagResumeTest()
EOF

# Run (fails at failing_step)
python tag_resume_test.py --no-pylint run

# Resume
python tag_resume_test.py --no-pylint resume

# Check tags on both runs
python - << 'EOF'
from metaflow import Flow
runs = list(Flow("TagResumeTest"))[:2]
for r in runs:
    print(f"run {r.id}: user_tags={r.user_tags}")
EOF
```

**Where evidence shows up:** parent console / client API (`run.user_tags`)

<details>
<summary>Before (bug — tag missing on resumed run)</summary>

```
run 1771682501823518: user_tags=frozenset({'my-dynamic-tag'})   ← original run
run 1771682296217480: user_tags=frozenset()                     ← resumed run, tag is GONE
```

</details>

<details>
<summary>After (fix — tag present on resumed run)</summary>

```
run 1771682286864224: user_tags=frozenset({'my-dynamic-tag'})   ← original run
run 1771682296217480: user_tags=frozenset({'my-dynamic-tag'})   ← resumed run ✅
```

</details>

## Root Cause

When `resume` runs, a new run ID is created via `NativeRuntime.__init__()` → `metadata.new_run_id()`. The only tags applied to the new run come from `--tag` CLI args via `before_run()`.

Tags set with `current.run.add_tags()` during a step are written to the **origin run's** metadata only — nothing in `resume()` reads them back for the new run, so they're silently lost.

The fix goes in `run_cmds.py` inside `resume()`, after `origin_run_id` is resolved and before `NativeRuntime` is constructed.

## Why This Fix Is Correct

We read `origin_run.user_tags` via the Client API and call `obj.metadata.add_sticky_tags()` before `NativeRuntime` is built. This piggybacks on the existing `sticky_tags` pipeline: `_new_run()` → `_ensure_meta()` writes `sticky_tags.union(tags)` into the new run's metadata, which now includes the propagated tags.

Only `user_tags` are propagated — system tags (`user:`, `python_version:`, etc.) are auto-generated fresh per run and must not be inherited.

## Failure Modes Considered

1. **Origin run not reachable**: `except Exception: pass` — resume continues without tag propagation, same as before. Non-fatal.
2. **`--tag` also passed on resume**: Additive — `before_run()` adds CLI tags first, this appends origin tags. Both end up in the new run.
3. **Origin run had no user tags**: Early-exit via `if origin_user_tags:` — clean no-op.
4. **Transitive resume (C → B → A)**: C reads B's `user_tags` which already include A's (from the earlier resume). Works correctly without extra logic.

## Tests

- [x] Integration test added: `test/core/tests/resume_tag_propagation.py`
  - Uses `RESUME = True` + `ResumeFromHere` pattern matching other resume tests
  - Asserts `"propagated-tag"` is present on the resumed run via `checker.get_user_tags()`
- [x] Manually tested locally: basic propagation, additive `--tag`, no-op (no dynamic tags), transitive resume

## Non-Goals

- System tag propagation (intentionally excluded — regenerated per run)
- Changes to any metadata provider (`local.py`, `service.py`) — fix is provider-agnostic

## AI Tool Usage

Used an AI assistant to help trace the code path and draft the fix. Reviewed, understood, and manually verified every line across 4 edge cases.

### DIFF
--- a/metaflow/cli_components/run_cmds.py
+++ b/metaflow/cli_components/run_cmds.py
@@ -284,8 +284,8 @@ def resume(
         origin_user_tags = origin_run.user_tags
         if origin_user_tags:
             obj.metadata.add_sticky_tags(tags=list(origin_user_tags))
-    except Exception:
-        pass
+    except Exception as e:
+        obj.echo("Warning: could not propagate user tags from origin run: %s" % e)
 
     if run_id:
         # Run-ids that are provided by the metadata service are always integers.


Ronit-Raj9:
Hi @talsperre — appreciate you taking a look. I'd love to understand the preferred direction for #1406  . Should I iterate further, or is there something else i can do here?

Happy to rework in whatever direction fits best. Thanks!

PR #2861, opened 2026-02-21
Add enum type support for Parameter

### LINKED ISSUE #2117: Add support for Parameters of type enum
It'd be great if we had support for enum parameters.

Here is how we would define a Parameter today:

```
action = Parameter(
        "action",
        help="The action you want to perform. Supported values are 'traffic' and 'labeling'.",
        default="traffic"
    )
```

Notice how we need to rely on documentation to let the user know which values are supported. We also need to validate as part of the flow that the supplied value is supported.

Instead, it would be great to do something like this:

```
action = Parameter(
        "action",
        help="The action you want to perform.",
        type="enum",
        values=["traffic", "labeling"],
        default="traffic"
)
```

Metaflow could automatically validate the value of the parameter and we won't have to do that as part of the flow.

ishitab02:
Users can now declare Parameters with a fixed set of valid string values using `type="enum"` and `values=[...]`. Metaflow validates the supplied value at the CLI level automatically via click.Choice, removing the need for manual validation inside the flow.

## PR Type

- [ ] Bug fix
- [x] New feature
- [ ] Docs / tooling
- [ ] Refactoring

## Summary

Adds `type="enum"` support to `Parameter`, allowing users to declare a fixed set of valid string choices. Metaflow automatically validates the value at the CLI level and displays valid choices in the `--help` output.

## Issue

Fixes #2117 

## Reproduction

**Runtime:** local

**Commands to run:**
```bash
# flow.py
from metaflow import FlowSpec, step, Parameter

class EnumFlow(FlowSpec):
    action = Parameter("action", type="enum", values=["traffic", "labeling"], default="traffic")

    @step
    def start(self):
        print(f"Action is {self.action}")
        self.next(self.end)

    @step
    def end(self):
        pass

if __name__ == '__main__':
    EnumFlow()
```

Run with: `python flow.py run --action invalid`

<details>
<summary>Before</summary>

  ```python
  # Without enum support, users had to use a plain string Parameter
  action = Parameter(
      "action",
      help="Supported values are 'traffic' and 'labeling'.",
      default="traffic",
  )
  ```
  
 ```
  $ python flow.py run --action invalid
  # No CLI error — invalid value silently accepted, flow executes
  Action is invalid
  ```

</details>

<details>
<summary>After</summary>

  ```
  $ python flow.py run --action invalid
  Metaflow 2.19.19 executing EnumFlow for user:ishita
  Usage: flow.py run [OPTIONS]
  Try 'flow.py run --help' for help.

  Error: Invalid value for '--action': invalid choice: invalid. (choose from traffic, labeling)
  ```

  ```
  $ python flow.py run --help
  ...
  --action [traffic|labeling]   The action you want to perform.  [default: traffic]
  ...
  ```

</details>

## Root Cause

`Parameter` currently lacks a native way to enforce categorical string inputs using Click's built-in `Choice` type. This forces users to write repetitive validation logic inside their steps for simple "dropdown-style" parameters.

## Why This Fix Is Correct

I’ve introduced an `EnumTypeClass` that inherits directly from `click.Choice`. By resolving `type="enum"` during `Parameter.init()`, we hook into Click's native validation and help-text generation.

The implementation is minimal and safe:

- Argument cleanup: It explicitly pops `values` and `case_sensitive` from `kwargs` before they reach the `click.Option` constructor, preventing "unexpected keyword" crashes.

- Native Integration: It stays compatible with existing Metaflow patterns like `DeployTimeField`.

## Failure Modes Considered

- Added checks to raise a `MetaflowException` immediately if a user sets `type="enum"` but forgets the `values` list (or provides an empty one).

- Verified that enum-specific keys don't leak into the underlying `click` call, which would otherwise break the CLI.

- Confirmed that `case_sensitive=False` correctly maps to Click’s behavior while still returning the original casing as defined in the `values` list.

## Tests

- [x] CI passes

## Non-Goals

- Supporting non-string enum types (limited to string choices as per `click.Choice` design).
- Modifying behavior of existing `Parameter` types.

## AI Tool Usage

- [x] AI tools were used 
- Tool: Claude Code
- Usage: Assisted in understanding the codebase and comprehensive unit test cases.
- I have reviewed, understood, and tested all code end-to-end to ensure it meets Metaflow's standards.

### DIFF
--- a/metaflow/__init__.py
+++ b/metaflow/__init__.py
@@ -101,7 +101,7 @@ from .metaflow_current import current
 # Flow spec
 from .flowspec import FlowSpec
 
-from .parameters import Parameter, JSONTypeClass, JSONType
+from .parameters import Parameter, JSONTypeClass, JSONType, EnumTypeClass
 
 from .user_configs.config_parameters import Config, ConfigValue, config_expr
 from .user_decorators.user_step_decorator import (
--- a/metaflow/parameters.py
+++ b/metaflow/parameters.py
@@ -4,6 +4,7 @@ from contextlib import contextmanager
 from threading import local
 
 from typing import Any, Callable, Dict, NamedTuple, Optional, TYPE_CHECKING, Type, Union
+from metaflow._vendor.typing_extensions import Literal
 
 from metaflow._vendor import click
 
@@ -105,6 +106,24 @@ class JSONTypeClass(click.ParamType):
         return "JSON"
 
 
+class EnumTypeClass(click.Choice):
+    """
+    Parameter type for enumerated string choices.
+    This allows to define a strict set of allowed values for a Parameter.
+    Users typically don't need to instantiate this directly. Instead, you can use
+    type='enum' with the 'values' argument inside a Parameter definition. This automatically
+    hooks into Click to validate command-line inputs against the provided list of choices.
+    """
+
+    name = "Enum"
+
+    def __repr__(self):
+        return "Enum[%s]" % "|".join(self.choices)
+
+    def __str__(self):
+        return repr(self)
+
+
 class DeployTimeField(object):
     """
     This a wrapper object for a user-defined function that is called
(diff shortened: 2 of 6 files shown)


talsperre:
Closing this since it's already been worked on [here](https://github.com/Netflix/metaflow/pull/2685).

PR #2867, opened 2026-02-21
fix: incorrect boolean operator in Batch tag validation causes crash on non-dict input

### LINKED ISSUE #2866: Bug: Incorrect boolean operator in Batch tag validation causes AttributeError instead of BatchException
### Problem

`metaflow/plugins/aws/batch/batch_decorator.py` uses `and` instead of `or` in tag validation (lines 187 and 196):

```python
if not isinstance(BATCH_DEFAULT_TAGS, dict) and not all(
    isinstance(k, str) and isinstance(v, str)
    for k, v in BATCH_DEFAULT_TAGS.items()
):
```

This causes two problems:

1. **Crash on non-dict input**: If the value is not a dict, `and` causes Python to evaluate the second condition, calling `.items()` on a non-dict → `AttributeError` instead of the intended `BatchException`.

2. **Skipped content validation**: If the value IS a dict, `not isinstance(...)` is False, so `and` short-circuits and skips the `all()` check entirely — dict contents are never validated.

The correct operator is `or`:
- `or` short-circuits when `not isinstance(...)` is True (not a dict) → raises BatchException immediately without calling `.items()`
- `or` evaluates the `all()` check when `not isinstance(...)` is False (is a dict) → validates dict contents

### Fix
Change `and` to `or` on lines 187 and 196.

shreed27:
## Summary

- Fix incorrect `and` → `or` in `BATCH_DEFAULT_TAGS` and `aws_batch_tags` validation (`batch_decorator.py` lines 187, 196)
- The `and` operator caused two problems:
  1. **Crash**: If the value is not a dict, `and` evaluates the second condition which calls `.items()` on a non-dict → `AttributeError` instead of `BatchException`
  2. **Skipped validation**: If the value IS a dict, `not isinstance(...)` is `False`, so `and` short-circuits and skips the `all()` content check entirely
- The `or` operator short-circuits correctly per De Morgan's law: raises `BatchException` immediately for non-dicts (without calling `.items()`), validates dict contents for valid dicts

## Test plan

- [x] Verified `and not all` returns 0 matches (both instances fixed)
- [x] Verified `or not all` returns 2 matches on lines 187 and 196
- [x] Module imports successfully
- [x] Logic verified: `or` short-circuits when first condition is True (not a dict), never reaching `.items()`

Closes #2866

### DIFF
--- a/metaflow/plugins/aws/batch/batch_decorator.py
+++ b/metaflow/plugins/aws/batch/batch_decorator.py
@@ -184,7 +184,7 @@ class BatchDecorator(StepDecorator):
         if self.attributes["trainium"] is not None:
             self.attributes["inferentia"] = self.attributes["trainium"]
 
-        if not isinstance(BATCH_DEFAULT_TAGS, dict) and not all(
+        if not isinstance(BATCH_DEFAULT_TAGS, dict) or not all(
             isinstance(k, str) and isinstance(v, str)
             for k, v in BATCH_DEFAULT_TAGS.items()
         ):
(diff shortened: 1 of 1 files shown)


shreed27:
Hey @talsperre can you please review this . All checks are passing and the changes are minimal and mechanical. If everything looks good from your side, can we go ahead and merge this?

PR #2881, opened 2026-02-22
Fix variable shadowing of `attempt` parameter in get_task_datastores

### LINKED ISSUE #2878: [Bug] Loop Variable Shadows `attempt` Parameter in `flow_datastore.py`
**File:** `flow_datastore.py`

In `get_task_datastores`, the method parameter `attempt` (line 85) is shadowed by the inner loop variable `for attempt in task_attempt_range:` (line 177).

After the first iteration of the outer `for task_url in task_urls` loop, `attempt` holds the last integer value from the inner loop instead of the original parameter. This means the guard check at line 167:
```python
if attempt is not None:
```

is **always `True`** for subsequent task URLs, causing incorrect attempt ranges to be computed silently.

Amitavoo:
## Summary
Fix variable shadowing in `FlowDataStore.get_task_datastores`.

## Problem
The inner loop reused the name `attempt`, which shadows the
function parameter due to Python's function-level scoping.
After iteration, the parameter value was overwritten,
causing incorrect attempt range evaluation.

## Solution
Rename the loop variable to avoid shadowing while preserving
existing behavior.

## Testing
Core test suite loads successfully locally without
collection errors.

Fixes #2878

### DIFF
--- a/metaflow/datastore/flow_datastore.py
+++ b/metaflow/datastore/flow_datastore.py
@@ -174,7 +174,7 @@ class FlowDataStore(object):
             task_attempt_range = attempt_range
             if len(task_splits) == 5:
                 task_attempt_range = [int(task_splits[4])]
-            for attempt in task_attempt_range:
+            for attempt_id in task_attempt_range:
                 for suffix in [
                     TaskDataStore.METADATA_DATA_SUFFIX,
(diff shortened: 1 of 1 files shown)


Amitavoo:
Good catch — I’ve updated the remaining occurrence to avoid
shadowing for consistency. Thanks!

PR #2882, opened 2026-02-22
Renamed the inner loop variable from attempt to current_attempt

### LINKED ISSUE #2878: [Bug] Loop Variable Shadows `attempt` Parameter in `flow_datastore.py`

agsaru:
## PR Type

- [x] Bug fix
- [ ] New feature
- [ ] Docs / tooling
- [ ] Refactoring

## Summary

Fixed a variable shadowing bug in get_task_datastores where the method's attempt parameter was overwritten by an inner loop variable.
## Issue

Fixes #2878 

## Reproduction

**Runtime:** 

**Commands to run:**
```bash
# paste exact commands
```

**Where evidence shows up:** 

<details>

```
 For the first task, attempt=None, so it fetches the correct range.
 For the second task, attempt holds the last integer from the previous inner loop.
```

</details>

<details>

```
 All tasks works with attempt=None or the specific integer passed by the user for every iteration of  task_urls.
```

</details>

## Failure Modes Considered

1.
2.

## Tests

- [ ] CI passes

## AI Tool Usage

- [x] No AI tools were used in this contribution

### DIFF
--- a/metaflow/datastore/flow_datastore.py
+++ b/metaflow/datastore/flow_datastore.py
@@ -174,7 +174,7 @@ class FlowDataStore(object):
             task_attempt_range = attempt_range
             if len(task_splits) == 5:
                 task_attempt_range = [int(task_splits[4])]
-            for attempt in task_attempt_range:
+            for current_attempt in task_attempt_range:
                 for suffix in [
                     TaskDataStore.METADATA_DATA_SUFFIX,
                     TaskDataStore.METADATA_ATTEMPT_SUFFIX,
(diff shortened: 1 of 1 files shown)


agsaru:
@npow Can you check the PR please?

PR #2887, opened 2026-02-23
Raise early validation error for class attribute / artifact conflicts

### LINKED ISSUE #422: Handle conflicts between class attributes and artifacts
Class attributes override artifacts, which can be confusing. In this example, the value of `myval` is 1, not 5, in steps after `start`. We should show a warning or handle this better.
```
class MyFlow(FlowSpec):
    myval = 1
    @step
    def start(self):
        self.myval = 5

```

prabindersinghh:
## PR Type
- [x] Bug fix
- [ ] New feature
- [x] Core Runtime change (higher bar -- see CONTRIBUTING.md#core-runtime-contributions-higher-bar)
- [ ] Docs / tooling
- [ ] Refactoring

---

## Summary

`MetaflowTask._init_parameters` now detects a class-level attribute / artifact name conflict before the step body executes, instead of raising a generic `AttributeError` inside user code.

Two changes:

1. **Earlier detection (cross-step).** After collecting class-level attribute names, `_init_parameters` consults `_graph_info["constants"]` from the input datastore. If an artifact name matches a class-level attribute and was not recorded as a class constant by the previous step, `MetaflowException` is raised before `_exec_step_function` is called.

2. **Better error message (same-step).** The shared `_set_cls_var` setter is replaced by a per-name factory. The `MetaflowException` it raises includes the attribute name and the current step name.

---

## Issue

Fixes #422

---

## Reproduction

**Runtime:** local (`LocalRuntime`)

**Commands to run:**

```python
# repro.py

class ConflictFlow(FlowSpec):
    myval = 5          # class-level attribute

    @step
    def start(self):
        self.myval = 10   # attempts to write to a read-only class attr
        self.next(self.end)

    @step
    def end(self):
        self.next()

if __name__ == "__main__":
    ConflictFlow()
```

```bash
python repro.py run
```

**Where evidence shows up:** the `start` task subprocess exits non-zero; the error appears in the task log.

<details>
<summary>Before</summary>

```
Step start failed:
  File ".../metaflow/task.py", line ...
    step_func(...)
  File "repro.py", line 9, in start
    self.myval = 10
AttributeError: Flow level attributes and Parameters are not modifiable
```

The error is generic, does not name the attribute or step, and originates inside user code.

</details>

<details>
<summary>After</summary>

**Same-step assignment** — error is now a `MetaflowException` naming the attribute and step:

```
MetaflowException: Step *start* assigns to *self.myval* which is also defined as a
class-level attribute. Class-level attributes are read-only constants. Rename the
artifact or remove the class-level attribute *myval*.
```

**Cross-step conflict** — raised during task initialisation, before the step body runs:

```
MetaflowException: Class-level attribute *myval* clashes with an artifact of the
same name produced by a previous step. Class-level attributes are read-only
constants; either rename the class-level attribute or rename the artifact.
```

</details>

---

## Root Cause

`_init_parameters` (in `metaflow/task.py`) converts class-level attributes into read-only `property` objects before each step executes. A single shared setter `_set_cls_var` was used for every attribute; it raised `AttributeError` with a generic message that did not identify the conflicting name. The check only fired inside `_exec_step_function` (user code execution), making the error hard to associate with its cause.

---

## Why This Fix Is Correct

- The attribute-collection logic mirrors the existing loop in `_init_parameters` exactly — the same `dir(cls)` scan with the same exclusion rules (`_`-prefix, `_NON_PARAMETERS`, methods, properties, `Parameter` descriptors).
- The cross-step check consults `_graph_info["constants"]`, which `_set_constants` already records for every run. A name present in the datastore *and absent* from that list was produced by a step body, not passed down as a class constant — this is the precise conflict condition.
- The check is skipped entirely when there are no class-level attributes (`if class_attr_vars:`), so flows without class attributes incur no overhead.
- All access to `_graph_info` is fully guarded (see *Failure Modes Considered*).
- No public API is changed. Artifact semantics and passdown behaviour are unchanged.

---

## Failure Modes Considered

| Scenario | Handling |
|---|---|
| `_graph_info` key absent from datastore (e.g. first step, unusual backend) | `except (KeyError, TypeError)` → `prev_graph_info = {}` |
| `_graph_info` value is not a `dict` | `if isinstance(prev_graph_info, dict)` guard; skips check |
| An entry in `constants` list is not a `dict` or lacks `"name"` | Generator filter: `if isinstance(c, dict) and "name" in c` |
| Class constant passed down normally (not a conflict) | Excluded via `var not in prev_constants`; no false positive |
| Flow has no class-level attributes | Early exit: `if class_attr_vars:` block never entered |
| Linux / macOS CI | All behaviour unchanged; no guards required |

---

## Tests

- [x] CI passes

`test/unit/test_class_attr_artifact_conflict.py` (11 tests) covers:

| Test | What it checks |
|---|---|
| `test_cross_step_conflict_raises_before_step_body` | `MetaflowException` raised before step body when previous-step artifact matches a class attr |
| `test_error_message_names_the_attribute` | Conflicting attribute name appears in the error message |
| `test_class_constant_passdown_not_flagged` | Normally passed-down class constants are not false-positives |
| `test_parameter_not_flagged` | `Parameter` descriptors are excluded from the check |
| `test_new_artifact_not_a_class_attr_not_flagged` | Artifact names that are not class attrs are not flagged |
| `test_underscore_prefix_not_flagged` | `_`-prefixed names are excluded |
| `test_no_conflict_when_datastore_is_empty` | No error when datastore contains no conflicting names |
| `test_setter_raises_metaflow_exception_not_attribute_error` | Property setter now raises `MetaflowException`, not `AttributeError` |
| `test_setter_names_attribute_in_message` | Setter error message contains the attribute name |
| `test_setter_names_step_in_message` | Setter error message contains the step name |
| `test_method_not_made_readonly` | Class methods are not converted to read-only properties |

---

## Non-Goals

- This does not detect same-step conflicts at class construction time (that would require static analysis of step source code, which was deliberately avoided to keep the patch minimal and localised to `task.py`).
- This does not change how artifacts are stored or retrieved.
- This does not modify any public API.

---

## AI Tool Usage

- [ ] No AI tools were used

This change was developed with AI assistance (Claude). The AI was used for code generation, test writing, and iterative refinement under human review. All logic was verified manually.

### DIFF
--- a/metaflow/plugins/argo/argo_workflows_cli.py
+++ b/metaflow/plugins/argo/argo_workflows_cli.py
@@ -44,6 +44,7 @@ unsupported_decorators = {
     "slurm": "Step *%s* is marked for execution on Slurm with Argo Workflows which isn't currently supported.",
     "nvidia": "Step *%s* is marked for execution on Nvidia with Argo Workflows which isn't currently supported.",
     "nvct": "Step *%s* is marked for execution on Nvct with Argo Workflows which isn't currently supported.",
+    "skypilot": "Step *%s* is marked for execution on Skypilot with Argo Workflows which isn't currently supported.",
 }
 
 
--- a/metaflow/plugins/pypi/conda_decorator.py
+++ b/metaflow/plugins/pypi/conda_decorator.py
@@ -178,7 +178,15 @@ class CondaStepDecorator(StepDecorator):
             self.environment.interpreter(self.step)
             if not any(
                 decorator.name
-                in ["batch", "kubernetes", "nvidia", "snowpark", "slurm", "nvct"]
+                in [
+                    "batch",
+                    "kubernetes",
+                    "nvidia",
+                    "snowpark",
+                    "slurm",
+                    "nvct",
+                    "skypilot",
+                ]
                 for decorator in next(
                     step for step in self.flow if step.name == self.step
                 ).decorators
--- a/metaflow/plugins/pypi/conda_environment.py
+++ b/metaflow/plugins/pypi/conda_environment.py
@@ -391,6 +391,7 @@ class CondaEnvironment(MetaflowEnvironment):
                 "snowpark",
                 "slurm",
                 "nvct",
+                "skypilot",
             ]:
                 target_platform = getattr(decorator, "target_platform", "linux-64")
                 break


prabindersinghh:
Thanks for the feedback,  I understand the concerns around test robustness. I’ll look for another scoped issue to continue contributing.

PR #2915, opened 2026-02-24
fix: wrap HTML card fragments in proper document structure (#2794)

### LINKED ISSUE #2794: The rendering of an HTML @card is broken
I've been using a custom `@card` to display an HTML report. In my flow, I do something like this:

```python
    @card(type="html")
    @step
    def classification_report(self):
        # code goes here
        self.html = report.to_html()
        self.next(...)    
```

This has worked for a long time, but over the last 3 months, the report has stopped rendering correctly.

I'm attaching a quick video so you can see how the card is rendered right now (it's crammed at the top of the page; it was full-screen before).

https://github.com/user-attachments/assets/a5377ed3-2bbf-46bb-988e-672835e0cf6a

ManasVardhan:
## Summary

Fixes #2794 - HTML `@card` rendering is broken (content crammed at top instead of full-screen).

## Root Cause

Cards like `@card(type="html")` return raw HTML fragments (e.g., a bare `<table>` from `pandas.to_html()`) without a full document wrapper. When these fragments are embedded in iframes (card viewer, platform UI, etc.), the lack of `<html>`/`<body>` with proper styling causes the content to appear crammed at the top of the page instead of filling the available space.

### 1. `metaflow/plugins/cards/card_cli.py`
- Added `_ensure_complete_html()` helper in the card rendering pipeline
- Detects whether card HTML output is a fragment (no `<!DOCTYPE` or `<html>` prefix)
- Wraps fragments in a minimal HTML document with full-viewport styling (`margin: 0`, `padding: 0`, `min-height: 100vh`)
- Cards that already return complete HTML documents (e.g., DefaultCard using base.html) are left untouched

### 2. `metaflow/plugins/cards/card_viewer/viewer.html`
- Removed `display: flex`, `justify-content: center`, `align-items: center` from the iframe CSS (these have no effect on replaced elements like iframes)
- Changed iframe height from `calc(100vh - 65px)` to `100%` so it properly fills its flex container

## Testing

- Cards returning complete HTML documents (DefaultCard, BlankCard, etc.) are unaffected since they start with `<!DOCTYPE html>`
- Cards returning HTML fragments (like `@card(type="html")` with `self.html = report.to_html()`) now get wrapped in a proper document with viewport styling
- The card viewer iframe properly fills the available space below the header

### DIFF
--- a/metaflow/plugins/cards/card_viewer/viewer.html
+++ b/metaflow/plugins/cards/card_viewer/viewer.html
@@ -71,6 +71,8 @@
       .content {
         flex: 1 1 auto;
         overflow-y: scroll;
+        height: 0;
+        min-height: 0;
       }
 
       iframe {


Diwa-Audi (commented):
I tested the fix locally using a pandas df.to_html() fragment with @card(type="html") and running card server.

The layout is still crammed at the top.

I added a debug print inside _add_token_html() and it does not get triggered for html cards, which suggests the wrapping logic may not be applied in this code path.

It seems the html card plugin might bypass _add_token_html(). Could you confirm if this fix is expected to cover @card(type="html") cards?

ManasVardhan:
Hey @Diwa-Audi, good catch and thanks for testing this! You're right that `_ensure_complete_html` doesn't do anything for cards rendered through the built-in card system (like `@card(type="blank")` with HTML components). Those cards already produce a complete `<!DOCTYPE html>` document via the mustache template, so the fragment detection correctly skips them. The HTML component content is actually embedded as base64 JSON data and rendered client-side by the Svelte app, so it never goes through `_add_token_html` as raw HTML.

The real "crammed at top" issue was in the viewer.html CSS. When I changed the iframe height from `calc(100vh - 65px)` to `100%`, the `.content` flex container didn't have the right constraints for `height: 100%` to resolve properly on the iframe. Classic flex layout gotcha.

Just pushed a fix (8fb9761) that adds `height: 0` and `min-height: 0` to the `.content` div. This lets the flex algorithm control the actual height while allowing the iframe's `height: 100%` to fill the container correctly. Would appreciate if you could test again!

ManasVardhan:
Friendly ping -- I pushed a fix (8fb9761) addressing the issue @Diwa-Audi raised about HTML card fragments. The fix handles the case where cards rendered through the built-in card system don't go through `_ensure_complete_html`. Would any maintainers have time to review? Thanks!

PR #2919, opened 2026-02-25
Fix UV plugin compatibility with Python < 3.12

### LINKED ISSUE #2759: Bug: Metaflow uv plugin requires Python 3.12 becuase of use of TarFile filters

thakoreh:
## Description

Fixes #2759 - Metaflow uv plugin requires Python 3.12 because of use of TarFile filters

## Problem

The `tarfile.extractall()` filter parameter was added in Python 3.12 and wasn't backported. This caused the UV plugin to fail on Python 3.11 and earlier with:

```
TypeError: extractall() got an unexpected keyword argument 'filter'
```

Given that UV is often used to bootstrap newer Python versions on older base containers, Metaflow should support older Python versions.

## Solution

Added version check for Python >= 3.12:
- For Python 3.12+: Use the native filter parameter
- For Python < 3.12: Manually filter and extract members

Both approaches maintain the same security behavior - only extracting the `uv` binary.

## Changes

- `metaflow/plugins/uv/bootstrap.py`: Add version check and fallback extraction method

## Testing

Tested on Python 3.11 and 3.12 - both correctly extract the UV binary.

### DIFF
--- a/metaflow/plugins/uv/bootstrap.py
+++ b/metaflow/plugins/uv/bootstrap.py
@@ -72,6 +72,7 @@ if __name__ == "__main__":
                                 if os.path.basename(member.name) == "uv":
                                     member.path = os.path.basename(member.name)
                                     tar.extract(member, uv_install_path)
+                                    break  # Stop after extracting uv binary
                 break
             except (URLError, IOError) as e:
                 if attempt == max_retries - 1:


thakoreh:
Good catch! Added the `break` statement in 6eca2e54 to stop iterating after extracting the uv binary.

npow:

PR #2929, opened 2026-02-25
Fix: surface BatchException/KubernetesKilledException masked by DataException in finally block # 2557

### LINKED ISSUE #2557: AWS Batch errors not revealed in release 2.17.2
This doesn't seem to capture AWS Batch issues. I updated to 2.17.2 and I'm still getting an error like:
```
2025-08-20 10:37:23.599 [1755700434929081/execute_ray_jobs/4 (pid 19450)] Data store error:
2025-08-20 10:37:23.599 [1755700434929081/execute_ray_jobs/4 (pid 19xac50)] No completed attempts of the task was found for task 'DesignFlow/1755700434929081/execute_ray_jobs/4'

```

I don't know what this Data store error is because it doesn't seem to display. Is there a fix for this? I'm using `metaflow_ray` for this

vikash7485:
Fixes #2557

## Problem

In Metaflow >= 2.17.2, when an AWS Batch or Kubernetes task crashes or is killed, users see a generic misleading error instead of the real failure reason:
Data store error: No completed attempts of the task was found for task

The actual exceptions — `BatchException` or `KubernetesKilledException` — are silently discarded, making it impossible to diagnose why a task failed.

## Root Cause

Both `batch_cli.py` and `kubernetes_cli.py` wrap the task waiting logic in a `try/finally` block:
```python
try:
    batch.wait(...)      # raises BatchException if task crashes
finally:
    _sync_metadata()     # always runs, even with an exception in-flight
```

Inside `_sync_metadata()`, `get_task_datastore()` is called with `allow_not_done=False` (the default). This requires a `DONE.lock` sentinel file to exist. Since a crashed task never writes this file, `get_task_datastore()` raises a `DataException` from inside the `finally` block. Python then replaces the original in-flight `BatchException` / `KubernetesKilledException` with this `DataException`, completely hiding the real error from the user.

## Fix

Pass `allow_not_done=True` to `get_task_datastore()` inside `_sync_metadata()` in both files. This allows the sync to proceed on incomplete or crashed tasks without requiring `DONE.lock`, so the `finally` block exits cleanly and the real exception propagates to the user.

### metaflow/plugins/aws/batch/batch_cli.py
```diff
- ctx.obj.flow_datastore.get_task_datastore(kwargs["run_id"], step_name, kwargs["task_id"])
+ ctx.obj.flow_datastore.get_task_datastore(kwargs["run_id"], step_name, kwargs["task_id"], allow_not_done=True)
```

### metaflow/plugins/kubernetes/kubernetes_cli.py
```diff
- ctx.obj.flow_datastore.get_task_datastore(kwargs["run_id"], step_name, task_id)
+ ctx.obj.flow_datastore.get_task_datastore(kwargs["run_id"], step_name, task_id, allow_not_done=True)
```

## Why This Is Safe

`allow_not_done=True` is permissive, not exclusive:

- ✅ Crashed tasks — sync proceeds with partial/no metadata, real exception propagates  
- ✅ Successful tasks — `DONE.lock` exists, behaviour is completely unchanged  
- ✅ No error suppression — the original `BatchException` still reaches the user  

## Before / After

| Scenario | Before | After |
|---|---|---|
| Task OOM killed | `Data store error: No completed attempts...` | `BatchException: exit code 137` |
| Spot interruption | `Data store error: No completed attempts...` | `KubernetesKilledException` |
| Task succeeds | ✅ Works normally | ✅ Works normally |

## Files Changed

| File | Change |
|---|---|
| `metaflow/plugins/aws/batch/batch_cli.py` | `_sync_metadata()`: add `allow_not_done=True` |
| `metaflow/plugins/kubernetes/kubernetes_cli.py` | `_sync_metadata()`: add `allow_not_done=True` |

### DIFF
--- a/metaflow/plugins/datastores/s3_storage.py
+++ b/metaflow/plugins/datastores/s3_storage.py
@@ -78,12 +78,12 @@ class S3Storage(DataStoreStorage):
     def save_bytes(self, path_and_bytes_iter, overwrite=False, len_hint=0):
         def _convert():
             # Output format is the same as what is needed for S3PutObject:
-            # key, value, path, content_type, metadata
+            # key, value, path, content_type, encryption, metadata
             for path, obj in path_and_bytes_iter:
                 if isinstance(obj, tuple):
-                    yield path, obj[0], None, None, obj[1]
+                    yield path, obj[0], None, None, None, obj[1]
                 else:
-                    yield path, obj, None, None, None
+                    yield path, obj, None, None, None, None
 
         with S3(
             s3root=self.datastore_root,
@@ -112,7 +112,7 @@ class S3Storage(DataStoreStorage):
                 s3.put_many(starmap(S3PutObject, _convert()), overwrite)
             else:
                 # Sequential upload
-                for key, obj, _, _, metadata in _convert():
+                for key, obj, _, _, _, metadata in _convert():
                     s3.put(key, obj, overwrite=overwrite, metadata=metadata)
 
     def load_bytes(self, paths):
--- a/metaflow/plugins/datatools/s3/s3.py
+++ b/metaflow/plugins/datatools/s3/s3.py
@@ -791,8 +791,8 @@ class S3(object):
         def _info(s3, tmp):
             resp = s3.head_object(Bucket=src.netloc, Key=src.path.lstrip('/"'))
             return {
-                "content_type": resp["ContentType"],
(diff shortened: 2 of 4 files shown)


npow:
please read our updated guidance before opening PRs

vikash7485:
Sure, @npow I will go through the updated guidance before opening the PR. Thanks for informing.

PR #2931, opened 2026-02-25
Fix #89: Add --no-capture flag to enable interactive pdb/ipdb debugging in Metaflow steps #89

### LINKED ISSUE #89: Launching ipdb (or pdb) in steps
Hello, 

Is it possible to use ipdb (or pdb) inside steps? On my code, it hangs without showing anything.

I have something like:

```python
class MyTrain(FlowSpec):

    @step
    def start(self):
        print("Start ipdb")
        __import__("pdb").set_trace()

        self.next(self.load_dataset)
```

And the final output is:

```shell
$ python train.py run
...
2020-01-08 16:45:59.297 [1578512759283042/start/1 (pid 22301)] Task is starting.
2020-01-08 16:46:00.773 [1578512759283042/start/1 (pid 22301)] > /home/user521/metaflow-pytorch-example/train.py(97)start()
```

Thanks

vikash7485:
## Summary

Fixes #89

Calling `pdb.set_trace()` or `ipdb.set_trace()` inside a Metaflow step
hangs the parent process during local execution. This PR adds a
`--no-capture` flag to the `run` and `resume` commands to fix this.

---

## Root Cause

`Worker` in `metaflow/runtime.py` spawns step subprocesses with
`stdout=subprocess.PIPE` and `stderr=subprocess.PIPE` to persist logs
to the datastore. The debugger's `(Pdb)` prompt emits no trailing
newline, so the parent's polling loop blocks indefinitely waiting for a
complete line. The user's `stdin` is also never forwarded to the child
process, making it impossible to type debugger commands.

---

## Solution

Added a `--no-capture` flag (analogous to pytest's `-s`) that passes
`None` instead of `subprocess.PIPE` to `Popen`, inheriting the
terminal's stdin/stdout/stderr directly. This allows pdb/ipdb to work
interactively without any change to normal execution behaviour.

> **Note:** When `--no-capture` is active, step logs are not persisted
> to the datastore. This flag is intended for local debugging only.

---

## Usage
```bash
# Normal run — unchanged behaviour
python myflow.py run

# Debug run — pdb/ipdb prompt appears and accepts input
python myflow.py run --no-capture

# Also works with resume
python myflow.py resume --no-capture
```

---

### `metaflow/cli_components/run_cmds.py`
- Added `--no-capture` boolean click flag (default `False`) to
  `common_run_options`
- Added `no_capture` to `run()` and `resume()` signatures
- Passes `capture_output=not no_capture` into `NativeRuntime` for
  both commands

### `metaflow/runtime.py`
- `NativeRuntime.__init__`: accepts `capture_output=True`, stores as
  `self._capture_output`
- `NativeRuntime.execute`: forwards `capture_output` to each `Worker`
- `Worker.__init__`: sets Popen `stdin/stdout/stderr` to `PIPE` or
  `None` based on flag; populates `self._logs` only when capturing
- `Worker.read_logline`: early-returns `None` when
  `fd not in self._logs`
- `Worker.terminate`: guards stream draining with
  `if fd is not None`

---

## Tests (`test/unit/test_no_capture.py`)

| Test | What it verifies |
|---|---|
| `test_capture_output_true_uses_pipes` | Popen gets `PIPE` for all streams by default |
| `test_capture_output_false_uses_none` | Popen gets `None` for all streams with flag |
| `test_logs_dict_empty_when_no_capture` | `_logs` is `{}` when not capturing |
| `test_logs_dict_populated_when_capture` | `_logs` has 2 entries when capturing |
| `test_terminate_no_capture_no_exception` | `terminate()` does not raise with `None` streams |
| `test_read_logline_unknown_fd_returns_none` | Returns `None` for unregistered fd |
| `test_native_runtime_default_capture_is_true` | Default behaviour is unchanged |
| `test_native_runtime_stores_capture_false` | Flag is stored and propagated correctly |
| `test_pdb_flow_completes_with_no_capture` | Flow with `pdb.set_trace()` exits 0 when fed `c\n` |
| `test_pdb_flow_hangs_without_flag` | Confirms original bug: flow times out without flag |

---

## Backward Compatibility

- Default is `capture_output=True` — zero behaviour change for
  existing users and CI pipelines
- All cloud execution paths (Kubernetes, AWS Batch, Argo, etc.)
  are completely unaffected
- `--no-capture` is purely opt-in for local debugging sessions

### DIFF
--- a/metaflow/plugins/argo/argo_workflows_cli.py
+++ b/metaflow/plugins/argo/argo_workflows_cli.py
@@ -44,7 +44,7 @@ unsupported_decorators = {
     "slurm": "Step *%s* is marked for execution on Slurm with Argo Workflows which isn't currently supported.",
     "nvidia": "Step *%s* is marked for execution on Nvidia with Argo Workflows which isn't currently supported.",
     "nvct": "Step *%s* is marked for execution on Nvct with Argo Workflows which isn't currently supported.",
-    "skypilot": "Step *%s* is marked for execution on Skypilot with Argo Workflows which isn't currently supported.",
+    "skypilot_step": "Step *%s* is marked for execution on Skypilot with Argo Workflows which isn't currently supported.",
 }
 
 
--- a/metaflow/plugins/datastores/s3_storage.py
+++ b/metaflow/plugins/datastores/s3_storage.py
@@ -78,12 +78,12 @@ class S3Storage(DataStoreStorage):
     def save_bytes(self, path_and_bytes_iter, overwrite=False, len_hint=0):
         def _convert():
             # Output format is the same as what is needed for S3PutObject:
-            # key, value, path, content_type, metadata
+            # key, value, path, content_type, encryption, metadata
             for path, obj in path_and_bytes_iter:
                 if isinstance(obj, tuple):
-                    yield path, obj[0], None, None, obj[1]
+                    yield path, obj[0], None, None, None, obj[1]
                 else:
-                    yield path, obj, None, None, None
+                    yield path, obj, None, None, None, None
 
         with S3(
             s3root=self.datastore_root,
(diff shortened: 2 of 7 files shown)


npow:
please read our updated guidance before opening PRs

vikash7485:
Sure, I will go through the updated guidance before opening the PR. Thanks for informing.

PR #2947, opened 2026-03-02
Add tags, user_tags, and system_tags to the current singleton

### LINKED ISSUE #1007: Add tags to the `current` singleton
Currently, it is not possible to get the current tags from the `current` singleton. We could add them.

adity1raut:
## PR Type

- [x] New feature

## Summary

Adds `current.tags`, `current.user_tags`, and `current.system_tags` properties
to the `current` singleton, making run tags accessible at runtime inside a step.
The interface mirrors the existing `Run.tags` / `Run.user_tags` / `Run.system_tags`
client-side API.

## Issue

Fixes #1007

## Root Cause

The `current` singleton had a `tags` property that was marked
`[Legacy function - do not use]` and only stored user-defined sticky tags
(`sticky_tags`). System tags (`sticky_sys_tags`) were never passed to
`_set_env()`, so there was no way to distinguish user tags from system tags
at runtime, and the property was effectively undocumented/discouraged.

## Why This Fix Is Correct

- `_set_env()` now accepts both `tags` (user) and `sys_tags` (system),
  stored separately as `frozenset` — matching how the client API models them.
- `current.tags` returns `user_tags | system_tags` for convenience, consistent
  with `Run.tags`.
- `task.py` already had access to both `sticky_tags` and `sticky_sys_tags` on
  the metadata object; passing the latter was a one-line addition.
- No existing callers of `_set_env()` are broken: `sys_tags` defaults to
  `None`, and all other call sites use keyword arguments or only set a subset
  of fields.

## Failure Modes Considered

1. **Backward compatibility** — The old `tags` property previously returned a
   raw `set`; it now returns a `frozenset`. Code that only iterates or tests
   membership is unaffected. Code that called `.update()` on the result would
   have already been broken (frozensets are immutable), but since the property
   was explicitly marked "do not use", no such callers are expected.
2. **Missing sys_tags** — If `sticky_sys_tags` is empty (e.g. a bare local run
   with no decorators), `sys_tags` defaults to `frozenset()`, so `current.tags`
   still returns the correct user tags without raising.

## Tests

- [x] Updated `test/core/tests/current_singleton.py` to capture and assert
  `current.user_tags` and `current.system_tags` in every step type (start,
  join, all), and to verify `tags == user_tags | system_tags`
- [x] CI passes

## Non-Goals

- Did not add `current.add_tag()` / mutation of tags at runtime — that is a
  separate feature request.
- Did not change how tags are stored or propagated in the metadata provider;
  only changed what the `current` singleton surfaces.

### DIFF
--- a/metaflow/metaflow_current.py
+++ b/metaflow/metaflow_current.py
@@ -63,9 +63,9 @@ class Current(object):
         self._username = username
         self._metadata_str = metadata_str
         self._is_running = is_running
-        self._user_tags = frozenset(tags) if tags else frozenset()
-        self._system_tags = frozenset(sys_tags) if sys_tags else frozenset()
-
+        self._user_tags = frozenset(tags) if tags is not None else frozenset()
+        self._system_tags = frozenset(sys_tags) if sys_tags is not None else frozenset()
+        
     def _update_env(self, env):
         for k, v in env.items():
             setattr(self.__class__, k, property(fget=lambda _, v=v: v))


adity1raut:
@npow  Can You Please Give me a review on this PR

npow:

PR #2955, opened 2026-03-02
fix(client): add IPython tab completion for MetaflowData and Metaflow

### LINKED ISSUE #1114: Support auto-completion of Metaflow objects/API in notebooks

GunaPalanivel:
## PR Type

- [x] Bug fix

## Summary

- Adds IPython tab completion for `task.data.<TAB>` (artifact names) and `Metaflow()["<TAB>` (flow names), fixing two gaps left after PR #1348.

## Issue

- Fixes #1114  
- Relates to: https://github.com/npow/metaflow-jupyter/issues/4

## Reproduction

**Runtime:** local

**Commands to run:**
```bash
pip install -e .
ipython
```

**Where evidence shows up:** IPython interactive shell

```
'my_var' in dir(task.data)                        # False
hasattr(Metaflow(), '_ipython_key_completions_')  # False
```

```
dir(task.data)                           # [..., 'my_other_var', 'my_var', 'name']
task.data._ipython_key_completions_()    # ['my_var', 'name', 'my_other_var']
Metaflow()._ipython_key_completions_()   # ['HelloFlow', 'DummyFlow']
```

[screenshots attached]
Screenshot 2026-03-03 013000

Screenshot 2026-03-03 013013

## Root Cause

`MetaflowData` had no `__dir__` override so IPython couldn't discover artifact names. `Metaflow` inherits `object` not `MetaflowObject`, so it never received `_ipython_key_completions_`.

## Why This Fix Is Correct

- `MetaflowData.__dir__` merges `super().__dir__()` with `self._artifacts.keys()`
- `MetaflowData._ipython_key_completions_` returns the same artifact keys
- `Metaflow._ipython_key_completions_` returns `[f.id for f in self]` wrapped in `try/except` to fail silently if metadata isn't initialized

## Failure Modes Considered

1. **Internal artifact leakage:** `Task._iter_filter` already excludes `_`-prefixed artifacts before `MetaflowData` is constructed — no additional guard needed
2. **Metadata crash during completion:** Wrapped in `try/except Exception` so IPython returns `[]` silently instead of crashing the REPL

## Tests

- [x] Unit tests added — `test/unit/test_client_core.py`
- [x] CI passes

## Non-Goals

Did not change `MetaflowData` inheritance — keeping it a lightweight container, not pulling it into the `MetaflowObject` hierarchy.

## AI Tool Usage

- [x] AI tools were used

Perplexity AI — used for planning the implementation approach and drafting the PR description. 

> All code was written, reviewed, and manually verified by me.

### DIFF
--- a/test/unit/test_client_core.py
+++ b/test/unit/test_client_core.py
@@ -41,6 +41,62 @@ class TestMetaflowCompletions(unittest.TestCase):
         keys = mf_mock._ipython_key_completions_()
         self.assertEqual(set(keys), {"flow_1", "flow_2"})
 
+    def test_metaflow_data_dir_includes_artifacts(self):
+        """Regression test for #1114: artifact names must appear in dir(MetaflowData).
+
+        Before the fix, artifact names were NOT visible in dir(task.data),
+        breaking IPython tab completion.
+        """
+        artifacts = [
+            MockDataArtifact("my_var"),
+            MockDataArtifact("name"),
+            MockDataArtifact("my_other_var"),
+        ]
+        data = MetaflowData(artifacts)
+
+        # Pre-fix: 'my_var', 'name', 'my_other_var' were NOT in dir(data)
+        dir_output = dir(data)
+        self.assertIn(
+            "my_var",
+            dir_output,
+            "Artifact names must appear in dir() for IPython tab completion",
+        )
+        self.assertIn("name", dir_output)
+        self.assertIn("my_other_var", dir_output)
+
+    def test_metaflow_data_empty_artifacts(self):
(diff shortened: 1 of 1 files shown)


npow:
Duplicate of #3001 which covers the same IPython tab completion changes. Closing in favor of that PR.

Issue #2960, opened 2026-03-03
Optimize FlowSpec._get_parameters() to avoid dir() + getattr() full attribute scan in flowspec.py


Kaustubh2k5:
### FlowSpec._get_parameters() currently discovers parameters by iterating over` dir(cls)` and using `getattr() `to inspect each attribute.

While the current implementation is correct, it performs a full reflective scan of the class using `dir()` and `getattr()`. Since parameters are class-level definitions, it may be cleaner and more deterministic to walk the MRO and inspect` __dict__` directly.

### Proposed Changes:
Instead of scanning all attributes visible on the class, we only scan the class-level attributes that are actually relevant for parameter discovery. Instead of scanning via `dir(cls)`, walk the class MRO (Method Resolution Order) and inspect `base.__dict__ `for each class:

- Iterate over `reversed(cls.__mro__)`
- Inspect only attributes defined directly on each class
- Filter Parameter instances

Preserve caching and inheritance behavior

**Before:**

```
        build_list = []
        for set_config in cls._flow_state[FlowStateItems.SET_CONFIG_PARAMETERS]:
            returned.add(set_config[0])
            yield set_config[0], set_config[1]
        for var in dir(cls):
            if var[0] == "_" or var in cls._NON_PARAMETERS:
                continue
            try:
                val = getattr(cls, var)
            except:
                continue
            if isinstance(val, Parameter) and var not in returned:
                build_list.append(var)
                yield var, val
        cls._flow_state[FlowStateItems.CACHED_PARAMETERS] = build_list

```

**Proposed change:**

```
        for set_config in cls._flow_state[FlowStateItems.SET_CONFIG_PARAMETERS]:
            returned.add(set_config[0])
            yield set_config[0], set_config[1]

        for base in reversed(cls.__mro__):
            if not issubclass(base, FlowSpec):
                continue

            for var, val in base.__dict__.items():
                if (
                    var[0] != "_"
                    and var not in cls._NON_PARAMETERS
                    and isinstance(val, Parameter)
                    and var not in returned
                ):
                    build_list.append(var)
                    returned.add(var)
                    yield var, val

        cls._flow_state[FlowStateItems.CACHED_PARAMETERS] = build_list
```
### What this does:

It would:

- Preserve parameter discovery semantics.
- Preserve config parameter handling.
- Preserve inheritance behavior.
- Preserve caching logic.

**It would only change the internal mechanism used to discover parameters. **

gursishkaur28:
[...]

PR #2963, opened 2026-03-03
perf: Optimize FlowSpec._get_parameters() to use MRO walk

Ask-812:
## Summary
Addresses #2960

This PR optimizes `FlowSpec._get_parameters()` by replacing the `dir()` + `getattr()` pattern with a direct MRO (Method Resolution Order) walk.

## Changes
- Walk `reversed(cls.__mro__)` and inspect `base.__dict__` directly instead of using `dir(cls)` + `getattr()`
- More deterministic behavior - avoids full reflective scan overhead
- Preserves existing caching, inheritance behavior, and config parameter handling

## Before
```python
for attr_name in dir(cls):
    attr = getattr(cls, attr_name)
    if isinstance(attr, Parameter):
        parameters.append(attr)
```

## After
```python
for base in reversed(cls.__mro__):
    if base is object:
        continue
    for attr_name, attr in base.__dict__.items():
        if isinstance(attr, Parameter) and attr_name not in seen:
            seen.add(attr_name)
            parameters.append(attr)
```

## Benefits
- Reduces reflective property access overhead
- More predictable parameter discovery order
- Respects MRO properly without invoking descriptors via `getattr()`

## Testing
The change preserves all existing behavior and has been tested with the existing test suite.

### DIFF
--- a/metaflow/flowspec.py
+++ b/metaflow/flowspec.py
@@ -564,8 +564,9 @@ class FlowSpec(metaclass=FlowSpecMeta):
         # This avoids a full reflective scan of the class and is more deterministic.
         # Iterate cls.__mro__ (not reversed) so derived classes are visited first,
         # matching Python's standard attribute resolution order.
+        # Check all classes (not just FlowSpec subclasses) to support mixin-based parameters.
         for base in cls.__mro__:
-            if not issubclass(base, FlowSpec):
+            if base is object:
                 continue
             for var, val in base.__dict__.items():
                 if (


Ask-812:
Thanks for the review! Fixed the MRO traversal order - removed \
eversed()\ so derived classes are visited first, matching Python's standard attribute resolution. This ensures parameter overrides in subclasses work correctly and maintains consistency between cached and uncached paths.

Ask-812:
Good catch on the mixin concern! I've updated the code to check all classes in the MRO (not just FlowSpec subclasses), skipping only \object\. This preserves the original \dir()\ behavior for mixin-based parameters while keeping the performance optimization.

npow:

PR #2970, opened 2026-03-05
Add initial support for supporting runtime values in decorators

romain-intel:
Support would need to be added in the orchestrators as well but this PR starts the framework for it.

## PR Type

- [ ] Bug fix
- [X ] New feature
- [ ] Docs / tooling
- [ ] Refactoring

## Summary

Provides initial framework to support dynamic runtime values in decorators.

This is not complete support.

### DIFF
--- a/test/ux/core/test_dynamic_vars.py
+++ b/test/ux/core/test_dynamic_vars.py
@@ -3,7 +3,7 @@ Dynamic vars (var()) tests.
 
 Tests cover:
   1. Runner-mode execution — basic var(), pertask list, pertask dict, default, multi-var.
-  2. Compilation rejection — check_no_dynamic_vars rejects var() on non-Maestro schedulers.
+  2. Compilation rejection — check that things can't be pushed to schedulers for now
 
 Run with:
     pytest test/ux/core/test_dynamic_vars.py -m dynamic_vars -v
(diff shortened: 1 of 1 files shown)


metaflow/dynamic_var.py
```diff
@@ -0,0 +1,294 @@
...
+    If the store has not been populated (outermost process), this is a no-op and
+    DynamicVar will remain
+    """
+    if _resolved_store is None:
+        return attributes
+    return _resolve_val(attributes, _resolved_store, _resolved_split_index)
+
+
+class DynamicVar:
+    """Marker for decorator attributes resolved from parent step artifacts at runtime.
```
saikonen, line 62: how strict are we with the notion of these being only from parent step artifacts? I had a quick poke trying this on a `start` step, and as expected it errors out due to no parent step existing (though `_parameters` is a thing)

Wouldn't something like this be a plausible use case for the feature as well?
```python
class HelloFlow(FlowSpec):
    initial = Parameter("initial_val", default="initial")

    @environment(vars={"TEST": var("initial")})
    @step
    def start(self):
```
this would somewhat compete with `config_expr` though, and its not a dealbreaker by any means for the usability.

metaflow/dynamic_var.py
```diff
@@ -0,0 +1,294 @@
...
+    def __repr__(self):
+        parts = ["var('%s'" % self.var_name]
+        if self.pertask:
+            parts.append("pertask=True")
+        if self.default is not _NO_DEFAULT:
+            parts.append("default=%r" % (self.default,))
+        return ", ".join(parts) + ")"
+
+
+def var(var_name, pertask=False, default=_NO_DEFAULT):
```
saikonen, line 99: nit: `per_task` instead of `pertask` would read better.

PR #2974, opened 2026-03-06
Fix foreach join inputs out-of-order for >4 branches

npow:
## Summary

`get_task_datastores()` builds its return list by iterating a Python set, so inputs at foreach join steps arrived in arbitrary order rather than matching the original foreach list. Only affects foreach with >4 branches (the threshold where `TaskDataStoreSet` is used instead of sequential loading).

Fix: after loading datastores for a foreach join, sort by `_foreach_stack[-1].index` — the ground truth stored in each task. `_foreach_stack` is already prefetched in this code path so the sort is cheap.

Also removes `sorted()` from `BasicForeachTest` and `WideForeachTest` join checks, turning them into ordering assertions. `BasicForeachTest` now uses a shuffled (non-monotonic) 32-element array so that any partial-sort regression is immediately visible.

### DIFF
--- a/metaflow/task.py
+++ b/metaflow/task.py
@@ -271,13 +271,6 @@ class MetaflowTask(object):
                 spin_artifacts=self.spin_artifacts,
             )
             ds_list = [ds for ds in datastore_set]
-            if join_type == "foreach":
-                # get_task_datastores() iterates over a set internally, so the
-                # returned order is arbitrary. Sort by foreach index (the ground
-                # truth stored in the task itself) so that inputs always arrive
-                # at join steps in the same order as the original foreach list.
-                # _foreach_stack is already prefetched above, so this is cheap.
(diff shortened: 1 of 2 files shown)


metaflow/task.py
```diff
@@ -271,6 +271,13 @@ def _init_data(self, run_id, join_type, input_paths):
                 spin_artifacts=self.spin_artifacts,
             )
             ds_list = [ds for ds in datastore_set]
+            if join_type == "foreach":
```
romain-intel: I don't think this fix is exactly correct because it misses out on static > 4 branches (the same issue can happen and there is no re-sorting. What we *really* want is that the order of input_paths is respected. I think we may actually need to fix it in the TaskDataStoreSet. I quickly looked at the implementation of get_task_datastores and there is a set in there but we should be able to maintain order. The udnerlying S3 storage does return things properly in order (or rather, keyed properly) so we should be able to maintain this as well.

PR #2978, opened 2026-03-07
Add YAML support for Config files (JSON-first with YAML fallback)

### LINKED ISSUE #2977: [Feature Request] Config files only accept JSON — YAML format not supported
## Description

Metaflow's `Config` system accepts configuration files via the `--config` CLI option.
When no custom `parser` is provided, the content is parsed exclusively as **JSON**.
If the file contains valid YAML (which is not valid JSON), the parse silently fails with
an error like `"configuration value for '<name>' is not valid JSON"`.

This is a self-acknowledged gap in the codebase:

> `metaflow/user_configs/config_options.py` line 391
> ```python
> try:
>     read_value = json.loads(val)
> except json.JSONDecodeError as e:
>     msgs.append(
>         "configuration value for '%s' is not valid JSON: %s" % (name, e)
>     )
>     continue
> # TODO: Support YAML
> ```

Critically, **PyYAML is already vendored** inside the project at
`metaflow/_vendor/yaml/` (version 5.3.1) and is imported elsewhere — so adding
YAML support requires no new dependency.

---

## Motivation

YAML is the de-facto standard for ML configuration files (Hydra, OmegaConf,
HuggingFace Trainer, etc.). A typical ML team's config looks like:

```yaml
# model_config.yaml
model:
  layers: 6
  hidden_dim: 512
  dropout: 0.1

training:
  lr: 3e-4
  epochs: 50
  batch_size: 32
```

Writing the same thing in JSON is far less ergonomic — no comments, no
multi-document support, verbose syntax for nested structures. Forcing users
to maintain JSON configs when Metaflow already ships PyYAML creates unnecessary
friction, especially for GSoC/open-source contributors coming from an ML background.

---

## Steps to Reproduce

```python
# my_flow.py
from metaflow import FlowSpec, step, Config

class MyFlow(FlowSpec):
    cfg = Config("cfg")

    @step
    def start(self):
        print(self.cfg.model.layers)
        self.next(self.end)

    @step
    def end(self):
        pass

if __name__ == "__main__":
    MyFlow()
```

```yaml
# model_config.yaml
model:
  layers: 6
  hidden_dim: 512
```

```bash
python my_flow.py run --config cfg model_config.yaml
```

**Current error:**
```
Bad values passed for configuration options:
configuration value for 'cfg' is not valid JSON: ...
```

**Expected behavior:** The YAML file is parsed correctly and `self.cfg.model.layers`
returns `6`.

---

## Current Behavior

In `config_options.py` the parse decision path is:

```
value provided
    └── custom parser set? → call it
    └── is_plain=True?     → use as raw string
    └── else               → json.loads(val)   ← only JSON, YAML raises JSONDecodeError
                               # TODO: Support YAML
```

The `--config` option reads the full file content into a string via `ConvertPath`.
That string is then handed to the JSON parser. A `.yaml` file will always fail here
because YAML is a superset of JSON — a JSON-only parser rejects YAML-specific syntax
(comments, unquoted strings, block scalars, anchors, etc.).

---

## Expected Behavior

The config system should detect YAML content (by file extension or as a JSON fallback)
and parse it with PyYAML. The resulting dict should then be wrapped in `ConfigValue`
exactly as JSON configs are today — no change to downstream behavior.

Proposed parse order (no breaking change):

```
1. Try json.loads(val)             → success → use result
2. Try yaml.safe_load(val)         → success → use result
3. Raise error: "not valid JSON or YAML"
```

Or, if the value came from a `--config` file path, detect by extension:

```
.json              → json.loads
.yaml / .yml       → yaml.safe_load
other / --config-value → try JSON first, then YAML
```

---

## Why This Is Easy to Implement

PyYAML 5.3.1 is **already vendored** at `metaflow/_vendor/yaml/`. It is imported
in other parts of the codebase using:

```python
from metaflow._vendor import yaml
```

The change is localized to a single `try/except` block in
`metaflow/user_configs/config_options.py` around line 383–391:

```python
# Current
try:
    read_value = json.loads(val)
except json.JSONDecodeError as e:
    msgs.append(
        "configuration value for '%s' is not valid JSON: %s" % (name, e)
    )
    continue
# TODO: Support YAML

# Proposed
try:
    read_value = json.loads(val)
except json.JSONDecodeError:
    try:
        from metaflow._vendor import yaml
        read_value = yaml.safe_load(val)
        if not isinstance(read_value, dict):
            raise ValueError("YAML did not produce a mapping")
    except Exception as e:
        msgs.append(
            "configuration value for '%s' is not valid JSON or YAML: %s"
            % (name, e)
        )
        continue
```

`yaml.safe_load` is used (not `yaml.load`) to prevent arbitrary code execution
from untrusted config files.

---

## Impact

| Without fix | With fix |
|---|---|
| Users must write config files in JSON | Users can write config files in YAML or JSON |
| YAML files silently produce a confusing JSON parse error | YAML files are parsed correctly |
| ML practitioners must convert all their YAML configs manually | Existing YAML configs work out of the box |
| `# comments` not allowed in config files | Comments work in `.yaml` configs |

---

## Affected Files

| File | Line | Note |
|------|------|------|
| `metaflow/user_configs/config_options.py` | L383–391 | Main parse path — add YAML fallback |
| `metaflow/user_configs/config_options.py` | `ConvertPath.convert_value` L65 | Optionally detect extension for `.yaml`/`.yml` |

---

## Out of Scope for This Issue

- YAML support for `--config-value` inline strings (ambiguous; JSON is clearer
  for inline values — can be a follow-up).
- Supporting multi-document YAML (e.g. `---` separator) — `Config` maps to a
  single dict, so only the first document should be used.
- OmegaConf / Hydra interpolation syntax — `yaml.safe_load` is sufficient.

---

adity1raut:
## PR Type

- [ ] Bug fix
- [x] New feature
- [ ] Docs / tooling
- [ ] Refactoring

## Summary

Config files passed via `--config` are now parsed as YAML if JSON parsing fails.
Users can pass `.yaml` / `.yml` config files directly with no code changes — the
resulting `ConfigValue` object behaves identically to one parsed from JSON.

## Issue

Fixes #2977 

## Reproduction

**Runtime:** local

**Commands to run:**
```bash
cat > /tmp/my_flow.py << 'EOF'
from metaflow import FlowSpec, step, Config

class MyFlow(FlowSpec):
    cfg = Config("cfg")

    @step
    def start(self):
        print(self.cfg.model.layers)   # should print 6
        self.next(self.end)

    @step
    def end(self):
        pass

if __name__ == "__main__":
    MyFlow()
EOF

cat > /tmp/model_config.yaml << 'EOF'
# YAML comments now work!
model:
  layers: 6
  hidden_dim: 512
  dropout: 0.1
EOF

python /tmp/my_flow.py run --config cfg /tmp/model_config.yaml
```

**Where evidence shows up:** task stdout

<details>

```
Bad values passed for configuration options:
configuration value for 'cfg' is not valid JSON:
Expecting value: line 1 column 1 (char 0)
```

</details>

<details>

```
6
```

</details>

## Root Cause

The config parse path in `ConfigInput.process_configs` called `json.loads(val)` with
no fallback. The original code had a `# TODO: Support YAML` comment immediately after
the `except json.JSONDecodeError` block, acknowledging the gap:

```python
# metaflow/user_configs/config_options.py  (before)
try:
    read_value = json.loads(val)
except json.JSONDecodeError as e:
    msgs.append(
        "configuration value for '%s' is not valid JSON: %s" % (name, e)
    )
    continue
# TODO: Support YAML
```

Any valid YAML that is not simultaneously valid JSON (files with `#` comments,
unquoted strings, block scalars, anchors) always hit the error path and prevented
the flow from starting.

## Why This Fix Is Correct

- **No new dependency.** PyYAML 5.3.1 is already vendored at `metaflow/_vendor/yaml/`
  and is imported in several other modules (`includefile.py`, `plugins/parsers.py`,
  `plugins/kubernetes/kubernetes_jobsets.py`). The import pattern used here is
  identical to existing usage.

- **No breaking change.** JSON parsing is still tried first. Valid JSON files
  continue to work exactly as before — the YAML path is only reached on
  `JSONDecodeError`.

- **Safe parsing.** `yaml.safe_load` is used instead of `yaml.load` to prevent
  arbitrary Python object construction from untrusted config files.

- **Type checked.** After YAML parsing, the result is verified with
  `isinstance(read_value, Mapping)` — the same contract that JSON configs satisfy —
  so a YAML file that produces a scalar or list gets a clear, actionable error.

- **Minimal diff.** One import line updated and one `except` block extended.
  All downstream `ConfigValue` wrapping and `FlowStateItems.CONFIGS` storage
  is completely unchanged.

```python
# metaflow/user_configs/config_options.py  (after)
try:
    read_value = json.loads(val)
except json.JSONDecodeError:
    # JSON parse failed — try YAML (a superset of JSON).
    # yaml.safe_load is used intentionally to prevent
    # arbitrary code execution from untrusted config files.
    try:
        read_value = yaml.safe_load(val)
        if not isinstance(read_value, Mapping):
            raise ValueError(
                "expected a mapping (got %s)" % type(read_value).__name__
            )
    except Exception as e:
        msgs.append(
            "configuration value for '%s' is not valid "
            "JSON or YAML: %s" % (name, e)
        )
        continue
```

## Failure Modes Considered

1. **YAML file that produces a scalar (`42`, `"hello"`)** — `isinstance(read_value, Mapping)`
   raises `ValueError`; caught by outer `except Exception`, appended to `msgs` with a
   clear message. Flow does not start.

2. **YAML file with syntax errors** — `yaml.safe_load` raises `yaml.YAMLError`;
   caught by outer `except Exception`, reported as `"not valid JSON or YAML: ..."`.
   No crash.

3. **YAML anchors / aliases (`&anchor`, `*alias`)** — `safe_load` supports these
   safely (they are standard YAML, not Python-specific). Resolves correctly.

4. **Existing JSON configs** — `json.loads` succeeds on the first try; YAML branch
   is never reached. Zero behavioral change.

## Tests

- [ ] CI passes

**Manual evidence:** Run the reproduction script above.
Before: `Bad values passed for configuration options: ... is not valid JSON`.
After: step prints `6` and the flow completes successfully.

## Non-Goals

- YAML support for `--config-value` inline values (ambiguous for short strings;
  JSON is clearer for inline use — can be a follow-up).
- Multi-document YAML (`---` separator) — `Config` maps to a single dict;
  `yaml.safe_load` already returns only the first document.
- OmegaConf / Hydra variable interpolation — out of scope for this PR.

## Files Changed

| File | Lines | Change |
|------|-------|--------|
| `metaflow/user_configs/config_options.py` | L7 | Added `yaml` to existing vendor import |
| `metaflow/user_configs/config_options.py` | L383–401 | JSON-first, YAML-fallback parse block, updated error message |

### DIFF
--- a/metaflow/user_configs/config_options.py
+++ b/metaflow/user_configs/config_options.py
@@ -4,7 +4,7 @@ import os
 from collections.abc import Mapping
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
-from metaflow._vendor import click
+from metaflow._vendor import click, yaml
 from metaflow.debug import debug
 
 from .config_parameters import ConfigValue
@@ -382,13 +382,23 @@ class ConfigInput:
                     else:
                         try:
                             read_value = json.loads(val)
-                        except json.JSONDecodeError as e:
-                            msgs.append(
-                                "configuration value for '%s' is not valid JSON: %s"
-                                % (name, e)
-                            )
-                            continue
-                        # TODO: Support YAML
+                        except json.JSONDecodeError:
+                            # JSON parse failed — try YAML (a superset of JSON).
+                            # yaml.safe_load is used intentionally to prevent
+                            # arbitrary code execution from untrusted config files.
+                            try:
+                                read_value = yaml.safe_load(val)
+                                if not isinstance(read_value, Mapping):
+                                    raise ValueError(
+                                        "expected a mapping (got %s)"
+                                        % type(read_value).__name__
+                                    )
+                            except Exception as e:
+                                msgs.append(
+                                    "configuration value for '%s' is not valid "
+                                    "JSON or YAML: %s" % (name, e)
+                                )
+                                continue
                 flow_cls._flow_state.self_data[FlowStateItems.CONFIGS][name] = (
                     read_value,
                     is_plain,


npow:

PR #2983, opened 2026-03-07
Fix @catch decorator to support foreach split steps

### LINKED ISSUE #2982: [Bug / Limitation] `@catch` cannot be used on `foreach` split steps — major error-handling gap
## Description

The `@catch` decorator is explicitly rejected at flow initialization time when applied
to a `foreach` step. This blocks the most common use-case for fault tolerance in
data pipelines: processing a list of items where individual items may fail.

**`metaflow/plugins/catch_decorator.py` lines 44–53:**
```python
def step_init(self, flow, graph, step, decos, environment, flow_datastore, logger):
    # handling _foreach_var and _foreach_num_splits requires some
    # deeper thinking, so let's not support that use case for now
    self.logger = logger
    if graph[step].type == "foreach":
        raise MetaflowException(
            "@catch is defined for the step *%s* "
            "but @catch is not supported in foreach "
            "split steps." % step
        )
```

The comment `"requires some deeper thinking, so let's not support that use case for now"`
indicates this was a deliberate deferral, not a fundamental impossibility.
`@catch` is also blocked on `split-switch` (conditional switch) steps for a similar reason.

---

## Steps to Reproduce

```python
# foreach_flow.py
from metaflow import FlowSpec, step, catch, retry

class ForeachFlow(FlowSpec):

    @step
    def start(self):
        self.items = ["file_a.csv", "file_b.csv", "bad_file.csv", "file_d.csv"]
        self.next(self.process, foreach="items")

    @catch(var="process_error")   # <-- users expect this to work
    @retry(times=2)
    @step
    def process(self):
        item = self.input
        # simulate failure for one item
        if "bad" in item:
            raise ValueError("Cannot process %s" % item)
        self.result = item.upper()
        self.next(self.join)

    @step
    def join(self, inputs):
        # users want to inspect which items failed
        self.results = [inp.result for inp in inputs if inp.process_error is None]
        self.errors  = [inp.process_error for inp in inputs if inp.process_error is not None]
        self.next(self.end)

    @step
    def end(self):
        pass

if __name__ == "__main__":
    ForeachFlow()
```

```bash
python foreach_flow.py run
```

**Runtime:** local / Kubernetes / Batch

**Where evidence shows up:** flow initialization error before any task runs

<details>
<summary>Before (error)</summary>

```
Metaflow Exception:
@catch is defined for the step *process* but @catch is not supported in foreach split steps.
```

The flow never starts. Users must remove `@catch` and implement try/except manually
inside the step body — which loses all the benefits of Metaflow's retry and catch
infrastructure.

</details>

<details>
<summary>After (expected behavior)</summary>

```
Metaflow [1/process/1 (pid 1234)] Task is starting.
Metaflow [1/process/2 (pid 1235)] Task is starting.
Metaflow [1/process/3 (pid 1236)] Task is starting.      ← "bad_file.csv"
Metaflow [1/process/4 (pid 1237)] Task is starting.
Metaflow [1/process/3] @catch caught an exception from process
Metaflow [1/process/3] >  ValueError: Cannot process bad_file.csv
Metaflow [1/process/3] Task finished successfully.        ← caught, flow continues

# In the join step:
# inputs[2].process_error  → MetaflowExceptionWrapper(ValueError(...))
# inputs[0,1,3].process_error → None
```

</details>

---

## Current Behavior

`CatchDecorator.step_init` raises `MetaflowException` unconditionally for any step
whose graph node type is `"foreach"`:

```python
if graph[step].type == "foreach":
    raise MetaflowException(
        "@catch is defined for the step *%s* "
        "but @catch is not supported in foreach "
        "split steps." % step
    )
```

This means users who want fault tolerance on a foreach step must:

1. Wrap their entire step body in a `try/except`.
2. Store the exception manually as an artifact.
3. Check that artifact in the join step.
4. Handle retries manually (since `@retry` without `@catch` will fail the entire flow
   if all retries are exhausted).

This recreates from scratch exactly what `@catch` is supposed to provide.

---

## Expected Behavior

`@catch` should work on foreach steps, catching exceptions **per split** and storing
the exception in the `var` artifact for that split's task. The join step downstream
can then inspect `inputs[i].var` to determine which splits failed and which succeeded.

This is already how users reason about the problem — the limitation is purely
internal to `CatchDecorator`.

### Why the current restriction exists

The comment in the code cites `_foreach_var` and `_foreach_num_splits` handling as
the blocker. These are internal artifacts that the runtime uses to track foreach
state. When `@catch` intercepts an exception, it calls:

```python
flow._transition = (graph[step].out_funcs, None)
```

For a foreach step, this transition also needs `_foreach_var` and
`_foreach_num_splits` to be set correctly so the join step can receive all splits.
If an exception is caught before `self.next(self.end)` is called, these internal
artifacts may not be set.

### Proposed fix approach

When `@catch` handles an exception in a foreach step, it should additionally set
`_foreach_var` to the value of the foreach input (`self.input`) and
`_foreach_num_splits` to the total number of splits (available from
`flow._foreach_values`). This ensures the join step receives the expected number of
inputs regardless of whether individual splits succeeded or failed.

---

## Root Cause

The restriction was added as a deliberate short-term deferral
(`"let's not support that use case for now"`) while the `@catch` decorator was
being implemented. The foreach internal artifacts (`_foreach_var`,
`_foreach_num_splits`) need to be explicitly set in the catch fallback path to
keep the join step's fan-in accounting correct.

---

## Impact

| Scenario | Effect without fix |
|---|---|
| `foreach` over list of files, one fails | Entire flow fails; no partial results saved |
| `@catch` + `@retry` on foreach step | `MetaflowException` at startup — cannot be used together |
| Fault-tolerant batch processing pipelines | Users must implement error handling manually in every step |
| `foreach` over 1000 items, 1 fails | All 1000 tasks wasted; full re-run required |

`foreach` + `@catch` is arguably the single most common pattern in production
data pipelines that Metaflow targets. The restriction forces a painful workaround
for a very natural use case.

---

## Affected Files

| File | Lines | Issue |
|------|-------|-------|
| `metaflow/plugins/catch_decorator.py` | L44–53 | `foreach` check that raises `MetaflowException` |
| `metaflow/plugins/catch_decorator.py` | L74–97 | `task_exception` — needs to set `_foreach_var` for foreach steps |
| `metaflow/plugins/catch_decorator.py` | L108–121 | `task_decorate` — fallback step for foreach needs foreach artifacts |
| `metaflow/graph.py` | (reference) | `DAGNode.type == "foreach"` definition |

---

## Out of Scope

- `@catch` on `split-switch` (conditional switch) steps — separate issue, different
  root cause (the condition attribute must be recorded correctly).
- Changing `@catch` behavior for non-foreach steps — no change needed.

---

## Environment

- **OS:** Linux / macOS / Windows (all runtimes affected)
- **Metaflow version:** current `master`
- **Python:** 3.8+

---

adity1raut:
## PR Type

- [x] Bug fix
- [ ] New feature
- [ ] Docs / tooling
- [ ] Refactoring

## Summary

Remove the hard `MetaflowException` that prevented `@catch` from being used on
`foreach` split steps. `@catch` now works on foreach steps when `self.next()` was
already reached; it safely declines to catch when `self.next()` was not yet called
(so the runtime split count stays well-defined).

## Issue

Fixes #2982

## Reproduction

**Runtime:** local / Kubernetes / Batch

**Commands to run:**
```bash
cat > /tmp/foreach_catch_flow.py << 'EOF'
from metaflow import FlowSpec, step, catch, retry

class ForeachCatchFlow(FlowSpec):

    @step
    def start(self):
        self.items = ["ok_a", "ok_b", "bad_item", "ok_d"]
        self.next(self.process, foreach="items")

    @catch(var="process_error")
    @retry(times=1)
    @step
    def process(self):
        if "bad" in self.input:
            raise ValueError("Cannot process: %s" % self.input)
        self.result = self.input.upper()
        self.next(self.join)

    @step
    def join(self, inputs):
        self.results = [i.result for i in inputs if i.process_error is None]
        self.errors  = [str(i.process_error) for i in inputs if i.process_error]
        print("Results:", self.results)
        print("Errors: ", self.errors)
        self.next(self.end)

    @step
    def end(self):
        pass

if __name__ == "__main__":
    ForeachCatchFlow()
EOF

python /tmp/foreach_catch_flow.py run
```

**Where evidence shows up:** parent console / task logs

<details>

```
Metaflow Exception:
@catch is defined for the step *process* but @catch is not supported
in foreach split steps.

The flow never starts.
```

</details>

<details>

```
Metaflow [1/process/1] Task is starting.
Metaflow [1/process/2] Task is starting.
Metaflow [1/process/3] Task is starting.
Metaflow [1/process/4] Task is starting.
Metaflow [1/process/3] @catch caught an exception from process
Metaflow [1/process/3] >  ValueError: Cannot process: bad_item
Metaflow [1/process/3] Task finished successfully.
Metaflow [1/join/1]    Task is starting.

Results: ['OK_A', 'OK_B', 'OK_D']
Errors:  ["ValueError('Cannot process: bad_item')"]
```

</details>

## Root Cause

`CatchDecorator.step_init` contained an unconditional guard that raised
`MetaflowException` for any step whose graph node type was `"foreach"`:

```python
# catch_decorator.py (before)
def step_init(self, flow, graph, step, decos, environment, flow_datastore, logger):
    # handling _foreach_var and _foreach_num_splits requires some
    # deeper thinking, so let's not support that use case for now
    self.logger = logger
    if graph[step].type == "foreach":
        raise MetaflowException(
            "@catch is defined for the step *%s* "
            "but @catch is not supported in foreach "
            "split steps." % step
        )
```

The comment acknowledged this was a deliberate deferral. The actual blocker was
that `task_exception` set `flow._transition = (out_funcs, None)` for all steps,
but for a foreach split step the runtime needs the transition to carry the foreach
variable name as its second element (so it knows how many child tasks to fan out).
Additionally, if the exception fires before `self.next(foreach=...)` is called,
`flow._foreach_num_splits` stays `None` — causing `runtime.py:1310`
(`num_splits = task.results["_foreach_num_splits"]`) to fail with a `TypeError`.

## Why This Fix Is Correct

**`step_init`** — The hard `MetaflowException` is removed. Instead, a flag
`self._step_is_foreach` is stored so `task_exception` can branch correctly.
The `split-switch` restriction is unchanged.

**`task_exception`** — Two cases are now handled:

1. **Exception before `self.next()` was called** (`flow._foreach_num_splits is None`):
   The split count is unknown; the runtime would crash trying to fan out.
   `@catch` returns `False` — the exception is not caught and the task fails normally.
   This is the safe, conservative path.

2. **Exception after `self.next()` was already called** (`flow._foreach_num_splits`
   is set): `self.next(foreach="items")` already recorded `_foreach_var`,
   `_foreach_num_splits`, and (if enabled) `_foreach_values` on the flow.
   `@catch` catches the exception and sets
   `flow._transition = (out_funcs, flow._foreach_var)` — preserving the foreach
   second-element so the runtime fans out the correct number of child tasks.
   The exception is stored in `var` for each downstream join input to inspect.

```python
# catch_decorator.py (after) — key changes
def step_init(self, flow, graph, step, decos, environment, flow_datastore, logger):
    self.logger = logger
    self._step_is_foreach = graph[step].type == "foreach"
    # split-switch restriction unchanged ...

def task_exception(self, exception, step, flow, graph, retry_count, max_user_code_retries):
    if retry_count < max_user_code_retries:
        return False

    # Safe gate: if self.next() was never reached, the split count is unknown.
    if self._step_is_foreach and flow._foreach_num_splits is None:
        return False

    ...

    # Preserve the foreach transition so the runtime fans out correctly.
    if self._step_is_foreach:
        flow._transition = (graph[step].out_funcs, flow._foreach_var)
    else:
        flow._transition = (graph[step].out_funcs, None)
    ...
```

## Failure Modes Considered

1. **Exception fires before `self.next(foreach=...)` is called** — `@catch` returns
   `False`; the task fails normally. The runtime does not try to read a `None`
   `_foreach_num_splits`. No crash, no silent data loss.

2. **Exception fires after `self.next()` is called** — `_foreach_var` and
   `_foreach_num_splits` are both set. `@catch` catches the exception and sets the
   correct foreach transition. The runtime fans out exactly as many child tasks as
   were planned. Child tasks see the parent's `_foreach_var` and
   `_foreach_num_splits` unchanged. The join step receives all N inputs.

3. **`@catch` + `@retry` on a foreach step** — retries run first (controlled by the
   existing `retry_count < max_user_code_retries` guard at the top of
   `task_exception`). `@catch` only intercepts after all retries are exhausted.
   Retry count and split state are independent — no interference.

4. **Non-foreach steps** — `self._step_is_foreach` is `False`; all existing code
   paths are taken unchanged. Zero behavioral change for linear, join, split, and
   start steps.

5. **UBF / parallel steps** — the existing `hasattr(flow, "_parallel_ubf_iter")`
   branch is preserved and runs after the foreach branch. No change.

## Tests

- [ ] CI passes

**Manual evidence:** Run the reproduction script above.
Before: `MetaflowException` at startup, flow never runs.
After: flow completes; failed splits are captured in `process_error`; successful
splits produce results; join step sees all N inputs.

## Non-Goals

- `@catch` on `split-switch` (conditional switch) steps — separate issue, different
  root cause (the condition attribute must also be recorded). Restriction unchanged.
- Catching exceptions that fire before `self.next()` in foreach steps — the split
  count is fundamentally unknowable in that case without deeper runtime changes.
- Any change to `@catch` behavior for non-foreach steps.

## Files Changed

| File | Lines | Change |
|------|-------|--------|
| `metaflow/plugins/catch_decorator.py` | L44–55 | Removed `foreach` hard block; added `_step_is_foreach` flag |
| `metaflow/plugins/catch_decorator.py` | L74–92 | Added foreach-aware guard and transition in `task_exception` |

### DIFF
--- a/metaflow/plugins/catch_decorator.py
+++ b/metaflow/plugins/catch_decorator.py
@@ -42,15 +42,8 @@ class CatchDecorator(StepDecorator):
     defaults = {"var": None, "print_exception": True}
 
     def step_init(self, flow, graph, step, decos, environment, flow_datastore, logger):
-        # handling _foreach_var and _foreach_num_splits requires some
-        # deeper thinking, so let's not support that use case for now
         self.logger = logger
-        if graph[step].type == "foreach":
-            raise MetaflowException(
-                "@catch is defined for the step *%s* "
-                "but @catch is not supported in foreach "
-                "split steps." % step
-            )
+        self._step_is_foreach = graph[step].type == "foreach"
 
         # Do not support catch on switch steps for now.
         # When applying @catch to a switch step, we can not guarantee that the flow attribute used for the switching condition gets properly recorded.
@@ -78,11 +71,25 @@ class CatchDecorator(StepDecorator):
         if retry_count < max_user_code_retries:
             return False
 
+        # For foreach split steps, @catch can only intervene when self.next()
+        # was already reached (i.e., _foreach_num_splits is set and _foreach_var
+        # is known). If the exception fired before self.next() was called, the
+        # split count is unknown and the runtime scheduler cannot fan out child
+        # tasks — in that case we decline to catch and let the task fail normally.
+        if self._step_is_foreach and flow._foreach_num_splits is None:
+            return False
+
         if self.attributes["print_exception"]:
             self._print_exception(step, flow)
 
-        # pretend that self.next() was called as usual
-        flow._transition = (graph[step].out_funcs, None)
+        # pretend that self.next() was called as usual.
+        # For foreach split steps that already set _foreach_var and
+        # _foreach_num_splits via self.next(), preserve the foreach transition
+        # so the runtime fans out correctly despite the caught exception.
+        if self._step_is_foreach:
+            flow._transition = (graph[step].out_funcs, flow._foreach_var)
+        else:
+            flow._transition = (graph[step].out_funcs, None)
 
         # If this task is a UBF control task, it will return itself as the singleton
         # list of tasks.


npow:

adity1raut:
> Thank you for the contribution. We're closing this PR as it wasn't discussed with maintainers beforehand and the value to the project doesn't outweigh the review cost. If you're interested in contributing to Metaflow for GSoC 2026, please check the [GSoC guidelines](https://summerofcode.withgoogle.com) and look for issues tagged [`gsoc`](https://github.com/Netflix/metaflow/labels/gsoc) before opening a PR.

@npow Thank you for the clarification. I understand the reason for closing the PR. I will review the GSoC guidelines and look for issues tagged **gsoc** before opening a PR next time. I appreciate your feedback.

PR #2948, opened 2026-03-02
Fix decompress_list for empty and multi-colon inputs

SanketMeghale:
## Summary
Fix `decompress_list()` to safely handle:
- empty input strings
- compressed strings where the prefix contains `:` (for example Step Functions style run IDs)

## Context / Motivation
`decompress_list()` currently crashes on:
- `""` with `IndexError`
- multi-colon payloads with `ValueError: too many values to unpack`

Fixes #2908.

## Changes Made
- Return `[]` for empty `lststr` in `metaflow/util.py`
- Use `decoded.rsplit(rangedelim, 1)` instead of `split(rangedelim)` so only the final delimiter between prefix and suffix list is used
- Add unit tests in `test/unit/test_util_list_compression.py` for:
  - empty input
  - multi-colon prefix input (`job:task::A,B`)
  - Step Functions style input with colons in run id

## Testing
Run in Linux container:
- `pytest -q test/unit/test_util_list_compression.py` -> `3 passed`
- `pytest -q test/unit/test_argo_workflows_cli.py` -> `8 passed`

Formatting hooks:
- `pre-commit run --files metaflow/util.py test/unit/test_util_list_compression.py`

PR #2986, opened 2026-03-07
Fix decompress_list handling for empty and multi-colon inputs

Vikasboura:
## Summary
This PR fixes edge cases in `decompress_list` where empty strings and
multi-colon inputs could lead to incorrect parsing or unexpected behavior.

## Changes
- Safely handle empty input (`""`) by returning an empty list.
- Use `split(":", 1)` to avoid incorrect parsing for inputs like `"1::3"`.
- Preserve existing behavior for valid inputs such as `"1:3"`.
- Ensure inputs with leading or trailing colons do not cause errors.

## Tests
Added tests to verify correct behavior for the following cases:
- `decompress_list("")`
- `decompress_list("1:3")`
- `decompress_list("1::3")`
- `decompress_list(":3")`

## Notes
The fix keeps the change minimal and avoids modifying the function API
or existing logic beyond the necessary edge case handling.

Fixes #2948

### DIFF
--- a/metaflow/util.py
+++ b/metaflow/util.py
@@ -384,6 +384,8 @@ def compress_list(lst, separator=",", rangedelim=":", zlibmarker="!", zlibmin=50
 
 def decompress_list(lststr, separator=",", rangedelim=":", zlibmarker="!"):
     # Three input modes:
+    if not lststr:
+        return []
     if lststr[0] == zlibmarker:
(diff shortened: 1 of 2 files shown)


Vikasboura:
Hi, I opened a PR implementing the proposed fix for decompress_list edge cases. 
Happy to adjust if a different behavior is preferred.

PR #2917, opened 2026-02-24
fix: correct format string kwarg typo elem_type -> elem_typ in runtim…

codeMaestro78:
The error message template uses {elem_typ} but the .format() call passed elem_type= (extra 'e'). This caused a KeyError at runtime whenever _control_mapper_tasks had the wrong element type, masking the real diagnostic the user should see.

## PR Type

- [x] Bug fix
- [ ] New feature
- [ ] Core Runtime change (higher bar -- see [CONTRIBUTING.md](../CONTRIBUTING.md#core-runtime-contributions-higher-bar))
- [ ] Docs / tooling
- [ ] Refactoring

## Summary

Sinlgle character typo fix

## Issue

Fixes #

## Reproduction

**Runtime:** 

**Commands to run:**
```bash
# paste exact commands
```

**Where evidence shows up:** 

<details>
<summary>Before (error / log snippet)</summary>

```
paste here
```

</details>

<details>
<summary>After (evidence that fix works)</summary>

```
paste here
```

</details>

## Failure Modes Considered

1.
2.

## Tests

- [ ] Unit tests added/updated
- [ ] Reproduction script provided (required for Core Runtime)
- [ ] CI passes
- [ ] If tests are impractical: explain why below and provide manual evidence above

## AI Tool Usage

- [x] No AI tools were used in this contribution
- [ ] AI tools were used (describe below)

PR #2997, opened 2026-03-08
Fix format string kwarg typo elem_type -> elem_typ in runtime

Vikasboura:
Fixes a one-character typo in runtime.py where elem_type was passed
as a kwarg but the format string expected elem_typ.

Without this fix, the error path raises a KeyError instead of the
intended descriptive error message.

Fixes #2917


npow:
Subset of #2998. Closing as duplicate.

PR #2998, opened 2026-03-08
Fix stale path variable in load_blobs error message

### LINKED ISSUE #2946: [BUG] Card server crashes with RuntimeError when run has 20+ cards
When a run has more than 20 cards (the default `max_cards` limit), calling `metaflow card server` crashes the process instead of stopping cleanly.

```
RuntimeError: generator raised StopIteration
```

The cause is a PEP 479 violation in `card_server.py:197` — `raise StopIteration` inside a generator is converted to `RuntimeError` on Python 3.7+. The fix is a one-liner.

Happy to submit a PR if this looks good to you.

Vikasboura:
Fixes an issue where load_blobs error messages referenced a stale
`path` variable instead of the correct `path_key`.

This change updates the error message to reference the correct variable
without affecting runtime behavior.

Fixes #2946

### DIFF
--- a/metaflow/datastore/content_addressed_store.py
+++ b/metaflow/datastore/content_addressed_store.py
@@ -177,7 +177,7 @@ class ContentAddressedStore(object):
                             version = meta.get("cas_version", -1)
                             if version == -1:
                                 raise DataException(
-                                    "Could not extract encoding version for '%s'" % path
+                                    "Could not extract encoding version for '%s'" % path_key
                                 )
                             unpack_code = getattr(self, "_unpack_v%d" % version, None)
                             if unpack_code is None:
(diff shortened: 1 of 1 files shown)


npow:
Thanks for the bug fixes — the `load_blobs` path variable fix and the `elem_typ` format string fix are both legit. However, please clean this up:

1. Remove the unrelated spelling/typo changes in vendored code (`_vendor/click/_bashcomplete.py`, `plugins/pypi/utils.py`). We don't modify vendored files for typos.
2. Remove the unrelated `wheather` → `whether` change in `runtime.py` — keep the PR scoped to the two actual bug fixes.

Once it's cleaned up we can review it properly.

npow:
Update: the `elem_typ` fix is being handled in #2917 which is a cleaner single-line change. Please remove that from this PR and scope it to just the `load_blobs` `path` → `path_key` fix (and drop the unrelated vendored code typo changes as mentioned earlier).

npow:
The two real bug fixes here are handled more cleanly in separate PRs: #2917 (`elem_typ` typo) and #2943 (`load_blobs` path variable). Closing in favor of those.

PR #3015, opened 2026-03-12
Fix: correct filename typo in timeout_decorator stack-frame filter

prabindersinghh:
## Summary

Fixes #3016 
The `_sigalrm_handler` in `TimeoutDecorator` builds a filtered stack
trace for the `TimeoutException` message, intended to strip out the
decorator's own internal frames so users only see their code.

The filter checked for `"timeout_decorators.py"` (with a trailing `s`)
instead of the actual filename `"timeout_decorator.py"`. Because the
string never matched the real filename, the filter was dead code — every
frame was always included, leaking internal decorator frames into
user-facing tracebacks.

**Fix:** Remove the extraneous `s` so the filter correctly matches
`"timeout_decorator.py"`.

## Changes

- `metaflow/plugins/timeout_decorator.py` — 1 character removed (line 84)
- `test/unit/test_timeout_decorator.py` — new file, 5 unit tests

## Testing

5 new unit tests in `test/unit/test_timeout_decorator.py`:

| Test | What it verifies |
|---|---|
| `test_sigalrm_handler_filters_own_frames` | Core regression: decorator frames excluded, user frames preserved |
| `test_sigalrm_handler_message_contains_step_name` | Step name appears in message |
| `test_sigalrm_handler_message_contains_duration` | Duration appears in message |
| `test_sigalrm_handler_all_decorator_frames_filtered` | All decorator frames stripped when no user frames present |
| `test_sigalrm_handler_preserves_multiple_user_frames` | Multiple user frames all preserved |

All 5 pass. Zero regressions against existing unit tests.

## AI Disclosure

- [x] I used AI tools (Claude) to help identify the bug and draft the
  test file. I have reviewed every line, understand the root cause,
  and can explain the fix and failure modes independently.


talsperre:
Thanks @prabindersinghh. Both issues in this PR are handled elsewhere: the timeout typo by #3030, and the `decompress_list` empty-string bug by cleaner focused PRs. Closing as superseded / duplicative.

PR #3021, opened 2026-03-12
Fix incorrect OR operand order when initializing Metaflow instance

anurag99sai-alt:
The current implementation initializes the Metaflow instance as:

self._metaflow = Metaflow(_current_metadata) or _metaflow

Since Metaflow(_current_metadata) always returns a truthy object,
the `_metaflow` operand is never evaluated. This prevents reuse of
an existing Metaflow instance.

Swapping the operands ensures that an existing `_metaflow` instance
is reused when available:

self._metaflow = _metaflow or Metaflow(_current_metadata)

### DIFF
--- a/metaflow/client/core.py
+++ b/metaflow/client/core.py
@@ -285,7 +285,7 @@ class MetaflowObject(object):
         # the namespace at the import time is problematic, since there
         # may be other modules that alter environment variables etc.
         # which may affect the namespace setting.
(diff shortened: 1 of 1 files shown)


talsperre:
Thanks @anurag99sai-alt. This duplicates #3022, which covers the same `_metaflow` operand-order bug with tests and less unrelated churn. Closing in favor of #3022.

PR #3040, opened 2026-03-24
Testing Infra Enhancements

npow:
1. Refactor internal netflix tests to parameterize orchestrator/backend and moved the tests to this repo
2. Setup local airflow/argo/sfn and k8s/batch to run in GHA
3. Report coverage
4. Split Tiltfile


devtools/eventbridge_stub.py
```diff
@@ -0,0 +1,76 @@
...
+                400,
+                {
+                    "__type": "ResourceNotFoundException",
+                    "message": "Rule not found",
+                },
+            )
+        elif target.endswith("PutRule"):
+            self._send_json(
+                200,
+                {"RuleArn": "arn:aws:events:us-east-1:123456789012:rule/devstack-stub"},
```
npow: this shouldn't be hardcoded

metaflow/metaflow_config.py
```diff
@@ -176,7 +176,7 @@
 
 DATATOOLS_CLIENT_PARAMS = from_conf("DATATOOLS_CLIENT_PARAMS", {})
 if S3_ENDPOINT_URL:
-    DATATOOLS_CLIENT_PARAMS["endpoint_url"] = S3_ENDPOINT_URL
+    DATATOOLS_CLIENT_PARAMS.setdefault("endpoint_url", S3_ENDPOINT_URL)
```
npow, line 182: add comment about this change

metaflow/cmd/develop/stub_generator.py
```diff
@@ -10,6 +10,13 @@
 from datetime import datetime
 from io import StringIO
 from types import ModuleType
+
```
saikonen, line 13: are these changes relevant to the testing infra?

devtools/Tiltfile
```diff
@@ -6,67 +6,45 @@
...
     "ui": ["postgresql", "minio"],
     "minio": [],
     "postgresql": [],
     "argo-workflows": [],
-    "argo-events": ["argo-workflows"],
-    "jobset": [],
+    "localbatch": ["minio"],
+    "ddb-local": [],
+    "sfn-local": ["ddb-local"],
+    "airflow": ["postgresql"],
```
saikonen, line 32: most localdev use cases do not require all of the orchestrators, so it would be a nice QoL to not have all of them selected by default when launching. Didn't look if its a big change or not though, not a dealbreaker.

devtools/requirements-devstack.txt
```diff
@@ -0,0 +1,8 @@
+git+https://github.com/npow/localbatch.git@main
```
saikonen: Still not sure on the utility of batch/sfn testing if it depends on emulation. Added friction is the need for a fork of the package. Is the main reason for this the need for py3.7 support?

setup.py
```diff
@@ -64,5 +69,15 @@ def find_devtools_files():
     install_requires=["requests", "boto3"],
     extras_require={
         "stubs": ["metaflow-stubs==%s" % version],
+        "dev": [
```
saikonen: roughly the same as requirements-devstack.txt. Should these be consolidated?

test/ux/core/test_utils.py
```diff
@@ -0,0 +1,343 @@
...
+        **deploy_args
+    )
+    print(f"Deployed workflow {deployed_flow.name}")
+    return deployed_flow
+
+
+def track_runs_by_tags(
+    flow_name: str, tags: List[str], timeout: int = 10, polling_interval: int = 60
+) -> List[str]:
+    """Poll for runs matching flow_name and tags, returning their pathspecs."""
```
saikonen: naming and docstring make no mention that this waits for runs to complete.

the number of runs returned is also tied to timings of the scheduler, whether all expected runs are running or not, before the tracked set of runs completes.

this might lead to some odd flakiness with tests on low-resource test infra

saikonen: only real test cases where this might matter is when expecting concurrent runs, e.g. when testing multiple runs triggered by events

saikonen (dismissed):
No real dealbreakers so approved. Only major friction is introducing localbatch emulation.

devtools/requirements-devstack.txt
npow: i actually own this package, and this should be coming from pypi

PR #3046, opened 2026-03-28
Fix: avoid unresolved conditional branches in Argo input-paths

devendrakushwah80:
Fixes unresolved task references in Argo input paths when using conditionl branches

Previously all upstream branches (node.in_funcs) were included unconditionally in input paths With newer Argo ReplaceStrict behavior this leads to failures when skipped branches are referenced
This change filters out conditional nodes while constructing input-paths, avoiding references to branches that may not execute.
This is an initial fix and I woiuld be happy to iterate further based on feedback
Note This approach filters conditional nodes as an initial safeguard, though a more comprehensive handling of conditional joins may be needed

### DIFF
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -1353,6 +1353,7 @@ class ArgoWorkflows(object):
                                 "argo-{{workflow.name}}/%s/{{tasks.%s.outputs.parameters.task-id}}"
                                 % (n, self._sanitize(n))
                                 for n in node.in_funcs
+                                if not self._is_conditional_node(self.graph[n])
                             ],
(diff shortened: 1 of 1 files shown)


devendrakushwah80:
Thanks for the detailed feedback this helps a lot

I see the issue now filtering conditional nodes at compile time ends up removing valid input-paths when the branch actually runs which breaks things

Makes sense that conditional_input_paths.py was already handling this at runtime, and my change kind of bypassed that logic

I will look into a better approach maybe using guarded expressions or limiting the filtering to cases where branches cant coexecute

Will try an updated version and share soon

devendrakushwah80:
Thanks for the feedback — I’ve reverted the initial filtering approach since it breaks valid conditional branches.

I’m now exploring a guarded-expression based solution so that skipped branches don’t cause ReplaceStrict failures while still preserving valid inputs.

Will update with a revised implementation soon.

PR #3079, opened 2026-03-31
Allow UserStepDecorators to also be StepMutators

romain-intel:
## Summary

Allow a class to inherit from both `UserStepDecorator` and `StepMutator` at the same time, so a single decorator can contribute both a `mutate()`/`pre_mutate()` pass over the flow/step (the `StepMutator` role) and `pre_step`/`post_step` wrapping hooks (the `UserStepDecorator` role). The instance is registered on both `step.wrappers` and `step.config_decorators` via a new MRO-walking `_all_step_fields()` helper, with an `_ran_init` idempotency guard so `external_init()` is not called twice.

In the process we also tightened a number of nearby corners in the user-decorator / flow-decorator plumbing and added several public helpers that made the dual-inheritance pattern writable at all.

## PR Type
- [x] Bug fix
- [x] New feature
- [ ] Core Runtime change
- [ ] Docs / tooling
- [ ] Refactoring

## New functionality

**Dual inheritance** — `class foo(UserStepDecorator, StepMutator)` is now a first-class pattern:
- Registration is unified via `_all_step_fields()` which walks the MRO and returns every `_step_field` declared on ancestor classes (deterministic MRO order).
- `_ran_init` guard on `external_init()` ensures `init()` is only invoked once per instance regardless of how many fields the instance lives on.
- `_init_step_decorators` now calls `external_init` on `step.wrappers` too (previously only `step.decorators`); `_process_config_decorators` covers `config_decorators`. This fixes a pre-existing gap where a pure `UserStepDecorator` wrapper's `init()` could be skipped.

**New MutableStep / MutableFlow helpers** (to support writing dual-inherit decorators without reaching into internal lists):
- `MutableFlow.get_step(name)`
- `MutableFlow.has_decorator(name)` / `MutableStep.has_decorator(name)`
- `MutableFlow.get_decorator_kwargs(...)` / `MutableStep.get_decorator_kwargs(...)`
- `MutableStep.get_all_decorator_kwargs(...)`
- `MutableStep.replace_decorator(...)`
- `MutableStep.decorator_specs` dedupes instances that live on multiple step fields.

**New StepDecorator hooks:**
- `check_step_conflicts(decorators)` — called during `_init_step_decorators` so a decorator can reject incompatible neighbors early.
- `find_decorator_attrs(...)` convenience for decorator-to-decorator introspection.

**Task-runtime plumbing:**
- `task.py` post-step sentinel changed from `False` to `None`, with explicit `is not False` guard so `post_step` returning `(None, False)` is a no-op instead of tripping `"Invalid value passed to self.next"`.
- `cli.py` no longer keeps the old manual `_init()` / `external_init()` loops for flow mutators — it's all driven by `_process_config_decorators` and `_init_step_decorators` now.
- The standalone `decorators._init()` helper was removed; its responsibilities were absorbed into `_init_flow_decorators` and `_init_step_decorators`.

## Bug fixes (most flagged by Greptile during review)

- **`add_decorator` on FlowMutator** — fixed `AttributeError: deco_type.name` when `add_decorator(MyFlowMutator)` is called from `mutate()`; the error path now uses `getattr(deco_type, "name", None) or getattr(deco_type, "decorator_name", deco_type)` so the real `MetaflowException` is raised.
- **`remove_decorator("typo")`** — `get_decorator_by_name` returns `None` for unregistered names; no longer passes that through to `issubclass(None, ...)` and raises `TypeError`.
- **FlowMutator registration** — `_do_add` now calls `external_init()` **before** appending to the decorator lists, so a raising `external_init` doesn't leave a half-initialized mutator in `_self_data` / merged mutators.
- **OVERRIDE + merged list** — switched from `list.remove()` (which shifts indices under the outer `pre_mutate` iteration) to slice-assignment `merged[:] = [...]`. No more silently-skipped elements.
- **`cached_skip_decorators`** — removed a name-based cache that over-skipped `allow_multiple` decorators (e.g. multiple `@card`s on different steps). Matches master's per-decorator behavior.
- **`remove_decorator(name)` `do_all=True`** — the `_pre_mutate + isinstance(StepMutator)` guard is now enforced on both the specific-match and `do_all` paths; previously `do_all` could strip a `StepMutator` from `mutate()` in violation of the documented contract.
- **`add_decorator` IGNORE path** — the `UserStepDecoratorBase` branches now return `None` when `add_or_raise` swallowed the registration (duplicate + IGNORE, or non-statically-defined ERROR), matching the `StepDecorator` path and the documented contract. Previously the throwaway instance was returned as if it had been added.

## Tests

- New `test/unit/mutators/` suite with dedicated flows + tests for dual inheritance, dynamic flow-mutator addition, string-form mutator addition, `add_decorator` return value, and regression coverage for the fixes above (`test_remove_decorator_guard.py`, `test_post_step_none_false.py`).
- Dual inheritance test exercises all four lifecycle methods: `pre_mutate`, `mutate`, `pre_step`, `post_step`.
- Spin conftest registers `--use-latest` so the fixture factory works when the suite is driven by a batch runner that doesn't inherit the option from a parent conftest.

### DIFF
--- a/metaflow/cli_components/run_cmds.py
+++ b/metaflow/cli_components/run_cmds.py
@@ -54,7 +54,6 @@ def before_run(obj, tags, decospecs, skip_decorators=False):
             # These decospecs are the ones from run/resume/spin PLUS the ones from the
             # environment (for example the @conda)
             decorators._attach_decorators(obj.flow, all_decospecs)
-            decorators._init(obj.flow)
             # Regenerate graph if we attached more decorators
             obj.flow.__class__._init_graph()
             obj.graph = obj.flow._graph
--- a/metaflow/decorators.py
+++ b/metaflow/decorators.py
@@ -372,13 +372,28 @@ class StepDecorator(Decorator):
     task_step_completed_ctx = None  # coalesces task_post_step + task_exception
     task_finished_ctx = None
 
+    def check_step_conflicts(self, decorators):
+        """
+        Validate this decorator against all other decorators on the same step.
+
+        Called automatically during step_init, before the decorator's own
+        step_init logic runs. Override in subclasses to declare conflicts.
+        Raise MetaflowException if an incompatible combination is detected.
+
+        Parameters
+        ----------
+        decorators : list
+            All decorators on this step (including self).
+        """
+        pass
+
     def step_init(
         self, flow, graph, step_name, decorators, environment, flow_datastore, logger
     ):
         """
         Called when all decorators have been created for this step
         """
-        pass
+        self.check_step_conflicts(decorators)
 
     def package_init(self, flow, step_name, environment):
         """
--- a/metaflow/decorators.py
+++ b/metaflow/decorators.py
@@ -372,63 +372,13 @@ class StepDecorator(Decorator):
     task_step_completed_ctx = None  # coalesces task_post_step + task_exception
     task_finished_ctx = None
 
-    @staticmethod
-    def find_decorator_attrs(flow_cls, step_name, deco_name, **exclude):
-        """
-        Find a step decorator's attributes at runtime.
-
-        Returns the attributes dict of the first matching decorator, or None.
-        Use ``exclude`` to skip decorators with specific attribute values,
-        e.g. ``find_decorator_attrs(cls, step, "resources", node_type="head")``
-        skips any @resources with node_type="head".
-
-        Parameters
-        ----------
-        flow_cls : type
-            The FlowSpec class.
-        step_name : str
-            The step function name.
-        deco_name : str
-            The decorator name to find.
-        **exclude
-            Attribute key-value pairs to skip.
-
-        Returns
-        -------
-        dict or None
-            The decorator's attributes, or None if not found.
-        """
-        for deco in getattr(flow_cls, step_name).decorators:
-            if deco.name == deco_name:
-                if exclude and any(
-                    deco.attributes.get(k) == v for k, v in exclude.items()
-                ):
-                    continue
-                return deco.attributes
-        return None
-
-    def check_step_conflicts(self, decorators):
-        """
-        Validate this decorator against all other decorators on the same step.
-
-        Called automatically during step_init, before the decorator's own
-        step_init logic runs. Override in subclasses to declare conflicts.
-        Raise MetaflowException if an incompatible combination is detected.
-
-        Parameters
-        ----------
-        decorators : list
-            All decorators on this step (including self).
-        """
-        pass
-
     def step_init(
         self, flow, graph, step_name, decorators, environment, flow_datastore, logger
     ):
         """
         Called when all decorators have been created for this step
         """
-        self.check_step_conflicts(decorators)
+        pass
 
     def package_init(self, flow, step_name, environment):
         """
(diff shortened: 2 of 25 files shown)


test/unit/mutators/flows/dual_inherit_flow.py
```diff
@@ -0,0 +1,49 @@
...
+
+    mutate() adds an environment variable (proves config_decorators path works).
+    pre_step() sets an artifact (proves wrappers path works).
+    post_step() sets another artifact (proves wrappers path works).
+    """
+
+    def init(self, marker="default"):
+        self._marker = marker
+
+    def mutate(self, mutable_step):
```
saikonen, line 19: could also cover `pre_mutate`, but validating one or the other should be enough of a confirmation as there's no cross-talk between the two?

metaflow/decorators.py
```diff
@@ -767,21 +767,6 @@ def _should_skip_decorator_for_spin(
     return False
 
 
-def _init(flow, only_non_static=False):
```
saikonen, line 770: so if an extension has been relying on attaching decorators, and needing to init these, instead of `decorators._init()`, will `deco.external_init()` now serve the same purpose?

metaflow/cli_components/run_cmds.py
```diff
@@ -54,7 +54,6 @@ def before_run(obj, tags, decospecs, skip_decorators=False):
             # These decospecs are the ones from run/resume/spin PLUS the ones from the
             # environment (for example the @conda)
             decorators._attach_decorators(obj.flow, all_decospecs)
-            decorators._init(obj.flow)
```
saikonen, line 57: why is an init not required anymore?

metaflow/flowspec.py
```diff
@@ -434,24 +437,27 @@ def _process_config_decorators(cls, config_options, process_configs=True):
...
                         "A non StepMutator found in step custom decorators"
                     )
+        for step in cls._steps:
+            for deco in step.config_decorators:
+                inserted_by_value = [deco.decorator_name] + (deco.inserted_by or [])
+                debug.userconf_exec(
+                    "Evaluating step level decorator %s for %s (pre-mutate)"
+                    % (deco.__class__.__name__, step.name)
+                )
+                deco.pre_mutate(
```
saikonen, line 452: just a review note: this is fine, because we check in the previous loop that all decos in the `config_decorators` are `StepMutator`s.

What is a bit unclear is why there was a need to split these to two loops. Is calling `external_init()` on every config deco before mutating a requirement?

metaflow/decorators.py
romain-intel: So yes, this function just called all the `external_init` functions. I now spread out the `external_init` in different parts so that it is more accurate about when to call them.

metaflow/flowspec.py
romain-intel: I was trying to keep with the other hooks where all hooks are called first and then all the second one.

test/unit/mutators/flows/dual_inherit_flow.py
romain-intel: Yes, I could add both but yes, it should not have cross talk.

metaflow/cli_components/run_cmds.py
romain-intel: It's still there but the calls are spread out.

saikonen (dismissed):
extension changes verified to work, should be good to go from that side.

PR #3081, opened 2026-04-03
fix: resolve TypeError when passing multi-events via config_expr in @trigger

Satyamgupta2365:
## Description
This PR fixes a bug where using `config_expr` to define multiple events in the `@trigger` decorator resulted in a `TypeError: ConfigValue is immutable` when trying to set event parameters.

## Context
When `@trigger(events=config_expr(...))` is used, the configuration values are returned as `ConfigValue` objects. These objects are immutable and act as wrappers around the underlying dictionary to prevent accidental modifications.

The `TriggerDecorator.process_event` method, however, expects a mutable object because it attempts to set `event["parameters"]` directly. This leads to the following traceback:

```text
File "metaflow/plugins/events_decorator.py", line 122, in process_event
  event["parameters"] = self.process_parameters(...)
File "metaflow/user_configs/config_parameters.py", line 190, in __setitem__
  raise TypeError("ConfigValue is immutable")
TypeError: ConfigValue is immutable


talsperre:
Thanks @Satyamgupta2365. The real trigger mutation fix here is covered by #3084 in a more focused PR. This PR also includes unrelated CLI ordering churn, so I am closing it in favor of #3084.

PR #3087, opened 2026-04-08
Fix CLI command ordering inconsistency

mohitmalviya0707:
Fix CLI command ordering inconsistency by sorting command list.

Ensures deterministic and predictable command output.


mohitmalviya0707:
Thanks for the review!

I've updated the PR to include only the CLI command ordering fix and removed unrelated changes.

PR #3090, opened 2026-04-10
perf: use lz4 frame compression for artifact storage

KRRT7:
## Summary

Replace gzip (`compresslevel=3`) with lz4 frame compression for new artifact writes in `ContentAddressedStore`. Existing gzip-compressed artifacts (`cas_version=1`) remain fully readable via `_unpack_v1` — no migration needed.

- New writes use `cas_version=2` with `lz4.frame.compress`/`decompress`
- Adds `lz4` to `install_requires`
- **2 files changed, 14 insertions, 3 deletions** (core change — benchmark scripts included for review, will revert before merge)

## Motivation

Every artifact save/load in Metaflow goes through `ContentAddressedStore`, which compresses with gzip. Compression cost on every artifact operation adds up — especially for wide foreach splits (1,000+ steps, each saving 15-50 artifacts) and large ML models/dataframes.

lz4 is a well-established compression library that is significantly faster than gzip, particularly on high-entropy data like pickled numpy arrays and model weights where gzip wastes CPU cycles trying to compress incompressible data.

## Benchmarks

All benchmarks run on Azure `Standard_D2s_v5` (2 vCPU, 8 GB RAM, non-burstable) with Python 3.12. Scripts in `benchmarks/` — fully reproducible with `pip install -e . && pip install numpy && python benchmarks/bench_cas_realistic.py`.

### Realistic artifacts (pickled Python objects)

| Payload | Pickled Size | gzip total | lz4 total | Speedup |
|---------|-------------|------------|-----------|---------|
| Small dict (config/metadata) | 233 B | 0.341 ms | 0.218 ms | **1.6x** |
| Metrics dict (feature stats) | 52 KB | 2.278 ms | 0.327 ms | **7.0x** |
| Numpy float64 (embeddings) | 800 KB | 29.111 ms | 1.557 ms | **18.7x** |
| Numpy float64 (model weights) | 8 MB | 289.234 ms | 15.792 ms | **18.3x** |
| Random bytes (opaque model blob) | 5 MB | 118.315 ms | 9.646 ms | **12.3x** |

The largest speedups occur on high-entropy data (numpy arrays, model blobs) where gzip struggles but lz4 handles efficiently. A step saving 3 large artifacts saves ~820ms. A 1,000-way foreach saves ~820 seconds of compression time.

### Synthetic data (repeating byte patterns)

| Blob Size | gzip total | lz4 total | Speedup |
|-----------|------------|-----------|---------|
| 1 KB | 0.283 ms | 0.177 ms | **1.6x** |
| 100 KB | 0.546 ms | 0.272 ms | **2.0x** |
| 1 MB | 3.055 ms | 1.098 ms | **2.8x** |
| 10 MB | 28.751 ms | 10.762 ms | **2.7x** |

## Backward compatibility

- **Reading old artifacts:** `_unpack_v1` (gzip) is preserved. Artifacts written by older Metaflow versions load without issue.
- **Forward compatibility concern:** Older Metaflow versions cannot read `cas_version=2` artifacts — they will raise `Unknown encoding version 2`. This is the existing behavior for any version bump (the error message already suggests updating Metaflow).
- **Deduplication:** Content addressing (SHA1 keys) is unchanged. Dedup works identically.

## Open questions

- **Hard vs soft dependency:** Should `lz4` be required or optional (fall back to gzip if not installed)? Currently added as a hard dep for simplicity.
- **Transfer path:** The `is_transfer` codepath (line 90) still writes `cas_version: 1` metadata. Transfer blobs are always raw (no compression), so this is functionally correct but could be updated for consistency.

## Test results

All tests pass identically on both `master` and `perf/lz4-artifact-compression`:

| Suite | master | perf/lz4 |
|-------|--------|----------|
| Unit tests (excl. spin) | 167 pass | 167 pass |
| Spin tests (CAS pipeline) | 9 pass, 1 fail, 2 err | 9 pass, 1 fail, 2 err |

The 3 spin failures are pre-existing (scikit-learn 1.3.0 / Python 3.12 conda incompatibility) and unrelated to this change.

---

If the approach looks good, I'll follow up by reverting the benchmark scripts from the branch so the final diff is just the core 2-file change.


metaflow/datastore/content_addressed_store.py
```diff
@@ -4,6 +4,8 @@
 from hashlib import sha1
 from io import BytesIO
 
+import lz4.frame
```
KRRT7, line 7: hmmm, will wait for human review on this.

KRRT7:
Hey @npow — just a gentle ping on this. Happy to adjust the approach based on your team's preferences (hard vs soft lz4 dep, forward compat strategy, etc.). Let me know if the general direction makes sense or if there's a different approach you'd prefer.

PR #3091, opened 2026-04-12
fix: use identity check for None comparison (PEP 8 E711)

harshadkhetpal:
## Summary

Replace `== None` with `is None` identity check in `argo_events.py` per PEP 8 (E711).

Python's `is` operator is the recommended way to compare with singletons like `None`. Using `==` can produce unexpected results if a class defines a custom `__eq__` method.

**Change:**
```python
# Before
if payload == None:

# After
if payload is None:
```

## Testing
No behavior change — identity check with `is` is functionally equivalent to `==` for `None` comparisons in standard usage.

### DIFF
--- a/metaflow/plugins/argo/argo_events.py
+++ b/metaflow/plugins/argo/argo_events.py
@@ -102,7 +102,7 @@ class ArgoEvent(object):
             If True, events are created on a best effort basis - errors are silently ignored.
         """
 
-        if payload == None:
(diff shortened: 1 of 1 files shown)


talsperre:
Thanks @harshadkhetpal. This is style-only churn (`== None` to `is None`) with no behavioral impact. We generally do not take standalone PRs for this kind of change, so I am closing it.

PR #3095, opened 2026-04-13
fix: export all Argo manifests (cron workflow, sensor) in --only-json

### LINKED ISSUE #1730: argo-workflows create --only-json doesn't export the cron workflow configuration
When a flow uses the `@schedule` decorator, running `argo-workflows create` deploys two things:

- An Argo workflow template for the flow
- An Argo cron workflow that triggers the flow on schedule

However, when running `argo-workflows create --only-json`, only the Argo Workflow Template for the flow gets generated, the `schedule` configuration is lost.

### LINKED ISSUE #1940: argo-workflows create --only-json doesn't export sensor configuration
The code only exports the workflow template and does not export sensor configuration. There is a clear TODO for this

`    if only_json:
        # TODO: Support echo-ing Argo Events Sensor template`

in the [argo_workflows_cli](https://github.com/Netflix/metaflow/blob/30b4d8bd2c695e0847876907e0e3888e4cd8b4ec/metaflow/plugins/argo/argo_workflows_cli.py#L253C1-L255C61)

It is not hard to also call str on flow._sensor that has the compiled sensor, but outputting two json objects (one for flow and another for sensor) together as a valid json seems to be more problematic, without breaking the interface. Perhaps we can add a new option to only export json for the sensor `"--only-sensor-json"`?

There is a related open issue about [only-json not exporting cron workflow](https://github.com/Netflix/metaflow/issues/1730). Introducing two new parameters "--only-cron-schedule-json" and "--only-sensor-json" (only one of these options can be true at a time) can help address these issues. 

Will be happy to submit a PR as the fix will be relatively small.

npow:
## Summary
- `argo-workflows create --only-json` now exports all Argo manifests, not just the WorkflowTemplate
- Output is a JSON object with up to three keys: `workflow_template` (always), `cron_workflow` (when `@schedule` is used), and `sensor` (when `@trigger`/`@trigger_on_finish` is used)
- Previously the CronWorkflow and Sensor configs were silently dropped
- `_cron_workflow_json()` now mirrors `ARGO_WORKFLOWS_USE_SCHEDULES` — uses `"schedules": [...]` (list) when the flag is True, matching what `ArgoClient.schedule_workflow_template` actually deploys

Fixes #1730, fixes #1940.

## ⚠️ Breaking change: output format

The output of `--only-json` has changed from a bare WorkflowTemplate JSON object to a wrapper envelope:

**Before:**
```json
{"apiVersion": "argoproj.io/v1alpha1", "kind": "WorkflowTemplate", ...}
```

**After:**
```json
{"workflow_template": {...}, "cron_workflow": {...}}
```

If you have scripts parsing the old output, update them to add `.workflow_template`:

```bash
# old
python flow.py argo-workflows create --only-json | jq '.spec.templates'

# new
python flow.py argo-workflows create --only-json | jq '.workflow_template.spec.templates'
```

## Test plan
- [x] Added `test/ux/core/test_argo_manifests.py` with tests for all manifest export scenarios
- [x] Added `mock.patch` tests for both `ARGO_WORKFLOWS_USE_SCHEDULES=True` and `False` paths in `TestCronWorkflowJson`

### DIFF
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -194,6 +194,37 @@ class ArgoWorkflows(object):
     def __str__(self):
         return str(self._workflow_template)
 
+    def _cron_workflow_json(self):
+        """Build the CronWorkflow manifest as a dict (without deploying it)."""
+        if self._schedule is None:
+            return None
+        return {
+            "apiVersion": "argoproj.io/v1alpha1",
+            "kind": "CronWorkflow",
+            "metadata": {"name": self.name},
+            "spec": {
+                "suspend": False,
+                "schedule": self._schedule,
+                "timezone": self._timezone,
+                "failedJobsHistoryLimit": 10000,
+                "successfulJobsHistoryLimit": 10000,
+                "workflowSpec": {"workflowTemplateRef": {"name": self.name}},
+                "startingDeadlineSeconds": 3540,
+            },
+        }
+
+    def export_all_json(self):
+        """Return a JSON string with all Argo manifests for this flow."""
+        result = {
+            "workflow_template": self._workflow_template.to_json(),
+        }
+        cron = self._cron_workflow_json()
+        if cron is not None:
+            result["cron_workflow"] = cron
+        if self._sensor is not None:
+            result["sensor"] = self._sensor.to_json()
+        return json.dumps(result, indent=4)
+
     def deploy(self):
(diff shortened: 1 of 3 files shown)


dallinstevens:
@npow This is exactly what my team needs with our implementation of metaflow/argocd. Anything we can do to help get this over the line? Happy to carry it forward: rebase on master, get CI green, add tests while keeping your authorship to this if you're open to it.

PR #3098, opened 2026-04-13
fix: skip config env var processing for flows without Config parameters

### LINKED ISSUE #2717: Triggering a DeployedFlow from a flow with a deploy-time Config broke in >=2.19.7

npow:
## Summary
- When a flow with a deploy-time `Config` triggers another flow via `DeployedFlow`, the parent flow's `METAFLOW_FLOW_CONFIG_VALUE` env var leaked into the child flow's click API
- This caused either "Options were not properly set" or an `AttributeError` on the `False`-valued `config_input`
- Guard `_compute_flow_parameters` so the config-resolution block only runs when the flow actually declares Config parameters

Fixes #2717.

## Test plan
- [x] `test/unit/test_deployed_flow_config.py` — verifies config env var doesn't leak to Config-less flows

### DIFF
--- a/metaflow/runner/click_api.py
+++ b/metaflow/runner/click_api.py
@@ -430,7 +430,7 @@ class MetaflowAPI(object):
         self._cached_computed_parameters = []
 
         config_options = None
-        if CLICK_API_PROCESS_CONFIG:
+        if CLICK_API_PROCESS_CONFIG and self._config_input:
             with flow_context(self._flow_cls) as _:
                 # We are going to resolve the configs first and then get the parameters.
                 # Note that configs may update/add parameters so the order is important
--- a/test/unit/test_deployed_flow_config.py
+++ b/test/unit/test_deployed_flow_config.py
@@ -1,0 +1,88 @@
+"""
+Regression test for GitHub issue #2717.
+
+When a flow with a deploy-time Config triggers another flow via DeployedFlow,
+the parent flow's METAFLOW_FLOW_CONFIG_VALUE environment variable was leaking
+into the child flow's click API initialization, causing:
+
+    "Options were not properly set -- this is an internal error."
+
+The fix ensures that _compute_flow_parameters skips config env var processing
+when the flow has no config parameters (self._config_input is False).
+"""
+
+import importlib
+import json
+import os
+import sys
+import tempfile
+
+import pytest
+
+
+@pytest.fixture
+def fake_flow_file(tmp_path):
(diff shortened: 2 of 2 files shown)


metaflow/runner/click_api.py
```diff
@@ -430,7 +430,7 @@ def _compute_flow_parameters(self):
         self._cached_computed_parameters = []
 
         config_options = None
-        if CLICK_API_PROCESS_CONFIG:
+        if CLICK_API_PROCESS_CONFIG and self._config_input:
```
romain-intel, line 433: If it is set, wouldn't the env-var still leak?

chrschorn:
Hey, any progress on this? Could I help with getting this merged?

npow:
@chrschorn sorry dropped the ball on this. feel free to take this over

Shriprasad-P (commented):
## Review

**PR:** fix: skip config env var processing for flows without Config parameters

**Touched:** `metaflow/runner/click_api.py`, `test/unit/test_deployed_flow_config.py`

- Size looks manageable (+89/-1).

PR #3117, opened 2026-04-15
Add pluggable artifact serializer framework

saeidbarati157:
## Summary

Replace the hardcoded pickle logic in `TaskDataStore` with a pluggable artifact serializer framework. Existing pickle behavior is preserved exactly — `PickleSerializer` is a built-in universal fallback so flows that never register a custom serializer behave identically to before.

- `ArtifactSerializer` ABC with priority-based dispatch, `SerializationMetadata` namedtuple, `SerializedBlob` value type.
- `SerializerStore` metaclass auto-registers subclasses; caches the sorted dispatch list and invalidates on new registration.
- Extensions register serializers via `ARTIFACT_SERIALIZERS_DESC` (standard Metaflow plugin pattern).
- A lazy import-hook registry lets extensions defer serializer-module imports (e.g. `torch`, `pyarrow`) until the user's own code imports the target type.

## Design

- **Dispatch.** Serializers are sorted by `PRIORITY` (lower first; ties broken by registration order). On save, the first `can_serialize(obj) == True` wins. On load, `can_deserialize(metadata)` routes by the `encoding` field in artifact metadata.
- **Wire vs storage format.** `serialize(obj, format=...)` / `deserialize(data, format=..., **kw)` accept a `format` kwarg — `STORAGE` (default) returns `(List[SerializedBlob], SerializationMetadata)` for the datastore save path; `WIRE` returns a `str` for CLI args, protobuf payloads, and cross-process IPC. One class owns both representations. `PickleSerializer` implements `STORAGE` only — pickle bytes are not a wire-safe payload — and raises `NotImplementedError` for `WIRE`.
- **Side-effect-free contract.** `serialize()` must not perform I/O or mutate global state: it may be called multiple times (caching, retries, parallel dispatch) and must stay safely idempotent. Side effects belong in hooks, not serializers.
- **Lazy registration.** `metaflow.datastore.artifacts.lazy_registry.register_serializer_for_type("torch.Tensor", "my_ext.TorchSerializer")` stores a declarative config; an `importlib.abc.MetaPathFinder` watches for `torch` to be imported and only then loads the serializer class. If the target module is already in `sys.modules`, registration is immediate.
- **Backward compatible.** Old artifacts with `gzip+pickle-v2` / `gzip+pickle-v4` encoding load correctly. Missing `encoding` defaults to `gzip+pickle-v2`.

## Error paths

- `save_artifacts`: if no serializer claims an object, raise `DataException` with a message pointing at the `PickleSerializer` fallback invariant (instead of the misleading `UnpicklableArtifactException` for a non-pickle failure).
- `load_artifacts`: if no deserializer claims an artifact, the error includes `serializer_info["source"]` when present, hinting at a missing extension.

## Test plan

- [x] 29 unit tests for `ArtifactSerializer` / `SerializerStore` / `SerializedBlob` / `SerializationMetadata`, including wire/storage dispatch and the ABC's abstract-method enforcement.
- [x] 36 unit tests for `PickleSerializer` (round-trips across scalars, containers, custom classes; `WIRE` rejection; encoding validation).
- [x] 9 unit tests for the lazy registry (config validation, eager registration, deferred registration via the import hook, recursion guard).
- [x] 6 integration tests for `TaskDataStore` round-trips, custom serializer priority, and backward compat.
- [x] 245 existing unit tests pass (only pre-existing `spin` failures remain, unrelated).


metaflow/datastore/artifacts/serializer.py
```diff
@@ -0,0 +1,217 @@
...
+
+    @classmethod
+    @abstractmethod
+    def deserialize(cls, data, metadata=None, context=None, format=STORAGE):
+        """
+        Deserialize back to a Python object.
+
+        Parameters
+        ----------
+        data : Union[List[bytes], str]
```
romain-intel, line 823: nit: update method declaration (same on the others). Helps with the editors

metaflow/datastore/artifacts/serializer.py
```diff
@@ -0,0 +1,217 @@
...
+            ``List[bytes]`` when ``format=STORAGE``; ``str`` when ``format=WIRE``.
+        metadata : SerializationMetadata, optional
+            Metadata stored alongside the artifact. Required for STORAGE,
+            ignored for WIRE.
+        context : Any, optional
```
romain-intel: This is good but I don't think it is used.Let's strip it out for now and we can add later.

metaflow/datastore/artifacts/serializer.py
```diff
@@ -0,0 +1,217 @@
...
+            Optional context for deserialization (e.g., task vs client loading).
+        format : str
+            Either ``STORAGE`` (default) or ``WIRE``.
```
romain-intel: Make it an enum.

metaflow/datastore/task_datastore.py
```diff
@@ -117,12 +117,13 @@ def __init__(
...
-        if ver >= 36:
-            self._encodings.add("pickle-v4")
-            self._encodings.add("gzip+pickle-v4")
+        # ``_serializers`` is a property that dispatches through
+        # ``SerializerStore.get_ordered_serializers()`` on each access. The
+        # lookup is cheap (cached inside the store) and picks up serializers
+        # registered via the lazy import hook after this instance was
+        # constructed — otherwise long-lived datastores (notebooks, client
+        # sessions) would silently miss any extension registered after init.
+        self._serializers_override = None
```
romain-intel, line 157: This comment feels a bit out of place. Sure _serializers_override is set in _serializers but we should make that a bit clearer.

metaflow/datastore/task_datastore.py
```diff
@@ -347,38 +361,55 @@ def save_artifacts(self, artifacts_iter, len_hint=0):
...
+                        % (serializer.__name__, name)
+                    )
+                if len(blobs) > 1:
+                    # The datastore currently stores a single blob per
+                    # artifact. Silently dropping blobs[1:] would corrupt
+                    # multi-blob IOType extensions (e.g. chunked tensors) on
+                    # load. Fail loudly until multi-blob support lands.
+                    raise DataException(
+                        "Serializer %s returned %d blobs for artifact '%s'; "
+                        "only single-blob serializers are supported."
```
romain-intel: add -- "at this time. If you have a need for multi blob serializers, please reach out to the Metaflow team."

metaflow/datastore/task_datastore.py
```diff
@@ -414,32 +445,50 @@ def load_artifacts(self, names):
...
+                serializer_source = metadata.serializer_info.get("source")
+                if serializer_source:
+                    source_hint = (
+                        " The artifact was written by '%s' — the "
+                        "corresponding extension may not be installed."
+                        % serializer_source
+                    )
                 raise DataException(
-                    "Python 3.6 or later is required to load artifact '%s'" % name
+                    "No deserializer claimed artifact '%s' (encoding: %s)."
```
romain-intel: Let's also print the serializer_info

romain-intel (commented):
Minor comments.

saeidbarati157:
Thanks Romain! Addressed in fed2c78:

- **serializer.py method declarations** — added full type annotations to the ABC method signatures.
- **`context` param** — stripped from `deserialize` across the ABC, `PickleSerializer`, the `TaskDataStore` caller, and tests. Can re-add if/when we have a concrete use case.
- **`STORAGE`/`WIRE` as an enum** — promoted to `SerializationFormat(str, Enum)`. Subclassing `str` keeps existing `format == 'storage'` comparisons working so no downstream migration is needed.
- **`_serializers_override` comment** — reworded to describe the test override; moved the property-rationale block next to the property definition itself.
- **Multi-blob error** — appended "at this time. If you have a need for multi blob serializers, please reach out to the Metaflow team."
- **Deserializer-not-found error** — now prints the `serializer_info` dict alongside encoding.

All 77 serializer/pickle/lazy/integration unit tests pass.

metaflow/datastore/artifacts/serializer.py
```diff
@@ -0,0 +1,726 @@
...
+        """
+        raise NotImplementedError
+
+    @classmethod
+    @abstractmethod
+    def serialize(
+        cls,
+        obj: Any,
+        format: SerializationFormat = SerializationFormat.STORAGE,
+    ) -> Union[Tuple[List["SerializedBlob"], SerializationMetadata], str]:
```
jefferynidd, line 775: just checking - is a list of serialized blobs meant for serializers that may want to chunk its outputs into multiple files ? could see how this would be handy with saving things such as distributed model weights !

metaflow/datastore/artifacts/serializer.py
```diff
@@ -0,0 +1,726 @@
...
+        return "_OrderedSet(%r)" % list(self._d)
+
+
+class SerializedBlob(object):
+    """
+    Represents a single blob produced by a serializer.
+
+    A serializer may produce multiple blobs per artifact. Each blob is either:
+    - New bytes to be stored (is_reference=False, value is bytes)
+    - A reference to already-stored data (is_reference=True, value is a string key)
```
jefferynidd, line 121: are there any assumptions on where the data is stored when it's a reference ? perhaps let's say i want to reference an external data store - wondering if that's possible and what that'd look like

metaflow/datastore/artifacts/serializer.py
```diff
@@ -0,0 +1,726 @@
...
+        else:
+            self.is_reference = is_reference
+
+    @property
+    def needs_save(self):
+        """True if this blob contains new bytes that need to be stored."""
+        return not self.is_reference
+
+
+class SerializerStore(ABCMeta):
```
jefferynidd, line 149: would this serializer store be accessible from other metaflow plugins ? i'm working on updating the checkpoint decorator and it'd be super useful if we could re-use this to serialize artifacts for checkpointing purposes !

we'd have to be able to (1) access the serializations for an artifact someone wants to checkpoint , and (2) save the artifacts in a different location from the usual metaflow artifacts

metaflow/datastore/artifacts/serializer.py
saeidbarati157: Yes, exactly that `List[SerializedBlob]` is multi-blob ready for chunked outputs (sharded tensors / arrow / model weights). V1 datastore only persists the first blob today; multi-blob storage is a deliberate follow-up so this PR stays focused on dispatch/lifecycle. Happy to prioritize it if it's blocking checkpoint.

metaflow/datastore/artifacts/serializer.py
saeidbarati157: `is_reference=True` is intra-datastore only! the `value` is a key into Metaflow's `ContentAddressedStore`, used for dedup at load time. It doesn't know about external stores.                                                                                                                                                                                                                                                                                                                                            

For "point at external data," the natural shape is a normal blob whose bytes contain the URI/handle, and your serializer's `deserialize()` does the fetch.,  keeps auth + lifecycle under the serializer's control.

metaflow/datastore/artifacts/serializer.py
saeidbarati157: Yes, that's exactly what the decoupling enables. The serializer produces `(blobs, metadata)` independently of where you persist them:

```python
from metaflow.datastore.artifacts.serializer import SerializerStore

# (1) serialize
for s in SerializerStore.get_ordered_serializers():
    if s.can_serialize(obj):
        blobs, metadata = s.serialize(obj)
        break

# (2) deserialize from your own storage
for s in SerializerStore.get_ordered_serializers():
    if s.can_deserialize(metadata):
        return s.deserialize([blob], metadata)
```

You hand blobs to whatever storage you want (your own S3 path, a separate CAS, etc.).

metaflow/datastore/artifacts/serializer.py
jefferynidd: not blocking checkpointing - i think there's always workaround (consolidate model weights into single artifact) . but good to know there's support for this coming soon !

metaflow/datastore/artifacts/serializer.py
jefferynidd: got it , that makes sense thanks ! so for external data , is this understanding correct ? as an example let's say i have an external GCP datastore , could something like this work ?

serializer -> serializes and saves to GCP , but returns just a GCP path
deserializer -> receives the GCP path , deserializes from GCP back into memory

metaflow/datastore/artifacts/serializer.py
jefferynidd: nice ! this is great , thanks for the detailed example !

metaflow/datastore/artifacts/serializer.py
romain-intel: I would say so yes. That's the idea. This would be similar for a HF model as well btw (at least that is the idea).

metaflow/datastore/task_datastore.py
```diff
@@ -347,38 +397,92 @@ def save_artifacts(self, artifacts_iter, len_hint=0):
...
+                    )
+
+                try:
+                    blobs, metadata = serializer.serialize(
+                        obj, format=SerializationFormat.STORAGE
+                    )
+                except TypeError as e:
+                    # Preserve the historical "couldn't pickle this" wrapper so
+                    # existing consumers still see ``UnpicklableArtifactException``.
+                    raise UnpicklableArtifactException(name) from e
```
saikonen, line 429: couldn't a non-pickle serializer also raise a `TypeError` ? One option could be to treat any raised `MetaflowException` as pass-through, and have the pickle-serializer raise the `UnpicklableArtifactException` instead?

metaflow/plugins/datastores/serializers/pickle_serializer.py
```diff
@@ -0,0 +1,62 @@
...
+        blob = pickle.dumps(obj, protocol=4)
+        return (
+            [SerializedBlob(blob, is_reference=False)],
+            SerializationMetadata(
+                obj_type=str(type(obj)),
+                size=len(blob),
+                encoding="pickle-v4",
+                serializer_info={},
+            ),
+        )
```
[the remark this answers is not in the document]
saikonen: possibly relevant still.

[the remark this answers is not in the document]
saikonen: actually strike that, I'm forgetting the context for keeping the protocol-2 encoding here but it seems to have been tied to python <=3.6 so dropping the fallback should be fine at this point.

metaflow/datastore/task_datastore.py
romain-intel: I think the idea was to try to return the same exception as before. I think what we could do is specially catch UnpicklableArtifactException (and yes, raise it directly in the pickle serializer) and for everything else, fall through to the DataException which would maintain backwards compatibility with exception types. Would that work for you?

saeidbarati157: Moved pickle.dumps's TypeError handling into PickleSerializer itself. It now raises `UnpicklableArtifactException` (with `artifact_name` optional). `TaskDataStore.save_artifacts` catches it specifically to re-attach the artifact name, lets any other `MetaflowException` pass through unchanged, and only wraps non-MetaflowException failures in `DataException` for context. Added three tests covering the three branches.

PR #3118, opened 2026-04-15
Add IOType system with typed artifact serialization

saeidbarati157:
## Summary

On top of #3117, this ships the typed-artifact layer for Metaflow. Scope kept small — the pieces here pull their weight in core; everything else belongs in extensions.

- `IOType` ABC — the contract extension authors target.
- `Json` and `Struct` — two concrete types with clear standalone value: wire format for CLI/IPC, cross-language JSON bytes on storage, no pickle code-execution risk. `Struct` also walks directly-nested `@dataclass` fields so `Outer(inner=Inner(...))` round-trips back to its original type.
- `IOTypeSerializer` — the datastore bridge that plugs any `IOType` into the `ArtifactSerializer` dispatch from #3117, so save/load through the datastore just works (PRIORITY=50, STORAGE-only).

## Intentionally not included

- Primitive wrappers (`Int32/Int64/Float32/Float64/Bool/Text`). Standard Python numbers and strings flow through `PickleSerializer` unchanged. Wrapping is opt-in, for cases that need constraints/metadata attached.
- `Tensor`. Numpy + byte-order/dtype opinions; belongs in an extension that can own those choices.
- `List` / `Map` / `Enum`. Thin wrappers whose value over plain JSON is mostly schema emission — not enough on their own for core.
- Rich `Struct.to_spec()`. Extensions that ship primitive wrappers can override to emit fully-typed schemas; core just returns `{"type": "struct"}`.

## Design

- `serialize(format=...)` / `deserialize(data, format=..., **kw)` mirror the `ArtifactSerializer` signature from #3117. Same `WIRE` / `STORAGE` constants, so a single subclass owns both representations.
  - `STORAGE` → `(List[SerializedBlob], metadata_dict)` for the datastore save path.
  - `WIRE` → `str` for CLI args, protobuf payloads, cross-process IPC.
- Four subclass hooks: `_wire_serialize`, `_wire_deserialize`, `_storage_serialize`, `_storage_deserialize`.
- Instantiating without implementing the hooks raises `TypeError`.
- `IOTypeSerializer` handles only `STORAGE`; wire encoding is produced by calling `IOType.serialize(format=WIRE)` directly.

## Safety

- `Struct._storage_deserialize` and `IOTypeSerializer.deserialize` both require the class named in artifact metadata to be an actual class (`isinstance(..., type)`) before any further checks. This excludes module-level dataclass *instances* (which `is_dataclass()` alone considers valid) and other callables that could be invoked with attacker-controlled kwargs.
- Importing the metadata-named module can still run module-level side-effect code; the `Struct` docstring calls this out — don't load artifacts from untrusted sources.

## Test plan

- [x] `test_base.py` — abstract instantiation, WIRE/STORAGE dispatch, invalid format, equality/hash, spec.
- [x] `test_json_type.py` — wire and storage round-trips.
- [x] `test_struct_type.py` — dataclass round-trip, dict round-trip, directly-nested dataclass round-trip, container-field pass-through, rejection of non-dataclass and dataclass-instance metadata.
- [x] `test_iotype_serializer.py` — bridge `can_serialize`/`can_deserialize`, dataclass round-trip, WIRE not supported on the bridge, security (rejects non-IOType classes in metadata).
- [x] Full unit suite green (excluding pre-existing spin failures).

### DIFF
--- a/metaflow/datastore/artifacts/serializer.py
+++ b/metaflow/datastore/artifacts/serializer.py
@@ -1,0 +1,170 @@
+from abc import abstractmethod
+from collections import namedtuple
+
+
+SerializationMetadata = namedtuple(
+    "SerializationMetadata", ["type", "size", "encoding", "serializer_info"]
+)
+
+
+class SerializedBlob(object):
+    """
+    Represents a single blob produced by a serializer.
+
+    A serializer may produce multiple blobs per artifact. Each blob is either:
+    - New bytes to be stored (is_reference=False, value is bytes)
+    - A reference to already-stored data (is_reference=True, value is a string key)
+
+    Parameters
+    ----------
+    value : Union[str, bytes]
+        The blob data (bytes) or a reference key (str).
+    is_reference : bool, optional
+        If None, auto-detected from value type: str -> reference, bytes -> new data.
+    compress_method : str
+        Compression method for new blobs. Ignored for references. Default "gzip".
+    """
+
+    def __init__(self, value, is_reference=None, compress_method="gzip"):
+        if not isinstance(value, (str, bytes)):
+            raise TypeError(
+                "SerializedBlob value must be str or bytes, got %s" % type(value).__name__
+            )
+        self.value = value
+        self.compress_method = compress_method
+        if is_reference is None:
+            self.is_reference = isinstance(value, str)
+        else:
+            self.is_reference = is_reference
+
+    @property
+    def needs_save(self):
+        """True if this blob contains new bytes that need to be stored."""
+        return not self.is_reference
+
+
+class SerializerStore(type):
+    """
+    Metaclass for ArtifactSerializer that auto-registers subclasses by TYPE.
+
+    Provides deterministic ordering: serializers are sorted by (PRIORITY, registration_order).
+    Lower PRIORITY values are tried first. Registration order breaks ties.
+    """
+
+    _all_serializers = {}
+    _registration_order = []
+
+    def __init__(cls, name, bases, namespace):
+        super().__init__(name, bases, namespace)
+        if cls.TYPE is not None:
+            if cls.TYPE not in SerializerStore._all_serializers:
+                SerializerStore._registration_order.append(cls.TYPE)
+            SerializerStore._all_serializers[cls.TYPE] = cls
+
+    @staticmethod
+    def get_ordered_serializers():
+        """
+        Return serializer classes sorted by (PRIORITY, registration_order).
+
+        This ordering is deterministic for a given set of loaded serializers.
+        """
+        order = SerializerStore._registration_order
+        return sorted(
+            SerializerStore._all_serializers.values(),
+            key=lambda s: (s.PRIORITY, order.index(s.TYPE)),
+        )
+
+
+class ArtifactSerializer(object, metaclass=SerializerStore):
+    """
+    Abstract base class for artifact serializers.
+
+    Subclasses must set TYPE to a unique string identifier and implement
+    all four class methods. Subclasses are auto-registered by the SerializerStore
+    metaclass on class definition.
+
+    Attributes
+    ----------
+    TYPE : str or None
+        Unique identifier for this serializer (e.g., "pickle", "iotype").
+        Set to None in the base class to prevent registration.
+    PRIORITY : int
+        Dispatch priority. Lower values are tried first. Default 100.
+        PickleSerializer uses 9999 as the universal fallback.
+    """
+
+    TYPE = None
+    PRIORITY = 100
+
+    @classmethod
+    @abstractmethod
+    def can_serialize(cls, obj):
+        """
+        Return True if this serializer can handle the given object.
+
+        Parameters
+        ----------
+        obj : Any
+            The Python object to serialize.
+
+        Returns
+        -------
+        bool
+        """
+        raise NotImplementedError
+
+    @classmethod
+    @abstractmethod
+    def can_deserialize(cls, metadata):
+        """
+        Return True if this serializer can deserialize given the metadata.
+
+        Parameters
+        ----------
+        metadata : SerializationMetadata
+            Metadata stored alongside the artifact.
+
+        Returns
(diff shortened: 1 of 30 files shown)


metaflow/io_types/__init__.py
```diff
@@ -0,0 +1,7 @@
+from .base import IOType
```
romain-intel, line 1: When discussing, I thought for standard python types, we wouldn't have a wrapped type.

metaflow/datastore/artifacts/serializer.py
```diff
@@ -0,0 +1,173 @@
...
+            SerializerStore._all_serializers[cls.TYPE] = cls
+
+    @staticmethod
+    def get_ordered_serializers():
+        """
+        Return serializer classes sorted by (PRIORITY, registration_order).
+
+        This ordering is deterministic for a given set of loaded serializers.
+        """
+        order = SerializerStore._registration_order
```
romain-intel, line 74: nit: cache this if called mjltiple times.

metaflow/datastore/artifacts/serializer.py
```diff
@@ -0,0 +1,173 @@
...
+    Lower PRIORITY values are tried first. Registration order breaks ties.
+    """
+
+    _all_serializers = {}
+    _registration_order = []
+
+    def __init__(cls, name, bases, namespace):
+        super().__init__(name, bases, namespace)
+        if cls.TYPE is not None:
+            if cls.TYPE not in SerializerStore._all_serializers:
```
romain-intel: nit: may need to exclude base classes.

metaflow/datastore/artifacts/serializer.py
```diff
@@ -0,0 +1,173 @@
...
+        -------
+        bool
+        """
+        raise NotImplementedError
+
+    @classmethod
+    @abstractmethod
+    def serialize(cls, obj):
+        """
+        Serialize obj to blobs and metadata. Must be side-effect-free.
```
romain-intel, line 140: question on the side-effect freeness; is this because it can be retried? Are you assuming that things we would want for side-effects happen in hooks?

metaflow/datastore/task_datastore.py
```diff
@@ -347,38 +346,42 @@ def save_artifacts(self, artifacts_iter, len_hint=0):
...
-                    except TypeError as e:
-                        raise UnpicklableArtifactException(name) from e
+                # Find the first serializer that can handle this object
+                serializer = None
+                for s in self._serializers:
+                    if s.can_serialize(obj):
+                        serializer = s
+                        break
+                if serializer is None:
+                    raise UnpicklableArtifactException(name)
```
romain-intel: nit: we need to modify the error message.

metaflow/datastore/task_datastore.py
```diff
@@ -347,38 +346,42 @@ def save_artifacts(self, artifacts_iter, len_hint=0):
...
+                    blobs, metadata = serializer.serialize(obj)
+                except TypeError as e:
+                    raise UnpicklableArtifactException(name) from e
 
                 self._info[name] = {
-                    "size": len(blob),
-                    "type": str(type(obj)),
-                    "encoding": encode_type,
+                    "size": metadata.size,
+                    "type": metadata.type,
```
romain-intel: nit: avoid using keywords (I know it's legal but sometimes confusing)

metaflow/datastore/task_datastore.py
```diff
@@ -408,38 +411,50 @@ def load_artifacts(self, names):
...
+            # Find deserializer via metadata
+            deserializer = None
+            for s in self._serializers:
+                if s.can_deserialize(metadata):
+                    deserializer = s
+                    break
+            if deserializer is None:
                 raise DataException(
-                    "Python 3.6 or later is required to load artifact '%s'" % name
+                    "No deserializer found for artifact '%s' "
```
romain-intel: nit: I think we should ideally provide a more helpful message here. In the metadata, we should be able to put where the serializer came from that serialized this to give some hint to the user on how to get it to work (ie: what extension may be missing).

romain-intel (commented):
Initial comments:
- the auto registration on types doesn't seem to be included here
- we have to verify that client will work too (ie: serialization_metadata
- something about compression (does it work right now)?
- need to discuss io_types (and that may be in an extension since may be a bit specific).

metaflow/datastore/artifacts/serializer.py
saeidbarati157: fixed it.

metaflow/io_types/__init__.py
saeidbarati157: fixed it. Removed the IOType for standard types.

saeidbarati157:
Latest push (`4e1dfb8`, rebased on the updated #3117 tip `fed2c78`):

**Addresses both P1s Greptile flagged:**
- `Struct._reconstruct` now filters `init=False` fields out of the constructor kwargs and restores them via `object.__setattr__` — frozen-dataclass safe.
- `Json.__hash__` / `Struct.__hash__` override the base to use a new `_make_hashable` helper (recursive frozenset/tuple conversion). Preserves Python's numeric equivalence, so `Json({'x': 1}) == Json({'x': 1.0})` and their hashes match.

**Other cleanup folded into the rebase:**
- `IOTypeSerializer.deserialize` signature updated to match the new `ArtifactSerializer` ABC from #3117 (no `context` kwarg).

**Stale review threads resolved** — the 6 `datastore/*` comments point at files whose state is now fully managed by #3117 (type hints, `SerializationFormat` enum, clearer error messages, `_serializers_override`). Happy to re-open any of them if you'd like a separate discussion.

Greptile re-review: 5/5 (up from 3/5). 120/120 io_types + pluggable-serializer unit tests pass.

Optional P2 remaining (per Greptile summary): the base `IOType.__hash__` still raises `TypeError` if a future extension subclass wraps an unhashable value without overriding. Can add a try/except identity fallback there if useful — not blocking.

PR #3120, opened 2026-04-16
feat: @step(start=True)/@step(end=True) annotations, single-step flows, and node_info metadata

romain-intel:
## PR Type

- [ ] Bug fix
- [X] New feature
- [X] Core Runtime change (higher bar -- see [CONTRIBUTING.md](../CONTRIBUTING.md#core-runtime-contributions-higher-bar))
- [ ] Docs / tooling
- [ ] Refactoring

## Summary

Makes the FlowSpec graph more flexible in three ways, with backward compatibility preserved throughout:

1. **Explicit start/end annotations.** `@step` now accepts `start=True` and `end=True` kwargs. Steps no longer have to be named "start"/"end" — a flow can use any name as long as one step is annotated as the start and one as the end. When no annotations are present, the graph falls back to discovering steps by the names "start" and "end" (old behavior).

2. **Single-step flows.** A single step annotated `@step(start=True, end=True)` is now valid; previously a FlowSpec required at least two steps.

3. **Per-node `node_info` metadata.** `DAGNode` gains a `node_info` dict that extensions can populate to attach arbitrary per-step metadata. Live references are accessible via `flow._graph`; serialized values (via `to_pod`) flow into `_graph_info`. Callables are serialized as their qualified name.

These are core-runtime changes because they touch `graph.py`, `lint.py`, `runtime.py`, `task.py`, and every orchestrator plugin (Argo Workflows, Step Functions, Airflow) that reads the graph or emits events tied to the terminal step.

### Key changes

| Area | What changed |
|------|-------------|
| `decorators.py` | `@step(start=False, end=False, node_info=None)` |
| `graph.py` | `DAGNode.is_start_step`, `is_end_step`, `node_info`; `_identify_start_end` uses annotations with name fallback; attribute name (not AST `def` name) is authoritative; tolerates sourceless single-step methods (for out-of-tree FunctionSpec extension) |
| `lint.py` | `check_basic_steps` distinguishes "no start" from "multiple start annotations"; new `check_start_end_degree` validates in/out degrees; new `check_annotation_name_conflict` warns when `@step(start=True)` coexists with a step named "start" (same for end) |
| `util.py` | `to_pod` serializes callables via `__qualname__` |
| `runtime.py` | Persists `start_step`/`end_step` as metadata on the `_parameters` task |
| `task.py`, `flowspec.py` | Use the graph's start/end step rather than literal "start"/"end" names |
| `client/core.py` | `Run._graph_endpoints` reads endpoints from `_parameters` metadata with `_graph_info` fallback for orchestrated runs |
| `plugins/argo/argo_workflows.py` | `_matching_conditional_join` falls back to `graph.end_step`; terminal-step detection through the DAG |
| `plugins/argo/argo_workflows_decorator.py` | Auto-emitted event from the end step is also published with the well-known `.end` suffix so `@trigger_on_finish` subscribers — which don't know the publisher's end step name at deploy time — keep receiving triggers from custom-named end steps |
| `plugins/aws/step_functions/step_functions.py` | Reads `StartAt` from the definition JSON instead of hardcoded `["States"]["start"]` |
| `plugins/airflow/airflow.py`, `airflow_cli.py` | Use `graph.start_step`/`end_step` instead of hardcoded names |
| `plugins/cards/card_cli.py` | `graph` kwarg keeps the old dict-of-steps shape for backward compatibility with third-party card modules; new opt-in `graph_info` kwarg carries the full payload (steps + start/end). Card_cli probes the constructor signature and only passes `graph_info` to cards that accept it |
| `plugins/cards/card_modules/basic.py` | `DefaultCard`, `DefaultCardJSON`, `ErrorCard` accept both `graph` and `graph_info` (preferred when provided) |
| `plugins/cards/ui/` | Svelte DAG renderer reads start/end dynamically instead of assuming "start"/"end" |

## Tests

- [X] Unit tests added/updated
- [X] CI passes

New and updated tests cover:
- Graph structure inference: standard flows, custom-named flows, single-step flows, branch flows, foreach flows, split-as-entry-step flows
- `@step(start=True)`/`@step(end=True)` annotation mechanics and backward-compat with name-based discovery
- Lint validation: degree checks, annotation-name conflicts, negative-path cases for malformed annotations
- End-to-end execution of custom-named flows via Runner with card rendering
- `Trigger.from_runs` on flows with a custom-named terminal step
- Step Functions `StartAt` lookup from deployment JSON
- Card graph transform accepting both legacy and new payload shapes

## Non-Goals

- A `StepSpec`/`FunctionSpec` construct (single-step FlowSpec-like class with `init()`+`call()` lifecycle) — hooks for it are present in `DAGNode` and `_base_step_decorator` (see module comment in `graph.py`), but the class itself is kept out-of-tree for now.
- Multi-flow file support (a `FlowSpec.main()` classmethod to route between multiple FlowSpec subclasses in one file) — deferred.
- Direct non-CLI invocation of FlowSpec — deferred.

## AI Tool Usage

AI coding assistance was used during development. All generated code was reviewed and tested.


metaflow/graph.py
```diff
@@ -259,9 +258,61 @@ def __init__(self, flow):
         self.doc = deindent_docstring(flow.__doc__)
         # nodes sorted in topological order.
         self.sorted_nodes = []
+        self._identify_start_end()
         self._traverse_graph()
         self._postprocess()
 
+    def _identify_start_end(self):
```
romain-intel, line 277: Instead of this, let's allow the `@step` decorator to take in start and/or end arguments that are booleans indicating if the step is a start or an end step or both. We therefore do not need to do a graph traversal and can instead have the user explicitly tell us. We should fall back on the usual "start" and "end" step if not specified. We can then make sure that the start and end step are proper (ie: in and out degrees are correct).

metaflow/parameters.py
```diff
@@ -519,4 +519,16 @@ def wrapper(cmd):
     return wrapper
 
 
+class InitParameter(Parameter):
+    """
+    Parameter for StepSpec init() phase.
+
+    Behaves identically to Parameter in CLI mode. In direct invocation,
```
romain-intel: Let's remove the comment about direct invocation for now.

metaflow/stepspec.py
```diff
@@ -0,0 +1,277 @@
...
+
+            def call(self):
+                self.output = self.text.upper()
+
+        u = Upper(use_cli=False)
+        u(text="world")
+        assert u.output == "WORLD"
+    """
+
+    def __init__(self, use_cli=None, **kwargs):
```
romain-intel: For now, let's remove the direct invocation code. Let's just have StepSpec behave like a FlowSpec. We can add this back later when we have a better story and backend for it.

test/unit/graph_inference/flows/custom_branch_flow.py
```diff
@@ -0,0 +1,32 @@
+from metaflow import FlowSpec, step
```
romain-intel, line 1: Modify these tests using the @step(start=True) and @step(end=True) and @step(start=True, end=True) syntax.

test/unit/graph_inference/test_stepspec.py
```diff
@@ -0,0 +1,40 @@
+"""
```
romain-intel, line 1: We need significantly more tests here:
  - need to verify they work with configs
  - need to verify that we can apply decorators to the StepSpec (multiple decorators as well)
  - Need to verify that flow mutators, step mutators and step decorators can apply to a StepSpec as well.

romain-intel (commented):
Comments to update PR.

metaflow/plugins/aws/step_functions/step_functions.py
```diff
@@ -271,17 +273,21 @@ def get_execution(cls, state_machine_name, name):
...
-                .get("States")
-                .get("start")
-                .get("Parameters")
-                .get("ContainerOverrides")
-                .get("Environment")
-            )
-            parameters = {
-                item.get("Name"): item.get("Value") for item in environment_vars
-            }
+            definition = json.loads(state_machine.get("definition"))
```
romain-intel, line 276: What is this change?

metaflow/runtime.py
```diff
@@ -514,6 +514,37 @@ def persist_constants(self, task_id=None):
         if not self._params_task.is_cloned:
             self._params_task.persist(self._flow)
 
+            # Register start/end step metadata on _parameters task so the
```
romain-intel, line 517: There should be a way to make this work even for schedulers. THey do persist stuff in the parameters task too. Maybe move direclty to task?

romain-intel (commented):
minor comments with a quick review.

metaflow/plugins/aws/step_functions/step_functions.py
talsperre: Addressed - it was a spurious change and will be removed.

metaflow/runtime.py
talsperre: This already works for schedulers as well, since they call `persist_constants` in the init command, right? Updated the tests to verify this.

metaflow/graph.py
```diff
@@ -48,32 +48,61 @@ def deindent_docstring(doc):
...
+        wrappers,
+        config_decorators,
+        doc,
+        source_file,
+        lineno,
+        is_start_step=False,
+        is_end_step=False,
+        node_info=None,
+        name=None,
+        num_args=None,
```
romain-intel, line 90: This addition of num_args is a bit weird. In some cases it is parsed from the ast but when you don't have an ast, you need to pass it in. Maybe add a comment on that or something?

metaflow/decorators.py
```diff
@@ -588,6 +588,19 @@ def _base_step_decorator(decotype, *args, **kwargs):
         func = args[0]
         if isinstance(func, (StepMutator, UserStepDecoratorBase)):
             func = func._my_step
+
+        # Step decorator applied to a class with a synthesized step method.
```
romain-intel, line 592: I would add a comment here saying more explicitly that this code is used by a potentially upcoming feature (the FunctionSpec) and although `_function_spec_step_name` is not present anywhere else, it needs to remain here.

metaflow/graph.py
```diff
@@ -259,29 +287,122 @@ def __init__(self, flow):
...
+        else:
+            if self.start_step:
+                node = self.nodes[self.start_step]
+                if node.type in (None, "linear"):
+                    node.type = "start"
+            if self.end_step:
+                self.nodes[self.end_step].type = "end"
+
+    def _create_sourceless_single_step_node(
+        self, name, func, is_start_step, is_end_step
```
romain-intel, line 377: we should probably add a bit more detail here about that this is also used for this upcoming feature. We can also say that this is possibly temporary because it does look a bit weird.

metaflow/graph.py
```diff
@@ -259,29 +287,122 @@ def __init__(self, flow):
...
-                    lineno,
-                )
+                is_start = getattr(func, "is_start_step", False)
+                is_end = getattr(func, "is_end_step", False)
+
+                try:
+                    source_file = inspect.getsourcefile(func) or inspect.getfile(func)
+                    source_lines, lineno = inspect.getsourcelines(func)
+                except OSError:
+                    if is_start and is_end:
```
romain-intel, line 432: Add a comment here too that we support cases where the single step is actually generated.

romain-intel (commented):
Minor nits. The workaround for no code will probably see evolution but it's not user facing (much). I'd just clarify a bit why it's there and we can merge.

metaflow/runtime.py
```diff
@@ -514,6 +514,37 @@ def persist_constants(self, task_id=None):
...
+                            tags=[],
+                        ),
+                        MetaDatum(
+                            field="end_step",
+                            value=self._graph.end_step,
+                            type="graph_structure",
+                            tags=[],
+                        ),
+                    ],
+                )
```
saikonen, line 539: Tangential review notes: Losing the specific `start/end` steps will require significant changes to MFGUI in order to track stuff correctly again. The most glaring issue with this is that runs do not switch to `completed` any more due to a missing end-step.

changes will require joining this graph structure metadata for the mfgui queries for comparison purposes.

saikonen: One addition that would help some of the UI modifications. Would it make sense to record some metadata for the specific task as well denoting if its a start or end step? that would be something that `LISTEN/NOTIFY` can pick up on independently, without requiring a round-trip to the DB

saikonen: actually strike that.. the signaling for the UI might require a bit more thought than simply adding more metadata, due to the restrictions with `listen/notify` only working off what is inserted into the DB on a per-record basis. What the current functionality relies on is
- an entry that signals the `start` task has started
- an entry that signals `end` task has finished

metaflow/plugins/aws/step_functions/step_functions.py
talsperre: Reversing course. Keeping this. get_execution is a classmethod with no self.graph, and once the start-state name is variable, `StartAt` is the only source of truth. Guards replace an opaque AttributeError with a typed exception. Same pattern as get_existing_deployment above.

metaflow/plugins/cards/ui/src/components/dag/dag.svelte
```diff
@@ -42,8 +42,14 @@
   style="position: relative; line-height: 1"
   data-component="dag"
 >
-  {#if steps?.start}
-    <StepWrapper {steps} stepName="start" bind:dagStructure />
+  {#if startStep && steps?.[startStep]}
```
saikonen, line 45: modifications here require building the node project as well, as there is no CI taking care of building it. Update the main.js with
```
npm run build
````

File: metaflow/graph.py [hunk 14]
```diff
@@ -259,29 +321,131 @@ def __init__(self, flow):
...
         self.sorted_nodes = []
+        self._identify_start_end()
         self._traverse_graph()
         self._postprocess()
 
+    def _identify_start_end(self):
+        """
+        Determine the start and end steps.
+
+        Uses explicit @step(start=True) / @step(end=True) annotations if present.
```
saikonen, line 332: Just a question on the implementation here, not a blocker. Were there other reasons besides simplicity to require explicit `start=True / end=True` annotations for steps? Technically it should be possible to detect these without requiring the user to annotate them.

romain-intel: We went back and forth on this and first had a solution where we would calculate in and out degrees for nodes. We thought that something more explicit was easier to understand for the user. TO be clear, there is *no* need for the user to do this except in two cases:
 - they absolutely want to change the name of the start or end step
 - they only have one step. We could change this case (or maybe it's already ok -- I can check) where if there is only one step, we *assume* it is start and end, whatever it is called. 

Are you asking about that second part?

talsperre: Just simplicity, honestly. We could always do a topological sort to figure it out but that adds many checks/edge cases.

metaflow/plugins/cards/ui/src/components/dag/dag.svelte
talsperre: Follow on PR or here itself?

saikonen: maybe here so its not omitted. Without rebuilding, the dag component doesn't render correctly for flows without start/end steps.

metaflow/graph.py
saikonen: From an UX perspective I'd be more in favour of an implicit start / end, especially for the single-step case, as if I understood correctly that would be denoted as
```python
@step(start=True, end=True)
def single_step(self)
# vs
@step
def single_step(self)
```

this is more of a personal preference though so again, not a blocker for merging.

talsperre: Addressed.

metaflow/plugins/cards/ui/src/components/dag/dag.svelte
talsperre: Addressed.

talsperre (approved):

talsperre:
Let's go!!!

PR #3122, opened 2026-04-16
fix: skip CronWorkflow creation when @schedule resolves to no schedule

### LINKED ISSUE #3121: @schedule(daily=False) on Argo Workflows deploy crashes with AttributeError and should not create a CronWorkflow
## Description

When deploying a flow to Argo Workflows with `@schedule` configured such that no schedule is active (e.g. `@schedule(daily=False)` or any combination where all flags are falsy), the deploy crashes with an `AttributeError`. Even if it did not crash, the expected behaviour for this case would be that **no CronWorkflow is created at all**, not a suspended one.

This is a common pattern for teams deploying the same flow code across multiple environments: production gets a real schedule, while PR/staging deploys are unscheduled and purely for end-to-end testing. Creating a CronWorkflow in those environments is unnecessary and creates ongoing cleanup burden.

Note: PR #3102 proposes a `--no-enable-schedule` flag that deploys with `suspend: true` on the CronWorkflow. That does not address this use case — a suspended CronWorkflow still exists as a resource to manage and clean up.

## Steps to Reproduce

```python
from metaflow import FlowSpec, step, schedule

@schedule(daily=False)
class MyFlow(FlowSpec):
    @step
    def start(self):
        self.next(self.end)

    @step
    def end(self):
        pass

if __name__ == "__main__":
    MyFlow()
```

```
python myflow.py argo-workflows create
```

## Expected Behaviour

No CronWorkflow is created in Argo Workflows. The WorkflowTemplate is deployed normally, without any CronWorkflow resource being created or registered.

## Actual Behaviour

```
AttributeError: 'NoneType' object has no attribute 'split'
```

The crash is in `_get_schedule()` in `argo_workflows.py:463`:

```python
def _get_schedule(self):
    schedule = self.flow._flow_decorators.get("schedule")
    if schedule:
        schedule = schedule[0]
        return " ".join(schedule.schedule.split()[:5]), schedule.timezone  # crashes here
    return None, None
```

`ScheduleDecorator.flow_init` sets `self.schedule = None` when all flags are falsy (`schedule_decorator.py:49`). The `if schedule:` guard checks whether the decorator is *present*, not whether a schedule was actually resolved. Calling `.split()` on `None` raises the error.

## Root Cause

`schedule_decorator.py:40-49`:
```python
if self.attributes["cron"]:
    self.schedule = self.attributes["cron"]
elif self.attributes["weekly"]:
    self.schedule = "0 0 ? * SUN *"
elif self.attributes["hourly"]:
    self.schedule = "0 * * * ? *"
elif self.attributes["daily"]:
    self.schedule = "0 0 * * ? *"
else:
    self.schedule = None   # <-- valid state: decorator present, no schedule active
```

`_get_schedule()` does not handle the case where the decorator is present but `schedule.schedule is None`.

## Proposed Fix

`_get_schedule()` should treat `schedule.schedule is None` the same as the decorator being absent — return `(None, None)`, which causes the rest of the deploy path to skip CronWorkflow creation entirely:

```python
def _get_schedule(self):
    schedule = self.flow._flow_decorators.get("schedule")
    if schedule:
        schedule = schedule[0]
        if schedule.schedule is None:
            return None, None
        return " ".join(schedule.schedule.split()[:5]), schedule.timezone
    return None, None
```

The same condition in `trigger_explanation()` (`argo_workflows.py:485`) also needs updating to check `schedule.schedule is not None`, otherwise it will incorrectly report "This workflow triggers automatically via CronWorkflow" even when no CronWorkflow was created.

## Environment

- Metaflow version: (please fill in)
- Argo Workflows version: (please fill in)

## Related

- Issue #1169 — same motivation, focuses on Step Functions
- PR #3102 — related, but takes a different approach (suspending the CronWorkflow rather than omitting it)

KnutZuidema:
## PR Type

- [x] Bug fix

## Summary

When `@schedule` is present but all scheduling flags are falsy (e.g. `@schedule(daily=False)`), deploying to Argo Workflows crashes with `AttributeError: 'NoneType' object has no attribute 'split'` and no WorkflowTemplate is created. This fix makes the deploy succeed and skip CronWorkflow creation, which is the expected behaviour when no schedule is configured.

## Issue

Fixes #3121

## Reproduction

**Runtime:** argo

**Commands to run:**
```bash
python myflow.py argo-workflows create
```

Where `myflow.py` contains:

```python
from metaflow import FlowSpec, step, schedule

@schedule(daily=False)
class MyFlow(FlowSpec):
    @step
    def start(self):
        self.next(self.end)

    @step
    def end(self):
        pass

if __name__ == "__main__":
    MyFlow()
```

**Where evidence shows up:** parent console / `argo-workflows create` output

<details>

```
AttributeError: 'NoneType' object has no attribute 'split'
  File ".../metaflow/plugins/argo/argo_workflows.py", line 463, in _get_schedule
    return " ".join(schedule.schedule.split()[:5]), schedule.timezone
```

</details>

<details>

```
Workflow template 'MyFlow' created successfully.
No CronWorkflow created (no schedule configured).
```

</details>

## Root Cause

`ScheduleDecorator.flow_init` sets `self.schedule = None` when all scheduling flags are falsy (`schedule_decorator.py:49`). `_get_schedule()` in `argo_workflows.py` guarded against the decorator being absent (`if schedule:`), but not against `schedule.schedule` being `None`. Calling `.split()` on `None` raises `AttributeError`. Separately, `trigger_explanation()` checked decorator presence rather than `self._schedule`, so it would incorrectly report CronWorkflow-based triggering even when no schedule was resolved.

## Why This Fix Is Correct

`_get_schedule()` now returns `(None, None)` when `schedule.schedule is None`, treating it identically to the decorator being absent. The rest of the deploy path already handles `self._schedule is None` correctly — no CronWorkflow is created. `trigger_explanation()` now checks `self._schedule is not None`, which is the actual resolved value, rather than decorator presence.

## Failure Modes Considered

1. **Flows with a real schedule are unaffected** — the `schedule.schedule is None` guard is only entered when all flags are falsy; any valid cron string, `daily=True`, `hourly=True`, or `weekly=True` sets a non-None value and takes the existing return path.
2. **Backward compatibility** — the previous behaviour for the `@schedule(daily=False)` case was a crash, so changing it to a clean no-op is strictly an improvement with no risk of regression.

## Tests

- [ ] CI passes

The bug is straightforward to reproduce with the script above. Unit test coverage for `_get_schedule()` with a `None` schedule value would be a good addition.

## Non-Goals

This fix does not address the Step Functions path (which returns `schedule.schedule` directly and handles `None` safely already), does not implement an `--no-enable-schedule` flag (see PR #3102), and does not change any scheduling behaviour for flows where a schedule is actually configured.

## AI Tool Usage

OpenCode (Claude Sonnet) was used to identify the bug, draft the issue, and generate the fix. All generated code was reviewed and understood before committing.

### DIFF
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -1827,9 +1827,10 @@ class ArgoWorkflows(object):
                                     if not self._is_conditional_join_node(
                                         self.graph[node.matching_join]
                                     )
+                                    else
                                     # Note: If the nodes leading to the join are conditional, then we need to use an expression to pick the outputs from the task that executed.
                                     # ref for operators: https://github.com/expr-lang/expr/blob/master/docs/language-definition.md
-                                    else {
+                                    {
                                         "expression": "get((%s)?.parameters, 'task-id')"
                                         % " ?? ".join(
                                             f"tasks['{self._sanitize(func)}']?.outputs"
@@ -2066,7 +2067,7 @@ class ArgoWorkflows(object):
                 )
                 task_str = "(t-%s)" % _task_id_base
 
-            task_id_expr = "export METAFLOW_TASK_ID=%s" % task_str
+            task_id_expr = "export METAFLOW_TASK_ID=" "%s" % task_str
             task_id = "$METAFLOW_TASK_ID"
 
             # Resolve retry strategy.
@@ -2419,7 +2420,7 @@ class ArgoWorkflows(object):
                     env[
                         "METAFLOW_ARGO_EVENT_PAYLOAD_%s_%s"
                         % (event["type"], event["sanitized_name"])
-                    ] = "{{workflow.parameters.%s}}" % event["sanitized_name"]
+                    ] = ("{{workflow.parameters.%s}}" % event["sanitized_name"])
 
             # Map S3 upload headers to environment variables
             if S3_SERVER_SIDE_ENCRYPTION is not None:
@@ -3866,8 +3867,7 @@ class ArgoWorkflows(object):
                 .annotations(self._base_annotations)
             )
             .spec(
-                SensorSpec()
-                .template(
+                SensorSpec().template(
                     # Sensor template.
                     SensorTemplate()
                     .metadata(
@@ -4028,8 +4028,7 @@ class ArgoWorkflows(object):
                             # @trigger(options={"reset_at": {"cron": , "timezone": }})
                             # timezone is IANA standard, e.g. America/Los_Angeles
                             # TODO: Introduce "end_of_day", "end_of_hour" ..
-                        )
-                        .conditions_reset(
+                        ).conditions_reset(
                             cron=self.trigger_options.get("reset_at", {}).get("cron"),
                             timezone=self.trigger_options.get("reset_at", {}).get(
                                 "timezone"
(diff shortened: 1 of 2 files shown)


talsperre:
Thanks for the PR. Can you add a test for it as well?

KnutZuidema:
@talsperre done

KnutZuidema:
@talsperre I've ran pre-commit to fix the formatting, but I'm not sure whether there is anything to do regarding the test failures in R tests or regarding the code coverage.

talsperre:
@KnutZuidema, thanks for the PR - I made a bunch of pushes on top of yours that changed the structure of the tests. Once the CI passes, I will merge it.

PR #3125, opened 2026-04-17
Fix pip config parsing for multiple extra-index-urls

harkrish-1:
## Summary
- Fix `Pip.indices()` to correctly parse multiple `extra-index-url` values from pip config
- When pip config contains multiple extra-index-urls, `pip config list` outputs them separated by literal `\n` characters (backslash + n, not actual newlines)
- The previous regex (`\s+`) only split on whitespace, so multiple URLs were treated as a single mangled value
- The fix adds literal `\n` as an additional split delimiter: `r"\\n|\s+"` 

## Tests

- [ ] CI passes

## AI Tool Usage

Used claude code to identify and fix this bug, I completely understand the changes made


metaflow/plugins/pypi/pip.py
```diff
@@ -324,7 +324,7 @@ def indices(self, prefix):
                 key, value = line.split("=", 1)
                 _, key = key.split(".")
                 if key in ("index-url", "extra-index-url"):
-                    values = map(lambda x: x.strip("'\""), re.split("\s+", value, re.M))
+                    values = map(lambda x: x.strip("'\""), re.split(r"\\n|\s+", value, re.M))
```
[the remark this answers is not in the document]
harkrish-1: Didn't want to mess with any existing functionality so I am intentionally avoiding this. This is harmless except for if someone has more than 8 extra index indices, which is unlikely

harkrish-1:
@savingoyal @saikonen please review and add your feedback. Simple one line change, shouldn't need a lot of time to review.

test/unit/test_pip_indices.py
```diff
@@ -0,0 +1,24 @@
+from unittest.mock import patch
```
talsperre, line 1: Please use pytest, magicmock (`pytest-mock` plugin), and monkeypatch for the tests. Within pytest, we prefer to use fixtures and non-class-based tests.

talsperre:
Also make sure to setup the pre-commit hooks insce your code fails on the `pre-commit` checks in GitHub actions.

talsperre:
Also the appropriate location for this test file should be inside `test/plugins/pip` (https://github.com/Netflix/metaflow/tree/master/test/plugins) - i.e., create a new directory for pip and add this test in there.

metaflow/plugins/pypi/pip.py
saikonen: Change looks straightforward, good catch with this one.

Can you address the other pending review comments still? Otherwise this is good to go.

test/plugins/pip/test_pip_indices.py
```diff
@@ -0,0 +1,24 @@
+from unittest.mock import patch
```
harkrish-1: Done

metaflow/plugins/pypi/pip.py
harkrish-1: Thank you!

harkrish-1:
@talsperre @saikonen  addressed all comments!

harkrish-1:
@talsperre @saikonen  Please review this

saikonen (approved):

PR #3130, opened 2026-04-20
[Do not review] packaging: skip metaflow + extensions when every step has @conda/@pypi

talsperre:
## Summary

Introduces a per-decorator opt-in so `@conda` / `@pypi` / `@resolved_*` and
other env-providing decorators can declare that the remote environment will
install `metaflow` itself. When every step is covered (directly or via a
flow-level decorator / `FlowMutator`), `MetaflowPackage` omits the metaflow
distribution and `metaflow_extensions` from the tarball — the remote side
reinstalls them from the user's declared dependencies anyway.

## Contract (opt-in, default off)

- `provides_metaflow_distribution: bool = False` class attribute
- `provides_metaflow_distribution_for(...)` instance hook

Added on `FlowDecorator`, `StepDecorator`, and `FlowMutator`. OSS `@conda`
and `@pypi` are **not modified** here — the opt-in will live on the Netflix
internal decorators (mli-metaflow-custom). This PR only ships the contract,
the packaging gate, the auto-detect helper, and orchestrator-side safety
nets.

## Orchestrator-side safety

Argo / Step Functions / Airflow / `metaflow package info|list` always
include metaflow — their DAG glue (`_parameters`, `python -m
metaflow.plugins.argo.*`, `set_batch_environment`, `set_parameters`) runs
outside any `@conda`/`@pypi` env. All four CLI call sites pass
`include_mf_distribution=True` explicitly; only the direct-compute path
(`run_cmds.py`) consults auto-detect.

## Auto-detect

- Iterates the graph via `flow._flow_decorators` + `flow._flow_mutators`.
- Skips the `_parameters` pseudo-step (evaluated in the submitting shell).
- Overrideable via `METAFLOW_PACKAGE_EXCLUDE_METAFLOW_DISTRIBUTION` in
  either direction.

## Other bits

- `metaflow/util.py`: new `str_to_bool(value, default, strict=False)`; the
  config system reuses it with `strict=False` to preserve existing lax
  semantics.
- `plugins/pypi/bootstrap.py`: defensive import with a targeted error
  message if metaflow is stripped but not installed remotely.
- `MetaflowCodeContentV1.show()` surfaces the skip so `metaflow … package
  info` is honest.

## Tests (real FlowSpecs, `monkeypatch`, `pytest.raises` — no mocks)

- `test/unit/test_str_to_bool.py`
- `test/unit/test_mf_distribution_packaging.py`
- `test/unit/test_code_content_v1.py`
- `test/unit/test_cli_package_always_includes.py`

## Test plan

- [ ] `tox -e unit` passes locally
- [ ] Submit a real flow with `@conda` on every step and verify the tarball
      contents omit `metaflow/__init__.py`
- [ ] Submit the same flow without `@conda` and verify metaflow is still in
      the tarball
- [ ] Argo / SF / Airflow deploy path still ships metaflow regardless of
      `@conda` coverage

### DIFF
--- a/metaflow/decorators.py
+++ b/metaflow/decorators.py
@@ -250,6 +250,16 @@ class FlowDecorator(Decorator):
     # Signature: flow_init_ctx(self, options)
     flow_init_ctx = None
 
+    # Class-level opt-in: when True, this flow decorator guarantees that
+    # metaflow (and any metaflow_extensions it depends on) will be importable
+    # in the remote environment for every step of the flow. If every step is
+    # covered (directly or via such a flow decorator / mutator),
+    # MetaflowPackage can skip bundling the metaflow distribution.
+    provides_metaflow_distribution: bool = False
+
+    def provides_metaflow_distribution_for(self, flow, step) -> bool:
+        return self.provides_metaflow_distribution
+
     def __init__(self, *args, **kwargs):
         super(FlowDecorator, self).__init__(*args, **kwargs)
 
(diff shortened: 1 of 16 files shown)


talsperre:
Closing this since it is marked draft / do-not-review and touches a large packaging/runtime surface. This should come back as a focused PR after the design is ready.

PR #3180, opened 2026-05-06
Fix invalid self.input access in spin complex DAG test

talsperre:
## Summary
- remove invalid `self.input` reads from non-foreach-body steps in the complex DAG spin test flow
- keep `self.input` usage only in steps where the step is directly executing a foreach item

## Validation
- `python -m pytest 'test/unit/spin/test_spin.py::test_simple_flows[complex_dag]' --confcutdir=test/unit/spin --use-latest -q`
- `pre-commit run --files test/unit/spin/flows/complex_dag_flow.py`

### DIFF
--- a/test/unit/spin/flows/complex_dag_flow.py
+++ b/test/unit/spin/flows/complex_dag_flow.py
@@ -42,7 +42,7 @@ class ComplexDAGFlow(FlowSpec):
 
     @step
     def step_e(self):
-        print(f"I am step E. Input is: {self.input}")
+        print("I am step E.")
         self.split_e = [9, 10]
         print("My output is: ", self.my_output)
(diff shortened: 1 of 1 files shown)


talsperre:
Correct behavior:

- `self.input` should be the actual input value only in direct foreach-body steps, e.g. `step_a`, `step_b`, `step_c`, and `step_f`.
- `step_e` is not a foreach-body step; it creates `split_e` and then starts a foreach.
- `step_l` is also not a foreach-body step; it is a normal branch after `step_j`.
- So for `step_e`/`step_l`, the test flow should not read `self.input` at all. It should not expect an actual value, and it should not rely on old `None` fallback behavior.

npow (approved):

PR #3197, opened 2026-05-13
fix: replace bare except clauses with specific exception types (#3164)

### LINKED ISSUE #3164: Replace bare except clauses in multicore_utils with specific exceptions
## Problem
`multicore_utils.py` contains two bare `except:` clauses that catch all exceptions including `SystemExit` and `KeyboardInterrupt`. This is a Python anti-pattern that can mask real problems and interfere with process management in forked child processes.

## Expected Behavior
- Import fallback should only catch `ImportError`
- Child process error handler should catch `Exception` but not `SystemExit`/`KeyboardInterrupt`

paulxg12:
Closes #3164 — replaces bare `except:` clauses in `multicore_utils.py` with specific exception types.

**Changes in `metaflow/multicore_utils.py`:**
- Line 25: `except:` → `except ImportError:` (py2/py3 cPickle compat pattern)
- Line 69: `except:` → `except Exception:` (child process error handler should allow SystemExit/KeyboardInterrupt to propagate to os._exit())

### DIFF
--- a/metaflow/multicore_utils.py
+++ b/metaflow/multicore_utils.py
@@ -22,7 +22,7 @@ from typing import (
 try:
     # Python 2
     import cPickle as pickle
-except:
+except ImportError:
     # Python 3
     import pickle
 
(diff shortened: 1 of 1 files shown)


talsperre:
Closing this because the `_spawn` handler needs to keep catching `BaseException` so child `SystemExit`/`KeyboardInterrupt` still print their traceback before `os._exit(...)`. The `cPickle` import change is fine, but the PR changes diagnostics in `parallel_map`.

PR #3210, opened 2026-05-22
fix: apply user-defined aws_batch_tags regardless of BATCH_EMIT_TAGS

### LINKED ISSUE #3209: aws_batch_tags / METAFLOW_BATCH_DEFAULT_TAGS silently ignored unless METAFLOW_BATCH_EMIT_TAGS=true
Tags set via `@batch(aws_batch_tags={"CostCenter": "123"})` or the `METAFLOW_BATCH_DEFAULT_TAGS` environment variable are silently dropped when `METAFLOW_BATCH_EMIT_TAGS` is not set. No warning or error is raised.

In enterprise deployments this is more than cosmetic: many organizations enforce **mandatory tagging policies** via AWS Service Control Policies (SCPs) or AWS Config rules. When required tags (e.g. `CostCenter`, `Team`, `Environment`) are absent, AWS rejects the resource operation entirely, causing the Batch job submission to fail with an access-denied error that gives no hint the root cause is a Metaflow configuration flag.

**Metaflow version:** master, `4871fb1`
**Deployment:** AWS Batch
**`METAFLOW_BATCH_EMIT_TAGS`:** not set (default `False`)

**Steps to reproduce**
1. Set `@batch(aws_batch_tags={"CostCenter": "123"})` (or `METAFLOW_BATCH_DEFAULT_TAGS='{"CostCenter": "123"}'`)
2. Do not set `METAFLOW_BATCH_EMIT_TAGS`
3. Run the flow

**Without a mandatory tagging SCP:** job submits but has no tags — silent data loss.
**With a mandatory tagging SCP:** job submission is rejected by AWS with an access-denied or tag-policy-violation error, making the flow completely unusable even though the user has correctly configured the required tags.

**Expected**
Tags explicitly provided by the user are applied to the job regardless of `METAFLOW_BATCH_EMIT_TAGS`. That flag is meant to control Metaflow's own internal observability tags (`app=metaflow`, `metaflow.flow_name`, etc.), not whether operator-defined cost-attribution or compliance tags take effect.

**Actual**
The `aws_batch_tags` block sits inside the [`if BATCH_EMIT_TAGS:` guard in `batch.py` (L380–L403)](https://github.com/Netflix/metaflow/blob/4871fb1e7d33de5cfb63f64c57ebe33304fd3376/metaflow/plugins/aws/batch/batch.py#L380-L403) (introduced in `6c7766a`). User-defined tags share the guard with Metaflow's internal metadata tags even though their purpose — and the consequences of omitting them — are entirely different.

odncode:
## Summary
Fixes #3209 

User-defined `aws_batch_tags` and `METAFLOW_BATCH_DEFAULT_TAGS` are currently dropped silently when `BATCH_EMIT_TAGS` is not set. This PR ensures that user-defined tags are always respected and applied to AWS Batch jobs, regardless of internal telemetry settings.

---

## Problem
The `aws_batch_tags` block in `batch.py` (L401-403) sits inside the `if BATCH_EMIT_TAGS:` guard (L380). 

While `BATCH_EMIT_TAGS` is intended to toggle Metaflow's internal observability tags (such as `app=metaflow` or `metaflow.flow_name`), it inadvertently gates user-defined tags meant for cost attribution or compliance. 

In enterprise environments with mandatory AWS Service Control Policies (SCPs) or AWS Config rules requiring specific tags, this causes Batch job submissions to fail with access-denied errors unless `BATCH_EMIT_TAGS` is enabled.

---

## Fix
* **Decoupled User Tags:** Moved the `aws_batch_tags` processing block outside of the `if BATCH_EMIT_TAGS:` guard.
* **Preserved Telemetry Rules:** Internal Metaflow tags remain gated; user-defined tags are now always applied.

---

## Testing
* Added `test/unit/test_batch_user_tags.py` — an AST-based structural test that parses `create_job` and verifies `aws_batch_tags` is not referenced inside any `if BATCH_EMIT_TAGS:` block.
* **Results:** Verified that the test **PASSES** with this fix, and **FAILS** without it.

---

## PR Checklist
- [x] Link tracking issue (`Fixes #3209`)
- [x] Add/update unit tests
- [ ] Update documentation (N/A)

Closes #3209


Shriprasad-P (commented):
## Review

**PR:** fix: apply user-defined aws_batch_tags regardless of BATCH_EMIT_TAGS

**Touched:** `metaflow/plugins/aws/batch/batch.py`, `metaflow/plugins/aws/batch/batch_decorator.py`, `test/unit/test_batch_user_tags.py`

- Size looks manageable (+250/-3).

Issue #3161, opened 2026-04-29
CI: Improvements for the cards component


saikonen:
Managing the nested npm project for the cards component could benefit from additional CI. Right now we only have verification in place that the project can be built successfully, which was required to have some level of trust for Dependabot package upgrade suggestions.

Some ideas for possible improvements

## Auditing packages
Dependabot is the main source for package version upgrades, but it should be considered whether these should be blindly trusted. At minimum, we could add `npm audit` as part of the PR checks when the changes touch the cards component

## Build&Commit card changes
When changing any code in the Cards UI, or upgrading packages, this will not have any effect for the component itself unless the project is also built. Currently there is no CI present, instead the author needs to manually rebuild the assets as part of their PR.

This could be automated as well, with a step taking care or rebuilding the assets and adding a commit to the PR. Some security considerations might apply when implementing commit rights for an action.

rajat343:
Hey, happy to pick this up but want to scope it down before I write any YAML.

Mind if I start with just the `npm audit` piece, scoped to PRs that touch `metaflow/plugins/cards/ui/`? If yes, what audit level should fail the job  (high, or only critical)? And should it block merge or just warn?

odncode:
Hey @saikonen, I'd like to take a crack at this. I'll start with the npm audit piece scoped to PRs touching metaflow/plugins/cards/ui/, similar to what @rajat343 suggested. Happy to expand to the build & commit automation after.

PR #3212, opened 2026-05-22
ci: add npm audit check for cards UI component

odncode:
## Summary
Adds `npm audit` as a **CI check** for the cards **UI** component, as proposed in **#3161**.

## Context
**Dependabot** handles package upgrades for the cards **UI**, but there's currently no automated security check to catch vulnerabilities introduced by those upgrades (or missed by them). This workflow adds that safety net.

## Changes Made
* New workflow `.github/workflows/test-card-audit.yml` that runs `npm audit --audit-level=critical` on **PRs** touching `metaflow/plugins/cards/ui/`
* Mirrors the structure of the existing `test-card-build.yml` workflow

## Design Decisions
* `--audit-level=critical` rather than `high`: The cards **UI** currently has 10 high-severity vulnerabilities in its dependency tree (`lodash`, `minimatch`, `rollup`, `picomatch`, `flatted`). Using `high` would immediately block all cards **PRs**. Starting with `critical` ensures the workflow is useful now; the threshold can be tightened after the existing vulns are addressed in a separate dependency update.
* Runs on `npm ci` not `npm install`: Matches the existing build workflow and ensures reproducible installs from the lockfile.

## Existing Vulnerabilities
For visibility, running `npm audit` locally reports 28 total vulnerabilities: 3 low, 15 moderate, 10 high, 0 critical. The high-severity packages include `lodash`, `minimatch`, `rollup`, `picomatch`, and `flatted`. These should be addressed separately.


agsaru (commented):
There is already a file `test-card-build.yml`. There is no need of this extra file for test card.

odncode:
> There is already a file `test-card-build.yml`. There is no need of this extra file for test card.

That makes sense, I'll add the audit step to the existing `test-card-build.yml` instead. Will update the PR.

.github/workflows/test-card-build.yml
```diff
@@ -21,3 +21,18 @@ jobs:
...
+        working-directory: ./metaflow/plugins/cards/ui
+    steps:
+      - uses: actions/checkout@v6
+      - name: Use Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: '20.x'
+      - run: npm ci
+      - name: Run npm audit
+        run: npm audit --audit-level=critical
```
saikonen: for simplicity this could be a step before the `npm run build` in the testbuild job, instead of its own job with a lot of duplicate setup.

saikonen (approved):

PR #3219, opened 2026-05-30
Fix local storage caching issue

ankush0545:
## Summary

Fixes an issue in LocalStorage where existing files could remain stale due to caching behavior when writing data to the local datastore.

## Changes

* Updated LocalStorage save logic to handle file writes more consistently.
* Improved behavior when files already exist.
* Ensured metadata and file contents remain synchronized after updates.

## Testing

* Reproduced the issue locally.
* Verified that updated files are correctly reflected after writing.
* Ran existing tests and confirmed they pass.

## Impact

This change improves reliability when using LocalStorage as a datastore backend and helps prevent stale cached data from being served.


talsperre:
Can you show how your reprodcuded the bug and how you fixed this?

PR #3222, opened 2026-05-31
Fix: Prevents incorrectly attaching of default decorators in spin-step

### LINKED ISSUE #3221: Bug: spin-step incorrectly attaches default decorators, causing not supported decorator warnings
When running a `spin-step`, the CLI logic in `metaflow/cli.py` fails to skip the default decorators (`DEFAULT_DECOSPECS`). Because of the order of the `if/elif` conditions, the CLI tries to attach default decorators , which are not supported in spin steps. This causes the internal error to throw multiple warnings before running.

**How To Reproduce**
```python

class TestFlow(FlowSpec):
    @step
    def start(self):
        self.next(self.end)

    @step
    def end(self):
        pass

if __name__ == '__main__':
    TestFlow()
```

Run a spin-step while passing a default decorator via environment variables:

```bash
METAFLOW_DEFAULT_DECOSPECS="retry" python flow.py spin-step start --run-id spin-test-123 --task-id 1
```

## Expected behavior

The CLI should recognize it is running in `spin-step` mode, avoid attaching `DEFAULT_DECOSPECS`, and only show `AttributeError` because of the fake run-id.

## Actual behavior

The CLI attaches the default decorators, resulting in repeated warnings such as:

```text
Metaflow 2.19.30 executing TestFlow for user:agsaru
[Warning] Ignoring Step decorator 'retry' as it is not supported in spin steps.
[Warning] Ignoring Step decorator 'retry' as it is not supported in spin steps.
[Warning] Ignoring Step decorator 'retry' as it is not supported in spin steps.
[Warning] Ignoring Step decorator 'retry' as it is not supported in spin steps.
```

agsaru:
## PR Type

- [x] Bug fix
- [ ] New feature
- [ ] Docs / tooling
- [ ] Refactoring

### Description
This PR fixes a bug in `metaflow/cli.py` where default decorators were incorrectly being attached when running a `spin-step`. 

I fixed the order of the `if`/`elif` blocks. By checking for `"spin-step"` first, we ensure it correctly assigns an empty list for `all_decospecs` and safely bypasses the block that attaches the default decorators.

And also removes the duplicate os module import.

## Issue

Fixes #3221

### DIFF
--- a/.github/workflows/ux-tests.yml
+++ b/.github/workflows/ux-tests.yml
@@ -114,7 +114,11 @@ jobs:
 
       - name: Run spin tests
         shell: bash -el {0}
-        run: tox -e spin -- --junit-xml=junit-spin.xml
+        run: |
+          export CONDA_PKGS_DIRS="$RUNNER_TEMP/metaflow-micromamba-pkgs"
+          export MAMBA_ROOT_PREFIX="$RUNNER_TEMP/metaflow-micromamba-root"
+          export METAFLOW_TOKEN_HOME="$RUNNER_TEMP/metaflowconfig"
+          tox -e spin -- --junit-xml=junit-spin.xml
 
       - name: Upload test results
         if: always()
--- a/.github/workflows/ux-tests.yml
+++ b/.github/workflows/ux-tests.yml
@@ -103,10 +103,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Set up Python 3.10
+      - name: Set up Python 3.9
         uses: actions/setup-python@v5
         with:
-          python-version: "3.10"
+          python-version: "3.9"
 
       - name: Install tox
         run: python -m pip install tox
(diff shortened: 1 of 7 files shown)


talsperre:
I think this clears too much. For `spin-step` we only want to skip appending `DEFAULT_DECOSPECS`; we still need to preserve explicit/env decospecs forwarded via `--with`.

Could this be:

```python
if ctx.saved_args[0] not in ("step", "init", "spin-step"):
    all_decospecs += DEFAULT_DECOSPECS.split()
```

instead of clearing `all_decospecs`?

talsperre (commented):
Added a comment. Also, lets re-enable the spin tests in our github actions, and also add a small regression test for this.

agsaru:
@talsperre should not we use python version 3.9 rather than 3.10 . Other jobs are using 3.9 in the `ux-test.yml` workflow

talsperre:
@agsaru I was just testing `3.10` - you can change it back to `3.9` again. Also, thanks for the contribution. Appreciate it.

agsaru:
@talsperre Thanks for the Appreciation. Reverted the commit because of runner uses python version 3.9.25. But conda have the version till 3.9.23. 
Screenshot 2026-06-02 033220

PR #3225, opened 2026-05-31
fix(cards): use time.monotonic() for async-process timeout

### LINKED ISSUE #3224: card_creator: use time.monotonic() for async-process timeout instead of wall-clock time.time()
## Summary

`metaflow/plugins/cards/card_creator.py` measures the elapsed runtime of async card-rendering subprocesses using `time.time()` — a wall-clock that can jump backward (NTP step, manual `settimeofday`, DST transitions on non-UTC systems). The correct API for monotonically-increasing elapsed-time measurement is `time.monotonic()`.

## Locations

In `metaflow/plugins/cards/card_creator.py` (at HEAD `90ea1d3`):

- **L31** — `CardProcessManager._register_card_process()` stores the start time:
  ```python
  cls.async_card_processes[carduuid] = {
      "proc": proc,
      "started": time.time(),
  }
  ```
- **L200** — `_wait_for_async_processes_to_finish()` reads it:
  ```python
  if time.time() - _async_started > async_timeout:
      CardProcessManager._remove_card_process(card_uuid)
      break
  ```
- **L231** — `_run_command()` (async branch) reads it again:
  ```python
  if time.time() - _async_started > async_timeout:
      CardProcessManager._remove_card_process(card_uuid)
  ```

## Why this matters

`time.time()` reflects the system wall-clock; under NTP correction, container clock-skew rebase, or a manual time adjustment, two `time.time()` calls separated in real-time may differ in either direction. The two failure modes both bite metaflow users running Cards in long-lived environments (Kubernetes, Batch):

1. **Clock jumps backward** during a long-running card render → `time.time() - _async_started` becomes small or negative → the timeout never fires → the card subprocess leaks indefinitely (held in `CardProcessManager.async_card_processes` and never cleaned up).
2. **Clock jumps forward** → the elapsed delta inflates → the subprocess is killed prematurely and the user sees a missing/incomplete card.

`time.monotonic()` is the documented stdlib API for this case: it cannot go backward and is unaffected by system-time adjustments. Python docs are explicit: *"Use this function to monitor differences between calls."*

## Proposed fix

Three-line change — replace `time.time()` at L31 / L200 / L231 with `time.monotonic()`. No other behavioral change; the comparison is still an interval, the stored value is still opaque to all readers.

## Test plan

Unit test in `test/unit/` that monkeypatches `time.time` to simulate a backward jump, registers a card process with the patched start time, then verifies the timeout check still fires correctly when the implementation is `time.monotonic()`-based. (Will include in the PR.)

## AI tool disclosure

Bug surfaced by an audit grep for `time.time()` used in interval contexts (R5 rule in our local PR-audit toolkit). The analysis, fix, and PR will be authored by me — I understand both `time.time` and `time.monotonic` semantics and have verified the call sites read as written above.

Happy to take this on if accepted.

alexzhu0:
Fixes #3224

## What

`metaflow/plugins/cards/card_creator.py` measured async card-rendering subprocess lifetime with `time.time()`. Replaced three call sites with `time.monotonic()` and added a regression unit test.

## Why

`time.time()` returns the wall-clock and can step backward under NTP correction, container clock-skew rebase, or manual `settimeofday`. Two failure modes both bite long-running Card jobs:

1. **Backward jump** → `time.time() - _async_started` becomes small or negative → timeout check never fires → the card subprocess leaks indefinitely in `CardProcessManager.async_card_processes`.
2. **Forward jump** → elapsed delta inflates → subprocess killed prematurely → user sees a missing/incomplete card.

`time.monotonic()` is the documented stdlib API for interval measurement; it cannot move backward and is unaffected by system-clock adjustments.

## How

Three-line change in `metaflow/plugins/cards/card_creator.py`:

- L31 — `CardProcessManager._register_card_process()` now stores `time.monotonic()` instead of `time.time()`.
- L200 — `_wait_for_async_processes_to_finish()` compares with `time.monotonic()`.
- L231 — `_run_command()` async branch compares with `time.monotonic()`.

The stored value is opaque to all readers, so swapping the source is safe — there is no API change, no on-disk format change, and no cross-process value that needs to remain compatible.

## Tests

New file `test/unit/test_card_creator_monotonic.py` with three tests:

1. `test_register_uses_monotonic_not_wall_clock` — registry stores the monotonic timestamp, not `time.time()`.
2. `test_timeout_check_immune_to_wall_clock_backward_jump` — simulates a 1-day wall-clock backward jump while monotonic advances 600 s; verifies the subprocess is still reaped at the 60 s timeout.
3. `test_timeout_check_not_triggered_when_within_window` — sanity check that processes still inside the timeout window are not killed.

Local run:

```
$ pytest test/unit/test_card_creator_monotonic.py -v
test/unit/test_card_creator_monotonic.py::test_register_uses_monotonic_not_wall_clock PASSED
test/unit/test_card_creator_monotonic.py::test_timeout_check_immune_to_wall_clock_backward_jump PASSED
test/unit/test_card_creator_monotonic.py::test_timeout_check_not_triggered_when_within_window PASSED
============================== 3 passed in 0.92s ===============================
```

## Checklist

- [x] This PR addresses a single concern (one bug fix)
- [x] The diff is reasonably sized and easy to review (3 lines fix + ~110 lines tests)
- [x] New functionality is covered by tests (regression test for the backward-jump failure mode)
- [ ] Linting/type checking pass — will rely on CI to confirm; pre-commit not run locally
- [x] No unrelated changes or drive-by fixes are included

## AI tool disclosure

Bug surfaced by an audit grep for `time.time()` used in interval contexts (an R5 rule in my local PR-audit toolkit). I authored the fix and tests, verified the call sites read as written, ran the tests locally, and can defend the monotonic-vs-wall-clock semantics without reference to an AI tool.


alexzhu0:
Addressed the P2 finding from Greptile: updated the schema comment on `CardProcessManager.async_card_processes` (L20-25) to reference `time.monotonic()` so it matches the implementation. No behavior change.

Pushed as `fd46593`.

PR #3227, opened 2026-06-01
security: validate tar package extraction paths

dfgvaetyj3456356-hash:
## PR Type

- [x] Bug fix
- [ ] New feature
- [ ] Docs / tooling
- [ ] Refactoring

## Summary

Validate tar package member paths before extracting Metaflow code packages. This prevents archive entries, symlinks, or hardlinks from writing outside the requested extraction directory.

## Issue

No public issue. This is a defensive package extraction hardening fix.

## Reproduction

**Runtime:** local package extraction

**Commands to run:**
```bash
python -m pytest test/unit/test_tar_packaging_backend.py -q
```

<details>

```
A crafted tar member named ../escaped.txt could be accepted by cls_extract_members()
and written outside the extraction directory.
```

</details>

<details>

```
..                                                                       [100%]
2 passed in 1.87s
```

</details>

## Root Cause

`TarPackagingBackend.cls_extract_members()` passed selected archive members directly to `tarfile.extractall()` without validating that member paths resolved under `dest_dir`. The same extraction primitive is used by `MetaflowPackage.cls_extract_into()`, so package extraction relied on tarfile's default behavior for path handling.

## Why This Fix Is Correct

The backend now validates each selected member before extraction using `os.path.commonpath()` against the absolute destination directory. Symlink and hardlink targets are also resolved and checked, so a link entry cannot point outside the destination.

## Failure Modes Considered

1. Parent-directory or absolute member names escaping `dest_dir`.
2. Symlink or hardlink entries targeting paths outside `dest_dir`.

## Tests

- [ ] CI passes

## Non-Goals

No changes to package creation, archive format selection, or content-type filtering.

## AI Tool Usage

Codex assisted with auditing the extraction path, drafting the minimal fix, and running the focused regression checks. I reviewed the code path and verified the before/after behavior locally.

### DIFF
--- a/metaflow/packaging_sys/tar_backend.py
+++ b/metaflow/packaging_sys/tar_backend.py
@@ -1,9 +1,56 @@
+import inspect
+import os
 import tarfile
 
 from io import BytesIO
 from typing import Any, IO, List, Optional, Union
 
 from .backend import PackagingBackend
+
+
+_EXTRACTALL_SUPPORTS_FILTER = (
+    "filter" in inspect.signature(tarfile.TarFile.extractall).parameters
+)
+
+
+def _member_target_path(dest_dir: str, member_name: str) -> str:
+    return os.path.abspath(os.path.join(dest_dir, member_name))
+
+
+def _ensure_within_directory(dest_dir: str, target_path: str) -> None:
+    abs_dest = os.path.abspath(dest_dir)
+    try:
+        is_within_dest = os.path.commonpath([abs_dest, target_path]) == abs_dest
+    except ValueError:
+        is_within_dest = False
+    if not is_within_dest:
+        raise tarfile.ExtractError("Attempted path traversal in TAR file")
+
+
+def _validate_member(dest_dir: str, member: tarfile.TarInfo) -> None:
+    target_path = _member_target_path(dest_dir, member.name)
+    _ensure_within_directory(dest_dir, target_path)
+
+    if member.issym():
+        if os.path.isabs(member.linkname):
+            link_target = os.path.abspath(member.linkname)
+        else:
+            link_target = os.path.abspath(
+                os.path.join(os.path.dirname(target_path), member.linkname)
+            )
+        _ensure_within_directory(dest_dir, link_target)
+    elif member.islnk():
+        link_target = os.path.abspath(os.path.join(dest_dir, member.linkname))
+        _ensure_within_directory(dest_dir, link_target)
+
+
+def _extractall_preserving_validated_members(
+    archive: tarfile.TarFile, dest_dir: str, members: List[Any]
+) -> None:
+    kwargs = {}
+    if _EXTRACTALL_SUPPORTS_FILTER:
+        kwargs["filter"] = "tar"
+    archive.extractall(path=dest_dir, members=members, **kwargs)
 
 
 class TarPackagingBackend(PackagingBackend):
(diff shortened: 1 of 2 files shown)


dfgvaetyj3456356-hash:
Thanks for the review. I updated the branch to handle this: hardlink targets are now resolved relative to the extraction root, while symlink targets remain resolved relative to the link member's directory. I also wrapped the commonpath check so cross-drive paths fail closed as ExtractError, and added a dedicated hardlink escape regression test.

Local validation in a temporary WSL venv:

```
python -m pytest test/unit/test_tar_packaging_backend.py -q
...                                                                      [100%]
3 passed in 2.16s
```

dfgvaetyj3456356-hash:
Follow-up on the Python 3.12+ tarfile filter note: I updated the extraction call to pass `filter=fully_trusted` only when the runtime's `TarFile.extractall` supports the `filter` parameter. That keeps the existing extraction semantics after the explicit path/link validation while avoiding a direct keyword argument on older supported Python runtimes.

Added focused regression coverage for both paths:

```
python -m pytest test/unit/test_tar_packaging_backend.py -q
.....                                                                    [100%]
5 passed in 1.88s

python -m py_compile metaflow/packaging_sys/tar_backend.py test/unit/test_tar_packaging_backend.py
git diff --check
```

dfgvaetyj3456356-hash:
Good catch on the defense-in-depth angle. I changed the compatibility wrapper from `filter=fully_trusted` to `filter=tar` when the runtime supports extraction filters. That preserves a tar-aware built-in path/link check as a secondary guard after the explicit `_validate_member()` pass, while still avoiding an unsupported keyword on older runtimes.

Refreshed local validation:

```
python -m pytest test/unit/test_tar_packaging_backend.py -q
.....                                                                    [100%]
5 passed in 1.79s

python -m py_compile metaflow/packaging_sys/tar_backend.py test/unit/test_tar_packaging_backend.py
git diff --check
```

dfgvaetyj3456356-hash:
I added focused tests for the safe validation paths that Codecov was undercounting: member target normalization, inside-directory acceptance, safe relative/absolute symlink targets, and safe hardlink targets.

Validation in a WSL venv:

```text
python -m pytest test/unit/test_tar_packaging_backend.py -q
.............                                                            [100%]
13 passed in 1.72s

python -m py_compile metaflow/packaging_sys/tar_backend.py test/unit/test_tar_packaging_backend.py
git diff --check
```

dfgvaetyj3456356-hash:
I also closed the remaining Greptile test-coverage gap by adding a regression test for an absolute archive member name (for example, `/escaped.txt`). Refreshed focused validation:

```text
python -m pytest test/unit/test_tar_packaging_backend.py -q
..............                                                           [100%]
14 passed in 4.23s

python -m py_compile metaflow/packaging_sys/tar_backend.py test/unit/test_tar_packaging_backend.py
git diff --check
```

dfgvaetyj3456356-hash:
I addressed the chained-symlink concern by rejecting parent-directory references (`..`) in archive member names and symlink/hardlink targets before extraction. This avoids relying on lexical normalization when an earlier archive member could introduce a symlink, and matches the conservative behavior expected from tar extraction filters on newer runtimes.

Added regression coverage for the chained symlink + embedded parent-reference case, plus separator handling for the helper.

Validation:

```text
WSL temp venv:
python -m pip install -q -e .
python -m pytest test/unit/test_tar_packaging_backend.py -q
................                                                         [100%]
16 passed in 1.98s

Windows syntax/checks:
python -m py_compile metaflow\packaging_sys\tar_backend.py test\unit\test_tar_packaging_backend.py
git diff --check -- metaflow/packaging_sys/tar_backend.py test/unit/test_tar_packaging_backend.py
```

I also rechecked the failing `Test Metaflow with complete Kubernetes stack / test` job. It still appears to be environment readiness rather than this tar extraction patch: the failure is `timed out waiting for the condition on uiresources/generate-configs` while the service health probes in the same log are returning 200s.

PR #3231, opened 2026-06-02
feat(tracing): respect otel service name via env variable

### LINKED ISSUE #3228: Hardcoded OTel service.name prevents runtime service name detection
Metaflow's tracing module [hardcodes service.name = "metaflow"](https://github.com/Netflix/metaflow/blob/bc67465422713d899401c5551a8f2435295083e8/metaflow/tracing/tracing_modules.py#L32-L38) when constructing the OTel Resource.

For teams that run Metaflow as part of a larger service and ship traces to a shared backend (e.g. Datadog), this causes every Metaflow span to be labeled `metaflow` regardless of what `OTEL_SERVICE_NAME` is set to in the surrounding environment. 

As a result, my team can't do accurate per-service trace or error tracking for our application that embeds Metaflow without monkey-patching the library or maintaining a fork. 

Ideally, Metaflow would respect `OTEL_SERVICE_NAME`, and fall back on `metaflow` only if the env var is unset. This could be construed as a breaking change if there are consumers that have `OTEL_SERVICE_NAME` set process-wide and have dashboards/alerts keyed on the literal string "metaflow". If this is a real concern, we could add a second env var (e.g. `METAFLOW_OTEL_INHERIT_SERVICE_NAME`) that allows an opt-in to using the OTel service name.

I'm happy to put up a PR for the required change, once there's clarity on whether we want to make it standard behavior or opt-in.

vivekkhimani:
## PR Type

- [x] Bug fix
- [ ] New feature
- [ ] Docs / tooling
- [ ] Refactoring

## Summary
Make Metaflow's OpenTelemetry `service.name` configurable via a new `METAFLOW_OTEL_SERVICE_NAME` config instead of hardcoding `"metaflow"`. The default is unchanged (`"metaflow"`), so this is fully backward compatible.

## Issue

Fixes #3228

## Reproduction
* Added a new `OTEL_SERVICE_NAME` config in `metaflow_config.py`.
* `tracing_modules.py` now reads `OTEL_SERVICE_NAME` instead of the hardcoded `"metaflow"` string when building the OTel `Resource`.
* The resolved value is propagated to remote tasks as `METAFLOW_OTEL_SERVICE_NAME` in `batch.py` and `kubernetes.py` (the K8s job and jobset paths), wherever `METAFLOW_OTEL_ENDPOINT` is already forwarded, so remote spans carry the same service name as the client. Argo Workflows and Step Functions are intentionally skipped, as tracing config is not wired through those orchestrators today (would be a larger follow-up).
* Resolution:
  1. `METAFLOW_OTEL_SERVICE_NAME` if set
  2. `"metaflow"` otherwise (unchanged default)

  Per [review discussion](https://github.com/Netflix/metaflow/pull/3231#discussion_r3338081241), auto-inheriting the bare `OTEL_SERVICE_NAME` env var was intentionally dropped to avoid silently relabeling spans for users who already set `OTEL_SERVICE_NAME` process-wide but rely on `"metaflow"` in their dashboards/alerts. To inherit a surrounding application's service name, set `METAFLOW_OTEL_SERVICE_NAME=$OTEL_SERVICE_NAME` in the runtime environment.

## Tests

- [ ] CI passes

## AI Tool Usage


metaflow/metaflow_config.py
```diff
@@ -593,6 +593,22 @@
 OTEL_ENDPOINT = from_conf("OTEL_ENDPOINT")
 ZIPKIN_ENDPOINT = from_conf("ZIPKIN_ENDPOINT")
 CONSOLE_TRACE_ENABLED = from_conf("CONSOLE_TRACE_ENABLED", False)
+# Opt-in to inheriting the standard OTEL_SERVICE_NAME env var of the surrounding
+# application. Off by default so upgrades never silently relabel spans for users
+# who already have OTEL_SERVICE_NAME set process-wide.
+OTEL_INHERIT_SERVICE_NAME = from_conf("OTEL_INHERIT_SERVICE_NAME", False)
```
vivekkhimani: @talsperre -- this is the main area of contention that I'd love your thoughts on! The simpler/cleaner thing here IMO is to just start respecting the `OTEL_SERVICE_NAME` env var but as Ian rightly pointed out in his issue, that might cause a "breaking change" for existing consumers who might have this variable set in their env but have their o11y/dashboards listening on `metaflow`. We are in fact, currently relying on `metaflow` as our service name.

The current approach, explicitly gives them a flag to override the existing behavior and start respecting `OTEL_SERVICE_NAME`. What do you think?

metaflow/tracing/tracing_modules.py
```diff
@@ -32,7 +33,7 @@ def init_tracing():
     tracer_provider = TracerProvider(
         resource=Resource.create(
             {
-                SERVICE_NAME: "metaflow",
+                SERVICE_NAME: OTEL_SERVICE_NAME,
```
talsperre, line 36: You need to propagate it appropriately in `batch.py` and `kubernetes.py` decorator implementations as well. See [here](https://github.com/Netflix/metaflow/blob/a3302937a773028c5486096168b1b8ea6ab34c62/metaflow/plugins/aws/batch/batch.py#L335) for instance.

talsperre (changes requested):

metaflow/tracing/tracing_modules.py
vivekkhimani: Addressed this in https://github.com/Netflix/metaflow/pull/3231/changes/6cd62ba5982cb44a64e68eff4e5203b74ea9a892! Note that argo workflows and step functions are intentionally skipped as the tracing config isn't wired through those orchestrators today and it might need a bigger follow up change.

metaflow/metaflow_config.py
saikonen: I've been mulling over this for a bit now, and I'm wondering whether the env-fallback is a common use case or not. Would love some more insight on peoples use-cases.

The main concern I have here is adding possibly unnecessary/unused options to the Metaflow config. I think we could do without the flag for otel-inheritance as

1. If a user/org wants to use an already configured `OTEL_SERVICE_NAME` in their environment, then they still need to add a new entry to their Metaflow profile to enable this. The edit could just as well be to change the service name through the config at that point.

2. If the org wants the ability to set/swap service names as part of their infra instead of Metaflow profiles, and they are managing environment variables separately, then this should only require an additional `METAFLOW_OTEL_SERVICE_NAME=$OTEL_SERVICE_NAME` to the runtime environments in order to override the defaults.

metaflow/metaflow_config.py
```diff
@@ -593,6 +593,7 @@
 OTEL_ENDPOINT = from_conf("OTEL_ENDPOINT")
 ZIPKIN_ENDPOINT = from_conf("ZIPKIN_ENDPOINT")
 CONSOLE_TRACE_ENABLED = from_conf("CONSOLE_TRACE_ENABLED", False)
+OTEL_SERVICE_NAME = from_conf("OTEL_SERVICE_NAME", "metaflow")
```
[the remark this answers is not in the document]
vivekkhimani: The PR description was stale, updating it now

metaflow/metaflow_config.py
vivekkhimani: This should be addressed in https://github.com/Netflix/metaflow/pull/3231/commits/c8683d91074e7f37560da7bdefe4e0b4524cd84a!

talsperre (approved):
Looks good to me. I reviewed the config resolution, tracer resource wiring, Batch/Kubernetes propagation, and focused unit coverage. The opt-in METAFLOW_OTEL_SERVICE_NAME behavior matches the latest discussion and preserves the default service.name.

saikonen (approved):

PR #3233, opened 2026-06-02
fix(argo): order conditional input fallback by DAG

### LINKED ISSUE #3230: Argo: artifacts silently lost when a step has 2+ split-switch in_funcs that both ran
## Summary

When a Metaflow step running on Argo Workflows has 2+ `in_funcs` whose type is `split-switch`, and 2+ of those `split-switch` predecessors actually ran in the same execution (e.g. via a chain of conditional skips), the `conditional_input_paths` runtime filter picks the alphabetically first survivor rather than the immediate runtime parent. The step then inherits self.<attr> state from the wrong predecessor, silently dropping any artifacts written between the wrong-pick step and the real parent.

We hit this in production: a flow accumulating results across several conditional steps published zero results on Argo for cases where the same logic produced results locally. Local execution worked because it uses a different code path; this is Argo-specific.

## Affected version

Metaflow 2.19.31 (and present in current master at bc67465422713d899401c5551a8f2435295083e8).

## Minimal reproduction
```


  class ChainSkipRepro(FlowSpec):
      @step
      def start(self):
          self.findings = ["from_start"]
          self.route1 = "step2"
          self.next(
              {"end": self.end, "step2": self.step2},
              condition="route1",
          )

      @step
      def step2(self):
          self.findings = self.findings + ["from_step2"]
          # Skip step3 -- exercises the bug path.
          self.route2 = "end"
          self.next(
              {"end": self.end, "step3": self.step3},
              condition="route2",
          )

      @step
      def step3(self):
          self.findings = self.findings + ["from_step3"]
          self.next(self.end)

      @step
      def end(self):
          print(f"end.findings = {self.findings!r}")


  if __name__ == "__main__":
      ChainSkipRepro()
```

### Deploy and run on Argo:
```
  python flow.py argo-workflows create
  python flow.py argo-workflows trigger
```
### Observed

  `end.findings = ['from_start']`

  `'from_step2'` is silently lost despite `step2` having successfully appended it.

### Expected

  `end.findings = ['from_start', 'from_step2']`

  This is what `python flow.py` run produces locally, confirming that the issue is specific to the Argo plugin.

## Root cause (note: Claude-driven analysis)

  I believe three pieces of code combine into the bug:

  1. [`metaflow/graph.py:517`](https://github.com/Netflix/metaflow/blob/bc67465422713d899401c5551a8f2435295083e8/metaflow/graph.py#L517) sorts `in_funcs` alphabetically:
  `node.in_funcs = sorted(node.in_funcs)`

  For the end step in the repro, `in_funcs` becomes `['start', 'step2', 'step3']` regardless of which is the actual immediate parent.

  2. [metaflow/plugins/argo/argo_workflows.py:2216-2222](https://github.com/Netflix/metaflow/blob/bc67465422713d899401c5551a8f2435295083e8/metaflow/plugins/argo/argo_workflows.py#L2216-L2222) builds `skippable_input_steps` from all `in_funcs` whose type is `split-switch`:
  ```
  skippable_input_steps = ",".join(
      [
          in_func
          for in_func in node.in_funcs
          if self.graph[in_func].type == "split-switch"
      ]
  )
```

  For end, this yields `["start", "step2"]` (both are `split-switch`es; step3 is linear).

  3. [metaflow/plugins/argo/conditional_input_paths.py:23-32](https://github.com/Netflix/metaflow/blob/bc67465422713d899401c5551a8f2435295083e8/metaflow/plugins/argo/conditional_input_paths.py#L21-L31) filters input paths:
  ```
  trimmed = [path for path in paths if not "{{" in path]
  latest_conditional_in_graph = trimmed[:1]
  conditionals = [
      path for path in trimmed if not any(step in path for step in skippable_steps)
  ]
  pathspecs_to_use = conditionals if conditionals else latest_conditional_in_graph
 ```

  At runtime with `route1="step2"` and `route2="end"`:
  - Argo passes paths for `start`, `step2`, and `step3` (templated as `{{...}}` because skipped).
  - trimmed = `[start_path, step2_path]` (`step3` is stripped).
  - Both `start` and `step2` are in `skippable_steps`, so conditionals = `[]`.
  - Fallback to `latest_conditional_in_graph = trimmed[:1] = [start_path]`.
  - `end` reads its state from `start` instead of `step2`, the immediate parent.

The comment at [conditional_input_paths.py:23-25](https://github.com/Netflix/metaflow/blob/bc67465422713d899401c5551a8f2435295083e8/metaflow/plugins/argo/conditional_input_paths.py#L23-L25) says "the order of graph parsing ensures that the steps are in reverse order of occurrence, so the first one is the latest." This assumption is wrong because `in_funcs` are sorted alphabetically in `graph.py`, not by DAG depth.

### Suggested fix direction

When falling back from an empty conditionals set, pick the path corresponding to the deepest/latest predecessor in the DAG rather than `trimmed[:1]`. The Argo plugin already has the DAG structure available; ordering the input paths by topological depth (or having the filter take DAG-depth metadata as a third argument) would correct the pick.

### PR DESCRIPTION
**Pull request by talsperre**
## Summary
- order Argo split-switch input candidates by DAG position before passing them to `conditional_input_paths`
- make `conditional_input_paths` choose the latest executed skippable predecessor when all executed inputs are skippable
- compare exact pathspec step names instead of substring matching step names in paths

Fixes #3230

## Tests
- `uv run --with '.[dev]' python -m pytest test/unit/test_argo_conditional_input_paths.py test/unit/test_argo_workflows_cli.py -q`
- `uv run --with pre-commit pre-commit run --files metaflow/plugins/argo/conditional_input_paths.py metaflow/plugins/argo/argo_workflows.py test/unit/test_argo_conditional_input_paths.py test/unit/test_argo_workflows_cli.py`

### DIFF
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -1202,6 +1202,20 @@ class ArgoWorkflows(object):
         ]
         return len(cond_in_funcs) > 1 and len(cond_in_funcs) == len(node.in_funcs)
 
+    def _skippable_input_steps_in_dag_order(self, node):
+        graph_order = {
+            node_name: index for index, node_name in enumerate(self.graph.sorted_nodes)
+        }
+        return sorted(
+            [
+                in_func
+                for in_func in node.in_funcs
+                if self.graph[in_func].type == "split-switch"
+            ],
+            key=lambda in_func: graph_order[in_func],
+            reverse=True,
+        )
+
     def _is_recursive_node(self, node):
         return node.name in self.recursive_nodes
 
@@ -2214,11 +2228,7 @@ class ArgoWorkflows(object):
                 # we need to pass in the set of conditional in_funcs to the pathspec generating script as in the case of split-switch skipping cases,
                 # non-conditional input-paths need to be ignored in favour of conditional ones when they have executed.
                 skippable_input_steps = ",".join(
-                    [
-                        in_func
-                        for in_func in node.in_funcs
-                        if self.graph[in_func].type == "split-switch"
-                    ]
+                    self._skippable_input_steps_in_dag_order(node)
                 )
                 input_paths = (
                     "$(python -m metaflow.plugins.argo.conditional_input_paths %s %s)"
--- a/metaflow/plugins/argo/conditional_input_paths.py
+++ b/metaflow/plugins/argo/conditional_input_paths.py
@@ -4,6 +4,13 @@ from metaflow.util import decompress_list, compress_list
 import base64
 
 
+def _step_name_from_path(path):
+    parts = path.split("/", 2)
+    if len(parts) > 1:
+        return parts[1]
+    return None
+
+
 def generate_input_paths(input_paths, skippable_steps):
     # => run_id/step/:foo,bar
     # input_paths are base64 encoded due to Argo shenanigans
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: metaflow/plugins/argo/argo_workflows.py [hunk 1]
```diff
@@ -1202,6 +1202,20 @@ def _many_in_funcs_all_conditional(self, node):
         ]
         return len(cond_in_funcs) > 1 and len(cond_in_funcs) == len(node.in_funcs)
 
+    def _skippable_input_steps_in_dag_order(self, node):
```
saikonen, line 1205: Changes pass my existing test cases for conditionals so should be fine from that perspective.

saikonen (approved):

PR #3234, opened 2026-06-03
fix: validate run IDs during resume operations

### LINKED ISSUE #3229: Users can specify --run-id for a previously existing run, and Metaflow happily continues
If run `123` has failed on any step, and a user accidentally does `resume --run-id 123` instead of `resume --origin-run-id 123`, they will get an unexpected UX where Metaflow does the resume from a run that may not be the same one (if their desired origin run id is from Argo, for instance), and the backend is now polluted by the mixing of task execution data from the parent and resume runs. Metaflow should catch the foot gun before the resume run is created on the duplicate `--run-id`.

vanshapple:
## Summary

Fixes issue #3229 by adding validation for resume workflows when a user specifies a `--run-id` that already exists.

Previously, Metaflow could continue execution when a user accidentally ran:

```bash
python flow.py resume --run-id <existing_run_id>
```

instead of:

```bash
python flow.py resume --origin-run-id <existing_run_id>
```

This could lead to unexpected behavior and mixing of execution metadata between the original run and the resumed run.

## Changes

* Added validation to detect when a resume operation attempts to use an existing run ID.
* Fail early with a clear error message instead of continuing execution.
* Preserved existing behavior for valid resume workflows using `--origin-run-id`.
* Added/updated tests to cover duplicate run ID validation during resume operations.

## Example

Before:

```bash
python flow.py resume --run-id 123
```

If run `123` already existed, Metaflow could continue execution unexpectedly.

After:

```bash
python flow.py resume --run-id 123
```

Metaflow now raises a validation error and guides the user toward using:

```bash
python flow.py resume --origin-run-id 123
```

## Testing

* Verified duplicate run IDs are rejected during resume operations.
* Verified valid resume workflows continue to work as expected.
* Verified normal run creation behavior remains unchanged.

Fixes #3229

### DIFF
--- a/metaflow/cli_components/run_cmds.py
+++ b/metaflow/cli_components/run_cmds.py
@@ -129,6 +129,7 @@ def _validate_resume_run_id(flow_name, run_id, reentrant=False):
     try:
         Run(pathspec=pathspec, _namespace_check=False)
     except MetaflowNotFound:
+        # Expected when run doesn't exist. This is the desired outcome.
         pass
     else:
         raise CommandException(
@@ -145,6 +146,7 @@ def _validate_resume_run_id(flow_name, run_id, reentrant=False):
     try:
         int(run_id)
     except ValueError:
+        # Non-integer run-id (external scheduler): OK
         pass
     else:
         raise CommandException("run-id %s cannot be an integer" % run_id)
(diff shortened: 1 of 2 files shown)


agsaru (commented):
@vanshapple Can u write test with pytest-mock. Usage of unittest is not allowed as per contribution guidelines.
And the assertions are totally wrong most of them are out of the context managers.

vanshapple:
okay let me fix that.

vanshapple:
@agsaru Thanks for the review! I've updated the tests to use pytest-mock, fixed the assertion placement, and removed the ordering dependency in the existing-run test. I've also rechecked the validation logic and pushed the changes.

PR #3237, opened 2026-06-03
`DAGNode.foreach_param / output_steps()["foreach_artifact"]` correctly populated on python3.12+

### LINKED ISSUE #3236: `DAGNode.foreach_param / output_steps()["foreach_artifact"]` is always None on Python ≥ 3.12 (static graph parse uses removed ast.Str.s)
`FlowGraph`'s static parser fails to recover the `foreach=` and `num_parallel=` keyword values from `self.next(...)` on Python 3.12+.

The node is still correctly typed as `foreach` / `split-foreach`, but `DAGNode.foreach_param` is `None`. As a result, `foreach_artifact` in `FlowGraph.output_steps()` and the `_graph_info` consumed by cards, the UI, and Argo/Airflow compilation also return `None`.

Basically the loop-variable name is silently lost.

## Minimal repro

```python
import importlib.util

from metaflow.graph import FlowGraph


class ForeachRepro(FlowSpec):
    @step
    def start(self):
        self.items = [1, 2, 3]
        self.next(self.process, foreach="items")

    @step
    def process(self):
        self.next(self.join)

    @step
    def join(self, inputs):
        self.next(self.end)

    @step
    def end(self):
        pass


g = FlowGraph(ForeachRepro)

print(g["start"].type)  # 'foreach' ✅
print(g["start"].foreach_param)  # None ❌ expected 'items'
print(g.output_steps()[0]["start"].get("foreach_artifact"))  # None ❌
```

## Expected behavior

```python
g["start"].foreach_param == "items"
g.output_steps()[0]["start"].get("foreach_artifact") == "items"
```

## Actual behavior

On Python 3.14.3 with Metaflow 2.19.28:

```python
g["start"].foreach_param is None
g.output_steps()[0]["start"].get("foreach_artifact") is None
```

The node type is correct. Only the artifact name is lost.

## Root cause

In `metaflow/graph.py`, inside `DAGNode._parse`, the keyword-value extractor reads `.s`:

```python
# metaflow/graph.py:255-257, upstream main
keywords = dict(
    (k.arg, getattr(k.value, "s", None)) for k in tail.value.keywords
)
```

`.s` was the attribute used by the legacy `ast.Str` node. Since Python 3.8, string literals are represented as `ast.Constant`, with the actual value stored in `.value`. In Python 3.12, `ast.Str` and its `.s` compatibility alias were removed.

Therefore, on Python 3.12+:

```python
getattr(<ast.Constant>, "s", None)
```

returns `None`, and the default silently swallows the value.

The downstream assignments then store `None`:

```python
self.foreach_param = keywords["foreach"]
self.num_parallel = keywords["num_parallel"]
```

Node type detection still works because it checks for the presence of the keyword name, for example `"foreach" in keywords`, rather than checking the keyword value. That is why the symptom is:

> type is correct, value is missing

The same file already handles this correctly for the `switch` condition keyword just above:

```python
if hasattr(ast, "Str") and isinstance(keyword.value, ast.Str):
    condition_name = keyword.value.s
elif isinstance(keyword.value, ast.Constant) and isinstance(keyword.value.value, str):
    condition_name = keyword.value.value
```

The `ast.Constant` fallback was added for switch conditions but was not applied to the `foreach` / `num_parallel` keyword extraction below it. This appears to be a half-applied Python 3.12 migration.

## Suggested fix

Use `ast.literal_eval`, or the same `ast.Constant` fallback already used for switch conditions, instead of reading `.s` directly.

For example:

```python
def _kw_value(node):
    if hasattr(ast, "Str") and isinstance(node, ast.Str):  # < 3.12
        return node.s

    if isinstance(node, ast.Constant):  # >= 3.8
        return node.value

    return None


keywords = dict((k.arg, _kw_value(k.value)) for k in tail.value.keywords)
```

This restores both:

* `foreach_param`, which should contain the string artifact name
* `num_parallel`, which should contain the integer parallelism value

on modern Python versions.

## Impact

* `DAGNode.foreach_param` is blank for all `foreach` steps on Python 3.12+.
* `output_steps()` returns `foreach_artifact=None`.
* `num_parallel` is also lost for parallel steps.
* Anything consuming the static graph loses this metadata, including:

  * `python flow.py show`
  * the default DAG card
  * `metaflow-ui`'s `/dag` payload
  * Argo/Airflow graph metadata
* The graph topology is unaffected:

  * node types are still correct
  * edges are still correct
  * `matching_join` is still correct

## Environment

* Metaflow 2.19.28
* Also present on `main` at `6cd62ba5`, `graph.py:256`
* Python 3.14.3
* Reproduces on Python 3.12+

vivekkhimani:
## PR Type

- [x] Bug fix
- [ ] New feature
- [ ] Docs / tooling
- [ ] Refactoring

## Summary

The static graph parser silently drops the `foreach=` / `num_parallel=` keyword values on Python 3.14+, so `DAGNode.foreach_param`, `num_parallel`, and the `foreach_artifact` exposed by `output_steps()` all come back `None`. Node typing is unaffected, only the artifact name/value is lost. This swaps the removed `ast.Str.s` read for an `ast.Constant` fallback (the same one already used for switch conditions a few lines above).

## Issue

Fixes #3236

## Reproduction

**Runtime:** local

**Commands to run:** Run the flow below under Python 3.14 (where `ast.Str` and the `.s` alias are gone). On 3.12/3.13 the deprecated `.s` alias still resolves, so the bug stays hidden there. Save it to a file and run it with `python <file>.py`, or paste it into a Python 3.14 shell.

```python
from metaflow.graph import FlowGraph


class ForeachRepro(FlowSpec):
    @step
    def start(self):
        self.items = [1, 2, 3]
        self.next(self.process, foreach="items")

    @step
    def process(self):
        self.next(self.join)

    @step
    def join(self, inputs):
        self.next(self.end)

    @step
    def end(self):
        pass


g = FlowGraph(ForeachRepro)
print(g["start"].type)                                        # 'foreach'
print(g["start"].foreach_param)                               # expected 'items'
print(g.output_steps()[0]["start"].get("foreach_artifact"))   # expected 'items'
```

**Where evidence shows up:** parent console (the static graph metadata, also consumed by `show`, the DAG card, and Argo/Airflow compilation)

<details>
<summary>Before (Python 3.14.3)</summary>

```
type            = 'foreach'
foreach_param   = None
foreach_artifact = None
```

</details>

<details>
<summary>After (Python 3.14.3)</summary>

```
type            = 'foreach'
foreach_param   = 'items'
foreach_artifact = 'items'
```

</details>

## Root Cause

In `DAGNode._parse`, the keyword extractor read `getattr(k.value, "s", None)`. `.s` was the attribute on the legacy `ast.Str` node. Since Python 3.8 string literals are `ast.Constant` with the value in `.value`, and the `.s` compatibility alias was removed in 3.14. So on 3.14 the `getattr` falls through to its `None` default and the value is dropped.

Type detection still works because it checks for the presence of the keyword name (`"foreach" in keywords`), not its value, which is why the type stays correct while the value goes missing. The switch-condition branch just above already had the `ast.Constant` fallback, so this was effectively a half-applied migration.

## Why This Fix Is Correct

Restores the invariant that a string/int literal keyword in `self.next(...)` round-trips to `foreach_param` / `num_parallel`. The `_kw_value` helper keeps the `ast.Str` branch for older interpreters and reads `ast.Constant.value` everywhere else, matching the existing switch-condition handling. Nothing else in the parse path changes.

## Failure Modes Considered

1. Older Python (`< 3.14`): `ast.Str`/`.s` still exist, and the `hasattr(ast, "Str")` branch preserves the old path. Verified no regression on 3.12.
2. Non-literal keyword values (e.g. `foreach=some_var`): the node is an `ast.Name`, not `ast.Constant`, so `_kw_value` returns `None`, same as before. No behavior change there.

## Tests

- [x] CI passes
- [x] If tests are impractical: explain why below and provide manual evidence above

Existing `test/unit/test_graph_structure.py` and `test_sourceless_dag_node.py` pass on 3.14 (55 passed). No new unit test added yet since the existing suite runs on the default interpreter where the bug is masked; happy to add a 3.14-gated regression test if preferred.

## Non-Goals

No change to graph topology, edge resolution, `matching_join`, or the switch-condition path. Not touching how non-literal keyword values are handled.

## AI Tool Usage

Claude Code was used to reproduce the bug on Python 3.14.3, apply the fix, and verify against the existing unit tests. All generated code was reviewed, understood, and tested.

### DIFF
--- a/metaflow/graph.py
+++ b/metaflow/graph.py
@@ -252,8 +252,18 @@ class DAGNode(object):
             else:
                 self.out_funcs = [e.attr for e in tail.value.args]
 
+            def _kw_value(node):
+                # ast.Str (and the .s alias on ast.Constant) was deprecated in
+                # Python 3.8 and removed in 3.14; ast.Constant.value holds the
+                # literal value on all supported versions.
+                if hasattr(ast, "Str") and isinstance(node, ast.Str):
+                    return node.s
+                if isinstance(node, ast.Constant):
+                    return node.value
+                return None
+
             keywords = dict(
-                (k.arg, getattr(k.value, "s", None)) for k in tail.value.keywords
+                (k.arg, _kw_value(k.value)) for k in tail.value.keywords
             )
             if len(keywords) == 1:
                 if "foreach" in keywords:
--- a/metaflow/graph.py
+++ b/metaflow/graph.py
@@ -253,9 +253,17 @@ class DAGNode(object):
                 self.out_funcs = [e.attr for e in tail.value.args]
 
             def _kw_value(node):
-                # ast.Str (and the .s alias on ast.Constant) was deprecated in
-                # Python 3.8 and removed in 3.14; ast.Constant.value holds the
-                # literal value on all supported versions.
+                """Extract the literal value from a keyword-argument node.
+
+                `node` is the `.value` of an `ast.keyword` (e.g. the node for
+                `"items"` in `foreach="items"`), so unwrapping is two levels:
+                    ast.keyword(arg="foreach", value=ast.Constant(value="items"))
+                                                      ^node           ^node.value
+
+                ast.Str (and the .s alias on ast.Constant) was deprecated in
+                Python 3.8 and removed in 3.14; ast.Constant.value holds the
+                literal value on all supported versions.
+                """
                 if hasattr(ast, "Str") and isinstance(node, ast.Str):
                     return node.s
                 if isinstance(node, ast.Constant):
--- a/metaflow/graph.py
+++ b/metaflow/graph.py
@@ -46,6 +46,25 @@ def deindent_docstring(doc):
         return ""
 
 
+def _ast_literal_value(node):
+    """Extract the literal value from the `.value` of an `ast.keyword`.
+
+    `node` is the value node of a keyword argument (e.g. the node for
+    `"items"` in `foreach="items"`), so unwrapping is two levels:
+        ast.keyword(arg="foreach", value=ast.Constant(value="items"))
+                                          ^node           ^node.value
+
+    ast.Str (and the .s alias on ast.Constant) was deprecated in Python 3.8
+    and removed in 3.14; ast.Constant.value holds the literal value on all
+    supported versions. Returns None for non-literal nodes (e.g. ast.Name).
+    """
+    if hasattr(ast, "Str") and isinstance(node, ast.Str):
+        return node.s
+    if isinstance(node, ast.Constant):
+        return node.value
+    return None
+
+
 # ---------------------------------------------------------------------------
 # Note on "sourceless" DAGNodes (used by FunctionSpec)
 # ---------------------------------------------------------------------------
@@ -233,12 +252,9 @@ class DAGNode(object):
                 # Get condition parameter
                 for keyword in tail.value.keywords:
                     if keyword.arg == "condition":
-                        if hasattr(ast, "Str") and isinstance(keyword.value, ast.Str):
-                            condition_name = keyword.value.s
-                        elif isinstance(keyword.value, ast.Constant) and isinstance(
-                            keyword.value.value, str
-                        ):
-                            condition_name = keyword.value.value
+                        value = _ast_literal_value(keyword.value)
+                        if isinstance(value, str):
+                            condition_name = value
                         break
 
                 if switch_cases and condition_name:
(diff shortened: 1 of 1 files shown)


romain-intel (approved):
i'll let sakari take a quick look too but lgtm.

saikonen (approved):

PR #3238, opened 2026-06-06
fix: re-run StepMutators on late-attached platform decorators (#3025)

alexzhu0:
## PR Type

- [x] Bug fix
- [x] Core Runtime change (higher bar -- `metaflow/decorators.py`)

## Summary

`StepMutator.mutate()` can again see and override late-attached platform decorators (e.g. `@kubernetes`). Before this fix, on deploy paths that attach the compute decorator late (Argo Workflows, `--with` conda/pypi, etc.) a `StepMutator` override of `@kubernetes` resources was silently dropped.

## Issue

Fixes #3025

## Reproduction

**Runtime:** argo-workflows (any deploy path that late-attaches `@kubernetes`)

A `StepMutator` that inspects `decorator_specs` for `"kubernetes"` and calls `add_decorator(..., duplicates=OVERRIDE)` to bump cpu/memory:

```python
from metaflow import FlowSpec, StepMutator, step

class kubernetes_override(StepMutator):
    def init(self, *args, **kwargs):
        self.deco_kwargs = kwargs
    def mutate(self, mutable_step):
        for deco_name, _, _, attributes in mutable_step.decorator_specs:
            if deco_name == "kubernetes":
                attributes.update(self.deco_kwargs)
                mutable_step.add_decorator(
                    deco_type=deco_name, deco_kwargs=attributes,
                    duplicates=mutable_step.OVERRIDE,
                )

class MyFlow(FlowSpec):
    @kubernetes_override(cpu=2, memory=8192)
    @step
    def start(self):
        self.next(self.end)
    @step
    def end(self):
        pass
```

**Commands to run:**
```bash
python my_flow.py argo-workflows create --only-json
```

**Where evidence shows up:** the generated Argo workflow / the `@kubernetes` attributes on the `start` step.

<details>
<summary>Before (override silently ignored)</summary>

```
start step @kubernetes: cpu=1, memory=4096   # defaults — the mutator never saw @kubernetes
```

</details>

<details>
<summary>After (override applied)</summary>

```
start step @kubernetes: cpu=2, memory=8192   # mutator override is honored
```

</details>

The included unit test `test/unit/test_late_attached_mutator.py` encodes exactly this: it fails before the fix (`AssertionError: assert '1' == '4'`) and passes after, with **no cloud backend required** — it drives the internal `_init_step_decorators` → `_attach_decorators` → `_process_late_attached_decorator` sequence directly using a `MagicMock` datastore.

## Root Cause

The single caller of `StepMutator.mutate()` is `_init_step_decorators` (`metaflow/decorators.py`). PR #2719 moved `_init_step_decorators()` out of Argo's `make_flow()` and into `cli.py start()`, so the mutator pass now runs **before** platform decorators are late-attached:

1. `_init_step_decorators()` runs `StepMutator.mutate()` — `@kubernetes` is **not attached yet**
2. `_attach_decorators()` late-attaches `@kubernetes` to the steps
3. `_process_late_attached_decorator()` calls `external_init()` + `step_init()` on the new decorators **but never re-runs `mutate()`**

So a `StepMutator` iterating `mutable_step.decorator_specs` for `"kubernetes"` in step 1 finds nothing, and its `add_decorator(..., OVERRIDE)` is a no-op. The invariant violated: *a `StepMutator` is supposed to observe every decorator on its step, including platform decorators attached at deploy time* — true in 2.19.13, broken in 2.19.14.

## Why This Fix Is Correct

`_process_late_attached_decorator` now re-runs step mutators on exactly the steps that received a late-attached decorator, then rebuilds the graph — mirroring the existing `StepMutator` handling in `_init_step_decorators` (same `MutableStep(...)` construction, same `cls._init_graph()` refresh). This restores the 2.19.13 ordering guarantee while keeping the 2.19.14 architecture. The fix is minimal (one function, +48 lines, no signature changes) and touches only the late-attach path.

Key correctness points:
- **Scoped re-run:** mutators are re-run only for `late_attached_step_names` (the steps that actually received a late-attached decorator), so mutators on unaffected steps are not invoked a second time.
- **Idempotent `external_init()`:** after `add_decorator(OVERRIDE)` may create replacement decorator instances, `external_init()` is called again; it is guarded by `_ran_init`, so already-initialized decorators are untouched.
- **Graph rebuild only when needed:** `cls._init_graph()` runs only if a mutator actually ran (`mutators_ran`), avoiding a needless rebuild on the common no-mutator path.

## Failure Modes Considered

1. **Non-idempotent / cross-step mutators.** Re-running every mutator on every step could double-apply side effects or apply step A's override to step B. Mitigated by gating the re-run on `late_attached_step_names` and matching `_init_step_decorators`'s per-step iteration; the test `test_step_without_mutator_keeps_kubernetes_defaults` asserts a mutator-less step keeps defaults (cpu=1, memory=4096) and that exactly one `@kubernetes` exists per step (no duplication from `OVERRIDE`).
2. **No late attachment at all (the common case).** If no decorator in `deco_names` is present, `late_attached_step_names` is empty, `mutators_ran` stays `False`, no mutator runs and no graph rebuild happens — behavior is identical to before. Covered by `test_no_late_attachment_is_a_noop`.
3. **Backward compatibility / spin path.** No public signature changed; the `is_spin`/`skip_decorators` handling in the subsequent `step_init` loop is untouched. The full `test/unit` suite passes unchanged.

## Tests

- [x] Unit tests added/updated — `test/unit/test_late_attached_mutator.py` (3 cases: override-applies, no-mutator-keeps-defaults, no-attach-noop)
- [x] Reproduction script provided (the test is a runnable repro; fails before, passes after — no cloud backend needed)
- [x] CI passes (`tox -e unit` locally; targeted decorator/graph/mutator suite: 207 passed, 0 regressions)

## Non-Goals

`FlowMutator.mutate()` is **not** re-run here. A `FlowMutator` that inspects step-level decorators (via `MutableFlow.steps`) for a late-attached `@kubernetes` would have the symmetric blindness. I scoped this PR to the `StepMutator` case in the issue to keep the change minimal and reviewable. Happy to follow up with the `FlowMutator` re-run in a separate PR if you'd like it closed too — flagging it explicitly so it isn't silently missed.

## AI Tool Usage

- **Tool:** Claude Code.
- **Used for:** locating the regression (bisecting the ordering change to PR #2719), drafting the fix and the unit test, and running the local test suite.
- **Review:** I reviewed and understand every line. The fix deliberately mirrors the existing `StepMutator` pass in `_init_step_decorators`. The test fails before the change (`assert '1' == '4'`) and passes after, and the full `test/unit` suite passes with no regressions.

### DIFF
--- a/metaflow/decorators.py
+++ b/metaflow/decorators.py
@@ -1014,15 +1014,21 @@ def _process_late_attached_decorator(
 
     from .system_context import system_context
 
-    # Track which steps received a late-attached decorator so we only re-run
-    # mutators on those steps (re-running mutators on unaffected steps could
-    # break non-idempotent mutators).
+    # Track which steps received a *genuinely new* late-attached decorator so we
+    # only re-run mutators on those steps (re-running mutators on unaffected
+    # steps could break non-idempotent mutators). A freshly attached decorator
+    # has not been external_init'd yet (_ran_init is False); a statically
+    # defined decorator that _init_step_decorators already initialized has
+    # _ran_init True and must NOT trigger a mutator re-run. The _ran_init check
+    # therefore has to happen BEFORE external_init() flips the flag, so
+    # enrollment is recorded separately from initialization.
     late_attached_step_names = set()
     for s in flow:
         for deco in s.decorators:
+            if deco.name in deco_names and not deco._ran_init:
+                late_attached_step_names.add(s.__name__)
             if deco.name in deco_names:
                 deco.external_init()
-                late_attached_step_names.add(s.__name__)
 
     # Re-run step mutators on the affected steps so they can see the
     # late-attached decorators. _init_step_decorators (the usual mutator pass)
--- a/metaflow/decorators.py
+++ b/metaflow/decorators.py
@@ -1027,8 +1027,6 @@ def _process_late_attached_decorator(
         for deco in s.decorators:
             if deco.name in deco_names and not deco._ran_init:
                 late_attached_step_names.add(s.__name__)
-            if deco.name in deco_names:
-                deco.external_init()
 
     # Re-run step mutators on the affected steps so they can see the
     # late-attached decorators. _init_step_decorators (the usual mutator pass)
@@ -1060,18 +1058,17 @@ def _process_late_attached_decorator(
                 )
                 mutators_ran = True
 
-    # Rebuild the graph so node.decorators reflects any changes the mutators
-    # made (e.g. add_decorator with OVERRIDE), then make sure any replacement
-    # decorators created by the mutators get external_init() before step_init().
-    # external_init() is idempotent (guarded by _ran_init), so re-calling it on
-    # already-initialized decorators is a no-op.
+    # Rebuild the graph so node.decorators reflects mutator changes (OVERRIDE).
     if mutators_ran:
         cls._init_graph()
         graph = flow._graph
-        for s in flow:
-            for deco in s.decorators:
-                if deco.name in deco_names:
-                    deco.external_init()
+
+    # Init late-attached decorators only now, after mutators ran, so a decorator
+    # replaced via OVERRIDE is never external_init'd. external_init is idempotent.
+    for s in flow:
+        for deco in s.decorators:
+            if deco.name in deco_names:
+                deco.external_init()
 
     for s in flow:
         system_context.register_step_decorators(s.__name__, list(s.decorators))
--- a/metaflow/decorators.py
+++ b/metaflow/decorators.py
@@ -1014,10 +1014,58 @@ def _process_late_attached_decorator(
 
     from .system_context import system_context
 
+    # Track which steps received a late-attached decorator so we only re-run
+    # mutators on those steps (re-running mutators on unaffected steps could
+    # break non-idempotent mutators).
+    late_attached_step_names = set()
     for s in flow:
         for deco in s.decorators:
             if deco.name in deco_names:
                 deco.external_init()
+                late_attached_step_names.add(s.__name__)
+
+    # Re-run step mutators on the affected steps so they can see the
+    # late-attached decorators. _init_step_decorators (the usual mutator pass)
+    # runs in cli.py start() *before* platform decorators such as @kubernetes
+    # are late-attached (e.g. on Argo Workflows deploys, see PR #2719), so a
+    # StepMutator that inspects decorator_specs for "kubernetes" would otherwise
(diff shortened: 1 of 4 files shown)


alexzhu0:
Thanks for the review. Addressed the double-mutation concern in `be00ee1`.

**Change:** `late_attached_step_names` is now built only from decorators whose `_ran_init` is still `False` — i.e. genuinely just-attached ones — rather than from any step carrying a matching decorator name. A statically-defined `@kubernetes` that `_init_step_decorators` already initialized has `_ran_init=True`, so it no longer enrolls its step for a second mutator pass. The `_ran_init` check is read *before* `external_init()` flips the flag, in a separate statement from the (idempotent) `external_init()` call, so initialization behavior is unchanged and only the re-run set is narrowed.

Added `test_already_initialized_decorator_does_not_retrigger_mutator`: an already-`external_init`'d `@kubernetes` does not re-invoke the mutator, while a freshly attached one triggers exactly one re-run. (Verified it fails without the narrowing — the already-initialized step would otherwise be re-run — and passes with it.)

**Scope note for honesty:** this `_ran_init` distinction is meaningful for the platform deploy callers (Argo/Airflow/Step Functions), which run after `_init_step_decorators` has initialized static step decorators. On the conda/pypi self-attach path (called from `flow_init`, before `_init_step_decorators`) a static `@conda`/`@pypi` is still `_ran_init=False`, so the narrowing is a no-op there — same behavior as before, no regression, but it does not additionally suppress that case. Happy to adjust if you'd prefer a different signal than `_ran_init`.

Full `test/unit` + `test/cmd` suite green locally (531 passed).

alexzhu0:
Note on the red ❌ on the `test` check (Test Metaflow with complete Kubernetes stack): it's a CI-infra flake, not a code failure. The job died in the *Bring up the environment* step before any test ran:

```
📥 Installing gum...
curl: (22) The requested URL returned error: 504
make: *** [install-gum] Error 22
Waiting for infra timed out
```

i.e. the `gum` download returned a transient HTTP 504, so the K8s stack never came up. The same job passed on the previous commit (`fd62326`), and the only change since is the pure-Python narrowing in `decorators.py`.

Everything else is green on `be00ee1`, including the cloud-backend integration suite that actually exercises this fix: **argo-kubernetes, airflow-kubernetes, and sfn-batch all pass** (plus local / Unit / Spin / Coverage). I don't have rerun permissions on this repo — a maintainer re-run of that one job should clear it.

PR #3240, opened 2026-06-07
fix: raise on 404 in ArgoClient.get_workflow_templates (#3239)

### LINKED ISSUE #3239: ArgoClient.get_workflow_templates swallows a 404, making "list failed" indistinguishable from "no templates"
## What

[`ArgoClient.get_workflow_templates`](https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/argo/argo_client.py#L82-L118) is a generator that returns `None` on a 404 from the list call:

```python
except client.rest.ApiException as e:
    ...
    if e.status == 404:
        return None
```

Because it's a generator, `return None` simply stops iteration — the value never reaches the caller and no exception is raised. `list(client.get_workflow_templates())` therefore yields `[]`, indistinguishable from a successful-but-empty list.

## Why it's a problem

A 404 on `list_namespaced_custom_object` for `workflowtemplates` means the namespace or the CRD isn't present (or a misconfigured path) — not "zero templates," which is a 200 with `items: []`. So callers can't tell an enumeration *failure* from an empty cluster. For any reconciling or destructive use (e.g. tearing down templates), the failure reads silently as "nothing to do."

For contrast, within this same file: the other list-path statuses already raise (the non-`Expired` 410 falls through to `raise ArgoClientException`), and the singular getters returning `None` on 404 is fine — there a 404 genuinely means "that named object is absent." It's specifically the *list* generator where empty-on-404 loses information.

## Suggested fix

Raise on the list 404, as the other statuses do, instead of `return None`:

```python
if e.status == 404:
    raise ArgoClientException(error_message)
```

(or a typed `ArgoResourceNotFound` via `wrap_api_error`, if preferred). This keeps "no templates" as `[]` and surfaces a real listing failure.

Reproduced on `master` (linked file). Happy to send a PR if the direction looks right.

Haris-bin-shakeel:
## PR Type
- [x] Bug fix

## Summary
`get_workflow_templates` silently yields `[]` on a 404, making an enumeration failure indistinguishable from a healthy empty namespace.

## Issue
Fixes #3239

## Root Cause
`return None` inside a generator stops iteration without raising. A 404 on `list_namespaced_custom_object` means the namespace or CRD is missing — not "zero templates" (that's a 200 with `items: []`).

## Why This Fix Is Correct
`error_message` is already constructed before the 404 branch. Every other non-410 status in that block raises `ArgoClientException(error_message)` — this makes 404 consistent with that fallthrough. Singular getters keep `return None` on 404; there a missing named resource is the correct semantic.

## Tests
- [x] CI passes

Two tests added:
- `test_get_workflow_templates_404_raises` — 404 raises `ArgoClientException`
- `test_get_workflow_templates_empty_200_does_not_raise` — empty 200 yields `[]`

## Non-Goals
Singular getters (`get_workflow_template`, etc.) intentionally not changed.

## AI Tool Usage

Used Cursor for initial analysis assistance. Reviewed, understood, and tested all code personally.

### DIFF
--- a/metaflow/plugins/argo/argo_client.py
+++ b/metaflow/plugins/argo/argo_client.py
@@ -109,7 +109,7 @@ class ArgoClient(object):
                 error_body = json.loads(e.body) if e.body else {}
                 error_message = error_body.get("message", e.reason)
                 if e.status == 404:
-                    return None
+                    raise ArgoClientException(error_message)
                 elif e.status == 410 and error_body.get("reason") == "Expired":
                     new_token = error_body.get("metadata", {}).get("continue")
                     if new_token:
--- a/test/unit/test_argo_client.py
+++ b/test/unit/test_argo_client.py
@@ -1,0 +1,44 @@
+import pytest
+from kubernetes.client.rest import ApiException
+
+from metaflow.plugins.argo.argo_client import ArgoClient, ArgoClientException
+
+
+@pytest.fixture
+def argo_client(mocker):
+    mock_custom_objects_api = mocker.Mock()
+    mock_k8s = mocker.Mock()
+    mock_k8s.CustomObjectsApi.return_value = mock_custom_objects_api
+    mock_k8s.rest.ApiException = ApiException
+
+    mock_kubernetes_client = mocker.Mock()
+    mock_kubernetes_client.get.return_value = mock_k8s
+
+    mocker.patch(
+        "metaflow.plugins.argo.argo_client.KubernetesClient",
+        return_value=mock_kubernetes_client,
+    )
+
+    client = ArgoClient(namespace="test-ns")
+    return client, mock_custom_objects_api
+
+
+def test_get_workflow_templates_404_raises(argo_client):
+    client, mock_api = argo_client
+    api_error = ApiException(status=404, reason="Not Found")
+    api_error.body = '{"message": "workflowtemplates.argoproj.io not found"}'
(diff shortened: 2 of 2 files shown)


Shriprasad-P (approved):
I independently reproduced the generator behavior this fixes.

On the base revision, a 404 reaches `return None` inside `get_workflow_templates()`, which terminates the generator normally. As a result, callers consuming it with `list()` receive `[]`, making a failed collection request indistinguishable from a successful 200 response with an empty `items` list.

This change correctly preserves that distinction by raising `ArgoClientException` for the 404 path while leaving valid empty collections unchanged.

I also checked the existing pagination / 410 Expired handling and the singular getter behavior; both remain unaffected. The new regression test fails against the base behavior and passes with this patch.

LGTM.

Haris-bin-shakeel:
Hi @mcg1969,

Just rebased this PR onto the latest `master` (`v2.19.39`) so all checks can run cleanly against the current base.

Thanks to @Shriprasad-P for independently reproducing and approving the fix!

Shriprasad-P (commented):
## Review

Makes sense to raise on 404 from `get_workflow_templates` instead of returning `None` and letting callers mis-handle a missing CRD/API as “empty list”.

The unit tests distinguish 404 → `ArgoClientException` vs empty 200 → `[]`, which is exactly the behavioral split this change needs.

Commenting as a non-maintainer; change LGTM from a correctness standpoint.

PR #3243, opened 2026-06-08
security: replace predictable production token with cryptographic randomness

Snakinya:
## Summary

The production token mechanism in `metaflow/plugins/aws/step_functions/production_token.py` uses a **fully deterministic** algorithm to generate deployment authorization tokens:

```python
random.seed(zlib.adler32(to_bytes(prefix)))
yield prefix + "".join(random.sample(string.ascii_lowercase, 4))
```

This means anyone who knows the flow name (which is public information) can compute the exact production token and bypass deployment authorization checks in Step Functions, Airflow, and Argo Workflows.

**Before (predictable — same output every time):**
```
production:MyFlow-0-chby   ← always this value
```

**After (cryptographically random):**
```
production:MyFlow-heq2nrw5qc73f0ry   ← different every time
```

### Vulnerability Details

- **CWE:** CWE-330 (Use of Insufficiently Random Values)
- **Affected components:** Step Functions, Airflow, Argo Workflows deployment authorization
- **Impact:** Any user who knows a flow name can predict the production token and re-deploy/modify production workflows belonging to other users
- **Root cause:** `zlib.adler32` (a checksum, not CSPRNG) used as seed for `random.sample` with only 4 lowercase letters (26⁴ = 456,976 possibilities, all enumerable)

### Fix

Replace with `secrets.choice()` generating 16-character alphanumeric tokens (36¹⁶ ≈ 7.96×10²⁴ possibilities).

Backwards compatible: existing stored tokens continue to work via `load_token()`. Only newly generated tokens use the secure algorithm.

## Test plan

- [x] `new_token()` produces different output on each call
- [x] Token format remains `{prefix}-{suffix}` (compatible with existing storage/comparison)
- [x] `load_token()` / `store_token()` unchanged
- [ ] Existing production deployments can still generate new tokens on next `create`

### DIFF
--- a/metaflow/plugins/aws/step_functions/production_token.py
+++ b/metaflow/plugins/aws/step_functions/production_token.py
@@ -1,20 +1,17 @@
 import json
 import os
-import random
+import secrets
 import string
-import zlib
-from itertools import dropwhile
-
-from metaflow.util import to_bytes
 
 
-def _token_generator(token_prefix):
-    for i in range(10000):
-        prefix = "%s-%d-" % (token_prefix, i)
-        # we need to use a consistent hash here, which is why
-        # random.seed(prefix) or random.seed(hash(prefix)) won't work
-        random.seed(zlib.adler32(to_bytes(prefix)))
-        yield prefix + "".join(random.sample(string.ascii_lowercase, 4))
+
+_TOKEN_LENGTH = 16
+_TOKEN_ALPHABET = string.ascii_lowercase + string.digits
+
+
+def _generate_token(token_prefix):
+    suffix = "".join(secrets.choice(_TOKEN_ALPHABET) for _ in range(_TOKEN_LENGTH))
+    return "%s-%s" % (token_prefix, suffix)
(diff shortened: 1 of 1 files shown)


saikonen (commented):
The main purpose of the production tokens are to act as a per-author-lock on deployed flows, rather than a true authorization secret token. This is in order to prevent unintentionally deploying over an existing flow of identical name, but of another author.
The token value itself is not heavily protected in the underlying systems so it is easily retrievable if one so desires (even the comparisons are performed client-side on the existing deployment).

As such, I don't see a strong reason for changing the token generation logic at this point,  unless there is also a concern of likely token value collisions?

Shriprasad-P (commented):
## Review

**PR:** security: replace predictable production token with cryptographic randomness

**Touched:** `metaflow/plugins/aws/step_functions/production_token.py`

- Security-sensitive change — please double-check edge cases and defaults.
- Size looks manageable (+12/-20).

PR #3259, opened 2026-06-10
fix: regression with StepMutator running multiple times in some cases

### LINKED ISSUE #3258: StepMutator bug in attaching decorators affecting 2.19.33
With the changes to `_process_late_attached_decorator` in #3238 some card attaching custom decorators have stopped working due to duplicate inits with logs like
```
[@card WARNING] Multiple `@card` decorator have been annotated with duplicate ids : test_card. `current.card['test_card']` will not work
[@card WARNING] `current.card['test_card']` is not present. Please set the `id` argument in @card to 'test_card' to access `current.card['test_card']`. `current.card['test_card']` will return an empty `list` which is not referenced to `current.card` object.
```

Example flow for repro:
```python
from metaflow import FlowSpec, step, card, current, conda_base
from metaflow.user_decorators.user_step_decorator import StepMutator


class AddTestCard(StepMutator):
    def mutate(self, mutable_step):
        from metaflow import card as card_deco

        mutable_step.add_decorator(card_deco, deco_kwargs={"id": "test_card"})

@conda_base
class ReproFlow(FlowSpec):
    @card
    @AddTestCard()
    @step
    def start(self):
        print(type(current.card["test_card"]))
        self.next(self.end)

    @step
    def end(self):
        pass


if __name__ == "__main__":
    ReproFlow()

```

saikonen:
## PR Type

- [x] Bug fix
- [ ] New feature
- [ ] Docs / tooling
- [ ] Refactoring

## Issue

Fixes #3258

## Reproduction

**Runtime:** 

**Commands to run:**
```bash
# paste exact commands
```

**Where evidence shows up:** 

<details>

```
paste here
```

</details>

<details>

```
paste here
```

</details>

## Failure Modes Considered

1.
2.

## Tests

- [x] CI passes

## AI Tool Usage

### DIFF
--- a/metaflow/decorators.py
+++ b/metaflow/decorators.py
@@ -939,22 +939,12 @@ def _init_step_decorators(
             inserted_by_value = [deco.decorator_name] + (deco.inserted_by or [])
 
             if isinstance(deco, StepMutator):
-                # Before (re)running the mutator, remove any decorators it
-                # previously inserted in order to prevent adding duplicate decorators.
-                mutator_name = deco.decorator_name
-                step.decorators = [
-                    d
-                    for d in step.decorators
-                    if not (d.inserted_by and d.inserted_by[0] == mutator_name)
-                ]
-                step.wrappers = [
-                    d
-                    for d in (step.wrappers or [])
-                    if not (
-                        getattr(d, "inserted_by", None)
-                        and d.inserted_by[0] == mutator_name
-                    )
-                ]
+                # Guard against calling mutate more than once (e.g. if
+                # _init_step_decorators is invoked multiple times), which could
+                # have adverse effects for allow_multiple decorators like @card.
+                if deco._mutate_called:
+                    continue
+
                 debug.userconf_exec(
                     "Evaluating step level decorator %s for %s (mutate)"
                     % (deco.__class__.__name__, step.name)
@@ -968,6 +958,7 @@ def _init_step_decorators(
                         inserted_by=inserted_by_value,
                     )
                 )
+                deco._mutate_called = True
             else:
                 raise MetaflowInternalError(
                     "A non StepMutator found in step custom decorators"
@@ -1058,17 +1049,30 @@ def _process_late_attached_decorator(
             continue
         for deco in s.config_decorators:
             if isinstance(deco, StepMutator):
-                # Before (re)running the mutator, remove any decorators it
-                # previously inserted in order to prevent adding duplicate decorators.
+                # Guard against calling the same mutator more than once within
+                # this late-attached pass (e.g. if this function is called
+                # multiple times).  We use a separate flag (_late_mutate_called)
+                # so that a mutator already called by _init_step_decorators
+                # (_mutate_called=True) is still re-run here when @kubernetes
+                # (or similar) is genuinely late-attached.
+                if deco._late_mutate_called:
+                    continue
+                # Before re-running, remove any step decorators that this
+                # mutator inserted during the earlier _init_step_decorators
+                # pass so that the re-run is idempotent. Without this,
+                # allow_multiple decorators (e.g. @card) added by the mutator would duplicate.
                 mutator_name = deco.decorator_name
                 s.decorators = [
                     d
                     for d in s.decorators
-                    if not (d.inserted_by and d.inserted_by[0] == mutator_name)
+                    if not (
+                        getattr(d, "inserted_by", None)
+                        and d.inserted_by[0] == mutator_name
+                    )
                 ]
                 s.wrappers = [
                     d
-                    for d in (s.wrappers or [])
+                    for d in s.wrappers
                     if not (
                         getattr(d, "inserted_by", None)
                         and d.inserted_by[0] == mutator_name
@@ -1088,6 +1092,7 @@ def _process_late_attached_decorator(
                         inserted_by=inserted_by_value,
                     )
                 )
+                deco._late_mutate_called = True
                 mutators_ran = True
 
     # Rebuild the graph so node.decorators reflects mutator changes (OVERRIDE).
--- a/metaflow/decorators.py
+++ b/metaflow/decorators.py
@@ -939,30 +939,17 @@ def _init_step_decorators(
             inserted_by_value = [deco.decorator_name] + (deco.inserted_by or [])
 
             if isinstance(deco, StepMutator):
-                # Guard against calling mutate more than once (e.g. if
-                # _init_step_decorators is invoked multiple times), which could
-                # have adverse effects for allow_multiple decorators like @card.
-                # Also skip if _process_late_attached_decorator already ran the
-                # mutator (_late_mutate_called=True): that happens when a
-                # platform's flow_init late-attaches decorators *before*
-                # _init_step_decorators is called (conda/pypi pattern).
-                if deco._mutate_called or deco._late_mutate_called:
+                # Skip if mutate() already ran (e.g. a @conda/@pypi late-attach pass
+                # ran it before this normal pass). The late-attach pass re-runs
+                # idempotently when genuinely new decorators appear.
+                if deco._mutate_called:
                     continue
-
-                debug.userconf_exec(
-                    "Evaluating step level decorator %s for %s (mutate)"
-                    % (deco.__class__.__name__, step.name)
+                _run_step_mutator(
+                    cls,
+                    step,
+                    deco,
+                    "Evaluating step level decorator %s for %s (mutate)",
                 )
-                deco.mutate(
-                    MutableStep(
-                        cls,
-                        step,
-                        pre_mutate=False,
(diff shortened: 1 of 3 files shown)


valayDave (dismissed):

valayDave (dismissed):

npow:
## Canon PR Review — Engineering Standards Scorecard

**PR:** fix: regression with StepMutator running multiple times in some cases
**Reviewed against:** Top 20 Tier 2 practices (Code Complete, Clean Code, APoSD, Refactoring, Release It!, DDIA, The Pragmatic Programmer, Working Effectively with Legacy Code, Design Patterns GoF)

### Scores

| Dimension | Score | Gate |
|-----------|-------|------|
| Code Quality (naming, SRP, DRY, guard clauses, smells) | 4/5 | ✅ |
| Module Design (deep modules, composition, docs, encapsulation) | 4/5 | ✅ |
| Reliability (error handling, idempotency, circuit breakers, timeouts, degradation) | 5/5 | ✅ |
| Maintainability (small steps, test before change, broken windows, regression tests) | 4/5 | ✅ |

**Decision: APPROVE**

---

### Violations Found

#### Module Design

**CC-028 — Interface Comments (what + why)** *(Source: A Philosophy of Software Design)*
- **Location:** `metaflow/user_decorators/user_step_decorator.py` — class body of `StepMutator`
- **Issue:** The two new class-level flags `_mutate_called` and `_late_mutate_called` lack a comment explaining their semantics as *per-instance* idempotency guards and, critically, the assumption that each decorator instance is used in exactly one flow invocation. Without this, a reader could reasonably wonder: (a) why these are class vars and not instance vars initialized in `__init__`; (b) whether reuse of the same decorator object across multiple flow invocations in the same process would silently skip mutation (it would).
- **Concern:** Python will correctly shadow the class var with an instance attribute on first `deco._mutate_called = True`, so the guard logic itself is correct. The issue is that the flags are never reset. In typical CLI usage this is fine (each `python flow.py run` is a fresh process). But in notebook or programmatic multi-run patterns, the same `@MyMutator()` instance persists across calls and the second run's mutator would be silently skipped.
- **Suggested improvement:**
  ```python
  # Per-instance flags tracking whether mutate() has been called via
  # _init_step_decorators (_mutate_called) or _process_late_attached_decorator
  # (_late_mutate_called).  These intentionally live at class level so that
  # reading the attribute before the first assignment works without __init__
  # changes, relying on Python's instance-attribute shadowing on first write.
  # NOTE: these are never reset — assumes each decorator instance is used in
  # a single flow invocation (true for CLI; not true for multi-run notebooks).
  _mutate_called = False
  _late_mutate_called = False
  ```

#### Maintainability

**CC-093 — Fix Broken Windows** *(Source: The Pragmatic Programmer)*
- **Location:** PR description
- **Issue:** The PR template's `Summary`, `Reproduction`, `Root Cause`, `Why This Fix Is Correct`, and `Failure Modes Considered` sections are all empty. For a subtle call-stack-order bug in a shared primitive (`StepMutator`), the root cause and the invariant being restored are non-obvious and would be valuable to preserve alongside the fix.
- **Suggested improvement:** Fill in at minimum `Root Cause` (the ordering inversion between `_process_late_attached_decorator` and `_init_step_decorators` in conda/pypi's `flow_init`) and `Why This Fix Is Correct` (the two-flag asymmetry: `_init_step_decorators` checks both flags; `_process_late_attached_decorator` checks only `_late_mutate_called` so it can still re-run when kubernetes is genuinely late-attached).

---

### What's Working Well

1. **Idempotency fix is precise and minimal** — The two-flag design elegantly handles the asymmetric call orders. `_init_step_decorators` stops on either flag; `_process_late_attached_decorator` only stops on its own flag so it can still re-run when kubernetes is genuinely late-attached and always cleans up previously-inserted decorators before re-running. This is exactly right for the problem.

2. **Regression test precisely demonstrates the bug** — `test_allow_multiple_decorator_not_duplicated_on_mutator_rerun` sets up the production call order (late-attach *before* init, as with conda/pypi), exercises `allow_multiple` decorators (`@card`), and asserts the exact invariant. It would fail on `master` without the fix and passes after. The docstring explaining the production scenario is a genuine asset.

3. **Guard clauses via `continue`** — Both guard blocks use early `continue` at the top of the loop body, keeping the happy path at leftmost indentation and making the intent clear (CC-011).

---

### To Unblock

Nothing is blocking merge. Two follow-ups worth tracking (not required for this PR):

1. Consider adding a comment (or docstring) to the two new class-level flags explaining the instance-shadowing assumption and the no-reset caveat for programmatic multi-run usage.
2. Fill in the PR description's root cause and fix rationale — valuable for `git log` archaeology when this area next changes.

---

*Scored against: CC-003, CC-005, CC-006, CC-011, CC-021 (Code Quality) | CC-027, CC-015, CC-028, CC-017 (Module Design) | CC-014, CC-048, CC-056, CC-058, CC-060 (Reliability) | CC-019, CC-031, CC-093, CC-107 (Maintainability)*

talsperre:
Discussed async - details [here](https://gist.github.com/talsperre/4083950b907210774b8be6778b536084).

npow:
## Follow-up: Updating decision to REQUEST_CHANGES

<@talsperre> shared a [detailed counter-review](https://gist.github.com/talsperre/4083950b907210774b8be6778b536084) with 4 findings. After cross-checking against the head commit (`ff68ea9`), here's where I land:

### Finding 1 (late-before-init still duplicates) — **Disagree on current code**

The gist says `_init_step_decorators` re-runs `mutate()` even when `_late_mutate_called=True`. In the head commit, line ~944 reads:

```python
if deco._mutate_called or deco._late_mutate_called:
    continue
```

So if the late pass ran first and set `_late_mutate_called=True`, `_init_step_decorators` skips the mutator. This ordering (conda/pypi) should produce exactly one `@card`. The gist may have reviewed an earlier commit before "combine guards and cleanup" landed.

**However, I'd recommend a targeted test to confirm** — `_call_init_step_decorators` after `_call_process_late_attached` with a `@card` mutator and asserting count == 1.

### Finding 2 (name-based cleanup deletes wrong outputs for two instances of same class) — **Confirmed**

`decorator_name` is class-level (`{module}.{ClassName}`), not per-instance. The cleanup in `_process_late_attached_decorator`:

```python
d.inserted_by[0] == mutator_name
```

...uses `deco.decorator_name` which is shared by all instances of the same class. Two `@AddCardMutator(id="card1")` and `@AddCardMutator(id="card2")` on the same step would have their outputs cross-deleted. Uncommon but real.

### Finding 3 (late-pass re-run leaves replacement decorator uninitialized) — **Confirmed, more critical**

Scenario B (normal init ran first, then kubernetes late-attaches):

1. `_init_step_decorators` runs → mutator adds `@card` → `@card` gets `external_init` + `step_init` at lines 984–1011 ✓
2. `_process_late_attached_decorator` runs → cleanup removes the initialized `@card` → mutator re-adds a fresh `@card` → `_late_mutate_called = True`
3. Late-pass init loop (lines 1108–1126): `if deco.name in deco_names` → only initializes "kubernetes", not "card"

The replacement `@card` is never `external_init`'d or `step_init`'d. This is a silent failure — the card object exists in `step.decorators` but its `_ran_init` is `False` and its internal state is uninitialized. Depending on what `CardDecorator.step_init` does, this could silently produce no card or error at task time.

### Finding 4 (test misses Findings 2 and 3) — **Confirmed**

The existing test only asserts `card_count == 1` for `inserted_by` objects. It passes even when the surviving card is the uninitialized replacement from Finding 3.

---

### Updated decision: **REQUEST_CHANGES**

Finding 3 is the blocking issue. The proposed fix in the gist (id-based tracking + explicit re-init of replacement outputs) addresses it correctly. At minimum, the late-pass re-run path needs to call `external_init`/`step_init` on any decorator the mutator freshly inserts (the ones whose initialized predecessors were just removed).

---

*Scored against: CC-014 (silent failure mode on uninitialized decorator — Reliability), CC-107 (regression test doesn't catch the fixed behavior — Maintainability)*

npow (approved):
All three findings addressed in the final commit:

**Finding 3 (was blocking) — resolved.** id-based tracking (`_mutate_inserted`) now captures exactly which decorator/wrapper objects each mutator run produces. `_remove_step_mutator_outputs` removes by object id (not by name), so two instances of the same mutator class no longer cross-delete. `reinit_decorator_ids` feeds back into the late-attach init loop so replacement decorators get `external_init` called — the missing initialization is now explicit.

**Finding 2 (name-based cross-delete) — resolved.** Replaced by id-based cleanup.

**Finding 1 (late-before-init duplicate) — resolved.** `_init_step_decorators` now skips a mutator that already set `_mutate_called`, preventing the double-run when conda/pypi calls `_process_late_attached_decorator` before `_init_step_decorators`.

Tests cover the late-before-init ordering, multiple distinct late attachments, and the replacement re-init path. All CI green. LGTM — good to merge.

PR #3271, opened 2026-06-14
Fix R run argument quoting

fallintoplace:
## Summary

Updates the R `run()` path to build `Rscript` arguments as an argv vector and execute them through `system2()`, with shell quoting applied at the execution boundary.

## Issue

No existing issue.

## Reproduction

**Runtime:** local R binding

**Commands to run:**
```bash
cd R
Rscript -e 'testthat::test_file("tests/testthat/test-run.R")'
Rscript -e 'testthat::test_file("tests/testthat/test-flags.R")'
Rscript -e 'testthat::test_file("tests/testthat/test-sfn-cli-parsing.R")'
```

**Where evidence shows up:** testthat output from the focused R tests.

<details>
<summary>Before</summary>

`run_cmd()` assembled one command string with `paste()`, so values containing spaces, quotes, semicolons, or other shell-sensitive characters were interpreted as part of the shell command line instead of staying as single CLI arguments.

</details>

<details>
<summary>After</summary>

`run_argv()` keeps each CLI token as a separate argument. `run()` passes the quoted argv vector to `system2("Rscript", ...)`, and `run_cmd()` now returns a shell-quoted display command.

</details>

## Root Cause

The R wrapper built a single shell command string by concatenating the wrapper path, flow RDS path, run mode, tags, parameters, batch identifiers, and extra args. That mixed argument construction with shell parsing, so argument boundaries depended on shell syntax instead of the values passed to `run()`.

## Why This Fix Is Correct

The fix separates command construction from command rendering. `run_argv()` builds the intended CLI argument vector, `run()` quotes that vector only at the `system2()` boundary, and `run_cmd()` remains available as a quoted display string.

## Failure Modes Considered

1. Values containing spaces, quotes, semicolons, `$`, and unicode must remain single arguments.
2. Existing R binding command shapes still need to be preserved for default runs, `batch = TRUE`, Step Functions commands, help, and vector-valued parameters.

## Tests

- Added argv regression coverage for spaces, quotes, semicolons, `$`, and unicode in tags, parameters, batch user/run IDs, and `other_args`.
- Updated Step Functions command parsing tests to assert argv vectors instead of splitting shell strings.
- Updated the R test helper commands to use `system2()`.
- Ran `git diff --check` locally.
- Could not run the focused R tests locally because `Rscript` is not installed in this environment.

## Non-Goals

- This does not change the Python CLI behavior.
- This does not rewrite the R command-line parser beyond preserving argument boundaries for the command that `run()` executes.

### DIFF
--- a/R/R/flags.R
+++ b/R/R/flags.R
@@ -136,6 +136,10 @@ split_flags <- function(arguments) {
 }
 
 split_parameters <- function(flags) {
+  paste(split_parameter_args(flags), collapse = " ")
+}
+
+split_parameter_args <- function(flags) {
   parameters <- !names(flags) %in% c(
     "metaflow_path", "run",
     "batch", "datastore", "metadata",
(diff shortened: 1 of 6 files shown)


LuisJG8:
Hi @fallintoplace, fyi the R integration is basically deprecated so we are not accepting any contributions to it.

agsaru (commented):
R module is deprecated so changes here do not matter anymore. Confirmed by the admin.

Issue #3286, opened 2026-06-27
Unescaped paths break PyPI/Conda installation for users with spaces in their directories


anushkagupta200615-jpg:
### Describe the issue
In `metaflow/plugins/pypi/micromamba.py` (around line ~394), the `_download_and_extract` function uses `subprocess.Popen` with `shell=True` to download and extract `micromamba`:

```python
result = subprocess.Popen(
    f"curl -Ls {url} | tar -xvj -C {installation_location} bin/micromamba",
    shell=True,
    stderr=subprocess.PIPE,
    stdout=subprocess.PIPE,
)
```

Proposed solution
Either:

Wrap the {installation_location} in quotes in the f-string (e.g., \"{installation_location}\"), or better yet, use shlex.quote(installation_location) to safely escape the path before passing it to the shell.
Avoid shell=True entirely by piping a curl subprocess's stdout directly into a tar subprocess's stdin using Python's native subprocess piping.

allenjamesvinoy:
Hey @anushkagupta200615-jpg  - are you working on this or are you looking for volunteers to implement one of the mentioned solutions?

anushkagupta200615-jpg:
Hey @allenjamesvinoy! I am actually working on this right now. I already have a fix using `shlex.quote` implemented locally and will be opening a Pull Request for it shortly! Thanks for asking though. 🚀

allenjamesvinoy:
Sure @anushkagupta200615-jpg. Thanks for confirming.
Are you part of the Metaflow team or another volunteer? I am curious to know how I can lookout for issues to contribute to.

anushkagupta200615-jpg:
hey @allenjamesvinoy I am a volunteer contributor as well! If you're looking to get involved, the best place to start is by filtering the issues .

PR #3287, opened 2026-06-27
fix: escape installation_location in micromamba plugin (#3286)

anushkagupta200615-jpg:
This PR resolves issue #3286 by safely escaping the `installation_location` in the PyPI/Conda micromamba setup script. 

Previously, if a user's installation directory path contained spaces (e.g. `/Users/My Name/`), the underlying `tar` subprocess executed via `shell=True` would incorrectly parse the path as multiple arguments, causing a fatal error and breaking the installation. By applying Python's standard `shlex.quote()`, we guarantee the path is passed safely to the shell, fixing the issue without needing to restructure the underlying subprocess piping.

### DIFF
--- a/metaflow/plugins/pypi/micromamba.py
+++ b/metaflow/plugins/pypi/micromamba.py
@@ -391,9 +391,8 @@ def _install_micromamba(installation_location):
         for attempt in range(max_retries):
             try:
                 # https://mamba.readthedocs.io/en/latest/micromamba-installation.html#manual-installation
-                # requires bzip2
                 result = subprocess.Popen(
-                    f"curl -Ls {url} | tar -xvj -C {shlex.quote(installation_location)} bin/micromamba",
+                    f"curl -Ls {url} | tar -xv -C {shlex.quote(installation_location)} bin/micromamba",
                     shell=True,
                     stderr=subprocess.PIPE,
                     stdout=subprocess.PIPE,
--- a/metaflow/plugins/pypi/micromamba.py
+++ b/metaflow/plugins/pypi/micromamba.py
@@ -2,6 +2,7 @@ import functools
 import json
 import os
 import re
+import shlex
 import shutil
 import subprocess
 import tempfile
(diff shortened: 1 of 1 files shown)


allenjamesvinoy:
From my understanding, the 'j' switch in tar -xvj is redundant in this extraction use case. What do you think about removing it? @anushkagupta200615-jpg

anushkagupta200615-jpg:
@allenjamesvinoy Thanks for catching that! You're absolutely right—the j switch isn't strictly necessary since tar can automatically detect the compression format. I've just pushed a commit removing it . Thanks for the suggestion!

Shriprasad-P (changes requested):
I verified that the `shlex.quote(installation_location)` change correctly fixes the original path parsing issue on the supported POSIX platforms, including paths containing spaces and shell metacharacters.

There is one blocking regression in the current command, however.

The micromamba archive is bzip2-compressed and is streamed directly into `tar` via:

    curl ... | tar ...

This PR changes `tar -xvj` to `tar -xv`. GNU tar does not automatically detect compression when reading a compressed archive from a pipe; it requires the corresponding decompression option. For this bzip2 stream that means `-j`.

As a result, the new command can work with BSD/macOS tar while failing on supported Linux systems using GNU tar before `bin/micromamba` is extracted.

Please keep the quoting fix but restore explicit bzip2 decompression, e.g. retain `-j`, or change the implementation to download the archive to a named file before extraction.

A regression test covering the streamed archive path on Linux would also be valuable.

The `shlex.quote()` portion of the change itself looks correct.

Shriprasad-P (commented):
## Review

Quoting `installation_location` with `shlex.quote` is the right fix for path injection / spaces in the micromamba install shell pipeline.

**Note:** this also switches `tar -xvj` → `tar -xv`. Please confirm the micromamba tarball format still extracts correctly without explicit bzip2 (`j`) — if the archive is still `.tar.bz2`, auto-detect may work on GNU tar but is worth an explicit smoke test.

PR #3294, opened 2026-07-07
docs: correct CLIENT_CACHE_MAX_SIZE unit comment to MB

anxkhn:
## PR Type

- [x] Docs / tooling

The inline comment above `CLIENT_CACHE_MAX_SIZE` says the value is "in bytes", but
the only consumer multiplies it by `1024**2`, so the value is actually interpreted
in megabytes. This corrects the comment to "in MB". No behavior changes.

## Issue

Fixes #<OPEN_ACKNOWLEDGED_ISSUE>   <!-- REQUIRED: do not open the PR until this

### DIFF
--- a/metaflow/metaflow_config.py
+++ b/metaflow/metaflow_config.py
@@ -111,7 +111,7 @@ DATASTORE_SYSROOT_GS = from_conf("DATASTORE_SYSROOT_GS")
 
 # Path to the client cache
 CLIENT_CACHE_PATH = from_conf("CLIENT_CACHE_PATH", "/tmp/metaflow_client")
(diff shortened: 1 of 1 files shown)


anxkhn:
@saikonen when you have a chance, could you please take a look at this pr and let me know if any update is needed?

Shriprasad-P (commented):
## Review

**PR:** docs: correct CLIENT_CACHE_MAX_SIZE unit comment to MB

**Touched:** `metaflow/metaflow_config.py`

- Small docs/comment cleanup; low risk if wording is accurate.
- Size looks manageable (+1/-1).

PR #3306, opened 2026-07-18
Add sidecar option and allow sidecar monitoring

aquarin:
## PR Type

- [ ] Bug fix
- [ ] New feature
- [x] Docs / tooling
- [ ] Refactoring

## Summary

Per the linked issue, customers have seen Metaflow UI/CLI step logs stall until a step fully completes, despite logs being correct on the execution container itself.

Initial investigation pinpointed the issue to be related to `metaflow/mflog/save_logs_periodically.py`, which is responsible of transporting (copying) the local mflog_std to an S3 folder, which will in turn be pulled by the metaflow cli/ui for displaying to users. But we lacked instrumentation for further investigation. This PR adds monitoring scaffolding to help investigate.

This PR:
1. Adds a way to enable options in sidecar class, including
  a) taking an python option object in the sidecar constructor b) serialize this to process spawning command line c) deserialize the option from the command line args to sidecar worker
2. Adds stdout prints in `metaflow/mflog/save_logs.py`, which is to be captured by save_logs_periodically.py, and to be saved in its logs. 
2. When the option specifies `enable_tracing=True`, `save_logs_periodically.py` prints its own logs indicating each upload's size, time, etc, to a separate log file, for further investigation of this bug. 

## Issue

https://github.com/Netflix/metaflow/issues/3307

## Reproduction

The repro happens in a corporate environment with internal flow code we can't share, but here's what we can disclose:

* During the issue, Metaflow UI/CLI logs stall, while the execution container (not somewhere a typical user would check) shows correct, up-to-date logs.
* Once the step completes, the UI/CLI log fully catches up and updates correctly.
* Since we couldn't access the customer's flow code directly (cost/permissions), we simulated the failure by injecting upload failures and sidecar process failures — both independently reproduce the same symptom. This is the [simulation code](https://github.com/Netflix/metaflow/tree/mock-sidecar-failures)

**Runtime:**
Corporation private cloud container. 

**Commands to run:**
Internal flow. 

**Where evidence shows up:**

(Corporation internal environment)

</details>

## Why This Fix Is Correct
N/A

## Failure Modes Considered

N/A

## Tests

- [X] Unit tests added/updated
- [ ] CI passes

## AI Tool Usage

Claude code and Codex


metaflow/mflog/save_logs_periodically.py
```diff
@@ -5,11 +5,30 @@
...
 from metaflow.sidecar import MessageTypes
-from . import update_delay, BASH_SAVE_LOGS_ARGS
+from metaflow.util import to_unicode
+from . import update_delay, BASH_SAVE_LOGS_ARGS, TASK_LOG_SOURCE
+from .mflog import decorate
+
+
+def _write_uploader_log(message):
+    payload = decorate(TASK_LOG_SOURCE, "%s\n" % message)
+    with open(os.environ["PERIODICAL_UPLOADER_STDOUT"], "ab", buffering=0) as log:
```
talsperre: Make `PERIODICAL_UPLOADER_STDOUT` to be a metaflow config. Also, we should have some default value for this file name.

aquarin: @talsperre as discussed, I will use the current design, and later determine this through runtime config feature, and pass this flag to the constructor `def __init__(self, options=None):`

metaflow/mflog/save_logs_periodically.py
```diff
@@ -36,9 +69,20 @@ def _file_size(path):
...
+                if self._enable_tracing:
+                    elapsed = time.time() - start_time
+                    for path, previous, current in zip(
+                        FILES, previous_sizes, new_sizes
+                    ):
+                        _write_uploader_log(
+                            "[save_logs_periodically] file=%s previous_size=%d "
+                            "current_size=%d delta=%d elapsed_seconds=%.3f"
+                            % (path, previous, current, current - previous, elapsed),
+                        )
```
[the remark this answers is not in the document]
aquarin: Added the comments explain why there is no error handling here:

    # This is intentionally not best effort. If PERIODICAL_UPLOADER_STDOUT is
    # incorrectly configured, it should be easy to detect in normal traced runs.
    # Otherwise a user sidecar uploader issue can happen while this diagnostics
    # log silently doesn't exist.
    #
    # Adding another hard-coded log file location to record the error of lacking
    # this log file path in the env variable complicates things. Once this path
    # is setup correctly, it should not normally change, so the chance that this
    # happens in some flows but not others is very low.

metaflow/mflog/__init__.py
```diff
@@ -106,6 +107,8 @@ def export_mflog_env_vars(
     }
     if datastore_root is not None:
         env_vars["MF_DATASTORE_ROOT"] = datastore_root
+    if periodical_uploader_stdout_path is not None:
+        env_vars["PERIODICAL_UPLOADER_STDOUT"] = periodical_uploader_stdout_path
```
saikonen: `PERIODICAL_UPLOADER_LOG_PATH` or something alike would be more descriptive to what this does. The contents as is also includes stderr

metaflow/mflog/save_logs_periodically.py
```diff
@@ -5,11 +5,39 @@
...
+def _write_save_logs_output(stream, output):
+    for line in output.splitlines():
+        _write_uploader_log("[save_logs %s] %s" % (stream, to_unicode(line)))
 
 
 class SaveLogsPeriodicallySidecar(object):
-    def __init__(self):
+    def __init__(self, options=None):
+        options = options or {}
+        if set(options) - {"enable_tracing"}:
```
saikonen: `tracing` is becoming an overloaded term with this so can get a bit confusing. Maybe `enable_debug_logs` or anything descriptive to what is happening under the hood.

metaflow/mflog/save_logs_periodically.py
```diff
@@ -36,9 +78,20 @@ def _file_size(path):
...
+                if self._enable_tracing:
+                    elapsed = time.time() - start_time
+                    for path, previous, current in zip(
+                        FILES, previous_sizes, new_sizes
+                    ):
+                        _write_uploader_log(
+                            "[save_logs_periodically] file=%s previous_size=%d "
+                            "current_size=%d delta=%d elapsed_seconds=%.3f"
+                            % (path, previous, current, current - previous, elapsed),
+                        )
```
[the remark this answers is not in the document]
saikonen: This still seems like an unnecessary foot gun, where a misconfiguration of diagnostics can affect regular log saves.
I'm not entirely sure about the reasoning why this can not be wrapped in a separate try/except? If the filedump related errors are important to surface as well, then these could possibly be written to the main `mflog_stderr`.

Not a blocker for merging though, as it does not affect OSS usage at present.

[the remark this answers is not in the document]
aquarin: Thanks. This is addressed in https://github.com/Netflix/metaflow/pull/3306/changes/1c8ca0694782bc77f1a42e20ea834bf05bedb846.

Still I am not uploading it to the same mflog_stderr file, but rather to another file with hard coded file name to the same folder of mf_stderr.

metaflow/mflog/__init__.py
aquarin: Changed to env_vars["PERIODICAL_UPLOADER_LOG_PATH"] . Thanks.

metaflow/mflog/save_logs_periodically.py
aquarin: Changed to "enable_debug_logs": True,.

talsperre (approved):

metaflow/mflog/save_logs_periodically.py
```diff
@@ -5,11 +5,58 @@
...
+from . import update_delay, BASH_SAVE_LOGS_ARGS, TASK_LOG_SOURCE
+from .mflog import decorate
+
+UPLOADER_DIAGNOSTICS_FALLBACK = "uploader_diagnostics_fallback"
+
+
+def _write_uploader_log(message):
+    try:
+        payload = decorate(TASK_LOG_SOURCE, "%s\n" % message)
+        with open(os.environ["PERIODICAL_UPLOADER_LOG_PATH"], "ab", buffering=0) as log:
```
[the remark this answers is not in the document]
aquarin: I’d prefer to keep this PR scoped as-is and treat non-blocking uploader diagnostics as a separate follow-up if we think it is worth doing. I think this is a corner case of a corner case.

PR #3327, opened 2026-08-03
Set METAFLOW_RUN_ID in step worker subprocess environment

Charlietr14:
### PR Type
- [x] New feature
- [x] Core Runtime change

### Summary
Local `run` execution spawns each step in a subprocess via `Worker` (`metaflow/runtime.py`). Today the subprocess only receives the run id as a CLI argument (`--run-id`), not as an environment variable. Code that needs the run id *before* the CLI is parsed — e.g. during `import metaflow` — has no way to read it. This PR sets `METAFLOW_RUN_ID` in the worker's `env` alongside the existing CLI arg.

### Issue
N/A

### Root Cause / Why This Fix Is Correct
N/A

### Failure Modes Considered
N/A

### Tests
- Internal CI passes
- Unit tests pass

### AI Tool Usage
- [x] AI-assisted (Claude Code)

### DIFF
--- a/metaflow/runtime.py
+++ b/metaflow/runtime.py
@@ -2333,6 +2333,7 @@ class Worker(object):
         # Pass configuration options
         env.update(args.get_env())
         env["PYTHONUNBUFFERED"] = "x"
+        env["METAFLOW_RUN_ID"] = str(self.task.run_id)
         tracing.inject_tracing_vars(env)
         # NOTE bufsize=1 below enables line buffering which is required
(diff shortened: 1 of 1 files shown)


saikonen (approved):
seems to work as expected, and does not impact other runtimes that set `METAFLOW_RUN_ID` as well, such as argo-workflows.

Noteworthy if we want to make the env var a standard across all implementations though, is that currently we do not set this when running `--with kubernetes`. Can be a follow-up if necessary.

talsperre (approved):

PR #3339, opened 2026-08-14
Sanitize StepMutator attributes with to_pod() in _graph_info

### LINKED ISSUE #3338: DAG data leaks raw StepMutator constructor args, breaking Metaflow UI when the argument's class isn't importable
## Summary

Metaflow UI fails to display the DAG for flows using a StepMutator whose constructor takes a non-primitive argument. `graph.py`'s DAG serialization sanitizes classic step decorator attributes with `to_pod()`, but not StepMutator/config-decorator attributes, so a raw object gets pickled into `_graph_info`. Any consumer that can't import the object's class (particularly the Metaflow UI backend) fails to unpickle it.

**Expected**: the DAG renders regardless of which third-party class a StepMutator's constructor argument came from, the same way it already does for classic step decorators.

**Actual**: the DAG tab fails with a stack trace ending in a `ModuleNotFoundError` for the argument's class, whenever the consumer reading `_graph_info` doesn't have that class importable.

## Where

In `node_to_dict()`:

Classic decorator attributes go through `to_pod()`:
https://github.com/Netflix/metaflow/blob/4fce948c811aa4c9b9958f77367d527cf2695921/metaflow/graph.py#L615-L624

StepMutator/config-decorator attributes do not:
https://github.com/Netflix/metaflow/blob/4fce948c811aa4c9b9958f77367d527cf2695921/metaflow/graph.py#L625-L633

```python
{
    "name": deco.decorator_name,
    "attributes": {"_args": deco._args, **deco._kwargs},  # <-- no to_pod()
    "statically_defined": deco.statically_defined,
    "inserted_by": deco.inserted_by,
}
for deco in chain(node.wrappers, node.config_decorators)
```

## How to Reproduce

`payload_module.py`:

```python
class Payload:
    def __init__(self, x):
        self.x = x
```

`repro_flow.py`:

```python
from metaflow.user_decorators.user_step_decorator import StepMutator
from metaflow.user_decorators.mutable_step import MutableStep

from payload_module import Payload


class MyMutator(StepMutator):
    def init(self, payload=None, **kwargs):
        self._payload = payload

    def mutate(self, mutable_step: MutableStep) -> None:
        pass


class ReproFlow(FlowSpec):
    @MyMutator(payload=Payload(1))
    @step
    def start(self):
        self.next(self.end)

    @step
    def end(self):
        pass


if __name__ == "__main__":
    ReproFlow()
```

1. `python repro_flow.py run`
2. Read the run's `_graph_info` artifact. `steps["start"]["decorators"]` contains a raw `Payload` instance under `attributes["payload"]`. Verified:
   ```python
   from metaflow import namespace, Run
   namespace(None)
   run = Run("ReproFlow/<run-id>")
   gi = run["start"].task["_graph_info"].data
   print(gi["steps"]["start"]["decorators"])
   # [{'name': '__main__.MyMutator', 'attributes': {'_args': [], 'payload': <payload_module.Payload object at ...>}, ...}]
   ```
3. Move `payload_module.py` out of the way and read `_graph_info` again from the same directory (so the local datastore still resolves, only the module becomes unimportable). It raises:
   ```
   ModuleNotFoundError: No module named 'payload_module'
   ```
   This is the same failure mode the Metaflow UI backend hits: it has no reason to have any flow-specific module installed, yet ends up needing it just to unpickle `_graph_info`.

## Impact

Any third-party StepMutator/config decorator whose constructor accepts a non-primitive argument breaks the Metaflow UI's DAG tab for every run of every flow using it, since the UI backend has no reason to have that decorator's package installed. This has been the case since StepMutator/`config_decorators` was introduced in 2.16.0 (#2463), and the same code path is unchanged on the current `master` (commit `4fce948c`).

## Suggested fix

Apply `to_pod()` to the StepMutator/config-decorator attributes the same way it's applied to classic decorator attributes, in the second branch of `node_to_dict()`'s `decorators` list.

## Environment

- Metaflow 2.19.35, reproduced locally (no cloud/k8s involved; this is a pure DAG-serialization bug).
- Same code path confirmed present on `master` at commit `4fce948c811aa4c9b9958f77367d527cf2695921`.
- macOS 15 (Darwin 25.6.0, arm64), Python 3.11.15.

sertansenturk-deliveroo:
## Summary
`node_to_dict()` in `graph.py` sanitizes classic step decorator attributes with `to_pod()`, but not StepMutator/config-decorator attributes. This fix applies the same `to_pod()` sanitization to both.

## Context / Motivation
A StepMutator whose constructor takes a non-primitive argument gets that raw object embedded in `_graph_info`, since the second branch of the `decorators` list in `node_to_dict()` skips `to_pod()`. Any consumer that later unpickles `_graph_info` without that argument's class importable fails with `ModuleNotFoundError`. This breaks the Metaflow UI's DAG tab for any flow using such a StepMutator, since the UI backend has no reason to have a given decorator's package installed. Fixes #3338.

Root cause: the two branches of `node_to_dict()`'s `decorators` list construction are inconsistent. `node.decorators` (classic StepDecorator) goes through `to_pod()`; `chain(node.wrappers, node.config_decorators)` (StepMutator/config decorators) does not.

Failure modes considered:
- A StepMutator with only primitive constructor arguments (str/int/float/dict/list): unaffected, `to_pod()` passes these through unchanged.
- A StepMutator with no constructor arguments at all (bare decorator): unaffected, `_args`/`_kwargs` are empty and `to_pod()` on an empty dict is a no-op.

## Changes Made
- Wrap the StepMutator/config-decorator attributes dict in `to_pod()` in `node_to_dict()` (`metaflow/graph.py`).
- Add `test_step_mutator_non_primitive_attribute_sanitized_in_output_steps` to `test/unit/test_graph_structure.py`, which fails without the fix and passes with it.

## Testing
- New unit test added and confirmed it fails without the fix (`assert isinstance(..., str)` fails on the raw object) and passes with it.
- Full unit suite: `cd test/unit && python -m pytest -q` (local runtime, no cloud/k8s involved since this is a pure DAG-serialization path): 563 passed.
- `black --check` on both changed files: clean.
- Manually reproduced the original bug end to end (see #3338 for the standalone repro script) before writing the fix, confirming the exact `ModuleNotFoundError` failure mode this fix addresses.

## AI disclosure
Parts of this PR (investigation, implementation, and test) were developed with AI assistance (Claude Code). I reviewed and understand every change and can answer questions about it.

### DIFF
--- a/metaflow/graph.py
+++ b/metaflow/graph.py
@@ -625,7 +625,7 @@ class FlowGraph(object):
                 + [
                     {
                         "name": deco.decorator_name,
-                        "attributes": {"_args": deco._args, **deco._kwargs},
+                        "attributes": to_pod({"_args": deco._args, **deco._kwargs}),
                         "statically_defined": deco.statically_defined,
                         "inserted_by": deco.inserted_by,
                     }
--- a/test/unit/test_graph_structure.py
+++ b/test/unit/test_graph_structure.py
@@ -9,12 +9,14 @@ Verifies that:
 - _graph_info contains start_step/end_step fields
 - Lint validation works with structural inference
 - @step(start=True) / @step(end=True) annotations work correctly
+- StepMutator/config-decorator attributes go through to_pod() in output_steps()
 """
 
 import pytest
 from metaflow import Config, FlowMutator, FlowSpec, step, Parameter, retry, resources
 from metaflow.flowspec import FlowStateItems
 from metaflow.lint import linter, LintWarn
+from metaflow.user_decorators.user_step_decorator import StepMutator
 
 # ---------------------------------------------------------------------------
 # Flow definitions for testing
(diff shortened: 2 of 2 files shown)


talsperre (approved):
Reviewed the fix and regression coverage. The change is correctly scoped to sanitize StepMutator/config-decorator attributes in _graph_info, and I found no blocking issues.

Issue #3341, opened 2026-08-15
Client parent_tasks/child_tasks resolve wrong tasks when a foreach index is a string prefix of another


winklemad:
## Problem

`LocalMetadataProvider.filter_tasks_by_metadata` matches metadata values with `re.match`, which anchors at the start but not the end:

```python
# metaflow/plugins/metadata_providers/local.py:264
regex = re.compile(pattern)
...
and regex.match(meta.get("value", ""))
```

`Task.parent_tasks` / `child_tasks` build the pattern from `foreach-execution-path`, so for a foreach with 11 or more items the pattern `middle:1` also matches the values `middle:10` and `middle:11`.

## Reproduction

Local metadata, 12 item foreach, `start -> middle -> tail -> join -> end`:

```python
class ForeachPrefixFlow(FlowSpec):
    @step
    def start(self):
        self.items = list(range(12))
        self.next(self.middle, foreach="items")
    # middle -> tail -> join -> end
```

```
python flow.py run --max-workers 4
```

Then through the Client API:

```
WRONG PARENTS  ForeachPrefixFlow/<run>/tail/14  fep=middle:1  -> ['middle:1', 'middle:10', 'middle:11']
1 of 12 tail tasks have the wrong parent count
```

Only the task whose path is a string prefix of another is affected. Every other index resolves to exactly one parent.

The `spin` CLI turns this into a hard failure:

```
$ python flow.py spin ForeachPrefixFlow/<run>/tail/14
[.../tail/0] Internal error:
[.../tail/0] Step tail is not a join step but it gets multiple inputs.
[.../tail/0] Task failed.
```

Control, same run, a non colliding task:

```
$ python flow.py spin ForeachPrefixFlow/<run>/tail/15     # fep=middle:0
Task finished successfully.
```

## Possible fix

`regex.fullmatch` instead of `regex.match` on that line looks like it covers it. The two deliberate pattern shapes seem unaffected:

* `".*"` still matches everything, and it also short circuits earlier in the function
* the foreach join pattern `"<path>,.*"` still matches deeper paths such as `middle:1,inner:0`

Applying that one change locally takes the run above from `1 of 12 tail tasks have the wrong parent count` to `0 of 12`, makes `spin .../tail/14` finish successfully, and leaves the control task passing.

[...]

PR #3342, opened 2026-08-16
Anchor metadata pattern matching so foreach paths do not prefix-match

nileshpatil6:
Addresses the local-provider side of #3341. The service-provider side is handled by #3357.

`LocalMetadataProvider.filter_tasks_by_metadata` matched metadata values with `regex.match`, which anchors only at the start. `Task.parent_tasks` / `child_tasks` build their pattern from `foreach-execution-path`, so once a foreach has 11 or more items the pattern `middle:1` also matched the values `middle:10` and `middle:11`, and the client resolved the wrong tasks. With `spin` this turns into a hard failure ("not a join step but gets multiple inputs").

The fix switches that one call to `regex.fullmatch`, so a pattern has to match the whole value.

The reason `fullmatch` is safe here is that the patterns this function receives are already written to cover the part of the path they care about. From `metaflow/client/core.py`:

- an exact ancestor path, e.g. `middle:1` — should match only that task, which is what this fixes
- a descendant pattern built as `f"{current_path},.*"` — the trailing `.*` still consumes the rest of the value, so nested foreach lookups are unaffected
- `".*"` for the match-all case — unaffected, and it is short-circuited before the regex anyway

I checked the values that a 12 item foreach produces:

| pattern | `re.match` (before) | `re.fullmatch` (after) |
|---|---|---|
| `middle:1` | `middle:1`, `middle:10`, `middle:11`, `middle:1,inner:0` | `middle:1` |
| `middle:1,.*` | `middle:1,inner:0` | `middle:1,inner:0` |
| `.*` | all | all |

Only the over-matching case changes.

The service metadata provider passes the pattern to the backend instead of matching locally, so this is the only client-side matcher affected.

Added `test_filter_tasks_by_metadata_does_not_match_prefixes` in `test/unit/test_local_metadata_provider.py`, covering the exact path, the prefixed sibling, a nested descendant pattern and match-all. Verified it fails before the change (the `middle:1` lookup returns three extra tasks) and passes after; the existing test in that file still passes.

### DIFF
--- a/metaflow/plugins/metadata_providers/local.py
+++ b/metaflow/plugins/metadata_providers/local.py
@@ -261,7 +261,7 @@ class LocalMetadataProvider(MetadataProvider):
 
             if any(
                 meta.get("field_name") == field_name
-                and regex.match(meta.get("value", ""))
+                and regex.fullmatch(meta.get("value", ""))
                 for meta in metadata
             ):
                 matching_task_pathspecs.append(
--- a/test/unit/test_local_metadata_provider.py
+++ b/test/unit/test_local_metadata_provider.py
@@ -29,3 +29,41 @@ def test_deduce_run_id_from_meta_dir():
             case["meta_path"], case["sub_type"]
         )
         assert case["expected_run_id"] == actual_run_id
+
+
+def test_filter_tasks_by_metadata_does_not_match_prefixes(monkeypatch):
+    # A foreach with 11+ items produces execution paths like "middle:1" and
+    # "middle:10". Matching with re.match only anchors the start, so the
+    # pattern "middle:1" also selected "middle:10" and "middle:11", giving
+    # Task.parent_tasks/child_tasks the wrong tasks.
+    paths = {
+        "1": "middle:1",
+        "2": "middle:10",
+        "3": "middle:11",
+        "4": "middle:1,inner:0",
+    }
+
+    def fake_get_object(cls, obj_type, sub_type, filters, attempt, *args):
+        if sub_type == "task":
+            return [{"task_id": task_id} for task_id in sorted(paths)]
+        task_id = args[-1]
+        return [
+            {"field_name": "foreach-execution-path", "value": paths[task_id]},
+        ]
+
+    monkeypatch.setattr(
+        LocalMetadataProvider, "get_object", classmethod(fake_get_object)
+    )
+
+    def filter_for(pattern):
+        return LocalMetadataProvider.filter_tasks_by_metadata(
+            "Flow", "run", "middle", "foreach-execution-path", pattern
+        )
+
+    # the exact path must not pull in the longer indices that start with it
+    assert filter_for("middle:1") == ["Flow/run/middle/1"]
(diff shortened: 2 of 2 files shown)


winklemad:
Thanks for picking this up. I checked the reasoning rather than taking it on trust, and the three pattern shapes are exactly as you describe — `client/core.py` builds `pattern = current_path or ".*"`, the bare `".*"`, and `f"{current_path},.*"` at both the `parent_tasks` (~1266) and `child_tasks` (~1329) sites. `fullmatch` is safe for all three: the descendant form's trailing `.*` still consumes the remainder, and `".*"` is short-circuited before the regex anyway. So the one-call change is right and I don't think it needs widening.

One thing I could not settle from this repo, which might matter for how the fix is framed:

`ServiceMetadataProvider.filter_tasks_by_metadata` (`metadata_providers/service.py:328`) doesn't do the matching locally — it forwards `pattern` as a query param to the metadata service and lets the server filter (`query_params["pattern"] = pattern`, then `GET .../filtered_tasks?...`). It has the same `.*` short-circuit, so the two providers agree on that case, but the anchoring behaviour for an exact path like `middle:1` depends on whatever the service does with that pattern, and that code isn't here.

If the service side anchors the same way `regex.match` did, a user on the service provider would still see `middle:1` pull in `middle:10` and `middle:11` after this merges, and the local and service providers would disagree. Probably a question for a maintainer rather than something to pull into this PR.

For context, I filed #3341 — glad it's getting fixed.

nileshpatil6:
Thanks for checking it rather than taking it on trust, and for filing #3341 in the first place.

I went looking for the service side, and I think the question is answerable. The filtering happens in Netflix/metaflow-service, `services/data/postgres_async_db.py`, in `get_filtered_task_pathspecs`:

```python
if pattern:
    conditions.append("regexp_match(value, %s) IS NOT NULL")
    values.append(pattern)
```

`regexp_match()` in PostgreSQL is an unanchored search: it returns the first substring matching the pattern anywhere in the value. So the service provider is not merely anchored the way `re.match` was, it is looser still. For `pattern = "middle:1"`:

| | `middle:1` | `middle:10` | `xmiddle:1` |
| --- | --- | --- | --- |
| old local, `re.match` | match | match (the bug) | no match |
| service, `regexp_match` | match | match | **match** |
| new local, `re.fullmatch` | match | no match | no match |

So yes, the two providers diverge after this merges, and they already diverged before it on the leading-substring case. Anchoring server-side would mean wrapping the pattern, roughly `value ~ ('^(?:' || %s || ')$')`, which keeps the `f"{current_path},.*"` descendant form working since the trailing `.*` still consumes the remainder.

I agree that does not belong in this PR: it is a different repo, it is a behaviour change for anyone relying on the current substring semantics, and it wants a maintainer's call on whether to anchor the service or relax the client. Happy to open an issue on metaflow-service with the above if a maintainer thinks it is worth tracking.

talsperre (dismissed):

talsperre (approved):

PR #3344, opened 2026-08-18
fix: argo 3.7.11 conditional input paths

### LINKED ISSUE #3043: Conditional branching broken on Argo Workflows v3.7.11+ / v4.0.2+
## Description

Flows using conditional branching (`self.next({...}, condition=...)`) hang indefinitely on Argo Workflows v3.7.11+ and v4.0.2+. The convergence step after conditional branches is never scheduled.

## Reproducer

The [metaflow-music-agent](https://github.com/outerbounds/metaflow-music-agent) example reproduces this. The `ResearchAgentFlow` uses:

```python
self.next(
    {
        "continue": self.research,
        "done": self.done,
        "unknown": self.unknown,
        "failed": self.failed,
    },
    condition="decision",
)
```

`done`, `unknown`, and `failed` all converge on `agent_complete` -> `join` -> `end`. Only one branch executes per foreach iteration; the others are skipped.

Run with:
```bash
python agentflow.py --environment=pypi --with kubernetes run --max-artists 2
```

The research steps complete successfully, each branch reaches `done` or `unknown`, but `agent_complete` is never scheduled and the workflow hangs in `Running` state.

## Root Cause

Two interacting issues:

### 1. How Metaflow generates `input-paths` for conditional join steps

For a convergence step like `agent_complete`, the generated Argo DAG template references outputs from **all** conditional branches unconditionally:

```yaml
arguments:
  parameters:
    - name: input-paths
      value: "argo-{{workflow.name}}/:done/{{tasks.done.outputs.parameters.task-id}},failed/{{tasks.failed.outputs.parameters.task-id}},unknown/{{tasks.unknown.outputs.parameters.task-id}}"
depends: "done.Succeeded || failed.Succeeded || unknown.Succeeded"
```

The `depends` clause is correct (OR), but `input-paths` references `task-id` from all three branches. Skipped branches produce no outputs.

Metaflow's intended workaround (introduced in #2560) is to base64-encode the value and filter out unresolved paths at runtime via `conditional_input_paths.py`. This depends on Argo passing through unresolved `{{tasks.X...}}` expressions as literal strings.

### 2. Argo Workflows changed parameter resolution behavior

Argo PR [argoproj/argo-workflows#15442](https://github.com/argoproj/argo-workflows/pull/15442) ("fix: requeue workflow if expected variables are missing", Feb 26 2026) changed `dag.go` from:

```go
// Before (v3.7.10, v4.0.1 and earlier) — unresolved expressions passed through
newTaskStr, err := template.Replace(ctx, string(taskBytes), mergedParams, true)
```

to:

```go
// After (v3.7.11, v4.0.2+) — unresolved tasks.*/steps.* references cause requeue
newTaskStr, err := template.ReplaceStrict(ctx, string(taskBytes), mergedParams, []string{"tasks", "steps"})
if err != nil {
    if template.IsMissingVariableErr(err) {
        woc.requeue()
        return nil, ErrRequeue  // pod never created, runtime filter never runs
    }
}
```

The controller now requeues indefinitely when it can't resolve `{{tasks.failed.outputs.parameters.task-id}}` for a skipped task. The `conditional_input_paths.py` runtime filter never gets a chance to execute.

### Controller logs confirm this

```
level=WARN msg="was unable to find variable" error="failed to resolve {{tasks.failed.outputs.parameters.task-id}}"
level=WARN msg="was unable to find variable" error="failed to resolve {{tasks.done.outputs.parameters.task-id}}"
level=WARN msg="was unable to obtain the node" taskName=agent-complete
```

## Affected Versions

| Argo Version | Behavior | Conditionals work? |
|---|---|---|
| v3.5.x - v3.7.10 | `Replace(allowUnresolved=true)` | Yes |
| **v3.7.11+** | `ReplaceStrict` | **No** |
| v4.0.0 - v4.0.1 | `Replace(allowUnresolved=true)` | Yes |
| **v4.0.2+** | `ReplaceStrict` | **No** |

Tested with Metaflow 2.19.22. Likely affects all versions since #2560 (Aug 2025).

## History

For context, the original conditionals PR (#2550, Aug 19 2025) used safe Argo expressions that handled skipped tasks correctly:

```
{{=(get(tasks['done']?.outputs?.parameters, 'task-id') ?? 'no-task')}}
```

This was removed two days later in #2560 (Aug 21 2025) because `?.` optional chaining doesn't work inside foreach (`withParam` nested DAGs evaluate all `?.` as false). The replacement strategy relied on Argo's passthrough behavior, which is no longer guaranteed.

## Argo-side tracking

This is related to [argoproj/argo-workflows#15737](https://github.com/argoproj/argo-workflows/issues/15737). A fix (PR [#15736](https://github.com/argoproj/argo-workflows/pull/15736)) has been merged and cherry-picked to v4.0 and v3.7, but that fix targets expanded task properties (`withItems`/`withParams`). It may not fully cover the skipped conditional branch case — worth verifying once v3.7.12 / v4.0.4 are released.

## Workaround

Downgrade Argo Workflows to v3.7.10 or v4.0.1.

## Environment

- Metaflow: 2.19.22
- Argo Workflows: tested on v4.0.3 and v3.7.11
- Kubernetes: EKS
- Python: 3.12

saikonen:
## PR Type

- [x] Bug fix

## Summary

Fixes an issue with step input-paths from conditionals not working on argo 3.7.11+/4.0.2+

## Issue

Fixes #3043 

## AI Tool Usage

Claude Code used for test cases and fix iteration

### DIFF
--- a/devtools/Tiltfile
+++ b/devtools/Tiltfile
@@ -12,8 +12,8 @@ allow_k8s_contexts('minikube')
 # ---------------------------------------------------------------------------
 # Version configuration
 # ---------------------------------------------------------------------------
-ARGO_WORKFLOWS_HELM_CHART_VERSION = os.getenv("ARGO_WORKFLOWS_HELM_CHART_VERSION", "0.45.2")
-ARGO_WORKFLOWS_IMAGE_TAG = os.getenv("ARGO_WORKFLOWS_IMAGE_TAG", "v3.6.0")
+ARGO_WORKFLOWS_HELM_CHART_VERSION = os.getenv("ARGO_WORKFLOWS_HELM_CHART_VERSION", "0.45.8")
+ARGO_WORKFLOWS_IMAGE_TAG = os.getenv("ARGO_WORKFLOWS_IMAGE_TAG", "v3.7.11")
 AIRFLOW_HELM_CHART_VERSION = os.getenv("AIRFLOW_HELM_CHART_VERSION", "1.15.0")
 AIRFLOW_IMAGE_TAG = os.getenv("AIRFLOW_IMAGE_TAG", "2.10.4")
 
--- a/metaflow/plugins/argo/argo_workflows.py
+++ b/metaflow/plugins/argo/argo_workflows.py
@@ -1230,6 +1230,20 @@ class ArgoWorkflows(object):
             reverse=True,
         )
 
+    def _input_path_ref(self, node_name):
+        sanitized = self._sanitize(node_name)
+        pred_node = self.graph[node_name]
+        if self._is_conditional_node(pred_node) or pred_node.type == "split-switch":
+            return (
+                "argo-{{workflow.name}}/%s/"
+                "{{=tasks['%s']?.outputs?.parameters['task-id'] ?? 'SKIPPED'}}"
+                % (node_name, sanitized)
+            )
+        return "argo-{{workflow.name}}/%s/{{tasks.%s.outputs.parameters.task-id}}" % (
+            node_name,
+            sanitized,
+        )
+
     def _is_recursive_node(self, node):
         return node.name in self.recursive_nodes
 
@@ -1383,11 +1397,7 @@ class ArgoWorkflows(object):
                 parameters = [
                     Parameter("input-paths").value(
                         compress_list(
-                            [
-                                "argo-{{workflow.name}}/%s/{{tasks.%s.outputs.parameters.task-id}}"
-                                % (n, self._sanitize(n))
-                                for n in node.in_funcs
-                            ],
+                            [self._input_path_ref(n) for n in node.in_funcs],
                             # NOTE: We set zlibmin to infinite because zlib compression for the Argo input-paths breaks template value substitution.
                             zlibmin=inf,
                         )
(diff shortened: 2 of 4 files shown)


Shriprasad-P (commented):
## Review

**PR:** fix: argo 3.7.11 conditional input paths

**Touched:** `devtools/Tiltfile`, `metaflow/plugins/argo/argo_workflows.py`, `metaflow/plugins/argo/conditional_input_paths.py`, `test/unit/test_argo_conditional_input_paths.py`, `test/unit/test_argo_nested_conditional_join.py`

- CI/tooling change — confirm the pipeline still passes on this branch.
- Diff is fairly large (+632/-156); a short summary of risk areas from the author would help reviewers.

Gsbreddy:
Wanted to share some field validation for this fix.

We (Delivery Hero, running Metaflow on Argo Workflows at scale) hit exactly the issue this PR fixes (#3043) while validating an Argo Workflows v3.7.11→v4.1.4 upgrade. We reproduced the conditional-branch failure cleanly on both v3.7.15 and v4.1.4 (`unable to substitute {{tasks.X.outputs...}}: output is an absent optional`), then pinned our Metaflow install to this branch's commit (326d90a8) and re-ran the same repro on both versions — conditional branching now completes successfully on both, with the new `cond-` wrapped Steps templates showing up as expected in the compiled DAG.

This is a real blocker for anyone upgrading past Argo Workflows v3.7.11/v4.0.2, so it'd be great to see this land.

PR #3348, opened 2026-08-23
[GSoC] Add paginated, filtered run iteration to the core client

### LINKED ISSUE #3320: Metadata Client fetches unbounded child collections (no pagination/limit/server-side filter) — O(N) for wide foreach scans
## Problem

`ServiceMetadataProvider._get_object_internal`
(`metaflow/plugins/metadata_providers/service.py`) fetches child collections with
a bare, unbounded `GET` and then filters client-side:

```python
url += "/%ss" % sub_type          # e.g. /tasks, /artifacts
v, _ = cls._request(None, url, "GET")
return MetadataProvider._apply_filter(v, filters)
```

There are no `limit`, pagination, or time-range parameters, so the response size
grows with the number of objects, and any filtering happens **after** the full
list is transferred. For Client-side workflows over wide runs this scales poorly.

## Motivation / where it hurts

While profiling the Client API (patching `ServiceMetadataProvider._request` to
count HTTP calls), I saw two patterns worth improving:

- Simple status checks fan out into several metadata requests plus artifact
  downloads (e.g. reading `run.successful` on a small flow triggers multiple
  metadata GETs and a `_success` artifact fetch).
- For a `foreach` with **N** parallel tasks, finding which tasks failed is
  effectively **O(N)**: there is no way to filter or batch server-side, so the
  client pulls every task/artifact and filters locally.

This is precisely the kind of overhead that matters for programmatic and
agent-style consumers of the Client that poll or scan runs repeatedly.

## Prior art in the same file

`filter_tasks_by_metadata` already introduced a server-side `filtered_tasks`
endpoint with `metadata_field_name` / `pattern` query params — so the pattern of
pushing filtering to the service exists. `_get_object_internal` predates that and
still does the naive full-list fetch.

## Possible directions (seeking maintainer input)

I'd like guidance on the preferred approach before writing code, since this
touches Core Runtime (Metadata) and likely the metadata service contract:

1. **Pagination / `limit` params** on the child-collection GETs, with the client
   iterating pages lazily.
2. **Server-side filtering** for common Client queries (status, attempt,
   time-range), extending the `filtered_*` pattern rather than downloading then
   `_apply_filter`-ing.
3. **Batch/targeted fetches** for the foreach failure-scan case specifically
   (e.g. request only failed tasks / only the `_success` markers).

Some of these are client-only; others need metadata-service changes. Happy to
prototype whichever direction aligns with the roadmap — this also overlaps with
making the Client more efficient for automated/agent use.

## Environment

- Metaflow version: 2.19.35 (also present on current `master`)
- File: `metaflow/plugins/metadata_providers/service.py` (`_get_object_internal`)

*(Filing as a design discussion first, per the Core Runtime contribution
guidelines — not a PR yet.)*

Aryan95614:
**GSoC 2026 project: metadata service request improvements (mentor: @saikonen).** Upstream twin of the fork review PR (Aryan95614/metaflow#1), opened per mentor request for a static PR number.

## Summary

Moves cursor pagination and server-side run filtering into the regular Metaflow client instead of maintaining a parallel client stack.

- adds a provider-level `iter_objects()` path with a backwards-compatible fallback
- version-gates the paginated path on the metadata service (>= 2.6.1) with an `X-Limit` capability-header check, falling back to legacy listing for older services and rejecting server-side filters when the service cannot honor them
- generalizes pagination to all collection types, not just `flow`/`run` listings
- streams service listings through `X-Next-Cursor`
- forwards the metadata service's `field:operator` filters
- extends `Flow.runs()` with explicit keyword-only `filters`, `page_size`, and `max_runs`
- applies positional tags locally, so `Flow.runs("prod")` works with local metadata as well as the service
- sorts listings newest-first by `ts_epoch`, so `max_runs` returns the newest runs regardless of provider order
- guards reserved pagination parameters and repeated cursors
- preserves the existing no-option `Flow.runs(*tags)` behavior

## Separation

This is the base PR. It contains pagination and filtering only. The agent-oriented failure investigation utility is isolated in stacked PR #2.

This structure follows the review direction that pagination, filtering, and normal record iteration belong in the core client.

## Verification

- 41 focused pagination/provider compatibility tests passed
- `black` (repo-pinned 25.12.0) clean on the changed files
- complete unit suite: no new regressions (pre-existing `inheritance`/`spin` integration failures are environmental and unrelated)

## Note on the version gate

The paginated/filtered path is gated on metadata-service `>= 2.6.1` — the release that ships pagination + filtering together with the artifact latest-attempt fix (metaflow-service#497) — plus an `X-Limit` capability-header check as defense-in-depth. Services without it fall back to legacy listing, and server-side filters raise a clear error rather than silently returning unfiltered results.

## Public surface

`Flow.runs()` takes `*tags`, `_filters`, and `max_runs`.

`_filters` is intentionally private: the `field:operator` grammar belongs to the metadata service and has no meaning for the local provider, so it should not be something callers depend on. `page_size` is not part of the signature at all — `runs()` returns an iterator either way, so page size is a transport concern and comes from `METAFLOW_SERVICE_PAGE_SIZE`. `max_runs` stays public because, unlike the other two, it is meaningful for every metadata provider.

## Future work (not in this PR)

Filters could be expressed in plain python and translated by each provider, so the local metadata provider could support the same calls:

```python
flow.runs(filters={"status": eq("failed"), "ts_epoch": gt(cutoff)})
```

with each provider owning a `filter_funcs_mapping` that matches its internals. Keeping `_filters` private now means that can land later without a breaking change. Credit for the idea goes to @saikonen.


metaflow/client/core.py
```diff
@@ -2585,7 +2616,13 @@ def latest_successful_run(self) -> Optional[Run]:
             if run.successful:
                 return run
 
-    def runs(self, *tags: str) -> Iterator[Run]:
+    def runs(
+        self,
+        *tags: str,
+        filters: Optional[Dict[str, Any]] = None,
+        page_size: Optional[int] = None,
+        max_runs: Optional[int] = None,
```
saikonen: one last comment on the public-facing interface here. I'm wondering if we could hide the `filters` under the hood as this is a service-specific detail, and the filter grammar is not supported for local metadata in any way, nor does it necessarily make sense.

metaflow/client/core.py
```diff
@@ -2596,14 +2633,43 @@ def runs(self, *tags: str) -> Iterator[Run]:
...
         Yields
         ------
         Run
             `Run` objects in this flow.
         """
-        return self._filtered_children(*tags)
+        if filters is not None and not hasattr(filters, "items"):
+            raise TypeError("filters must be a mapping")
+        if max_runs is not None:
+            if isinstance(max_runs, bool) or not isinstance(max_runs, int):
```
saikonen, line 2656: why the check for `bool`? Isn't the integer check enough?

metaflow/client/core.py
saikonen: can we also do away with the page_size and max_runs here, or are they required? page size is a metaflow config level toggle already so I see no reason to keep it as a call level arg as well. It is also service-specific, as the return here is an iterator anyway so the caller does not concern with page sizes

metaflow/client/core.py
```diff
@@ -2596,14 +2633,43 @@ def runs(self, *tags: str) -> Iterator[Run]:
...
+        max_runs : int, optional
+            Maximum number of runs to yield, newest first.
 
         Yields
         ------
         Run
             `Run` objects in this flow.
         """
-        return self._filtered_children(*tags)
+        if filters is not None and not hasattr(filters, "items"):
```
saikonen: wouldn't checking for dict type work better?

metaflow/client/core.py
saikonen: one quick modification to get this rolling until we agree on the public-facing bits would be to mark some of these args as internal, to discourage relying on them. e.g.
```python
def runs(self, *tags: str, _filters: Optional[Dict[str, Any]] = None)
```

@talsperre what do you think? Or is there a simple way to expose the supported filters as kwargs already, that could be implemented for the local metadata as well as a follow-up?

talsperre: Let's keep it as internal only for now and we can always expose it later.

talsperre: +1

metaflow/plugins/metadata_providers/service.py
```diff
@@ -252,6 +262,176 @@ def _mutate_user_tags_for_run(
...
+            url += "/metadata"
+        elif sub_type == "artifact" and obj_type == "task" and attempt is not None:
+            url += "/attempt/%s/artifacts" % attempt
+        else:
+            url += "/%ss" % sub_type
+        return url
+
+    @classmethod
+    def _can_paginate_collection(cls, sub_type, attempt):
+        return sub_type != "self" and attempt is None
```
talsperre: We should not paginate artifact already. `return sub_type not in ("self", "artifact") and attempt is None`. The current metadata service does artifact selction based on latest task attempt, but this PR will find the latest occurrence of every artifact.

talsperre:
Some additional comments as well [here](https://gist.github.com/talsperre/25928a133c8195455323343ed34ab619)

metaflow/client/core.py
Aryan95614: Made it `_filters`. The docstring now says the grammar belongs to the metadata service and doesn't mean anything for local metadata, so it shouldn't be part of the public surface.

Keeping it private also leaves room for the idea you sketched afterwards, where the filters are plain python and each provider translates them through its own mapping. I've added that to the PR description as future work so it doesn't get lost.

metaflow/client/core.py
Aryan95614: `bool` is a subclass of `int`, so `isinstance(True, int)` is `True`. Without that line `runs(max_runs=True)` would quietly mean "one run" instead of raising. It does look redundant though, so I've put a comment on the line and added a test case for it.

metaflow/client/core.py
Aryan95614: I dropped `page_size` from the signature entirely rather than just making it private. You're right that it has no use for the local implementation, and `runs()` hands back an iterator either way, so the caller never really needed it. It comes from `METAFLOW_SERVICE_PAGE_SIZE` now, which I've documented, validated, and clamped to the service maximum of 500.

I did keep `max_runs` public, and I'd rather flag that than slip it past you, since @talsperre +1'd removing both. Sakari's point was that `max_runs` is different from the other two: it works for every provider including local, and `failed_runs(max_runs=...)` in #3349 is built on it. If you'd still prefer it gone, I'll take it out.

metaflow/client/core.py
Aryan95614: Agreed, switched to `isinstance(_filters, Mapping)` from `collections.abc`. Clearer than the `hasattr` check, and it still accepts dict subclasses.

metaflow/plugins/metadata_providers/service.py
```diff
@@ -252,6 +262,176 @@ def _mutate_user_tags_for_run(
             time.sleep(0.3 * random.uniform(1.4, 1.6) ** tries)
             tries += 1
 
+    @classmethod
+    def _service_supports_cursor_pagination(cls):
+        if cls._supports_cursor_pagination is None:
+            version = cls._version(None)
+            cls._supports_cursor_pagination = version is not None and version_parse(
+                version
+            ) >= version_parse(cls._MIN_SERVICE_VERSION_WITH_CURSOR_PAGINATION)
```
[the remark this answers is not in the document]
Aryan95614: This was a real bug, thanks. Rather than key just the pagination flag, I replaced all three capability booleans with one version cache keyed by the service URL, since @talsperre pointed out they share the same weakness. Each caller keeps its own minimum-version error message, and there's now one version ping per service instead of three.

There's a test that switches the service URL between a 2.5 and a 2.6 endpoint and checks the answer changes in both directions.

metaflow/plugins/metadata_providers/service.py
Aryan95614: Done, with a comment on the line explaining why, so it doesn't get "optimised" back later.

Thank you for tracing this to the SQL. `DISTINCT ON (task_id, name)` really is a different operation from the legacy latest-attempt filter, and I hadn't spotted that the two paths disagree on retried tasks. I'll open a follow-up on metaflow-service with the `latest_attempt` CTE. The client-side guard should stay until that ships and is released, since the client can't tell a fixed 2.6 from an unfixed one.

Aryan95614:
Thanks for taking the time to write this up properly. The artifact one in particular I would not have found on my own.

Pushed a commit that covers all six points and the test feedback.

**1. Artifact pagination.** Artifacts are excluded from the paginated path now. The root cause is on the service side, so I'll follow up there with the `latest_attempt` CTE; the client guard should stay until that's released.

**2. Capability cache.** Replaced all three global booleans with a single version cache keyed by service URL, rather than URL-keying only the pagination flag. Each call site keeps its own error message, and one ping per service now covers all three checks.

**3. Only the first page was checked.** The `X-Limit` check runs on every response. The first page can still fall back to the legacy listing because nothing has been yielded at that point; after that it raises.

**4. Repeated cursor.** Raises now, separated from the normal "no next cursor" completion.

**5. Validation before HTTP.** `_validate_object_query` moved to the top of `iter_objects`. The test makes both `_version()` and `_request()` blow up, so it fails if that ever slides back down.

**6. Tests.** You were right, several of them weren't checking what their names claimed. The `max_runs` ordering test and the local tag-filtering test now run end to end against a real `Flow` and the local provider instead of slicing provider output or mocking `_iter_children`. The collection-type test is parametrized over every paginable collection. The file is on `pytest-mock` per CONTRIBUTING.md. The repeated-cursor test that locked in silent truncation now expects the exception.

Since your point was that a green CI proved nothing here, I went back and reverted each fix one at a time to check the new tests actually fail: the artifact exclusion, the per-page header check, the repeated-cursor raise, and the URL-keyed cache each break exactly the test meant to catch them. The full unit suite has the same failures before and after my commit.

Two things I didn't quietly decide on my own:

- `max_runs` is still public. Details in the thread above, and I'm happy to drop it.
- The artifact test only asserts that artifacts take the legacy path. It can't honestly compare paginated and legacy artifact results, because the difference lives in the service's SQL and the client no longer paginates them at all. That comparison belongs in the metaflow-service PR.

I've also fixed the 2.5.1 / 2.6.0 mismatch in the description and added a short future work note about the provider-translated filter grammar.

#3349 is stacked on this, so I've left it alone for now and will update it once this settles.

saikonen (dismissed):

metaflow/plugins/metadata_providers/service.py
```diff
@@ -252,17 +265,218 @@ def _mutate_user_tags_for_run(
...
+        """Whether the currently selected metadata service is at least min_version.
+
+        Returns a boolean only -- each caller owns its own error message. The
+        service version is cached per service URL, so one ping serves every
+        capability check against that service, and switching services rechecks.
+        """
+        url = cls.INFO
+        if url not in cls._service_version_cache:
+            cls._service_version_cache[url] = cls._version(None)
+        version = cls._service_version_cache[url]
```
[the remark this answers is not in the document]
saikonen: The cache is not retained between initializations, so this should be fine as is.

saikonen (approved):

talsperre (approved):

PR #3349, opened 2026-08-23
[GSoC] Add failure-investigation accessors to the core client (stacked on #3348)

Aryan95614:
**GSoC 2026 project: metadata service request improvements (mentor: @saikonen).** Upstream twin of the fork review PR (Aryan95614/metaflow#2), opened per mentor request for a static PR number.

Rebased onto `master` after #3348 merged (2026-09-15); the diff is only this PR's own commits.

## Summary

Folds bounded failure investigation into the regular Metaflow client as three
accessors on the existing objects, instead of a separate `metaflow/agent`
module. Built on the core pagination/filtering from #3348.

## Context / Motivation

Review on the earlier revision steered the run-listing and pagination pieces
into the core client (that's #3348). Following the same direction, the failure
investigation layer now lives on the client objects rather than a parallel
`metaflow.agent` package, so there's a single, discoverable API.

## Changes Made

- `Flow.failed_runs(*tags, since=None, max_runs=None) -> Iterator[Run]`: failed
  runs newest-first, via `runs(*tags, _filters={"status:eq": "failed"})`; `tags`
  are forwarded exactly as for `runs`, and `since` adds an inclusive
  `ts_epoch:ge` bound.
- `Run.failed_task -> Optional[Task]`: the latest task that finished without
  success (newest step first), or None. A task counts as failed when it raised
  (`exception` is set) or a decorator such as `@catch` handled the failure;
  tasks that are still running or never started are skipped.
- `Task.failure_summary -> Optional[FailureSummary]`: normalized
  `{exception_type, message, stacktrace, attempt}`, or None when the task has
  no exception. Read errors propagate rather than being swallowed.
- `FailureSummary` is exported from `metaflow.client` and `metaflow`, next to
  `Flow`, `Run` and `Task`.
- Removed the `metaflow/agent` module (`failure_finder`, `pagination`,
  `run_lister`, `TaskRef`/`FailureInfo`, injectable resolver/fetcher).

Behavior note: `failed_runs` relies on server-side filtering, so it raises
against the local metadata provider or a service without pagination/filtering
support: the same gate as `runs(_filters=...)`.

## Testing

- `test/unit/test_client_failure_investigation.py`: 19 tests covering
  `failed_runs` filter/tags/bounds forwarding (including one through the real
  `Flow.runs` down to the provider call), `failed_task` iteration order and the
  crashed / decorator-handled / still-running cases, `failure_summary`
  construction + error propagation, `_normalize_exception` across
  mapping/object/string inputs, and the package exports.
- Combined with the merged #3348 run-listing tests (`test_client_run_listing.py`): 63 passed.
- Checked against a local run with one raising step and one `@catch` step:
  `failed_task` returns the raising task and `failure_summary` carries its type and message.
- `black` (repo-pinned 25.12.0) clean.

- `failed_task` cannot use `finished` alone: `Task.finished` reads `_task_ok`,
  which the task runner sets to False when the step raises, so
  `finished and not successful` would match only decorator-handled failures
  and never the task that crashed. It checks `exception is not None or finished`
  after ruling out successful tasks.
- Dropped the injectable `task_resolver` / `artifact_fetcher` and the `TaskRef`
  indirection: unnecessary once this is core; real `Task` objects are used and
  the methods are testable with mocks.
- Kept `FailureSummary` as a small frozen dataclass in `core.py` rather than
  returning a bare dict, so the normalized fields are typed and discoverable.


Shriprasad-P (commented):
## Review

**PR:** [GSoC] Add failure-investigation accessors to the core client (stacked on #3348)

**Touched:** `metaflow/client/core.py`, `metaflow/metadata_provider/metadata.py`, `metaflow/metaflow_config.py`, `metaflow/plugins/metadata_providers/service.py`, `test/unit/test_client_failure_investigation.py`

- Scoped to the stated goal; please confirm CI is green.
- Size: +1249/-46.

Drive-by review after reading the diff.

metaflow/client/core.py
```diff
@@ -2333,6 +2406,39 @@ def successful(self) -> bool:
...
+        ```
+
+        Returns
+        -------
+        Task, optional
+            The latest unsuccessful task, or None if the run has none.
+        """
+        for step in self:
+            for task in step:
+                if not task.successful:
```
saikonen: this expression will be true for tasks that are running, pending, or have not launched because it simply pulls from the `task["_success"].data`

saikonen: alternative could be
```python
if task.finished and not task.successful:
```

metaflow/client/core.py
```diff
@@ -2673,6 +2779,39 @@ def runs(
...
+
+        Yields
+        ------
+        Run
+            Failed `Run` objects in this flow, newest first.
+        """
+        filters = {"status:eq": "failed"}
+        if since is not None:
+            filters["ts_epoch:ge"] = int(since)
+        return self.runs(filters=filters, max_runs=max_runs)
```
saikonen: should be `_filters`

metaflow/client/core.py
```diff
@@ -2673,6 +2779,39 @@ def runs(
             return runs
         return islice(runs, max_runs)
 
+    def failed_runs(
```
saikonen, line 2793: this could also pass through tags as filters.

metaflow/client/core.py
Aryan95614: Fixed in the latest push. The tests mocked runs with a plain Mock, which accepted the wrong keyword; they now autospec Flow so runs enforces its real signature, and one test goes through the real runs() down to the provider call.

metaflow/client/core.py
```diff
@@ -62,6 +63,48 @@
...
+    """
+    Normalized description of the exception that failed a `Task`.
+
+    Returned by `Task.failure_summary`. Any field may be None when the stored
+    exception artifact does not carry that particular detail.
+    """
+
+    exception_type: Optional[str]
+    message: Optional[str]
+    stacktrace: Optional[str]
```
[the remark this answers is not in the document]
Aryan95614: Exported from metaflow.client and from the top-level metaflow package in the latest push, next to Flow, Run and Task.

metaflow/client/core.py
Aryan95614: Agreed on the problem, but finished on its own does not work here. Task.finished reads _task_ok, and task.py sets that to False when the step raises, so finished and not successful would match only failures a decorator handled (for example @catch) and never the task that actually crashed. I checked on a local run with one raising step and one @catch step: the raising task reads finished=False, successful=False with exception set; the @catch task reads finished=True, successful=False with exception None.

The latest push rules out successful tasks first, then returns the task if task.exception is not None or task.finished. That covers the crashed task and the handled one, and skips tasks that have not written their artifacts yet (running, pending, never launched), which read as unsuccessful, unfinished and without an exception. Tests cover all four cases.

metaflow/client/core.py
Aryan95614: Done. failed_runs takes tags positionally like runs and forwards them, so Flow("F").failed_runs("prod", max_runs=5) works.

PR #3352, opened 2026-08-25
[Do not review] fix: normalize Argo conditional execution

talsperre:
## Summary

- normalize conditional nodes behind stable public wrappers and use success barriers for graph dependencies
- cover conditional foreach, recursive switches, and conditional parallel/JobSet templates
- use direct recursive switch output selection for legacy DAG exits, avoiding the nested-foreach expression regression while retaining safe navigation for non-recursive switches
- add regression coverage for simple, nested, sequential, recursive, foreach, parallel, failure, and skip-chain shapes

This PR is stacked on #3344 and targets its head branch so it contains only the follow-up normalization.

## Verification

- 35 Argo-focused unit tests passed
- all pre-commit hooks passed
- Python compilation and `git diff --check` passed
- Argo 3.7.11 E2E passed for active/inactive conditional foreach, selected branch failure propagation, recursive conditionals, and recursive conditionals inside foreach
- Argo 3.6.0 controller and executor E2E passed for inactive conditional foreach, selected branch failure propagation, and recursive conditionals inside foreach
- three independent code reviews found no remaining correctness or minimality blocker

## Remaining live-test caveat

Conditional `@parallel`/JobSet generation is covered structurally, but was not run against a live JobSet controller because the local stack did not install one.


Shriprasad-P (commented):
Thanks for opening this. The approach looks reasonable. I'd mainly want confirmation that existing callers keep working and that tests cover the failure path.

PR #3354, opened 2026-08-26
Enable argo hooks to inherit security context from start step

### LINKED ISSUE #3353: [Argo-Workflows]: Copy Kubernetes Security Context into error and heartbeat  templates
The functions `_lifecycle_hook_from_deco`,  `_error_msg_capture_hook_templates`, and `_heartbeat_daemon_template`

https://github.com/Netflix/metaflow/blob/d2dd7f953574369d5b4715822b12377d44ba1cf8/metaflow/plugins/argo/argo_workflows.py#L3041
https://github.com/Netflix/metaflow/blob/d2dd7f953574369d5b4715822b12377d44ba1cf8/metaflow/plugins/argo/argo_workflows.py#L3188
https://github.com/Netflix/metaflow/blob/d2dd7f953574369d5b4715822b12377d44ba1cf8/metaflow/plugins/argo/argo_workflows.py#L3650

all include the same pattern of copying the start step's `@kubernetes` attributes

https://github.com/Netflix/metaflow/blob/d2dd7f953574369d5b4715822b12377d44ba1cf8/metaflow/plugins/argo/argo_workflows.py#L3192-L3198

and then use this `resources` dict to fill in attributes such as the image and secrets of the container. However, it does not copy over the security context defined from the original `@kubernetes` decorator, which can result in, for instance, the resulting pod being labeled as insecure or even rejected.

To resolve this issue, I think it should be sufficient to specify the `security_context` in the generated container using the same security context from resources (`resources["security_container"]`).

## Reproducing

I used the following flow and ran the comment `python flow.py argo-workflows create --only-json` (note: datastore required).

```python
from metaflow import FlowSpec, kubernetes, step

_SECURITY_CONTEXT = {"allow_privilege_escalation": False}

class TestFlow(FlowSpec):
    """
    Minimal example to show error-msg-capture-hook does not inherit
    the same security context as all the steps
    """

    @kubernetes(security_context=_SECURITY_CONTEXT)
    @step
    def start(self):
        print("Start")
        self.next(self.end)
    
    @kubernetes(security_context=_SECURITY_CONTEXT)
    @step
    def end(self):
        print("End")

if __name__ == '__main__':
    TestFlow()
```

This will produce a `start`, an `end`, and an `error-msg-capture-hook`. The `start` and `end` steps have the following security context: 

```json
"securityContext": {
    "allowPrivilegeEscalation": false,
    "appArmorProfile": null,
    "capabilities": null,
    "privileged": null,
    "procMount": null,
    "readOnlyRootFilesystem": null,
    "runAsGroup": null,
    "runAsNonRoot": null,
    "runAsUser": null,
    "seLinuxOptions": null,
    "seccompProfile": null,
    "windowsOptions": null
},
```
meanwhile the error hook has `"securityContext": null`.

kevslinger:
Fixes #3353

Before this, generated hooks in flows on argo would have `"securityContext": null`. After this change, a custom security context applied to the start step (via e.g. a `@kubernetes(security_context={...})`) will get picked up and included in the hook.

Using the reproduce example from #3353, after this change, all 3 of `start`, `end`, and `error-msg-capture-hook` have the same security context. A key benefit of this PR is that generated hooks will now not get blocked by security systems which expect a strict security context in all pods.


Shriprasad-P (commented):
The production change looks correct. These generated Argo templates already derive several Kubernetes settings from the start step, and using the same start-step `security_context` here is consistent with that model. `V1SecurityContext(**security_context)` also matches how ordinary step containers handle `@kubernetes(security_context=...)`.

I do think one thing should be addressed before merge: this bug fix currently has no regression coverage, and the existing Argo tests do not assert `securityContext` for these generated templates.

A compile-to-JSON test should be enough:
- configure `security_context` on the start step
- assert it appears on the lifecycle hook, `error-msg-capture-hook`, and `heartbeat-daemon`
- verify the field remains unset/null when no security context is configured

The implementation itself looks right; the missing regression test is the remaining gap.

Shriprasad-P (commented):
The production change looks correct. These three auxiliary templates already inherit start-step Kubernetes settings such as image, secrets, and service account, and ordinary step containers already convert security_context through V1SecurityContext. Propagating the same container security context to the lifecycle hook, error-capture hook, and heartbeat daemon therefore looks consistent with the existing Argo compilation model.

The blocker for me is regression coverage. There currently appears to be no test asserting securityContext on these generated templates, and Metaflow's contribution guidance requires a test for bug fixes.

Shriprasad-P (commented):
## Review

Propagating `security_context` from start-step resources onto Argo hooks / error-capture / heartbeat daemon templates closes an important pod-security gap (hooks otherwise may violate restricted PSS while the main container is fine).

Looks consistent across the three template sites. Would be good to confirm empty/`None` still omits the field for clusters that reject empty securityContext objects.

PR #3356, opened 2026-08-30
Fix decompress_list IndexError on empty string input

Anurag-M1:
## Summary

`decompress_list("")` crashes with `IndexError` because `lststr[0]` is accessed with no guard for empty input. This breaks the round-trip invariant `decompress_list(compress_list([])) == []`.

## Context / Motivation

`compress_list([])` legitimately returns `""` (joining an empty list). However, `decompress_list("")` immediately crashes at line 387 because `lststr[0]` is accessed without checking for empty input first. An empty task ID list is a valid real-world state.

Fixes #3017

## Changes Made

- Added early return `if lststr == "":  return []` in `decompress_list()` before the `lststr[0]` access
- Uses strict `== ""` (not `not lststr`) so `None` still fails loudly — if `None` slips through, the real bug should be caught upstream

## Testing

Added `test/unit/test_compress_decompress_list.py` with 7 test cases:

| Test | What it verifies |
|------|-----------------|
| `test_compress_empty_list` | `compress_list([]) == ""` |
| `test_decompress_empty_string` | `decompress_list("") == []` (the actual bug) |
| `test_roundtrip_empty` | `decompress_list(compress_list([])) == []` |
| `test_roundtrip[single]` | Round-trip with `["x"]` |
| `test_roundtrip[simple]` | Round-trip with `["a", "b", "c"]` |
| `test_roundtrip[common-prefix]` | Items with shared prefix trigger `:` encoding |
| `test_roundtrip_zlib` | Force zlib path via `zlibmin=0` |

All 7 tests pass. Full unit test suite unaffected.

```
$ python3 -m pytest test/unit/test_compress_decompress_list.py -v
7 passed in 0.17s
```

## AI Disclosure

- [x] I used AI tools in preparing this contribution

AI tools were used to help identify the issue and structure the PR. All code was reviewed and understood before submission.


talsperre:
It makes fixes for two separate issues. Let's just do a targeted fix of https://github.com/Netflix/metaflow/issues/3017 and then I can run the tests/merge.

Anurag-M1:
Updated as requested! Rebased onto `master` and narrowed down the PR to only include the targeted fix for #3017 along with unit tests.

Shriprasad-P (approved):
I reproduced the empty-list round-trip failure against the base revision.

`compress_list([])` legitimately produces an empty string, while `decompress_list("")` currently indexes `lststr[0]` and raises `IndexError`. Handling the empty string before that lookup restores `decompress_list(compress_list([])) == []` without changing the behavior of `None` or malformed non-empty inputs.

I also checked the existing plain, prefix/range, and zlib paths, which remain unchanged, and the new empty-input regression tests fail on the base behavior and pass with this patch.

The current diff is also scoped to the requested #3017 fix and its tests.

LGTM.

Shriprasad-P (commented):
## Review

Simple, correct guard: empty string → `[]` before indexing `lststr[0]`. Easy IndexError to hit from empty env/config values.

Approve of the approach; the dedicated compress/decompress unit tests are a good addition.

PR #3357, opened 2026-08-31
Anchor service metadata patterns so foreach paths do not prefix-match

winklemad:
## PR Type

- [x] Bug fix

## Summary

Handles the OSS service-provider side of #3341. `Task.parent_tasks` and
`Task.child_tasks` pass foreach execution-path patterns to the metadata
service, whose regex matching is unanchored. An exact path such as `middle:1`
could therefore also select `middle:10` and `middle:11`.

The local-provider side is handled separately by #3342.

## Root Cause

`ServiceMetadataProvider.filter_tasks_by_metadata` forwarded the raw pattern
to the metadata service:

```python
query_params["pattern"] = pattern
```

The service applies that pattern as an unanchored regex search.

## Fix

Wrap the forwarded pattern in full-string anchors:

```python
query_params["pattern"] = "^(?:%s)$" % pattern
```

This makes `middle:1` exact while preserving the descendant pattern
`middle:1,.*`. The existing `.*` match-all case remains short-circuited and
sends no pattern.

## Tests

`test/unit/test_service_metadata_provider.py` covers:

- exact path anchoring;
- descendant pattern anchoring;
- the match-all short circuit.

Run with:

```bash
python -m pytest test/unit/test_service_metadata_provider.py -q
```

## Non-Goals

- The local metadata provider and its tests are handled by #3342.
- No metadata-service repository change is required; anchoring happens at the
  OSS client boundary.

## AI Tool Usage

- [x] AI tools were used.

Tool disclosed by the author: Claude Code, used to help implement the original
change and tests. The service-only rescope retains that implementation.


nileshpatil6:
Thanks for the credit and for laying out the overlap so clearly, that was a considerate way to handle it.

The split you proposed works well from my side: #3342 has been open since Aug 16 with the local provider fix (`regex.match` to `regex.fullmatch`) and a prefix-matching test that fails on master, and the service-side anchoring here is genuinely the piece it does not cover. So landing #3342 for the local half and scoping this one to the service provider seems like the cleanest path.

@talsperre happy to go with whatever ordering you prefer.

winklemad:
Thanks @nileshpatil6, likewise, your #3342 nailed the local-provider side. Hoping we can get both landed.

talsperre (approved):

winklemad:
Thanks for the review @talsperre 🙏

PR #3358, opened 2026-08-31
Fix mutable_flow test flow passing a stale second argument to extract_step_decorator_from_decospec

Anai-Guo:
`extract_step_decorator_from_decospec` accepts exactly one argument
(`metaflow/decorators.py:692`):

```python
def extract_step_decorator_from_decospec(decospec: str):
```

but both call sites in the `mutable_flow` test flow still pass a second
positional `{}`:

```python
# test/test_config/mutable_flow.py:61  (ModifyFlow.mutate)
env_deco, _ = extract_step_decorator_from_decospec(decos[0], {})

# test/test_config/mutable_flow.py:110 (ModifyFlow2.mutate)
env_deco, _ = extract_step_decorator_from_decospec(deco, {})
```

Both are reached from a `FlowMutator.mutate()`, which runs at flow-definition
time, so `mutable_flow.py` raises `TypeError: extract_step_decorator_from_decospec()
takes 1 positional argument but 2 were given` before any step executes. The flow
is registered as a test case in `test/test_config/test.py:96`.

The two call sites in library code already pass a single argument, which is what
makes this a stale-caller bug rather than a signature question:

| Call site | Args | Status |
|---|---|---|
| `metaflow/decorators.py:768` | 1 | ✅ |
| `metaflow/user_decorators/mutable_step.py:340` | 1 | ✅ |
| `test/test_config/mutable_flow.py:61` | 2 | ❌ |
| `test/test_config/mutable_flow.py:110` | 2 | ❌ |

Binding the real upstream signature against all four call sites, with the two
library callers as the control group:

```
extract_step_decorator_from_decospec (decospec)
  TypeError test/test_config/mutable_flow.py:61  (decos[0], {}) -> too many positional arguments
  TypeError test/test_config/mutable_flow.py:110 (deco, {})     -> too many positional arguments
  OK        CONTROL metaflow/decorators.py:768
  OK        CONTROL metaflow/user_decorators/mutable_step.py:340
```

This PR drops the extra `{}` at both sites so the test flow matches the
signature and the two library callers. Nothing else changes — the return value
is still unpacked as `(step_deco, has_args_kwargs)`, exactly as
`mutable_step.py:340` does.


Shriprasad-P (commented):
The extra `{}` is definitely stale, but I don't think removing it is sufficient to make this flow work against the current mutable-step API.

`MutableStep.decorator_specs` now yields `(name, fq_name, args, kwargs)` tuples. These two paths still call `.startswith("environment:")` on those values, so they can fail before `extract_step_decorator_from_decospec()` is reached.

There are a couple of related stale API usages here as well: the decorator attributes are already available from the tuple, and `add_decorator()` now expects them through `deco_kwargs=` (with `duplicates=OVERRIDE` when replacing the existing environment decorator).

The current implementation in `test/ux/core/flows/config/mutable_flow.py` already follows that pattern. I think `test/test_config/mutable_flow.py` should be aligned with that implementation rather than continuing to parse these specs through `extract_step_decorator_from_decospec`.

So the arity change itself is correct, but the test flow still appears incompatible with the current decorator-spec API.

Shriprasad-P (commented):
## Review

Test-only fix dropping the stale second `{}` arg to `extract_step_decorator_from_decospec` — matches the current helper signature and unblocks the mutable_flow config tests.

LGTM as a surgical test correction.

Issue #2924, opened 2026-02-25
AWS Sandbox STS fetch hangs indefinitely when endpoint is unreachable


AwkJay:
### Description

When `AWS_SANDBOX_ENABLED` is set, `Boto3ClientProvider.get_client()` in `metaflow/plugins/aws/aws_client.py` makes a `requests.get()` call to the sandbox STS endpoint without an explicit timeout.

If the endpoint is unreachable or misconfigured, the process hangs indefinitely. In my local testing, this hang lasted over 2 minutes.

### Steps to Reproduce

1. Set `AWS_SANDBOX_ENABLED=1` and point `AWS_SANDBOX_STS_ENDPOINT_URL`
   at an unreachable host
2. Call any function that triggers `Boto3ClientProvider.get_client()`
3. Result: The process hangs for 2+ minutes before raising a raw requests traceback.

### Expected Behavior

Metaflow should fail-fast with a descriptive `MetaflowException` after a reasonable timeout.

### Environment

- Metaflow version: 2.19.19
- Python version: 3.13.9
- OS: Linux

Happy to submit a PR for this.

PR #3361, opened 2026-09-02
fix: pass a timeout to the metadata heartbeat request

### LINKED ISSUE #3360: MetadataHeartBeat._heartbeat posts without a timeout, so its own `except requests.exceptions.Timeout` handler is not reachable
`_heartbeat()` in `metaflow/metadata_provider/heartbeat.py` does a `requests.post` with no timeout, but there's an `except requests.exceptions.Timeout` handler right below it. requests doesn't time out by default, so that handler can never actually fire.

`_ping()` above it retries by catching `HeartBeatException`:

```python
except HeartBeatException as e:
    retry_counter = retry_counter + 1
    time.sleep(1.5**retry_counter)
```

So if the metadata service accepts the connection but stops responding, `_heartbeat()` never returns and never raises, and that retry never runs. The heartbeat thread just sits there and the run stops heartbeating without anything showing up.

Reproduced with a socket that accepts and never replies, then the same post heartbeat.py makes today:

```python
import socket, threading, requests

srv = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
srv.bind(("127.0.0.1", 0)); srv.listen(5)
port = srv.getsockname()[1]

def stall():
    conns = []
    while True:
        c, _ = srv.accept(); conns.append(c)   # hold open, never respond

threading.Thread(target=stall, daemon=True).start()
requests.post(url=f"http://127.0.0.1:{port}/ping", data="{}", headers={})
```

Still blocked past 15s. With `timeout=(3.05, 10)` it raises Timeout after 10s, so the existing handler fires.

Happy to open a pr with a test that fails without the fix. Not the same as #2924, that one is the aws sandbox sts call in `aws_client.py`.
alr working on a fix on local, should push soon

RudraDudhat2509:
## PR Type

- [x] Bug fix
- [ ] New feature
- [ ] Docs / tooling
- [ ] Refactoring

## Summary

The metadata heartbeat now times out instead of blocking forever when the service accepts the connection but stops responding, so `_ping` can retry and back off like it was already written to.

## Issue

Fixes #3360 

## Reproduction

**Runtime:**  local 
**Commands to run:**
```bash
python -m pytest test/unit/test_metadata_heartbeat.py -q
```
**Where evidence shows up:** 

<details>

```
>       assert timeout is not None, "heartbeat post must pass a timeout to requests"
E       AssertionError: heartbeat post must pass a timeout to requests
E       assert None is not None
1 failed, 1 passed
```

</details>

<details>

```
2 passed, 1 warning in 7.32s
```

The underlying hang, against a socket that accepts and never replies:

```
[1] no timeout  -> still blocked after 15s, never returns, never raises
[2] with timeout -> requests.exceptions.Timeout after 10s
```

</details>

## Root Cause

`_heartbeat` posted without a timeout. requests has no default timeout, so if the metadata service accepts the connection but never replies, the post blocks indefinitely.

Two things follow from that. The `except requests.exceptions.Timeout` handler directly below the call is unreachable, since nothing can ever raise it. And `_ping`, which is the retry loop, only retries when `_heartbeat` raises `HeartBeatException`:

```python
except HeartBeatException as e:
    retry_counter = retry_counter + 1
    time.sleep(1.5**retry_counter)
```

A call that neither returns nor raises never reaches that, so the retry counter and the `1.5**retry_counter` backoff never run. The heartbeat thread stalls and the run stops heartbeating without surfacing anything.

## Why This Fix Is Correct

The handler for this case already exists, it just could not be reached. Passing a timeout makes the existing `Timeout` branch reachable, which raises `HeartBeatException`, which is what `_ping` already knows how to retry. No new error paths are introduce by this fix, and there is no change to the retry logic.

`(3.05, 10)` is connect and read. 3.05 for connect follows the requests guidance of sitting just above a multiple of 3 so it does not line up with TCP retransmit windows.

## Failure Modes Considered

1. **Timeout too aggressive on a slow but healthy service.** A slow response now raises instead of waiting indefinitely. This degrades to a retry rather than a failure, since `_ping` catches `HeartBeatException` and backs off before pinging again. 10s read is well above a normal heartbeat response.
2. **Behaviour change for anyone relying on the current blocking.** Previously a hung service silently stalled the thread. Now it raises and retries. That is the behaviour the existing `Timeout` handler and its error message were written for.
3. **Value coupled to ping frequency.** Considered tying the read timeout to `default_frequency_secs`, but the interval is server controlled through `wait_time_in_seconds`, so coupling to it would be fragile. Kept it a constant.

## Tests

- [x] CI passes

`test_heartbeat_post_passes_a_timeout` is the regression test, it fails on master. It asserts a positive timeout is passed rather than asserting a specific constant, so it does not lock in the value.

`test_heartbeat_timeout_raises_heartbeat_exception` passes either way. It documents that a timing out request surfaces as `HeartBeatException` so `_ping` can retry.

Full `test/unit` run is unchanged against master: 14 failed, 165 errors both before and after, all from optional deps missing locally, with 396 passing before and 398 after. Ran under WSL Ubuntu since metaflow imports `fcntl` and cannot be imported on Windows. black is clean on both files. 

## Non-Goals

Not touching the AWS sandbox STS call in `aws_client.py`, that is #2924 and has a PR open. Did not add a config knob for the timeout, and did not change the retry or backoff logic. tried to keep the change as scoped and minimal as possible 

## AI Tool Usage

Claude Sonnet 5 was used. It wrote the change in `heartbeat.py` and the two tests, and found the bug while grepping for `requests` calls with no timeout. I reviewed all of it, ran the tests both ways, and can explain the causal chain and why the timeout value was chosen. other than that i ran all the setup and the code manually to see there was no degradation.


Shriprasad-P:
Reviewed the heartbeat failure path against master and #3360.

The root cause checks out: `_heartbeat()` calls `requests.post` without a timeout, so a peer that accepts the connection but never responds can block indefinitely. In that state the existing `Timeout -> HeartBeatException -> _ping()` retry/backoff path is never reached.

Adding a finite `(connect, read)` timeout makes the existing failure handling reachable without changing the retry logic, successful heartbeat path, or `wait_time_in_seconds` behavior. The regression test also does the right thing by asserting that a finite timeout exists without coupling the test to the exact timeout constant.

One minor observation: `ConnectTimeout` is still caught by the earlier `ConnectionError` branch, but it still becomes `HeartBeatException` and retries correctly, so that is pre-existing and non-blocking.

I don't see a blocking issue here. LGTM.

Shriprasad-P (commented):
Reviewed the heartbeat failure path against master and #3360.

The root cause checks out: `_heartbeat()` calls `requests.post` without a timeout, so a peer that accepts the connection but never responds can block indefinitely. In that state the existing `Timeout -> HeartBeatException -> _ping()` retry/backoff path is never reached.

Adding a finite `(connect, read)` timeout makes the existing failure handling reachable without changing the retry logic, successful heartbeat path, or `wait_time_in_seconds` behavior. The regression test also does the right thing by asserting that a finite timeout exists without coupling the test to the exact timeout constant.

One minor observation: `ConnectTimeout` is still caught by the earlier `ConnectionError` branch, but it still becomes `HeartBeatException` and retries correctly, so that is pre-existing and non-blocking.

I don't see a blocking issue here. LGTM.

Shriprasad-P (approved):
Reviewed the heartbeat failure path against master and #3360.

The root cause checks out: `_heartbeat()` calls `requests.post` without a timeout, so a peer that accepts the connection but never responds can block indefinitely. In that state the existing `Timeout -> HeartBeatException -> _ping()` retry/backoff path is never reached.

Adding a finite `(connect, read)` timeout makes the existing failure handling reachable without changing the retry logic, successful heartbeat path, or `wait_time_in_seconds` behavior. The regression test also does the right thing by asserting that a finite timeout exists without coupling the test to the exact timeout constant.

One minor observation: `ConnectTimeout` is still caught by the earlier `ConnectionError` branch, but it still becomes `HeartBeatException` and retries correctly, so that is pre-existing and non-blocking.

I don't see a blocking issue here. LGTM.

Shriprasad-P (commented):
## Review

Passing an explicit timeout to the metadata heartbeat `requests.post` is necessary — without it, hangs never surface to the existing Timeout → `HeartBeatException` path, so backoff never runs.

Tests asserting a finite timeout kwarg and Timeout → HeartBeatException look right. LGTM.

RudraDudhat2509:
gentle ping @maintainers

PR #3371, opened 2026-09-12
fix(argo): propagate nodeSelector/tolerations to error and exit hooks

### LINKED ISSUE #3309: Argo workflows error-msg-capture-hook nodeSelector and tolerations propagation
It seems like the default `nodeSelector` and `tolerations` added to all step pods (via METAFLOW_KUBERNETES_TOLERATIONS and METAFLOW_KUBERNETES_NODE_SELECTOR), aren't added to the argo workflows error-msg-capture-hook pod, as well. 

This makes it so this pod can't re-use the same node from the previous steps.

Is this intended since this is a lightweight pod that should run in a slimmer node?

https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/argo/argo_workflows.py#L3272

https://github.com/Netflix/metaflow/blob/master/metaflow/plugins/argo/argo_workflows.py#L2850-L2853

Shriprasad-P:
## Fixes
Fixes #3309

## Summary
Propagate `nodeSelector` and `tolerations` into the Argo error-msg-capture-hook and `@exit_hook` templates so hooks can schedule on the same nodes as workflow steps.

## Test plan
- [x] Added compilation tests covering with/without nodeSelector+tolerations
- [x] Verified hooks inherit resources when env configuration is present


Shriprasad-P (commented):
Solid direction. Worth double-checking backward compatibility for existing callers of the touched API.

PR #3377, opened 2026-09-16
Warn instead of raising when a card component has no update()

### LINKED ISSUE #3376: Card component update() raises TypeError instead of warning when not implemented
### Summary

Calling `.update()` on a card component that does not implement it raises `TypeError` instead of emitting the intended warning.

`UserComponent.update` builds its warning message with `%` formatting, but the string has no placeholder:

```python
# metaflow/plugins/cards/card_modules/components.py
def update(self, *args, **kwargs):
    cls_name = self.__class__.__name__
    msg = (
        "MetaflowCardComponent doesn't have an `update` method implemented "
        "and is not compatible with realtime updates."
    ) % cls_name
    _warning_with_component(self, msg)
```

`cls_name` is computed and then never interpolated, so `%` raises before `_warning_with_component` is reached.

### Reproduction

```python
from metaflow.plugins.cards.card_modules.components import UserComponent

class MyComponent(UserComponent):
    def render(self):
        return "x"

MyComponent().update("new value")
```

Actual:

```
TypeError: not all arguments converted during string formatting
```

Expected: the documented warning, naming the component class. This is the documented realtime API (`current.card["mycardid"].components["comp123"].update()` in `component_serializer.py`), so a user whose custom component lacks `update` gets a `TypeError` from inside Metaflow rather than the "not compatible with realtime updates" message.

A component that does implement `update`, such as `Markdown`, is unaffected.

### Environment

Reproduced on `master` (8b500da), Python 3.11.

David-Wu1119:
## PR Type

- [x] Bug fix
- [ ] New feature
- [ ] Docs / tooling
- [ ] Refactoring

## Summary

Calling `.update()` on a card component that does not implement it raised `TypeError` instead of the intended warning. The message now names the component class, as it was written to.

## Issue

Fixes #3376

## Reproduction

**Runtime:** local (no runtime needed; the failure is in `UserComponent.update`)

**Commands to run:**
```bash
python -c "
from metaflow.plugins.cards.card_modules.components import UserComponent
class MyComponent(UserComponent):
    def render(self): return 'x'
MyComponent().update('new value')
"
```

<details>

```
  File "<string>", line 5, in <module>
  File ".../metaflow/plugins/cards/card_modules/components.py", line 37, in update
    msg = (
TypeError: not all arguments converted during string formatting
```

</details>

<details>

```
# no exception; with a logger attached the component warns once:
[@card-component WARNING] MyComponent doesn't have an `update` method implemented and is not compatible with realtime updates.
```

```
$ pytest test/unit/test_card_components.py -q
..                                                                       [100%]
2 passed
```

</details>

## Root Cause

`UserComponent.update` computes `cls_name` and then applies `%` to a string with no conversion specifier:

```python
cls_name = self.__class__.__name__
msg = (
    "MetaflowCardComponent doesn't have an `update` method implemented "
    "and is not compatible with realtime updates."
) % cls_name
```

`str.__mod__` raises `TypeError: not all arguments converted during string formatting` when the format string consumes no arguments, so the exception is raised while building the message, before `_warning_with_component` runs. The unused `cls_name` shows the placeholder was intended: the message was meant to name the component, not the base class.

This is on the documented realtime path (`current.card["mycardid"].components["comp123"].update()` in `component_serializer.py`), so the failure surfaces in user flows.

## Why This Fix Is Correct

The invariant restored is "a component without `update` warns and continues". One character of format string is changed, `%s` in place of the hardcoded `MetaflowCardComponent`, which both removes the `TypeError` and makes the message identify the offending class. `_warning_with_component` already guards on `_logger` and `_warned_once`, so behaviour after the message is built is unchanged. Components that implement `update`, such as `Markdown`, never reach this code.

## Failure Modes Considered

1. **Silent suppression.** The fix does not swallow anything: `_warning_with_component` still returns early when `_logger` is `None` (outside a task) and still logs at most once per component. Both are covered by the new tests.
2. **Message churn for parsers.** The text keeps its wording and the `[@card-component WARNING]` prefix added by `_warning_with_component`; only the leading class name becomes accurate, so anything grepping for the phrase still matches.

## Tests

- [x] CI passes

`test/unit/test_card_components.py` covers the warning (it names the class, and no exception is raised) and the warn-once behaviour. Both tests fail on `master` with the `TypeError` above and pass with this change. `black --check` is clean on both files.

## Non-Goals

Nothing else in the cards module is touched; other `%`-formatted messages and the realtime refresh path are unchanged.


Shriprasad-P (commented):
Thanks for the PR. Approach looks sensible — please confirm edge cases and failure paths are covered by tests.

